Issue #3, opened 2020-08-12
Improve accessibility by using higher contrast on the website.


GabLotus:
In the spirit of accessibility we should be careful to not use difficult to read color formats/palettes on the website.

pppKin:
Can we also have light/dark mode switch? I have trouble reading texts on dark background and have always used in-browser reader on bevy websites. However the browser (Readed View in firefox at least) tend to mess up original page layout,  making it difficult to navigate through the article.

PR #5, opened 2020-08-12
Tweak color contrast and spacing for easier reading

tristanpemble:
Thought I'd put a little bit of time tonight into tweaking the landing page for a little bit of visual clarity/accessibility improvements. I tried to stay true to the spirit of the existing site. Let me know if you want me to undo something. Some things included:

- While the contrast of background+text passed a11y checkers like WAVE, I personally found it a little bit difficult on the eyes due to the combination of contrast + font weights, so I adjusted both.
- Adjusted contrast of the "Get Started" button. I'm not that happy with where it landed aesthetically. it at least passes a11y tests now
- Attempted to make spacing of elements a little bit more rhythmically consistent, because I have a compulsion
- Added alt tags to images for screen reading

Tackles a little bit of #3, there are still some small tweaks that can help with a11y checks -- easiest thing to do is just put the site into a [tool like this](https://wave.webaim.org/) and work through them. This PR knocks a few out.

Before
before

After
after

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -1,20 +1,21 @@
 html {
     overflow-y: scroll;
 }
+
 html, body {
     width: 100%;
     height: 100%;
     margin: 0 0 0 0;
     color: $default-color;
     font-family: 'Fira Sans', sans-serif;
-    background-color: #232326;
+    background-color: $background-color;
 }
+
 @media screen and (max-width: 650px) {
     html, body {
         font-size: 0.7rem;
     }
 }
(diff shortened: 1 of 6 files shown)


GabLotus:
I like it. I think the gray image boxes should be made darker too so that they are more seemless with the background like before. It would also have the added benefit of increasing contrast between the box and its content which is what is important here. (I dont think the contrast between the box and the background is an accessibility issue but I'm open to being corrected here)

PR #8, opened 2020-08-13
Adds OS dependencies section to Setup

### LINKED ISSUE #6: Add an "OS Dependencies" section to The Book's Setup page
Users (especially on linux) often don't have the required dependencies to build bevy. It would significantly reduce friction to provide the list for them.

This should extend the existing Setup page with something like:

```markdown
## Install OS Dependencies
* [Linux](https://github.com/bevyengine/bevy/docs/linux_dependencies.md)
* Windows: You should be good to go!
* MacOS: No dependencies here!
```

Then the bevy repo would contain `linux_dependencies.md` with one section for each linux variant.

MatteoGgl:
This commit resolves #6 

A commit on the main bevy repository has [already been submitted](https://github.com/bevyengine/bevy/pull/161) to create the linked document.

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -18,6 +18,11 @@ Install Rust by following the [Rust Getting Started Guide](https://www.rust-lang
 
 Once this is done, you should have the ```rustc``` compiler and the ```cargo``` build system installed in your path.
 
+### Install OS dependencies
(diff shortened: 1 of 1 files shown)


cart:
hmm CI failure isn't your fault from the looks of it

cart:
CI is clearly broken (probably because its trying to actually deploy .. scary). I'm merging

PR #9, opened 2020-08-13
Fix broken code example

Halfwhit:
Added `mut` to `position` in relation to [this issue.](https://github.com/bevyengine/bevy/issues/123)


cart:
Looks good to me. CI failure isn't your fault.

PR #13, opened 2020-08-22
Updated code in the tutorial

eXodiquas:
The code in the tutorial didn't compile anymore because the Timer::from_seconds() method needs two arguments to work. So I rewrote the code to work with the new version of the Timer::from_seconds() method.


cart:
Brilliant. Thank you!

PR #14, opened 2020-08-23
Fix Atom feed

hannobraun:
The current RSS feed configuration is broken as of the latest Zola version. See commits for details.


cart:
Yeah CI isn't your fault. We really need to fix that :)

cart:
Looks good. Thanks!

PR #15, opened 2020-08-23
Improve Ergonomics Example

CleanCut:
Add default plugins so `movement` system runs repeatedly.

`add_default_plugins()` was already introduced earlier, so it's not a new concept.  Without it, the `movement` system runs only once, which makes it hard to see the difference between `add_startup_system` and `add_system`.

If the extra line is still deemed too much or distracting, I'm happy to just close the pull request.


cart:
This seems reasonable to me. Thanks!

PR #20, opened 2020-08-26
Change windows to include using binutils

Dispersia:
It was changed to use rust-lld instead, but it doesn't specify that you need to install cargo-binutils, as rust-lld is not shipped by default for most platforms. Cargo-binutils adds rust-lld, and llvm-tools-preview adds the actual lld executable to your toolchain.


cart:
So is cargo-binutils actually required? I want to make sure this is a minimal as possible.

Dispersia:
I just tried on a fresh install and yes, this is still required. Thankfully though, both of the steps are very fast

PR #23, opened 2020-08-28
Tweak getting started

CleanCut:
While I was reading through it, I made a few tweaks to improve things I noticed in the "Getting Started" section.

- `--bin` is unnecessary on `cargo new` commands, since it is the default setting.
- I added an intermediate suggestion to configure _only the project_ to use nightly Rust, before considering changing the entire local toolchain to default to nightly.
- Adjusted a variable name to silence warnings for anyone actually running the example code.


cart:
Looks great!

PR #24, opened 2020-08-28
Only 'build' for pull requests. Build AND deploy for pushes (merges) to master.

CleanCut:
I noticed in https://github.com/bevyengine/bevy-website/pull/23 that the CI (luckily) failed to deploy the changed website to production.

Thrashing the web site with deployments from every commit of every pull request is _probably_ not what was intended, so this change makes it so that pull requests only attempt to _build_ the web site.  Later, when the pul request is merged to master then it will build _and_ deploy the web site.  Pushes straight to master will also build and deploy.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ name: CI
 #  or any pull request or pull request
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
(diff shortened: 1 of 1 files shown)


CleanCut:
If I'm reading [the CI error](https://github.com/bevyengine/bevy-website/pull/24/checks?check_run_id=1043287955#step:4:8) correctly, the `GITHUB_TOKEN` environment variable is not set in your repo or org Actions settings.

If that environment variable is absent on purpose to prevent pull requests from deploying the web site, then merging this PR should enable you to restore the environment variable safely.

If the environment variable _is_ present, then perhaps GitHub simply does not provide it to CI builds of public forks yet (I know [they do for forks of _private_ repositories](https://github.com/github/roadmap/issues/75)), in which case this change should still be beneficial for those with write access to this repository.

.github/workflows/main.yml
```diff
@@ -2,29 +2,38 @@
...
 
-    # Runs a single command using the runners shell
-    - name: shalzz/zola-deploy-action
+    - name: 'Build website'
       uses: shalzz/zola-deploy-action@master
       env:
         PAGES_BRANCH: gh-pages
+        BUILD_DIR: .
+        BUILD_ONLY: true
         TOKEN: ${{secrets.TOKEN}}
```
cart, line 26: can you try changing this to:

```
TOKEN: fake-secret
```

I just tried that and it worked for me.

cart (changes requested):

cart:
Awesome! I consider this a gap in the zola deploy github action. They shouldn't require a token for non-deployment builds

cart (approved):

CleanCut:
Weird! 🤣

Issue #2, opened 2020-08-11
Add a (curated?) section to the bevy website to include games created by the community


GabLotus:
A particular emphasis on open source games would be particularly useful to newcomers who want further examples of how to use the engine.

![image](https://user-images.githubusercontent.com/25371684/89936682-dadbc700-dbc8-11ea-8449-0f8836e1430c.png)

cart:
I love it. Lets start brainstorming what such a page would look like.

morr0ne:
Would it peraphs makes sense to add [awesome-bevy](https://github.com/bevyengine/awesome-bevy) to the site?

jamadazi:
We already have Awesome-Bevy as the official de-facto "hub" for featuring community work.

I'd rather feature that on the website, rather than create and maintain a new page, which would really just contain duplicate information.

cart:
Ultimately I would like to replace Awesome Bevy on github with a more searchable / graphical implementation in `bevy-website` (although it might still feed off of a schematized awesome-bevy repo). But yeah for now awesome-bevy fills that niche.

morr0ne:
Awesome-bevy is enough right now but we should probably think about the future. Someday bevy may (hopefully) become much bigger and awesome-bevy could become a mess (just look at [awesome-nodejs](https://github.com/sindresorhus/awesome-nodejs)). A more indexable and easier to navigate version on the website would be nice

alice-i-cecile:
This is now done as part of "Bevy Assets".

PR #26, opened 2020-08-29
Add Games section

guimcaballero:
I've created a new section called Games, as per #2. I'm not sure if we still need this, considering we have the Games section in awesome-bevy, but since the issue is still open I've gone ahead and created it anyway. 

Here's a screenshot of how it looks: 

Games section screenshot

Currently it only lists bevy-nbody, but more can be easily added by creating other folders in `content/games`.


cart:
I think long term we will want something like this, but I think I want to wait a bit for a few reasons:

1. We need to develop a page (and process) that provides "fair" visibility to everyone, but also gives us a chance to curate and put bevy's best foot forward. I think a static / flat list doesn't accomplish either of those goals. This will take time and I'm not sure its worth the investment at this point, given that there arent that many games yet.
2. awesome-bevy already has decent visibility and can tide us over until we can do (1).

cart:
Closing this as awesome-bevy currently fills this niche. Long term, we want to build an interactive/searchable website-integrated replacement for awesome-bevy, but that likely won't happen for awhile.

PR #29, opened 2020-08-30
Added dependency optimization suggestion

TheRawMeatball:
For [bevy#346](https://github.com/bevyengine/bevy/issues/346)

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -91,7 +91,6 @@ To enable fast compiles, install the nightly rust compiler and LLD. Then copy [t
 
 ### Add Bevy to your project's Cargo.toml
 
-
 Bevy is [available as a library on crates.io](https://crates.io/crates/bevy), the official Rust package repository. Find the latest version number ([![Crates.io](https://img.shields.io/crates/v/bevy.svg)](https://crates.io/crates/bevy)) and add it to your Cargo.toml file:
 
 ```toml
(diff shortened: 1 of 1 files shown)


content/learn/book/getting-started/setup/_index.md
```diff
@@ -105,6 +104,15 @@ edition = "2018"
 bevy = "0.1.2" # make sure this is the latest version
 ```
 
+### Make debug builds faster
+
+Debug Rust builds are notoriously slow, and this becomes an issue when making games. Luckily, you can set up rust so your dependendencies (including Bevy) are compiled with optimizations enabled, giving you faster runtime speeds without sacrificing debuggability or iteration speed with the ```--release``` flag. To enable this, append the following to your Cargo.toml file:
+
+```toml
+[profile.dev.packages."*"]
```
will-hart, line 112:
```suggestion
[profile.dev.package."*"]
```

`cargo run` seems to throw a warning when `packages` is used instead of `package`

TheRawMeatball: Yup, thats a typo. I'll fix it when I open my computer.

content/learn/book/getting-started/setup/_index.md
```diff
@@ -105,6 +104,15 @@ edition = "2018"
...
+opt-level = 3
```
stefee, line 113: Does this work on mac as well? I’ve found that I need to run with `debug = false` as well in order to get fast iterative compiles

content/learn/book/getting-started/setup/_index.md
```diff
@@ -105,6 +104,15 @@ edition = "2018"
...
+```
+
 Run ```cargo run``` again. The Bevy dependencies should start building. This will take some time as you are essentially building an engine from scratch. You will only need to do a full rebuild once. Every build after this one will be fast!
```
stefee, line 116: There’s no heading above this paragraph which might suggest it’s still part of the previous section. Maybe we should move the `make debug builds faster` section somewhere else. Maybe @cart can suggest a better place.

TheRawMeatball: There wasn't a title beforehand either, but perhaps there should. I'm open to suggestions

PR #30, opened 2020-08-31
Remove whitespace from getting started command

ablakey:
Not a big deal, but when you double click to copy the command and paste in your terminal, a bit of whitespace comes with it.


cart:
Still an improvement. Thanks :smile:

PR #36, opened 2020-09-18
Added troubleshooting section to "Getting Started"

sdfgeoff:
In particular with the solution to the issue https://github.com/bevyengine/bevy/issues/509


cart:
Good call. Sorry for taking so long to merge this!

PR #46, opened 2020-10-16
Added book nav toggle for mobile

### LINKED ISSUE #37: Book: Navigation is not visible on mobile devices
On small screen sizes or mobile devices the navigation (selector "nav.book-nav") is hidden by CSS. This way it isn't really possible to read the  book on mobile devices.

image

I would expect, that the navigation is either still visible on mobile devices, or an alternative way to navigate the book is provided.

nic96:
Here's my solution to fix the book navigation not being visible on mobile. Basically what my changes do is if the url path starts with `/learn/book` it replaces the bevy mobile logo with bars to toggle the nav. When the bars are pressed or clicked the book-nav will slide in. Currently I've coded it so that the book-content will slide to the right when the book-nav is open, but if you'd prefer another solution, for example, maybe the book-nav should cover the content instead of it sliding to the right, just let me know. I can change the code to do that.

Closes #37

### DIFF
--- a/sass/_book-nav.scss
+++ b/sass/_book-nav.scss
@@ -1,15 +1,24 @@
 $inactive-nav-color: #999999;
 
 @media screen and (max-width: 900px) {
+    #show-book-nav:not(:checked) + .book-nav {
+        transform: translateX(calc(0px - var(--book-nav-width)));
+    }
+
+    #show-book-nav:checked + .book-nav + .book-content {
+        transform: translateX(var(--book-nav-width));
+    }
+
     .book-nav {
-        display: none;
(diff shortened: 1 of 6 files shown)


nic96:
I just realized that this probably won't work on iOS since I'm using a JavaScript free way of toggling the book-nav. Should work on all other browsers though, but might have to add some JavaScript to make this work in iOS.

nic96:
I tested it on [appetize.io](https://appetize.io/demo) and it works there from iOS 9.3 and up so I don't think there is any need to add JavaScript to make the book-nav work.

cart:
Nice! Looks good to me. This is much appreciated :heart:

PR #58, opened 2020-11-04
Simplify previous/next links

Keats:


cart:
Very much appreciated!

1. Fixes a nasty bug
2. Removes a nasty implementation

This is so much nicer :heart:

PR #61, opened 2020-11-15
Small update on base.html

shirshak55:
- [x] Remove Unnecessary space
- [x] Add html5 support tag so it doesn't show red when you open view page source :D

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -6,6 +6,7 @@
 {%- else -%}
 {%- set page_title = "Bevy - A data-driven game engine built in Rust "-%}
 {% endif -%}
+
 {%- if section and section.path -%}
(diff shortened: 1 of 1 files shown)


cart:
I dig adding the support tag, but I'm not sure removing the space is an improvement. The space helps you visually parse sets of if statements. Removing the space makes it look like one big blob imo. Can you update the pr to only add the support tag?

shirshak55:
@cart Is it ok now?

cart:
Just removed the weird `-` that were added. now its good to go.

shirshak55:
that weird - just don't allow new lines :)  Anyway if you are fine I thing we are ok with it 💯

PR #66, opened 2020-11-29
Make using bevy as a git dependency more scary

amberkowalski:
I agree with [this](https://github.com/bevyengine/bevy/issues/890#issuecomment-730862427) comment, so I added some scary warnings.


cart:
I'm sold. Thanks!

PR #69, opened 2020-12-08
Change examples url

memoryruins:
This changes the link

from https://github.com/bevyengine/bevy/tree/master/examples
to https://github.com/bevyengine/bevy/tree/master/examples#examples

so that the user sees the warning about differences in master and previous release examples immediately instead of being hidden beyond the scroll. An alternative would be to make the link point to 0.3, but this avoids letting the url become outdated.


cart:
Sorry for the delay!

cart:
/rebase

sorry for the spam, im testing out a new "autorebase" action.

cart:
I think the last one would have worked if I was allowed to edit your branch. I'll stop spamming you now and merge this :smile:

PR #70, opened 2020-12-16
Added lld warning

morr0ne:
Related to [bevyengin/bevy#1062](https://github.com/bevyengine/bevy/issues/1062)

Since lld comes with some tradeoff it seems oportune to add a warning (maybe it should say disclaimer?) in the setup section.
I also don't have proper benchmark about the slower speed so unless someone is able to provide those it might be better to just state the size tradeoff and prompt the user to run his own benchmarks.
Maybe enabling it just for debug builds would be better?


memoryruins:
It would be nice to see an example demonstrating this in https://github.com/bevyengine/bevy/issues/1062 before adding the disclaimer.

cart:
Closing as this is pretty old and would need to be updated. Sorry for letting this sit for so long!

PR #84, opened 2021-01-16
Tell people to add Bevy to their Cargo.toml before suggesting Fast Compiles

Newbytee:
Enabling Bevy's Dynamic Linking Feature fails if Bevy isn't in Cargo.toml which might throw beginners off, so let's re-organise it like this to help those.

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -69,6 +69,23 @@ edition = "2018"
 [dependencies]
 ```
 
+### Add Bevy to your project's Cargo.toml
+
+Bevy is [available as a library on crates.io](https://crates.io/crates/bevy), the official Rust package repository. Find the latest version number ([![Crates.io](https://img.shields.io/crates/v/bevy.svg)](https://crates.io/crates/bevy)) and add it to your Cargo.toml file:
+
+```toml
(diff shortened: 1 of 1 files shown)


cart:
Its definitely worth talking about. The tutorial does "flow" better this way (and it avoids the failure mentioned above). The downside is that you need to re-compile Bevy if you ran `cargo run` before enabling fast compiles, which will waste a few minutes of people's lives (if they decide to change to nightly, enable new nightly features, or change linkers).

I think if we make this change, we'll want to hold of on asking people to run `cargo run` until after the "fast compiles" step.

Newbytee:
> Its definitely worth talking about. The tutorial does "flow" better this way (and it avoids the failure mentioned above). The downside is that you need to re-compile Bevy if you ran `cargo run` before enabling fast compiles, which will waste a few minutes of people's lives (if they decide to change to nightly, enable new nightly features, or change linkers).
> 
> I think if we make this change, we'll want to hold of on asking people to run `cargo run` until after the "fast compiles" step.

I updated the wording in a new commit. I think it sounds a bit awkward like this, but I'm not sure how to best word it. Suggestions welcome.

cart:
I just moved "cargo run" to the end of the page to avoid the need for clarification.

PR #88, opened 2021-01-30
update ecs.svg to not use obsolete syntax

### LINKED ISSUE #75: Screen shot on the home page uses obsolete for-each system syntax
The [main page](https://bevyengine.org/) screen shot of the syntax is now obsolete, as for-each systems have been deprecated.

The replacement code is:

```rust
struct Component;

fn system(query: Query<Entity, Component>){
}
```

But that doesn't demonstrate how to actually use a Query object smoothly.

Something like:

```rust
struct Component;

fn system(query: Query<Entity, Component>){
  for (e, c) in query.iter(){
  }
}
```

Is probably the most representative of actual use now.

jamadazi:
Keeping it as close as possible to the old version. Identical font, identical colors, identical style.

Closes #75


cart:
I pushed a new svg (and css changes) to make it use the space more effectively. Thoughts?

jamadazi:
I don't like the new SVG. I think it looks ugly and less readable, because it does not follow typical Rust formatting conventions. It also feels very dense/cluttered. It takes me more mental effort to parse the syntax.

I know the query line in my SVG was a little long, but I overall think it looked better.

I agree with the html/css change. It helps it fill the space better.

cart:
I 100% agree that "natural rust formatting" is more legible and pleasing to the eye. But the way it occupies the space is very unnatural, especially in the presence of the other "feature images". It feels like a pretty major downgrade from the previous version (ignoring the more complicated Query syntax)

Side-by-side comparison:

### 0. Old Syntax (original)

image

### 1. Natural Rust Formatting (first in pr)

image

### 2. Space Saving Formatting (second in pr)
image

### 3. Natural Rust Formatting (comma removed, single space indent, space in tuple removed, Query newline)

image

### 4. Natural Rust Formatting (comma removed, some spaces removed,  max image bounds hacked to 90% but only for this image)

image

### 5. Same as (4) but with the variable moved to the previous line

image

cart:
I'm thinking (4) or (5) are pretty good compromises.

jamadazi:
Oh wow you are spending so much time and effort on this :laughing: 

I like the 90% image bounds. It looks better with the smaller margins. So yes, do that.

For the actual SVG, either (1) or (4). I don't think it should deviate from typical rust formatting. Putting the parameter name on another line is weird.

I still prefer my original SVG (1), because removing that space (in (4)) makes it feel "squished", which IMO is uglier. But if you really want to remove the space, sure, (4) is fine too.

Just pick either (4) or (1 + 90% bounds) and merge it. Let's not waste more time bikeshedding. I don't think this is *that* important.

cart:
I think we have different opinions about what matters. This is literally the first thing people see when they go to bevyengine.org. I'm happy to spend time optimizing this (although in this case generating those permutations wasn't particularly time intensive).

I also think 4 is the call.

PR #93, opened 2021-02-07
Highlight navbar items when active

VasanthakumarV:
Noticed that the active items on the navbar weren't getting highlighted (except `Features`), this PR fixes that.

The change,
- Concatenating a `/` to the `name` variable so that the `starting_with` comparison passes for valid items


cart:
Ooh good catch. Much appreciated!

PR #94, opened 2021-02-12
Make this post more clearly historical for folks coming in from google

rparrett:
Inspired be a recent comment in discord

> are pages like this up-to-date with the latest releases? https://bevyengine.org/news/introducing-bevy/

But I have personally, in my early days with bevy, done something like

- Remember "hm, bevy can do ___ right? I think I read that somewhere"
- Googled "bevy ___"
- Stumbled on this post, and not immediately realize it's talking about a very old version of bevy.

Not a big issue obviously, opening this in case there's agreement. (And I was curious about zola)


cart:
This is a good call!

I just verified that the link wasn't broken by this change, so I think this is good to go.

PR #104, opened 2021-03-03
Community: feature awesome-bevy

jamadazi:
Add an "awesome-bevy" card to the "community" page.

I didn't know what logo to use, so I reused the github logo. I think it is a little ugly to have this duplication, so I hope we can come up with something else. Another suggestion: perhaps use the Bevy logo?

This should address #2 .

### DIFF
--- a/templates/community.html
+++ b/templates/community.html
@@ -82,5 +82,21 @@
             </div>
         </div>
     </a>
+    <a class="card" href="https://github.com/bevyengine/awesome-bevy">
+        <div class="card-image">
+            <img src="/assets/github.svg" class="centered-card-image community-icon" alt="Github logo"/>
+        </div>
+        <div class="card-text">
+            <div class="card-title">
+                Awesome Bevy
+            </div>
+            <div class="card-subtitle">
+                github.com/bevyengine/awesome-bevy
+            </div>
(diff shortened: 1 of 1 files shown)


cart:
This is a good call. You're right about the logo reuse, we should sort that out first. It would be nice if awesome-bevy had its own logo :smile: 

I guess we should probably just use the bevy logo for now (even though that is also a little confusing given that it uses the github logo in Learn, which i dont think we should change).

jamadazi:
I changed it to the Bevy logo.

jamadazi:
BTW, I think we should add a new card in "Learn" to link to the official Bevy examples. Then, it might make sense to use the Github logo for that, and the Bevy logo for awesome-bevy. That would resolve the inconsistency with the awesome-bevy card on this page.

Having a link to the official examples on the Learn page, linking to the github tag for the latest release, should also hopefully direct more people to their expected version of the examples (release rather than git main).

I will make a new PR for this.

cart:
> Having a link to the official examples on the Learn page, linking to the github tag for the latest release, should also hopefully direct more people to their expected version of the examples (release rather than git main).

I dig it!


> Discord message

Yup im on board :)

PR #110, opened 2021-03-16
Point people at the latest release in quick start section

rparrett:
Hopefully preventing one source of master/release examples confusion.

It wasn't entirely obvious to me what to do with the links/clone url in this section, but it didn't seem like a great situation to point a brand new user directly at the "main" examples, but then tell them to use 0.4.


rparrett:
Oops, this seems to be a duplicate of #109

PR #111, opened 2021-03-28
fix broken docs.rs links

B4rc1:
When clicking on any documentation that is not `std` in the book, you get `The requested version does not exist.`
This is because docs.rs changed their url scheme from ```https://docs.rs/std/<crate>``` to  ```https://docs.rs/latest/<crate>``` in order to obtain the docs for the newest version.

### DIFF
--- a/templates/shortcodes/rust_mod.html
+++ b/templates/shortcodes/rust_mod.html
@@ -4,7 +4,7 @@
 {% endif %}
 
 {% if not version %}
-{% set version = "std" %}
+{% set version = "latest" %}
 {% endif %}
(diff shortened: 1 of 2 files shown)


cart:
Bor r+

B4rc1:
excuse me but what do you mean?

cart:
Haha sorry about that. Bors is a "pull request" bot that we use in the main bevy repo. That command merges prs there. We don't have it set up on the bevy-website repo yet so nothing happened. I'll just need to merge this the old fashioned way :)

PR #113, opened 2021-04-05
news: release 0.5.0

cart:
This is a first draft of the 0.5.0 release blog post. Feel free to leave comments / suggest changes / help me fill in gaps I missed :smile:

If something is missing, it _probably_ wasn't intentional, theres just a lot to balance.

I'm posting this in draft mode because we're still missing the changelog and "whats next for bevy" section


scottmcm:
Convenience [RENDERED](https://github.com/cart/bevy-website/blob/release-0.5.0/content/news/2021-04-04-bevy-0.5/index.md) link.

cart:
(I recommend viewing this with a local zola server to get proper formatting)

Just download zola and run `zola serve`

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+
+In **Bevy 0.5**, I switched to using `ComponentId` instead of `ArchetypeComponentId`. This _is_ more constraining. `maybe_conflicts_system` will now always fail, but it will do it consistently at startup.
+
+Naively, it would also _disallow_ `filter_system`, which would be a significant downgrade in usability. Bevy has a number of internal systems that rely on disjoint queries and I expect it to be a common pattern in userspace. To resolve this, I added a new internal `FilteredAccess<T>` type, which wraps `Access<T>` and adds with/without filters. If two `FilteredAccess` have with/without values that prove they are disjoint, they will no longer conflict.
+
+This means `filter_system` is still perfectly valid in **Bevy 0.5**. I consider this a "best of both worlds" situation. We get most of the benefits of the old implementation, but with consistent and predictable rules enforced at app startup. 
+
+### EntityRef / EntityMut
+
+World entity operations in **Bevy O.4** require that the user passes in an `entity` id to each operation:
```
scottmcm, line 381: typo:
```suggestion
World entity operations in **Bevy 0.4** require that the user passes in an `entity` id to each operation:
```

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+})
+```
+
+This temporarily removes the `A` resource from `World`, provides mutable pointers to both, and re-adds A to World when finished. Thanks to the move to ComponentIds/sparse sets, this is a cheap operation.
+
+If multiple resources are required, scopes can be nested. We could also consider adding a "resource tuple" to the api if this pattern becomes common and the boilerplate gets nasty.
+
+### Query Conflicts Use ComponentId Instead of ArchetypeComponentId
+
+For safety reasons, systems cannot contain queries that conflict with each other without wrapping them in a QuerySet. In **Bevy 0.4**, we used ArchetypeComponentIds to determine conflicts. This was nice because it could take into account filters:
```
alice-i-cecile, line 356:
```suggestion
For safety reasons, systems cannot contain queries that conflict with each other without wrapping them in a `QuerySet`. In **Bevy 0.4**, we used `ArchetypeComponentIds` to determine conflicts. This was nice because it could take into account filters:
```

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+    )
+    .add_system(movement.system()
+        .label(PhysicsSystem::Movement)
+        .label(Physics)
+        .after(PhysicsSystem::UpdateVelocity)
+    )
+    .add_system(runs_after_physics.system().after(Physics));
+```
+
+Bevy plugin authors should export labels like this in their public apis to enable their users to insert systems before/after logic provided by the plugin.
```
alice-i-cecile, line 538:
```suggestion
Bevy plugin authors should export labels like this in their public APIs to enable their users to insert systems before/after logic provided by the plugin.
```

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+    .add_system_set(SystemSet::new()
+        .with_run_criteria(FixedTimestep::step(2.0))
+        .with_system(foo.system())
+        .with_system(bar.system())
+    )
+```
+
+### Improved Run Criteria
+
+Run Criteria are now decoupled from systems and will be re-used when possible. For example, the FixedTimestep criteria in the example above will only be run once per stage run. The executor will re-use the criteria's result for both the `foo` and `bar` system.
```
alice-i-cecile, line 573: Inconsistent capitalization of Run Criteria.

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+    for e in query.iter() {
+    }
+}
+```
+
+Global change detection was already a feature that set Bevy apart from other ECS frameworks, but now it is completely "fool proof". It works as expected regardless of system ordering, stage membership, or system run criteria.
+
+The old behavior was "systems detect changes that ocurred in systems that ran before them this frame". This was because we used a `bool` to track when each component/resource is added/modified. This flag was cleared for each component at the end of the frame. As a result, users had to be very careful about order of operations, and using features like "system run criteria" could result in dropped changes if systems didn't run on a given update.
+
+We now use a clever "world tick" design that allows systems to detect changes that happened at _any_ point in time since their last run.
```
alice-i-cecile, line 695: Might be fun to mention *just* how long we can wait between runs to impress the magnitude of their reliability.

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+    * Direct integration with the new scheduler 
+    * Improved "state lifecycle" events
+
+Read on for the details!
+
+## ECS Core Rewrite
+
+<div class="release-feature-authors">authors: @cart</div>
+
+Up until this point, Bevy used a heavily forked version of [hecs](https://github.com/Ralith/hecs) for our ECS core. Since Bevy's first release, we've learned a lot about Bevy's ECS needs. We've also collaborated with other ECS project leaders, such as [Sander Mertins](https://github.com/SanderMertens) (lead [flecs](https://github.com/SanderMertens/flecs) developer) and [Gijs-Jan Roelofs](https://github.com/gjroelofs) (Xenonauts ECS framework developer). As an "ECS community", we've started to zero in on what the future of ECS could be.
```
safasofuoglu, line 97: s/Mertins/Mertens

bjorn3:
2d_text.gif is rather large (3.79MB). Can you use an actual video format instead?

jamadazi:
@cart I have some pretty major criticisms about the current state of the blog post ([discord discussion](https://discord.com/channels/691052431525675048/692572690833473578/828586164122419200) with other people, from that message onwards).

The Core ECS Rewrite from https://github.com/bevyengine/bevy/pull/1525 is described in excessively verbose and unnecessary technical detail. It is currently somewhere between a third and half of all the content in the blog post. I agree that Bevy ECS is the most unique distinguishing feature of the Bevy project and should probably get the most attention, I just don't agree about this level of technical detail and rationale being appropriate for a release announcement post.

It also reads in a very different tone from the rest of it: everything else is covered briefly, with the user-facing aspects/changes showcased (which is a good style for a release announcement), while the stuff about the Core ECS Rewrite reads more like a technical post for ECS developers, full of jargon and implemetation details, with detailed rationale (much of the content feels similar to what was written/posted in the PRs themselves, for Bevy ECS devs). I don't think this is appropriate for a release announcement. I suspect a big part of the audience (who are not interested in ECS technicalities) would get bored or overwhelmed quickly, part way through that giant section, and lose their attention span for everything else.

It doesn't help that most other new features are covered *after* that. I think it would be a good idea to rearrange the content, so that as many of the cool new things in the release as possible get visibility first, while readers have the best attention span, and save the Core ECS Rewrite for last. It is the longest section that is the hardest to read. You could split away the benchmarks graphs to showcase the perf improvements early near the beginning of the blog post, while saving the detailed explanation for the end.

Also, much of the Core ECS content could really be slimmed down:
  - The Archetype Graph: this is really an internal implementation detail, that most bevy users will never be faced with (other than the awesome performance it results in :wink: ). Currently, it alone, has a detailed section that is larger than many other sections, including everything in the blog post about PBR. It feels unnecessary and out of place. I don't think it needs more than a brief mention.
  - Ditto for "Merged Resources into World"
  - and "Query Conflicts Use ComponentId Instead of ArchetypeComponentId" ... do we really need a detailed description of all the technicalities of how Bevy detects conflicting queries, in a Release Announcement?
  - much of the bullet-point content under "Archetypes"
  
You could replace all of that with a link to the PR, for anyone interested in the details.

jamadazi:
Also, with the "Orthographic Camera"  stuff, you forgot a very important aspect: it is now possible to use orthographic projection for 3D!

This enables people to make things like CAD applications, or games that use a 3D orthographic view.

This is probably more important than the fact that you can select different window scaling modes/behavior.

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+
+This example serves as a quick introduction to building 3D games in Bevy. It shows how to spawn scenes, respond to input, implement game logic, and handle state transitions. Pick up as many cakes as you can!
+
+![alien_cake_addict](alien_cake_addict.png)
+
+## Timer Improvements
+
+<div class="release-feature-authors">authors: @kokounet</div>
+
+Timers now internally use `Duration`s instead of using `f32` representations of seconds. This both increases precision and makes the api a bit nicer to look at.
```
stefee, line 1055: what is `Duration`? is this part of std or a bevy type? a link here to the docs would be nice!

stefee: I see it is part of std now - maybe just a small adjustment here for clarity

```suggestion
Timers now internally use `std::time::Duration`s instead of using `f32` representations of seconds. This both increases precision and makes the api a bit nicer to look at.
```

bjorn3: Maybe it could link to the official docs? https://doc.rust-lang.org/std/time/struct.Duration.html

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1254 @@
...
+
+![2d_text](2d_text.gif)
+
+## World To Screen Coordinate Conversions
+
+<div class="release-feature-authors">authors: @aevyrie</div>
+
+It is now possible to convert world coordinates to a given camera's screen coordinates using the new `Camera::world_to_screen()` function. Here is an example of this feature being used to position a UI element on top of a moving 3d object. 
+
+<video controls loop><source src="world_to_screen.mp4" type="video/mp4"/></video>
```
aevyrie, line 899: I made a small (63KB), smooth, looping video for this.

https://user-images.githubusercontent.com/2632925/113595672-2d517400-95ee-11eb-95d9-238aa0b624d2.mp4

mmahandev:
I may have missed it, but can you add what are the next steps and future plans are for Bevy?

content/news/2021-04-04-bevy-0.5/index.md
cart: I'm planning on doing a pass that will make all of the "types" link to docs (like we've done in past blog posts). I'll do that here.

cart:
@bjorn3 
> 2d_text.gif is rather large (3.79MB). Can you use an actual video format instead?

Good call. I got that from a PR / didn't record it myself so ill need to re-encode or re-record, but thats not too hard.

@jamadazi 
Yeah those are good points. I'll slim some things down and rethink the order a bit. Thanks!

@mmahandev
> I may have missed it, but can you add what are the next steps and future plans are for Bevy?
Yup those are missing. I called out that I'll be following up with those in the PR description.

content/news/2021-04-04-bevy-0.5/index.md
cart: Hmm i think the implied "it just works" is easier to convey and more approachable than something like "It is reliable up to a tick difference between last change and the last system run of 3 * u32::MAX/4 ticks, where each system run (and each frame) bumps the world tick by 1". While "true" and impressive, I think it muddles the message rather than driving it home.

cart:
I just cut a good portion of the ECS rewrite content, with a general focus on cutting down on implementation details that make people's eye's glaze over.

I also moved a good chunk of content to the end, which lets us "move on" to other things faster.

I did leave a bit of the "blog post-ey" introduction to storage types as I think it helps set the stage / tells a solid story that justifies the work being done. People have responded well to this style in the past so I'm comfortable rolling with it here.

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1215 @@
...
+    * System Labels
+    * System Sets
+    * Improved system "run criteria"
+    * Increased system parallelism
+* **"Reliable" change detection:**
+    * Systems will now always detect component changes, even across frames
+* **A rewrite of the State system:** 
+    * A much more natural "stack-based state machine" model
+    * Direct integration with the new scheduler 
+    * Improved "state lifecycle" events
```
bjorn3, line 89: Can these be links to the relevant sections of the post?

cart:
Added a small "whats next" section at the end. Feel free to suggest items (but only if they already have people committed to working on them / plans have already been made)

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -1119,6 +1119,19 @@ Naively, it would also _disallow_ `filter_system`, which would be a significant
...
 This means `filter_system` is still perfectly valid in **Bevy 0.5**. I consider this a "best of both worlds" situation. We get most of the benefits of the old implementation, but with consistent and predictable rules enforced at app startup. 
 
+## What's Next For Bevy?
+
+We still have a long road ahead of us, but the Bevy developer community is growing at a rapid pace and we already have big plans for the future. Expect to see progress in the following areas soon:
+
+* "Pipelined" rendering and other renderer optimizations
+* Bevy UI redesign
+* Animation: component animation and 3d skeletal animation
+* ECS: relationships/indexing, async systems
```
alice-i-cecile, line 1129: archetype invariants!

alice-i-cecile: Also Ratys's stageless work

DJMcNab: I believe in no more `.system()`. Maybe let's not promise it though

jamadazi:
@cart is there supposed to be a flight helmet picture that isn't showing? (near the beginning of the post, after the pbr balls)

The text suggests it, but I can't see one. (in the RENDERED link above)

cart:
@jamadazi it won't show videos in the preview. You'll need to use zola to render the html. It is a webm encoding, which doesn't work in safari (yet). I'm planning on re-encoding before release.

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1471 @@
...
+
+Two ECS storage paradigms have gained a lot of traction over the years:
+
+* **Archetypal ECS**:
+    * Stores components in "tables" with static schemas. Each "column" stores components of a given type. Each "row" is an entity.
+    * Each "archetype" has its own table. Adding/removing an entity's component changes the archetype.
+    * Enables super-fast Query iteration due to its cache-friendly data layout
+    * Comes at the cost of more expensive add/remove operations for an Entity's components, because all components need to be copied to the new archetype's "table"
+    * Parallelism-friendly: entities only exist in one archetype at a time so systems that access the same components but in different archetypes can run in parallel 
+    * Frameworks: Old Bevy ECS, hecs, legion, flecs, Unit DOTS
```
MinerSebas, line 111:
```suggestion
    * Frameworks: Old Bevy ECS, hecs, legion, flecs, Unity DOTS
```

content/news/2021-04-04-bevy-0.5/index.md
```diff
@@ -0,0 +1,1471 @@
...
+The old executor would have implicitly forced `increment_counter` to run first because it conflicts with `print_every_other_time` and it was inserted first. But the new executor requires you to be explicit here (which we believe is a good thing).
+
+To help detect this class of error, we built an opt-in tool that detects these ambiguities and logs them:
+
+```
+Execution order ambiguities detected, you might want to add an explicit dependency relation between some of these systems:
+ * Parallel systems:
+ -- "&app::increment_counter" and "&app::print_every_other_time"
+    conflicts: ["usize"]
+```
```
MinerSebas, line 404: This should mention how the ambiguity detection is activated.

cart: good call

RoyAwesome:
An idea based on what @jamadazi said... perhaps you can salvage the deep technical details and write a "deep dive" blog post using them.  I feel like that info is very valuable, but I agree that it's extremely verbose for a release announcement.

cart:
Yeah good call. We could definitely do a blog post on the details.

PR #120, opened 2021-04-07
minor fix to timer resource example so it compiles

fractaloop:
Tupled type GreetTimer needs `.0` to access `Timer.tick`


nextuponstream:
Also had the same problem, I fixed it by using a `Timer` instead of the `GreetTimer(Timer)` struct.

cart:
Nice catch!

PR #121, opened 2021-04-07
AppBuilder.add_resource was depreciated - fixed docs

PatrickDanker:
AppBuilder.add_resource() was depreciated in 0.5 - fixed the docs to work


cart:
Theres no need to use a startup system here. `add_resource` was just renamed to `insert_resource`. Fix that and i think we're good to go!

PatrickDanker:
Docs are up to date now - closing request

PR #122, opened 2021-04-07
Add gltf loading change to migration guide

iwikal:
I ran into this when upgrading and it wasn't obvious what was wrong.


cart:
Ah nice catch. Thanks!

Issue #131, opened 2021-04-13
Dynamic linking instructions broken for Windows


Alainx277:
Following the setup guide for fast builds on windows leads to errors when using dynamic linking. There seem to be two issues, which both have the solution of turning off "share-generics" for windows builds  (bevyengine/bevy#1126, bevyengine/bevy#1110).

Maybe it would be good to offer a solution in the setup guide, or perhaps even temporarily turn off shared generics in the cargo config template.

paul-hansen:
This should be working on Windows now afaik. I've been using it on Windows for awhile now.

The trick is that you _need_ these lines on Windows or it doesn't work:
```toml
# Enable a small amount of optimization in debug mode
[profile.dev]
opt-level = 1

# Enable high optimizations for dependencies (incl. Bevy), but not for our code:
[profile.dev.package."*"]
opt-level = 3
```

We should probably remove `Note that right now, this doesn't work on Windows` from the setup instructions from the book, and emphasize that the opt-level lines are needed for Windows.

See for more info see comments: https://github.com/bevyengine/bevy/issues/1110#issuecomment-1322645492 and https://github.com/bevyengine/bevy/issues/2921#issuecomment-1353395072

mnmaita:
Closing as per https://github.com/bevyengine/bevy-website/pull/628! Please re-open if you find the issue persists.

PR #137, opened 2021-04-23
Updated 'Enable Fast Compiles' section

lukors:
I found this section a bit confusing and ran into some issues when I tried following it. So my goal with this PR is to make it clearer and less error prone.

- Split it up into two sections to clearly show that you can either do this one simple step, or this one simple step AND these other more involved steps.
- Turned the 3 more involved steps into a numbered sequence to show clearly that you can't do just one or two, you have to do all three. Which technically you don't, but if you skip a step you would need to edit the config file in the end, and if you skip the config file (step 3) you won't get much benefit. So they make sense as a sequence.
- Added clang to the `sudo apt-get install` for Ubuntu as it was required for me on Pop!_OS 20.10 (based on Ubuntu 20.10).
- Cleared up the final step, its headline was "Generic Sharing", but that's not a step, the actual step is to configure cargo, which to my understanding both activates Generic Sharing as well as LLD.


content/learn/book/getting-started/setup/_index.md
```diff
@@ -85,34 +85,37 @@ bevy = "0.5" # make sure this is the latest version
 ```
 ### Enable Fast Compiles (Optional)
 
-Bevy can be built just fine using default configuration on stable Rust. However for maximally fast iterative compiles, we recommend the following configuration:
+Bevy can be built just fine using default configuration on stable Rust. However there are a few things you can do to significantly improve iterative compile times:
 
-* **Enable Bevy's Dynamic Linking Feature**: This is the biggest iterative compilation time win and requires no special setup. When you have `bevy` as a dependency you can run your app with dynamic linking like this:
+* **Bevy's Dynamic Linking Feature**: This is the biggest iterative compilation time win and requires no special setup. When you have `bevy` as a dependency you can run your app with dynamic linking like this:
```
MinerSebas, line 90: This should probably also mention that this Feature doesn't always work on Windows.
At least it doesn't when you also use the `Zshare-generics` config: https://github.com/bevyengine/bevy/issues/1126#issuecomment-750000945

cart: Lets make sure we close #131 if you make this change.

cart: I think it probably makes sense to add this to the "configure cargo" section instead of to the "dynamic linking" section, just to keep the dynamic linking section simple.

cart:
This looks good to me! Once we add the "windows + dynamic linking + share generics" stipulation i think its good to go.

lukors:
I've looked into this a bit closer and tested it on my Windows machine in an attempt to come up with an as simple solution as possible for people who follow the guide.

I made a pull request to turn off the "share generics" option in the config.toml for Windows users, since it doesn't seem to work (https://github.com/bevyengine/bevy/pull/2016). This removes the need to mention this in the actual guide since that file should then work for every OS.

"Share generics" doesn't work on Windows, so the only gains to get for Windows users are the LLD linker, bevy/dynamic, and any optimizations nightly brings. bevy/dynamic doesn't work without the config.toml file containing the setting that turns off "share generics", and that option requires nightly to be used. So my conclusion is that as a Windows user you might as well follow the entire guide.

So the only changes I've made to the guide is to add a note for windows users at the top saying they need to follow the numbered steps for bevy/dynamic to work, and I've noted where some something is different for Windows users.

content/learn/book/getting-started/setup/_index.md
lukors: Thanks! I believe my recent changes resolves this.

lukors:
Thanks cart and MinerSebas! I believe this is ready for another look!

content/learn/book/getting-started/setup/_index.md
```diff
@@ -85,34 +85,39 @@ bevy = "0.5" # make sure this is the latest version
 ```
 ### Enable Fast Compiles (Optional)
 
-Bevy can be built just fine using default configuration on stable Rust. However for maximally fast iterative compiles, we recommend the following configuration:
+Bevy can be built just fine using default configuration on stable Rust. However there are a few things you can do to significantly improve iterative compile times:
 
-* **Enable Bevy's Dynamic Linking Feature**: This is the biggest iterative compilation time win and requires no special setup. When you have `bevy` as a dependency you can run your app with dynamic linking like this:
+If you're using Windows, Bevy's dynamic linking feature won't work unless you've followed the three extra steps below first.
+
+* **Bevy's Dynamic Linking Feature**: This is the biggest iterative compilation time win and requires no special setup unless you use Windows. When you have `bevy` as a dependency you can run your app with dynamic linking like this:
```
cart, line 92: Does this actually require special setup on windows? Shouldn't it work with the default setup? If this is just a reference to the "share generics" issue, we should just cover that in the "fast compiles configuration section", either by removing it like you did in your pr or by documenting the weirdness / potential errors in the fast compiles section.

I'd prefer it if the "dynamic linking" section was "simple and asterisk free"

lukors: tl;dr: My understanding is that the problem is the share-generics issue. It seems like share-generics are on by default in debug compilation, so that's why it doesn't work on Windows without special setup.

---

I did some more testing to make sure if special setup is required, and in my tests I run into some variation of the errors linked in https://github.com/bevyengine/bevy-website/issues/131 when running with `--features bevy/dynamic` on Windows unless I run with..

**Option 1**
- `rustflags = ["-Zshare-generics=n"]`
- Nightly (required because of the -Z flag in rustflags)

**Option 2**
- `--release`

Like you write, `share-generics` *is* off by default, but only with `--release`, otherwise it's apparently on by default. This is where I got this information https://github.com/rust-lang/rust/issues/67276 and I've verified it on my machine by successfully doing `cargo +stable run --features bevy/dynamic --release` with no config.toml file. Incremental compiles with `--features bevy/dynamic --release` is slower for me than building in debug without bevy/dynamic.

So, following option 1, with nightly and the config file you've already followed 2 of the 3 "advanced" points, so you might as well do the remaining step of using LLD as well. If nothing else, just to keep the instructions simple.

I also want to keep these instructions as simple as possible, and it sucks to have the extra info for Windows users. But it seems to me like bevy/dynamic simply does not work on Windows unless you at least compile on nightly and with `rustflags = ["-Zshare-generics=n"]`.

cart: Thanks for the breakdown! I'll run a few tests on my machine just to get a feel for the various permutations / see if these behaviors are reproducible.

It does sound like we'll probably need to roll with the approach in this pr  :smile:

alice-i-cecile:
This should be rebased on top of #154 in the `new-book` branch. I've moved the fast compile instructions much later (to avoid tripping up beginners quite so aggressively), so you'll have to modify what I have there instead.

lukors:
@alice-i-cecile just so I don't do anything wrong, I should at this time rebase this PR's branch (https://github.com/lukors/bevy-website/tree/patch-1) onto your branch (https://github.com/alice-i-cecile/bevy-website/tree/book-setup)?

Including moving the changes in this PR to where they should be in the new book etc.

Won't that include all your changes into this PR? Or maybe that's fine?

alice-i-cecile:
> I should at this time rebase this PR's branch (https://github.com/lukors/bevy-website/tree/patch-1) onto your branch (https://github.com/alice-i-cecile/bevy-website/tree/book-setup)?

That's my understanding, yes.

> Won't that include all your changes into this PR? Or maybe that's fine?

Temporarily, yes. Those will get optimized away by Github once #154 is merged though, which will make this PR's commit history tidy again :) The other option of course is just to wait until #154 is merged, but that requires a bit more synchronous coordination.

lukors:
@alice-i-cecile I couldn't figure out how to preserve the history of my commits, so I just put the changes in a new commit on top of the branch and force pushed.

Also made sure formatting looked good after going through Zola and did some other minor improvements.

I hope this is ok and I did it right?

alice-i-cecile:
> @alice-i-cecile I couldn't figure out how to preserve the history of my commits, so I just put the changes in a new commit on top of the branch and force pushed.

Yep, that's totally fine here :) Thanks!

It looks like you need to change which branch you're making the PR to though; this is still targeting `master` (hence the giant diff).

lukors:
Great! :)

Ah ok, I think I did that now.

lukors:
A merge conflict appeared for some reason when #154 was merged. I fixed it, so this is ready on my end.

As a reminder for cart, this is what you wrote about it most recently, so you don't forget to run those tests:

> Thanks for the breakdown! I'll run a few tests on my machine just to get a feel for the various permutations / see if these behaviors are reproducible.
>
>It does sound like we'll probably need to roll with the approach in this pr 😄

If anyone else wants to test it that would be great as well! AFAIK I'm the only one who's checked thoroughly, so would be nice to make sure it's all good before it is merged.

SevenO2:
> If anyone else wants to test it that would be great as well! AFAIK I'm the only one who's checked thoroughly, so would be nice to make sure it's all good before it is merged.

lukors:
@SevenO2 Great news! Thanks a lot! :smile: 

Looks like some conflicts have popped up, I'll fix these soon, and then hopefully this will be ready to merge!

lukors:
I merged this on top of the new-book branch. I'm assuming that's what I was supposed to do, but please let me know if that's wrong. :)

I also want to add that I'm not 100% confident that I merged it correctly. I think I did, but I very rarely do these types of merges so I'm not sure.

alice-i-cecile:
> I merged this on top of the new-book branch. I'm assuming that's what I was supposed to do, but please let me know if that's wrong. :)

Yes please; that's the most useful spot for it right now. Hopefully we can merge in #176 soon, and then rebase this on that.

content/learn/book/development-practices/fast-compiles/_index.md
```diff
@@ -8,44 +8,47 @@ page_template = "book-section.html"
...
+      rustup component add llvm-tools-preview
+      ```
+
+   * **MacOS**: Modern LLD does not yet support MacOS, but we can use zld instead: `brew install michaeleisel/zld/zld`
+2. **Nightly Rust Compiler**: This gives access to the latest performance improvements and "unstable" optimizations
+
+   ```sh
+   # Install the nightly toolchain
+   rustup toolchain install nightly
+   # EITHER configure your current project to use nightly (run this mmand within the project)
```
djeedai, line 44: What's "mmand"? typo?

content/learn/book/development-practices/fast-compiles/_index.md
```diff
@@ -8,44 +8,47 @@ page_template = "book-section.html"
...
+   rustup override set nightly
+   # OR configure cargo to use nightly for all projects -- switch back with `rustup default stable`
+   rustup default nightly
+   ```
+
+   You can use `cargo +nightly ...` if you don't want to change the default to nightly.
```
djeedai, line 50:
```suggestion
   You can use `cargo +nightly ...` if you don't want to change the default to nightly but just want to use it once for the current command.
```

content/learn/book/development-practices/fast-compiles/_index.md
lukors: Yeah it's a typo, should be "command", fixed and thanks! :D

lukors:
Yeah that works for me, thanks! :D 

Should I close this PR now then? :)

Issue #135, opened 2021-04-20
"Donate" and "Get Started" buttons not visible on mobile


GabLotus:
Tested on:
- Edge for Android
- Firefox for Android

Would appreciate: 
- People confirming they are experiencing the issue on other platforms/browsers
- Someone fixing it :)

PR #138, opened 2021-04-24
Mobile hamburger menu

piedoom:
This also adds a _variables scss file. I placed breakpoints in that file, as the site was using a few specific ones and I think it is good to keep that to one source of truth. This is a JS free implementation. Solves #135

### DIFF
--- a/sass/_base.scss
+++ b/sass/_base.scss
@@ -1,3 +1,5 @@
+@import "_variables";
+
 html {
     overflow-y: scroll;
 }
@@ -9,7 +11,7 @@ html, body {
     font-family: 'Fira Sans', sans-serif;
     background-color: #232326;
 }
-@media screen and (max-width: 712px) {
+@media screen and (max-width: $width-medium) {
     html, body {
         font-size: 0.7rem;
     }
--- a/sass/_headerbar.scss
+++ b/sass/_headerbar.scss
@@ -1,3 +1,5 @@
+@import "_variables";
+
 .headerbar-container {
     width: 100%;
     background-color: $headerbar-color;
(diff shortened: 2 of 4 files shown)


cart:
This is awesome! Just two things before I think this is ready:

## Menubar Alignment

I'd prefer it if the menubar items were still "right aligned":

Old (right aligned):
image

New (left aligned):
image

## Consider a new icon for collapsed book sidebar

On mobile we now have two identical menubar icons for "collapsed menu items" and "collapsed book sidebar". Not sure what the best call is to disambiguate, but I think we should probably do something.

image

piedoom:
Thanks for reviewing. The nav being left aligned is an error when i cleaned up the commit, I have that fixed locally & will push something soon. I ended up removing that "pusher" element (with the auto margin) in favor of just making the nav use flexbox.

The book part is a little bit trickier, and I had a couple thoughts on how it could be done. 

My (unvalidated) hypothesis is that when users browse the book, they expect it to be a self-contained "resource" (like rustdoc) in which it's OK if the docs site has a different feel or navigation. The important action they might want to take is to get back to the main Bevy site. I think this means we can possibly just nix the full nav, and in place just have a link with "back to bevy's main site -->" (or something with better copy than that😅).  Basically rustdoc's pattern.

Alternatively, we could put the site nav items in the hamburger menu (likely at the bottom below the book nav). I think in any case I may push the menu icon to the left on all pages so it matches the book. 

I haven't considered these options in great detail, so if you have an alternative suggestion I'd love to hear it.

doup:
@alice-i-cecile this can be closed as it was fixed in #307

PR #148, opened 2021-05-08
skeptic setup

mockersf:
integration with skeptic to validate book!

Everything for the integration seems to work fine (Zola not bothered by skeptic annotations, skeptic loading templates). Two notes:
* I excluded the migration guides folder as it will contain invalid code
* skeptic want a language of `rust`, not `rs`

that said, it fails on my computer because:
```
error: linking with `cc` failed: exit code: 1
  |
  = note: "cc" "-m64" "-arch" "x86_64" "-L" [blablablabla]
  = note: ld: library not found for -lglslang.glsltospirv
          clang: error: linker command failed with exit code 1 (use -v to see invocation)

```
not sure how to move forward on that

PR #150, opened 2021-05-09
reuse code blocks from validated crate

mockersf:
sister PR to #148: the same thing, but with a crate that can be compiled, and code blocks included from that crate


alice-i-cecile:
@cart can we merge this or #148 so I can use it in #176 and its siblings?

cart:
Sure. I'll resolve conflicts

cart:
Heads up that we'll probably want to add support for configuring "file code block indents"
image

This could probably just be another `indent=4` parameter in the shortcode

cart:
Hmm i tried my hand at indents and so far no dice. No number of indents (either spaces or tabs) indents the code block as a whole (it just indents the interior code block). Also tried changing `markdown(inline=true)` to false.

```html
{% if not lang %}
    {% set lang = "" %}
{% endif %}

{% if not indent %}
    {% set indent = 0 %}
{% endif %}

{% set_global indent_str = "" %}
{% for i in range(end = indent) %}
    {% set_global indent_str = indent_str ~ " " %}
{% endfor %}

{% set code = load_data(path=path) %}

{% if anchor %}

    {% set code_lines = code | split(pat="\n") %}
    {% set code = "" %}
    {% set in_anchor = false %}

    {% for line in code_lines %}

        {% if line is ending_with("// ANCHOR_END: " ~ anchor) %}
            {% set_global in_anchor = false %}
        {% endif %}

        {% if in_anchor %}
            {% set_global code = indent_str ~ code ~ line ~ "
" %}
        {% endif %}

        {% if line is ending_with("// ANCHOR: " ~ anchor) %}
            {% set_global in_anchor = true %}
        {% endif %}

    {% endfor %}

{% endif %}

{% set data =  indent_str ~ "```" ~ lang ~ "
" ~ code ~ indent_str ~ "```" %}

{{data | markdown(inline=true) | safe}}
```

cart:
I do want to make this work first if possible. Without it we'll have ugly formatting in a number of places. I've gotta go back to renderer stuff for now. Lemme know if any of you can get this operational :)

mockersf:
#203 was chosen instead

Issue #153, opened 2021-05-24
Chapter folding for book


alice-i-cecile:
As I write many more chapters for the book, the sidebar is getting very long. I think being able to fold the chapters down is the best solution to improve the UX here.

PR #154, opened 2021-05-25
Book structure

alice-i-cecile:
I've created the skeleton of the new book following https://github.com/bevyengine/rfcs/pull/23, and added brief notes to each chapter on what I think should be covered and in what order.

Merging this in will allow us to start writing in a nice, distributed fashion. If you want to tackle a chapter, please mention it in #158 so we don't duplicate effort. I would particularly appreciate help with the platform-specific chapters, rendering and assets, as those are the areas I know least about.

There were two areas of content that were lost from the old book:

1. Contributing: moved to https://github.com/bevyengine/bevy/pull/2040
2. Migration guides: I strongly feel that this is a better fit to link at the top of each release blog post.

image
image
image
image


alice-i-cecile:
As you can see by the ridiculous screen shots, #153 is needed to make this expanded structure at all sensible to navigate 😆

content/learn/book/input/gamepad/_index.md
```diff
@@ -0,0 +1,13 @@
++++
+title = "Mouse"
```
cart, line 2: "Mouse" is already used in section 3. I'm assuming this was just a copy/paste error?

alice-i-cecile: Yep just a typo. There was a lot of tedium involved in the structure and all the files are called _index.md :(

cart: I haven't found a way yet, but id love to be able to flatten out this structure. I also like to be able to include a weight prefix in the name instead of defining it in the file ... that way its automatically sorted in IDEs / filesystems.

cart:
Solve the "mouse" thing and I think this is good to merge!

content/learn/book/input/gamepad/_index.md
alice-i-cecile: @mockersf any ideas? I would have loved this.

alice-i-cecile:
```suggestion
title = "Gamepad"
```

alice-i-cecile:
This is good to merge now @cart, likely along with #137 :)

PR #156, opened 2021-05-25
add Bevy awesome to website

mockersf:
`cd generate-assets && ./generate_assets.sh`

* clones awesome Bevy
* run a python script to transform existing `README.md` into toml files (this will be removed once awesome Bevy is in the new format)
* run a Rust script to transform those files into section and articles for Zola with valid frontmatter

With this version, everything is ordered alphabetically. Once we are ready to migrate awesome Bevy, we can add a custom order for some items if we want. Only information available from the readme are displayed, so no license, no picture, ...

It can be reached from the learn section, under "Awesome Bevy beta"

~~CI is not setup yet~~


mockersf:
Added CI:
* workflows for CI on PR and for deploy from master are now separated
* CI will trigger on PRs and only build
* Deploy will trigger:
  * on push to master (pr merged)
  * daily on a schedule (to catch updates from external sources like crates.io)
  * on demand from the UI/CLI/API: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow-using-the-rest-api (can be used to trigger automatically when a PR gets merged in https://github.com/bevyengine/awesome-bevy)

templates/awesome.html
```diff
@@ -0,0 +1,60 @@
...
+    <div class="card-list padded-content">
+        {% for post in section.pages %}
+        <div class="card" id="{{ post.title | slugify }}">
+            {% if post.extra.image %}
+            <div class="card-image card-image-default">
+                <img src="{{ get_url(path=post.extra.image) }}" />
+            </div>
+            {% endif %}
+            <div class="card-text">
+                <a class="card-title" href="{{ post.extra.link }}">
```
cart, line 20: I think clicking any part of the card should link to the `post.extra.link`. Can we wrap the whole card in an `<a>` element?

cart:
Great work! I love that you included migration automation.

I think this would benefit from a "table of contents sidebar". Maybe we can steal aspects of the Bevy Book impl?

Ultimately I'm thinking something along the lines of:

* Featured (default)
* Categories
  * Apps
    * Games
    * Tools
  * Plugins
     * 2D
     * 3D
     * etc
  * Learning
  * etc



> With this version, everything is ordered alphabetically.

Can we randomize the order (within each category) on each build for now? I definitely think in the near future we should build an "algorithm" to boost things (arbitrary curation, github stars, good licenses, etc). But in the short term randomizing feels fairer than alphabetizing.

cart:
I also think we should start considering alternative names. This will likely continue to evolve away from the "awesome-style" list format (which many people aren't familiar with in the first place).

Some ideas:

* Bevy Assets
* Bevy Asset Library
* Bevy Store
* Bevy Asset Store

I'm a _little_ hesitant to adopt "store" terminology as that implies "sales". We might get there eventually, but imo thats a separate step (and reserving "Store" for that step would help drive the point home). So I'm thinking "Bevy Asset Library" has my vote (with "Bevy Assets" as a close second)? Its a bit unfortunate that it overlaps with the `bevy_asset`, but I doubt that in practice this will cause confusion. "Asset Stores" are basically expected for general purpose engines these days, so not including "Asset" in the name would also cause more confusion than its worth.

cart:
I don't think we need to build a "featured" section in this PR, but I do think we should add a table of contents if we want this to be a replacement for the awesome-bevy readme. Right now finding "relevant" content involves a lot of scrolling. We should probably also make the card height a bit smaller.

mockersf:
* whole card is a link. Zola seems to have a bug here when the card description also contains a link, I'll check more
* added a toc
* order is now random in each section
* reduced card height

For the name, I'm not a fan of speaking of assets if we keep apps/games and learning in it

mockersf:
> Zola seems to have a bug here when the card description also contains a link, I'll check more

This is not a bug in Zola but a limitation of html: you can't have nested links. If we want the whole card to be clickable we'll need to forbid links in descriptions... (which I think we shouldn't)

cart:
>  This is not a bug in Zola but a limitation of html: you can't have nested links. If we want the whole card to be clickable we'll need to forbid links in descriptions... (which I think we shouldn't)

Hmmm. Ultimately I think we'll want to give each item its own "page". This would allow us to include more information:
* A long form markdown description of the thing (this could be readme content for crates)
* Github star counts
* Screenshot carousels and/or videos
* More label / tag information
* Crate version info

In that context, maybe its ok for descriptions to not include markup. They could be more like "github repo descriptions", which are short markup-free descriptions of a thing:
image

That frees us up to make the whole card a link.

cart:
> For the name, I'm not a fan of speaking of assets if we keep apps/games and learning in it

This is a fair point, but its worth pointing out that unity does put learning under the "asset store umbrella" (at least kind of  ... its just a link to their Unity Learn platform)

image

Defold also puts everything under the Assets umbrella:
image


Maybe we could adjust our categories to make this clearer?

* Apps
  * Games
  * Tools
* Learning
  * Tutorials
  * Migration Guides
* Assets
  * Plugins
  * Crates
  * 2D
  * 3D
  * Templates
  * Etc

I still think I prefer "Assets" over "Awesome Bevy" or just "Awesome" (and I want to avoid "Store" for the reasons mentioned above). "Library" is more generic, but also has more of a "reading" implication than a "tooling and content" implication. If you have other ideas, I'm all ears, but I think "awesome" needs to go.

mockersf:
To go further I think we would need to start changing the awesome-bevy repo (add a few weights for some things we want order, change categories, ...).

Do you want to do that now? Maybe start a new repo bevy assets where I could set it up?

cart:
Agreed!

I think we ultimately want to just rename the `awesome-bevy` repo to something like `bevy-asset`. This has the benefit of redirecting old awesome-bevy links and preserving github stars.
(although this could easily be confused with the `bevy_asset` crate, so I'm open to suggestions there). 

Maybe we could start by making a new branch and feeding off of that branch here?

cart:
Made a few tweaks to unify style / enable nav bar collapse/toggling. I also added `Assets` to the header bar.
(these are just proposals ... feel free to push back on them or make changes)

cart:
Also do you have thoughts on doing 3 cards per row by reducing the card width to 30%?

I think it improves information density /  makes scrolling through everything a bit less intense:
image

mockersf:
> Maybe we could start by making a new branch and feeding off of that branch here?

Ok for me 👍 

> Also do you have thoughts on doing 3 cards per row by reducing the card width to 30%?

I was worried it would leave less place for each card, but it should be OK with the plan to have dedicated pages later, so 👍

cart:
Cool cool. I think ultimately we'll adopt a more "vertical" card layout, which should help. I like the general direction of this Figma design `@giusdp` submitted on discord (although I would probably tweak things, such as removing the "view" button and adding the "description" to it):

image

https://www.figma.com/proto/BAWOVsxMB8JrQnfpGaIAIW/Bevy-Store?node-id=2%3A1&scaling=min-zoom&page-id=0%3A1
https://discord.com/channels/691052431525675048/695741366520512563/839584693832646738

cart:
I added a "hack" that enables us to pull categories to the top based on their title. I think Learning should probably come first, followed by Plugins and Crates.

mockersf:
for ordering, I would prefer to be able to specify a "weight" in the source repo in the section toml, and then pass that weight to the front matter for Zola ordering, defaulting to the computed weight based on alpha for section, random for content

cart:
Ooh yeah thats much better.

cart:
Added a new branch here: https://github.com/bevyengine/awesome-bevy/tree/bevy-asset

mockersf:
started a pr of awesome-bevy on that branch: https://github.com/bevyengine/awesome-bevy/pull/102
and updated this PR to work off that one

also added order for some things, not as pleased on how to do it as I was in my head before coding it but it works

sass/_assets.scss
```diff
@@ -0,0 +1,42 @@
...
+        max-width: 100%;
+    }
+
+    a {
+        text-decoration: none;
+    }
+
+    .book-content {
+        h3 {
+            font-size: 2.2rem;
```
mockersf, line 37: why did you add this one? it makes the `h3` very close to `h1` (no difference to my eye, `h1` is at 2.4rem)

PhaestusFox:
can I suggest resources as the name instead of assets? personally, when I hear assets, I'm imagining things directly used by the game being created (plugins, systems, textures, ext) and not necessarily things that may indirectly help the developer (tools, tutorials, guides). altho all these things would be an "asset" to have I personally don't feel that they should be called assets directly, that should be saved for things directly relating to the bevy_asset crate/the game dev use, they are all resources at our disposal tho. just my 2 cents on the name question

sass/_assets.scss
cart: I'm open to downsizing a little bit, but imo the "default" set for the book is "too small" for sub-categories in this context. I'm definitely open to downsizing a little bit because I agree that its probably too close to the parent size.

templates/assets.html
```diff
@@ -0,0 +1,77 @@
...
+                <a href="{{ post.extra.link }}">
+                    {% if post.extra.image %}
+                    <div class="card-image card-image-default">
+                        <img src="{{ get_url(path=post.extra.image) }}" />
+                    </div>
+                    {% endif %}
+                    <div class="card-text">
+                        <div class="card-title">{{ post.title }}</div>
+                        <div class="card-description">
+                            {{ post.description | markdown(inline=true) | safe }}
```
cart, line 65: I'm not sure allowing arbitrary markdown here is a good idea. I think it gives users too much power over formatting / visibility (ex: we shouldn't let them slap a `# LOOK AT ME` h1 in their description). As it stands, we've only removed links "by convention". Theres nothing stopping users from adding links to their descriptions, other than the pull request process.

sass/_assets.scss
mockersf: reduced the font a little, and added a ">" before headers. not sure about it though. What do you think?
Screenshot 2021-05-27 at 22 47 41

templates/assets.html
mockersf: I would like to keep markdown open, for at least code block, bold and italic, but that looks somewhat painful to validate automatically...

There's a manual review for PR now and the volume is ok, but I see how that won't work at scale and it wouldn't be nice to change the rule somewhere in the future to remove that.


I'l guess I'll remove it until I can find a way to validate that I like

sass/_assets.scss
cart: It definitely helps distinguish sub categories from categories, but I think it draws the eye a little bit too much and messes with alignment a bit too much. I think I'd prefer just bumping down the size another `0.1rem` or so.

cart:
> can I suggest resources as the name instead of assets? personally, when I hear assets, I'm imagining things directly used by the game being created (plugins, systems, textures, ext) and not necessarily things that may indirectly help the developer (tools, tutorials, guides). altho all these things would be an "asset" to have I personally don't feel that they should be called assets directly, that should be saved for things directly relating to the bevy_asset crate/the game dev use, they are all resources at our disposal tho. just my 2 cents on the name question

I do think Resources better encapsulates all "things", but I'm still pretty set on using Assets in the name, at the very least for crates, plugins, templates, and other things "usable from games". Its hard to ignore the "asset culture" around game engines. Using any other name fails to create the associations required to tap in to that culture (on first impression). The fact that we also put learning resources and full games/apps under that category is largely for convenience so we don't need to create multiple systems. Strategically, if we need to choose between dropping "non-asset" content entirely, and including Assets in the name, I'd rather drop non-asset content entirely.

I think ultimately it might make sense to build another feature to organize community generated learning material, but in the short term I'd prefer to just include it under the Assets umbrella.

templates/assets.html
cart: Things like codeblocks / bold / italic dont feel like much of a loss in the context of "short and sweet tagline describing a project". People shouldn't be including code examples in something like that. If we enable bold in a "short tagline" context, it incentivizes misuse to increase visibility (ex: just make the whole tagline bold to draw the user's attention away from non-bolded descriptions). 

I'd rather just keep this plain text, then enable markdown in the "long description" we show when someone clicks on an entry.

cart: I also think we should heavily limit character count for the "short description" / "tagline" / whatever we decide to call it. Stuff like this is too noisy / takes up too much space:

image

mockersf: ok 👍 

For long description: how do you feel about using `line-clamp` (https://caniuse.com/css-line-clamp)? It's pretty much the ellipsis text overflow for multiline text
Screenshot 2021-05-28 at 00 45 34

But it's not yet in the official css spec so it's supported by most browser with the `-webkit-` prefix. Most being all but IE

cart: Alternatively we could do it in Tera. We currently do this for the news section:

```rust
<div class="card-description">
  {{ page.summary | striptags | truncate(length=180) | safe}}
</div>
```

I think our "short description" should also have a max character limit enforced by the "bevy-asset repo toml validator". But theres no harm in truncating in the card, just to be doubly sure.

cart: (truncate does add ellipsis) https://tera.netlify.app/docs/#truncate

mockersf: Oh didn't even think to use the template engine for that 🤦 
But the limit in char count is not a guarantee that it wi display in a given multiline box due to line breaks

> I think our "short description" should also have a max character limit enforced by the "bevy-asset repo toml validator".

Then what happens fro existing assets that won't meet the new rules? we edit them ourselves? we remove them and ask for contributors to edit them?

cart: >  But the limit in char count is not a guarantee that it will display in a given multiline box due to line breaks

Rather than enforcing hard limits using line-clamp, I think we should probably just strictly define the size of the container div (for card size uniformity), disable overflow on that div, and then set character limit to something that prevents overflow in the first place. That ensures that worst case scenario, we allowed too many characters and they get cut off. Then we just need to either adjust the bounds to make those characters visible, or shrink the character limit.

Short term, I'm cool with just truncating in Zola and calling it a day. We can enforce card size uniformity when we rework the card layout.

> Then what happens fro existing assets that won't meet the new rules? we edit them ourselves? we remove them and ask for contributors to edit them?

We should edit them ourselves (and maybe leave a comment on the original pr to give the author a heads up in case they want to change something). Happy to do that work if nobody else wants to.

mockersf: Cutting at 100 chars seems to provide the best results 👍 

I added the task to update existing assets to the tracking issue

cart:
Alrighty one more thing before I think we should start focusing on migrating (ex: merge this, migrate awesome-bevy, update awesome-bevy readme to account for changes).

Currently the headerbar reads "assets.html". It should be "Assets":

image

PR #167, opened 2021-06-02
Mention dynamic linking not working on Windows

TheRawMeatball:


alice-i-cecile:
Meatball, give this a better title :p I agree with the content; this is an important stop gap.

PR #169, opened 2021-06-03
Recommend using the Rustup toolchain-file for nightly rust

ickk:
Using a ```rust-toolchain``` file is much simpler to set up, compared to manual rustup overrides. Also, using the ```rust-toolchain``` file allows other people that download a repository to get going immediately without any manual configuration of their local toolchain.

Finally, in my opinion the toolchain version should be checked into source control anyway, since a codebase written with the nightly compiler in mind may not compile on stable rust. This makes the toolchain basically a build-dependency, and we obviously value checking in other build-dependencies via ```Cargo.toml```.

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -108,15 +108,13 @@ Bevy can be built just fine using default configuration on stable Rust. However 
         ```
     * **MacOS**: Modern LLD does not yet support MacOS, but we can use zld instead: `brew install michaeleisel/zld/zld`
 * **Nightly Rust Compiler**: This gives access to the latest performance improvements and "unstable" optimizations
-    ```sh
-    # Install the nightly toolchain
-    rustup toolchain install nightly
-    # Configure your current project to use nightly (run this command within the project)
-    rustup override set nightly
-    # OR configure cargo to use nightly for all projects -- switch back with `rustup default stable`
-    rustup default nightly
+    
+    Create a ```rust-toolchain``` file in the root of your project, next to ```Cargo.toml```.
+    ```toml
+    [toolchain]
+    channel = "nightly"
     ```
-    * You can use `cargo +nightly ...` if you don't want to change the default to nightly.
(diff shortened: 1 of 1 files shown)


cart:
I honestly didn't know rust-toolchain files were a thing and I've been using rust for like 7 years!  This is a definite improvement.

cart:
Will this automatically install the required toolchain if it is missing?

cart:
Otherwise we'll still need to include the "install" rustup command.

bjorn3:
Yes, rustup will automatically install the right rust version if it is not yet installed.

cart:
Hmm to be clear: If I run `cargo run` in the root of a project containing a rust-toolchain file pointing to a missing toolchain, it will automatically run `rustup` and install the missing toolchain, then proceed with running `cargo run`?

cart:
Just want to make sure we're writing a complete guide.

bjorn3:
> Hmm to be clear: If I run cargo run in the root of a project containing a rust-toolchain file pointing to a missing toolchain, it will automatically run rustup and install the missing toolchain, then proceed with running cargo run?

Correct.

cart:
Brilliant!

DJMcNab:
I prefer `rust-toolchain.toml`, which makes it more clear that it's toml format

It might be a bit too recent after that stabilised for us to recommend it, however >:(

cart:
Eh if its stable, its fair game :)

ickk:
The ```rust-toolchain.toml``` file is only supported on the very latest rustup (v1.23.0), I actually ran into this issue a month or two ago and it wasn't obvious why it wasn't working (they had updated the docs, but hadn't released v.1.23.0 yet 😅). 

Weirder still, the ```rust-toolchain``` file (without toml extension) supports toml syntax in rustup v1.22.0 and possibly slightly earlier releases. 

I think if you are starting a new project it makes sense to use the toml variant, it might be worth adding a note about checking you have rustup >= v1.23.0.

cart:
Cool that all makes sense. We should probably let `.toml` cook for a bit before adopting it.

PR #170, opened 2021-06-04
Fix dependency error

MarenFayre:
Cargo looks for a "master" branch, but bevy changed it's convention to "main" causing the following error:
```
error: failed to get `bevy` as a dependency of package `xxx`

Caused by:
  failed to load source for dependency `bevy`

Caused by:
  Unable to update https://github.com/bevyengine/bevy

Caused by:
  failed to find branch `master`

Caused by:
  cannot locate remote-tracking branch 'origin/master'; class=Reference (4); code=NotFound (-3)
```
This change suggested by @mockersf fixes that.


cart:
Good call!

cart:
Its worth noting that this should be fixed by https://github.com/rust-lang/cargo/pull/9133

PR #173, opened 2021-06-08
display assets on parent categories

mockersf:
for https://github.com/bevyengine/bevy-assets/issues/112

assets that are not in a subcategory will now be displayed


cart:
Hmm just noticed that we now have a bunch of empty space when the parent category is empty:

image

mockersf:
fixed in #174 👍

PR #176, opened 2021-06-16
Welcome chapter

alice-i-cecile:
Welcome to the Welcome Chapter! This is good to go now from my perspective :) The code sample embedding can be a seperate PR: I'd rather not block forever on that.

TODO: 

- [ ] Automatic code sample embedding
- [x] Swap all docs links to short codes
- [x] Address https://github.com/bevyengine/bevy/issues/2773


content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,150 @@
...
+        group.add(bevy_gltf::GltfPlugin::default());
+
+        #[cfg(feature = "bevy_winit")]
+        group.add(bevy_winit::WinitPlugin::default());
+
+        #[cfg(feature = "bevy_wgpu")]
+        group.add(bevy_wgpu::WgpuPlugin::default());
+    }
+}
+```
```
Nilirad, line 81: I feel like this snippet is a bit too long to read. Since not every single detail is necessary here, maybe removing/hiding some lines could be a solution (I don't know if prepending a # to a line hides it like in rustdoc though).

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,150 @@
...
+```toml
+[dependencies]
+bevy = {"0.5", , default-features = false, features = ["..."]}
+```
+
+As shown in the [plugin_group.rs](https://github.com/bevyengine/bevy/blob/latest/examples/app/plugin_group.rs) example, you can also configure plugin groups from within Bevy itself.
+
 ## Writing your own plugins
 
+Plugins are simply collections of code that modify the `AppBuilder`.
```
Nilirad, line 98: This refers to the `AppBuilder`, but at this point of the book relevant talk about the app has not been done yet.

Nilirad (changes requested):
Found some typos + some observations.

content/learn/book/welcome/plugins/_index.md
alice-i-cecile: Yeah, I'm not sure I love linking this code directly.

Nilirad: Yeah, that would be awesome. It would also remove the code/docs synchronization problem.

content/learn/book/welcome/plugins/_index.md
alice-i-cecile: Provided some basic context here :)

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,90 @@
...
+struct App {
+  world: World
+  runner: Box<dyn Fn(App) + 'static, Global>
+  schedule: Schedule
+}
+```
+
+The `World` stores all of our data, both entity-component data and resources.
+The `runner` field contains a function that tells our app how to loop and execute.
+And the `Schedule` contains all of the systems that execute on our data, stored in a way that our scheduler can use to run them at the appropriate time.
```
Nilirad, line 46: This is better suited as API docs. Especially as it is stated later:

> Generally, you'll be operating at a higher-level of abstraction than this

We can transfer this knowledge to the API docs, which are currently scarce. There is also a benefit in doing this: pointing users to the API docs will encourage them to to it by their own.

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,90 @@
...
+
+Generally, you'll be operating at a higher-level of abstraction than this: thinking about data in terms of specific resources or components, adding systems to the right place rather than manipulating the `Schedule` yourself, and sticking to the default runner.
```
Nilirad, line 48: This is better suited as API docs. Especially as it is stated later:

> Generally, you'll be operating at a higher-level of abstraction than this

We can transfer this knowledge to the API docs, which are currently scarce. There is also a benefit in doing this: pointing users to the API docs will encourage them to to it by their own.

content/learn/book/welcome/installation/_index.md
```diff
@@ -95,6 +95,40 @@ Now run `cargo run` again. The Bevy dependencies should start building. This wil
...
+}
+
+fn hello(){
+    println!("Hello, Bevy!")
+}
+```
+
+Systems in Bevy are just ordinary Rust functions, which are registered in our app to operate on our world.
+
+Calling `cargo run` will run our schedule once, causing `hello` to print "Hello, Bevy!", allowing us to verify our installation of Bevy.
```
Nilirad, line 130: I think that the hello world example is better suited in the App section, where it is already fully quoted.

The installation guide is often partially or completely skipped by most users, since many are already familiar to the Rust ecosystem. Also, a basic app is not part of the installation.

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,106 @@
...
+
+pub struct MinimalPlugins;
+
+impl PluginGroup for MinimalPlugins {
+    fn build(&mut self, group: &mut PluginGroupBuilder) {
+        group.add(bevy_core::CorePlugin::default());
+        group.add(bevy_app::ScheduleRunnerPlugin::default());
+    }
+}
+```
```
Nilirad, line 76: IMO, it's better is source is always linked instead of being quoted, as explained in the main review message.

I think it's better to discuss *why* one should use `MinimalPlugins` and later help them decide whether `DefaultPlugin` suits better to their use case.

Nilirad (commented):
Round two 😝. Good job, as always.

The structure of the chapter looks good to me, especially since the addition of the App section.

The level of communication could be improved though. In many points the book explains the details of the engine, while the Diátaxis Framework states:

> [Explanation is] not a close-up view of the machinery, like reference material. [...] This allows explanation to _become discussion, a more relaxed, freer way to consider something_. Explanation joins things together. It’s documentation that it makes sense to read while away from the product itself.

I think this happens primarily because the API docs are quite neglected. So, I was thinking: since the next days I'm not extra busy, maybe I can open a parallel PR draft on bevy to enrich the API docs in the points where they are referenced by the book.

content/learn/book/welcome/installation/_index.md
alice-i-cecile: Hmm. I am uncomfortable ending the installation guide without a proper check that things "actually work". This isn't quite the right tool for that though; maybe it should use DefaultPlugins instead.

content/learn/book/welcome/plugins/_index.md
alice-i-cecile: I can agree with that.

content/learn/book/welcome/app/_index.md
alice-i-cecile: Sounds good to me.

alice-i-cecile:
> The structure of the chapter looks good to me, especially since the addition of the App section.

Agreed. This helps dramatically.

> So, I was thinking: since the next days I'm not extra busy, maybe I can open a parallel PR draft on bevy to enrich the API docs in the points where they are referenced by the book.

That sounds useful, thanks.

content/learn/book/welcome/installation/_index.md
alice-i-cecile: Improved in b0a4951c7a6423045a37dab05738240f56265167

content/learn/book/welcome/plugins/_index.md
alice-i-cecile: Fixed in b9dcd26d4ae4ee62abcfef184314d4edfa154960

Nilirad:
I don't know if you already know it, but you can link the source to docs.rs instead of GitHub.

Example: https://docs.rs/bevy_internal/latest/src/bevy_internal/default_plugins.rs.html#3

alice-i-cecile:
Hmm. The docs.rs source links are interesting. I'm not sure if I like them more or less than the GitHub links. They're more idiomatic, but make it harder to explore related code.

Nilirad:
I found a couple of pitfalls with the docs.rs source links, that can be fortunately avoided:

1. They relate to lines of code, so if the file gets updated the link may become outdated. (We can avoid this by just linking to the docs.rs page instead of directly to the source. This will also make easy to explore the source code of related items).
2. Subcrates don't have the src link (This can be solved by linking to the original crate docs, e.g. `bevy_app` for `App`)

Solving issue number 2 limits the docs navigation to only the subcrate though. This may not be a huge issue but it will be a bit frustrating for the user that has to manually go to docs.rs and select the main bevy crate to see stuff on other crates. This can be solved by hosting the docs on the website and building them with dependencies (which is by default). This will put on us some responsibility, but the benefits of having the complete docs outweights the cost IMO.

[Example of self-hosted API docs](https://api.rocket.rs/v0.5-rc/rocket/): Looks like it's possible to limit documenting dependencies to the local crates 🤔.

alice-i-cecile:
For 2, this is why I'm careful to link to the original definition :) Do you think it's better to link to docs.rs over GitHub then?

Nilirad:
To me, it looks better and it's more tightly coupled with the docs. The important thing though, is that we link to the API item and not the lines of code. Otherwise, we should manually check each link at each update to ensure it's still updated.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
 page_template = "book-section.html"
 +++
 
-TODO: Briefly discuss what Bevy is.
+Welcome to the Bevy game engine! A bevy may be a group of birds, but our engine is:
+
+- Rust-first: with all the safety, tooling and power that comes with a modern systems language
```
cart, line 11: I think bullet points should have consistent formatting. In this case, lets start each one with a captial letter.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+- completely free, open source and hackable with an architecture designed for modularity
+- powered by an ergonomic and innovative ECS that makes writing clear, performant gameplay and engine logic a joy
+- cross-platform: support all major desktop platforms with a single code-base, with limited mobile and web support
```
cart, line 14: I think each bullet should take the form '**Snappy Thing**: longer single-sentence description". I think that flows better and helps the user internalize each point (see the section that this replaces in the current book).

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+- designed for effortless parallelism with a data-oriented design and automatic parallel scheduler
+- built on a high-performance foundation
```
cart, line 16: These can probably be merged into a single **Fast** / **High Performance** bullet point.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+However, Bevy is still in its infancy, and there are very good reasons not to choose it for your next project.
+It is currently completely missing first-party solutions for:
+
+- animation
+- networking
+- console support
+
+Solid third-party (but not first-party) solutions currently exist for:
+
+- [audio](https://crates.io/crates/bevy_kira_audio)
```
cart, line 33: People will likely read this and think Bevy has no support for audio at all. I don't like those optics.

content/learn/book/welcome/_index.md
cart, line 14: "limited mobile and web support" in a list of features / design goals doesn't get the point across well enough (it reads  more like a fundamental limitation / consequence of the design). I think we should rephrase this more optimistically:

> publish Bevy Apps to Windows, MacOS, Linux, iOS*, Android*, and Web* using a single code base. Platforms with (*) are still a work in progress and are missing some features.

I also think iOS might be able to graduate to "first class" support. It has audio and asset support now and I'm not aware of any limitations.

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,90 @@
++++
+title = "Creating Bevy apps"
+weight = 2
+template = "book-section.html"
+page_template = "book-section.html"
++++
+
+When you're making a game in Bevy, you will have a single `App`, which stores and handles all of your game logic and data.
```
cart, line 8: Unless referring to a specific "game", I think we should refer to arbitary Bevy programs as "apps". So this should read:
"When you're making an app in Bevy, you will have a single `App`, which stores and handles ..."

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,90 @@
...
+We do so by add a simple system, which prints "Hello, Bevy!" when it is run.
+Finally once all of our work is complete, we call [`AppBuilder::run()`](https://docs.rs/bevy/latest/bevy/app/struct.AppBuilder.html#method.run) to actually make our app *do things*.
+
+## What makes an App?
+
+So, what sort of data does are `App` really store?
+Turns out: it's fairly simple.
+Looking at the docs for [`App`](https://docs.rs/bevy/latest/bevy/app/struct.App.html), we find three fields:
+
+```rust
```
cart, line 36: I would prefer to not immediately throw stuff like `Box<dyn Fn(App) + 'static, Global>` at people on the second page of the Bevy Book. We can introduce World/Schedule/runner concepts without this.

By not inlining the App struct, we also insulate against future changes (such as the infrastructure required to support deferred plugin init).

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,96 @@
...
+fn increment_score(score: ResMut<Score>){
+  score.0 += 1;
+}
+
+fn report_score(score: Res<Score>){
+  dbg!(score);
+}
+```
 
 ## `DefaultPlugins`
```
cart, line 61: Code blocks in headings currently show up "too small". We'll want to either remove the code block or improve the formatting before publishing this.

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,96 @@
...
+
+#[derive(Default, Debug)]
+struct Score(u8);
+
+fn increment_score(score: ResMut<Score>){
+  score.0 += 1;
+}
+
+fn report_score(score: Res<Score>){
+  dbg!(score);
```
cart, line 57: I think we should replace this with `info!(score)`. The "tracing" macros are "idiomatic bevy logging" and help us ensure proper cross platform logging.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
 page_template = "book-section.html"
```
cart, line 6: Quick heads up that I'm going to be a diva about this section :) It is a lot like the home page of the website and I have lots of opinions about what that should look like :smile:

cart: I think a couple of iterations here is probably good enough for now, but theres a good chance that I'll revisit / rework this.

content/learn/book/welcome/app/_index.md
Nilirad: Yes, I also talked with Alice about this. We ended up deciding to simply link to the API docs instead, where the user can also check the engine source code if they want. By the way, `App` fields are private implementation details and therefore should definitely not be in the Book.

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,90 @@
...
+}
+
+fn hello(){
+    println!("Hello, Bevy!")
+}
+```
+
+So, we first create our [`App`](https://docs.rs/bevy/latest/bevy/app/struct.App.html) by telling it to [`build`](https://docs.rs/bevy/latest/bevy/app/struct.App.html#method.build), performing initialization steps.
+This produces an [`AppBuilder`](https://docs.rs/bevy/latest/bevy/app/struct.AppBuilder.html), which we can use to configure the `App`.
+We do so by add a simple system, which prints "Hello, Bevy!" when it is run.
```
asleeponmykeyboard, line 27:
```suggestion
We do so by add our `hello()` function as a system through a **trait extension method**, which will print 'Hello, Bevy!' when the app is run.
```
Explaining what `.system()` does is pretty important.

alice-i-cecile: Definitely agree. It will likely be fully deprecated for most use cases (including here) though, so this is probably not the best place to describe how the magic works.

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,96 @@
...
 ## Third-party plugins
+
+Importing 3rd-party plugins is easy; they're just Rust code!
+
+1. Find a Bevy plugin (such as from our [collection of assets](https://bevyengine.org/assets/)).
+2. Add it to your `Cargo.toml` as a crate under `[dependencies]`.
+3. Import the code definitions from the crate.
+4. Add the plugin to your app!
+
+Follow the documentation and examples from the crates you're importing to make sure you have everything configured properly.
```
BlackPhlox, line 96: I think it would be most relevant to include the content of #195's mention of the plug 'n playiness here (if not clear enough at line 15-17) along with what to do if the user wants to publish their own crate/plugin.

alice-i-cecile: Added in https://github.com/bevyengine/bevy-website/pull/176/commits/6615a69236db3ccc7a8d27c5a28754e61e41aa8f

content/learn/book/welcome/plugins/_index.md
alice-i-cecile: I'd prefer to fix the formatting: it's nice to clearly signify that it's a proper programming noun.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
 +++
 
-TODO: Briefly discuss what Bevy is.
+Welcome to the Bevy game engine! A bevy may be a group of birds, but our engine is:
+
+- **Rust-first:** with all the safety, tooling and power that comes with a modern systems language
+- **Completely free:** open source and hackable with an architecture designed for modularity
```
mirkoRainer, line 12: This line makes me think of something like this SO https://softwareengineering.stackexchange.com/questions/21907/open-source-but-not-free-software-or-vice-versa

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+
+- audio
+- user interfaces
+
+Solid third-party solutions currently exist for:
+
+- [advanced audio](https://crates.io/crates/bevy_kira_audio)
+- [realistic physics](https://github.com/dimforge/bevy_rapier)
+
+While Bevy's modular architecture makes it relatively easy to integrate your own (or third-party) solutions, be mindful that Bevy does not provide a complete game solution out of the box yet.
```
mirkoRainer, line 42: imho, "yet" at the end should be emphasized. :)

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,82 @@
++++
+title = "Creating Bevy apps"
+weight = 2
+template = "book-section.html"
+page_template = "book-section.html"
++++
+
+Bevy programs store and execute all of their game logic and data with a single `App` data structure.
+Let's make a trivial Hello World app.
```
mirkoRainer, line 9: This seems like it should precede the first line.

I would like to see the flow if "Bevy programs store and execute all of their game logic and data with a single `App` data structure." was moved to below the code block, so the reader know's we're making a trivial game, sees the code, then sees the explanation for said code.

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,82 @@
...
+  .add_system(hello.system())
+  .run();
+}
+
+fn hello(){
+    println!("Hello, Bevy!")
+}
+```
+
+The process is straightforward: we first create a new [`App`](https://docs.rs/bevy/latest/bevy/app/struct.App.html).
```
mirkoRainer, line 25: "Bevy programs store and execute all of their game logic and data with a single `App` data structure."

The quoted text should go after this line, imho.

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,82 @@
...
+Then, we add a simple system, which prints "Hello, Bevy!" when it is run.
+Finally once we're done configuring the app, we call [`App::run()`](https://docs.rs/bevy/latest/bevy/app/struct.App.html#method.run) to actually make our app *do things*.
+
+## What makes an App?
+
+So, what sort of data does are `App` really store?
+Turns out: it's fairly simple.
```
mirkoRainer, line 32: Alternatively, or maybe more appropriately, "Bevy programs store and execute all of their game logic and data with a single `App` data structure." should go here?

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,82 @@
...
+Looking at the docs for [`App`](https://docs.rs/bevy/latest/bevy/app/struct.App.html), we find three fields: `world`, `runner` and `schedule`.
+You can read more about these by exploring the reference documentation linked just above.
```
mirkoRainer, line 34: This seems implied by the link present in the previous sentence.

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+Generally, you'll be operating at a higher-level of abstraction than these basic primitives: thinking about data in terms of specific resources or components, adding systems via convenient methods rather than manually supplying a `Schedule`, and sticking to the default runner which runs your game logic in a loop.
```
mirkoRainer, line 36: higher-level -> higher level

mirkoRainer:
```suggestion
Generally, you'll be operating at a higher-level of abstraction. You'll likely be thinking about data in terms of specific resources or components, or adding game logic systems via methods to be run in the [default runner](?url?).
```

Also, is "default runner" a Bevy specific term? Are there docs for this to be linked to?
url?? == https://docs.rs/bevy/0.5.0/bevy/app/struct.App.html#structfield.runner

content/learn/book/welcome/app/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+To do so, you'll customize your own `App` using various methods with the builder pattern.
```
mirkoRainer, line 38: Link to builder pattern? 

https://en.wikipedia.org/wiki/Builder_pattern#:~:text=The%20builder%20pattern%20is%20a,Gang%20of%20Four%20design%20patterns.

content/learn/book/welcome/plugins/_index.md
mirkoRainer: Would it also benefit to link to the respective doc if it's a proper noun in bevy context?

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,99 @@
...
+
+fn report_score(score: Res<Score>){
+  info!(score);
+}
+```
 
 ## `DefaultPlugins`
 
-## Writing your own plugins
+Bevy's `DefaultPlugins` are intended to get up and running with a "standard game" as quickly as possible. Let's take a look at the [source](https://github.com/bevyengine/bevy/blob/latest/crates/bevy_internal/src/default_plugins.rs) again.
```
mirkoRainer, line 63: Link to DefaultPlugin docs page?

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,99 @@
...
 
 ## Third-party plugins
+
+Importing 3rd-party plugins is easy; they're just ordinary Rust code that can be managed with `cargo`.
+Bevy's modular nature tends to result in simple plug-and-play interoperability and easy extensibility, so don't be afraid to try out plugins that seem interesting or useful for your game.
+
+1. Find a Bevy plugin (such as from our [collection of assets](https://bevyengine.org/assets/)).
+2. Add it to your `Cargo.toml` as a crate under `[dependencies]`.
+3. Import the code definitions from the crate (i.e. `using bevy_third_party::prelude::*`) to add the appropriate items to your workspace.
+4. Add the plugin to your app!
```
mirkoRainer, line 95:
```suggestion
4. Add the plugin to your app! (i.e. `App::build().add_plugins(bevy_third_party)...`)
```

mirkoRainer (commented):
OK. I read through this and made some opinion comments. Please let me know if they're way off.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+- **Rust-first:** with all the safety, tooling and power that comes with a modern systems language
+- **Completely free:** open source and hackable with an architecture designed for modularity
+- **Innovative ECS:** powered by an ergonomic, full-native entity-component-system architecture that makes writing clear, performant gameplay and engine logic a joy
+- **Cross-platform:** support Windows, MacOS, Linux, web*, Android* and iOS with a single code base
+- **Effortless parallelism:** a data-oriented design and automatic parallel scheduler provide a high-performance foundation for your game
+
+This book is designed for new and experienced users looking for a thoughtful, [explanation-first](https://diataxis.fr/explanation/) guide to the engine's essential features.
+If you just want to try it out hands-on, jump into one of our [quick-start guides](TODO: add link) instead!
+
+*: support for these platforms is still limited; you may encounter some missing features or a more involved setup process
```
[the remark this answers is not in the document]
alice-i-cecile: Hmm. The Cheatbook has a good link to the web setup. @jamadazi, would you prefer we fork those instructions or link directly?

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+
+## Stability warning
+
+Bevy aims to be a general-purpose game engine capable of handling arbitrary 2D or 3D games.
+However, Bevy is still in its infancy, and there are very good reasons not to choose it for your next project.
```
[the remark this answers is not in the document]
alice-i-cecile: Maybe toddler-hood now ;) It's still very immature, and if you're trying to release a Complete Game half of your time is going to end up doing engine dev still.

content/learn/book/welcome/_index.md
```diff
@@ -6,4 +6,44 @@ template = "book-section.html"
...
+It is currently completely missing first-party solutions for:
```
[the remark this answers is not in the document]
alice-i-cecile: Yes and no. Some of these (like networking) have bits and pieces of solutions, but nothing that I would call fully usable. By contrast, bevy_kira_audio and bevy_rapier basically Just Work at this point.

content/learn/book/welcome/app/_index.md
alice-i-cecile: I wanted a bit of a firmer nudge here, targeted towards beginners to programming. "Read the docs" is incredibly powerful but not always intuitive for new programmers or those coming from ecosystems with weak docs.

content/learn/book/welcome/app/_index.md
alice-i-cecile: I agree that this deserves a link. That wikipedia page is terrible though; let me try and dig up something better.

alice-i-cecile: https://doc.rust-lang.org/1.0.0/style/ownership/builders.html is much better, and Rust specific!

content/learn/book/welcome/app/_index.md
alice-i-cecile: I'd like to go explanation -> example in the book; the quick start guides will be better suited for readers who prefer the opposite direction.

content/learn/book/welcome/plugins/_index.md
alice-i-cecile: Link added just below :)

content/learn/book/welcome/_index.md
alice-i-cecile: Clarification added!

content/learn/book/welcome/_index.md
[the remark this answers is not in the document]
alice-i-cecile: Clarified to avoid use of first-party here.

content/learn/book/welcome/app/_index.md
alice-i-cecile: I think the default runner talk is mostly just confusing. I cut it.

content/learn/book/welcome/plugins/_index.md
```diff
@@ -0,0 +1,99 @@
...
+}
+
+fn report_score(score: Res<Score>){
+  info!(score);
+}
+```
+
+## `DefaultPlugins`
+
+Bevy's [`DefaultPlugins`](https://docs.rs/bevy/latest/bevy/struct.DefaultPlugins.html) are intended to get up and running with a "standard game" as quickly as possible. Let's take a look at the [source](https://github.com/bevyengine/bevy/blob/latest/crates/bevy_internal/src/default_plugins.rs) to see what's included.
```
cart, line 63: Any particular reason for using manual links instead of the short codes? They save space, cut down on redundancy, and ensure consistency:

```md
{{rust_type(type="struct" crate="bevy_core", name="CorePlugin")}}
[`CorePlugin`](https://docs.rs/bevy/latest/bevy/core/struct.CorePlugin.html)

{{rust_type(type="struct" crate="bevy_app", name="AppBuilder" method="add_system")}}
[`AppBuilder::add_system()`](https://docs.rs/bevy_app/latest/bevy_app/struct.AppBuilder.html#method.add_system)
```

cart: The "structured" approach also opens the doors to smarter formatting in the future (ex: different colors for traits vs structs)

alice-i-cecile: I just didn't know about the short codes :P

I'll do a cleanup pass on all of them...

content/learn/book/welcome/plugins/_index.md
```diff
@@ -0,0 +1,99 @@
...
++++
+
+In Bevy, virtually all of the functionality you might want is added via **plugins**, which are added to your app.
+These might provide basic windowing functionality, handle input or sound, calculate physics or provide game-specific logic like a tile map system.
+
+By combining Bevy's first-party plugins with third-party alternatives, you can modularly customize the behaviour of the Bevy game engine, and then add your own plugins to contain your own game-specific code in a well-organized way.
+
+## Writing your own plugins
+
+Plugins are collections of code that modify the `App` (which controls all of the data and logic of our game) using the builder pattern.
```
cart, line 15: Also as a general rule: if you are referencing a "rust symbol", I think it should use the "rust short code link", even if you've already provided a link elsewhere. I want the Bevy Book to feel "strongly typed".

cart:
Just pushed some style + phrasing tweaks:

* Shorter / more concise titles
  * I think this looks much better in the sidebar (given how compact it is) and makes the guide feel less overwhelming
* Tweaked the plugin section opener. 
* Replaced manual doc links with short codes in a couple of places and used them more (there are still places to change these)
* Moved the "default plugin features" to a new section to the end. This disconnects it from the DefaultPlugins introduction, but imo these is a super niche feature that most people dont need to worry about. It complicates the introduction of a topic that for most people should be very straightforward / simple. 
* Small rewords here and there according to taste

Happy to discuss any/all of these changes. It just seemed faster to make them inline instead of having a long back and forth explaining what I want.

content/learn/book/welcome/plugins/_index.md
cart: Awesome. Thanks!

alice-i-cecile:
> Happy to discuss any/all of these changes. It just seemed faster to make them inline instead of having a long back and forth explaining what I want.

Yep, I'm perfectly content to have you do tweaks like that directly :)

content/learn/book/welcome/_index.md
```diff
@@ -1,9 +1,48 @@
...
 page_template = "book-section.html"
 +++
 
-TODO: Briefly discuss what Bevy is.
+Welcome to the Bevy game engine! A bevy may be a group of birds, but our engine is:
+
+- **Rust-first:** all the safety, tooling and power that come with a modern systems language
+- **Completely free:** [free-of-charge](https://github.com/sponsors/cart), [open source](https://github.com/bevyengine/bevy/blob/main/LICENSE), and hackable with an architecture designed for modularity
+- **Innovative ECS:** powered by an ergonomic, full-native entity-component-system architecture that makes writing clear, performant gameplay and engine logic a joy
+- **Cross-platform:** support Windows, MacOS, Linux, web*, Android* and iOS with a single code base
```
blaind, line 14: What do you think of mentioning about future XR support? ("XR being under development" or similar)

content/learn/book/welcome/_index.md
```diff
@@ -1,9 +1,48 @@
...
+
+Solid third-party solutions currently exist for:
+
+- [advanced audio](https://crates.io/crates/bevy_kira_audio)
+- [realistic physics](https://github.com/dimforge/bevy_rapier)
+
+While Bevy's modular architecture makes it relatively easy to integrate your own (or third-party) solutions, be mindful that Bevy does not provide a complete game solution out of the box *yet*.
+[Contributions to the engine](https://github.com/bevyengine/bevy/) are extremely welcome, but if you want something fully-featured to make your dream game *today*, check out [Godot](https://godotengine.org/) instead!
+
+Unsurprisingly, Bevy's rapid pace of development means that our APIs will break, repeatedly, in both deep and pervasive ways.
```
blaind, line 45: One thing to consider: Rust is often considered to be in line with "if it compiles, it works" - however, I think bevy is not matured there yet. Maybe a good to mention in docs?

(920+ unwrap() or panic!'s in code currently - I've personally been caught by a surprise few times to runtime crashes)

content/learn/book/welcome/_index.md
Nilirad: Also, a brief explanation of what XR is would help people who don't know it's actually VR/AR.

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,107 @@
...
+fn increment_score(score: ResMut<Score>){
+  score.0 += 1;
+}
+
+fn report_score(score: Res<Score>){
+  info!(score);
+}
+```
+
+## `DefaultPlugins` and `PluginGroup`
```
Nilirad, line 60: Monospace text is rendered particularly small in headings. Also, the paragraph should talk about plugin groups in general (with an appropriate subtitle) and then introduce both `DefaultPlugins` and `MinimalPlugins`.

lol

content/learn/book/welcome/_index.md
alice-i-cecile: I think this is a good plan.

content/learn/book/welcome/_index.md
alice-i-cecile: That's probably a good caveat. Compared to other game engines, it's surprisingly "if it compiles it works". But much less so for people coming to games from the Rust ecosystem!

content/learn/book/welcome/plugins/_index.md
cart: I called the formatting problem out here too: https://github.com/bevyengine/bevy-website/pull/176#discussion_r661047979
We either need to:
1. fix "code formatted headings" to be the right size
2. remove code formatting in headings

I'm actually sort of in favor of (2) because code blocks in headings feel kind of "noisy" to me. But I'm happy to try (1) and see how we feel.

I don't think this needs to be resolved in this pr, but it needs to be resolved prior to "releasing" the new version of the book.

content/learn/book/welcome/_index.md
cart: I think "if it compiles, it works" is an unrealistic goal because Rust's static guarantees are just a subset of its capabilities. I also don't think that "number of unwraps" is a particularly good measure of code maturity. There are plenty of cases where the only real option is an internal unwrap.

That being said, except in a few specific cases, Bevy internals shouldn't crash from normal "userspace" code (and they do currently in a number of cases).

Its worth calling out that Bevy currently has more rough edges / bugs / crashes than a stable product would. However this page is also the first thing Bevy users will see and I consider it a key part of "Bevy marketing". We should do it in a way that doesn't (1) scare people away or (2) give the wrong impression about our goals and priorities. Feel free to add a line here, but (as I've already warned @alice-i-cecile) I have a lot of very specific goals for this page and I'll probably follow this pr up with a lot of changes.

mirkoRainer: > except in a few specific cases

Are these cases outlined elsewhere? While I agree that they shouldn't be listed on the first welcome page, perhaps a link to the "rough edges" would be appropriate.

content/learn/book/welcome/plugins/_index.md
Nilirad: Obviously (1) must be done sooner or later, but we really shouldn't abuse mentioning identifiers in the Book, especially in headings. I'm on the opinion that we should not strictly follow the structure of the API (otherwise we get a bad copy of the API docs), but that thoughts should be laid out in a way that give access to a comprehension level that reading source code or API docs can't do.

For the header, in this case, "Plugin groups" should be sufficient.

content/learn/book/welcome/_index.md
blaind: Something like:
- **Cross-platform:** support Windows, MacOS, Linux, web*, Android* and iOS with a single code base. XR** support under development.

** XR refers to extended reality, which is a broader concept for spatial computing such as augmented reality, virtual reality and mixed reality.

content/learn/book/welcome/_index.md
blaind: There was good discussion in Discord about the unwrap's as well yesterday. I agree that there will be unwraps, and some of them are totally okay as well (e.g. when starting app, during development work). Also, of the 920+ quite many seemed to be in the tests code, which I did not consider when grepping.

Good point about the marketing, maybe the extra warning should not be yet - the section header "Stability warning" already gives a hint about it. 

Also, over time the runtime panic's will get lower and lower. However, I would still be very mindful of this, since an unanticipated runtime panic at wrong point (e.g. demo time, or in production) will quickly turn new adopters away.

content/learn/book/welcome/apps/_index.md
```diff
@@ -0,0 +1,81 @@
...
+// A simple data structure to store the two current fibonacci numbers
+struct Fibonacci {
+    a: u64,
+    b: u64,
+}
+
+// This system requires mutable access to our Fibonacci resource,
+// so we add `ResMut` to its function parameters
+fn fibonacci_sum(mut fibonacci: ResMut<Fibonacci>) {
+    // This crashes fairly quickly, as we overflow our u64 data storage
```
mirkoRainer, line 70: Should we put the code in to stop the crash? As a beginner I would be hesitant to run something that crashes. Perhaps add a line with checks then comment that with something like "Learning exercise: What would happen if we removed this line?"

That may be too "tutorial" but just an idea.

content/learn/book/welcome/_index.md
cart: > Are these cases outlined elsewhere? While I agree that they shouldn't be listed on the first welcome page, perhaps a link to the "rough edges" would be appropriate.

They are not. It will be hard to maintain the list unless it integrates directly with the issue tracker. Maybe we should have a label for "UX speedbumps". We can link to that specific label. And that way whenever we resolve an issue, the list is automatically updated.

content/learn/book/welcome/apps/_index.md
cart: Yeah I'd rather just avoid having an example that requires special casing, crashes, or warning comments in the first place. Fibonacci (while extremely common in programming and math circles) might not be the best choice for some types of users. Maybe we should pick something a little more "generic" and approachable. Maybe something like a "greeting message" logger that prints a greeting message from a `struct GreetingMessage { message: String }` resource?

cart:
After handling @mirkoRainer's comments, I think we can probably merge this. Gotta start moving forward and we can always iterate on this later.

content/learn/book/welcome/_index.md
```diff
@@ -1,9 +1,48 @@
...
+
+- audio
+- user interfaces
+
+Solid third-party solutions currently exist for:
+
+- [advanced audio](https://crates.io/crates/bevy_kira_audio)
+- [realistic physics](https://github.com/dimforge/bevy_rapier)
+
+While Bevy's modular architecture makes it relatively easy to integrate your own (or third-party) solutions, be mindful that Bevy does not provide a complete game solution out of the box *yet*.
```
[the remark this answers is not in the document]
alice-i-cecile: I think the plugins section is probably best.

alice-i-cecile:
Done! I'm not thrilled with the new example (it mostly duplicates the line above and will spam the user's console), but we can modify it later in a small follow-up PR.

content/learn/book/welcome/apps/_index.md
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// This resource can store a string
+struct Message {
+    string: String,
+}
+
+// This system reads our Message resource,
+// so we add `Res<Message>` to its function parameters
+fn (mut message: Res<Message>) {
```
Protowalker, line 67: Might be worth adding a clarifying comment saying that bevy builds systems out of functions that take query values as parameters

cart: This function is also missing a name!

content/learn/book/welcome/apps/_index.md
```diff
@@ -0,0 +1,70 @@
...
+
+fn main() {
+    App::new()
+        // Plugins are App code that was written elsewhere,
+        // imported as a single unit for organization and clarity
+        .add_plugins(MinimalPlugins)
+        // Resources are global singleton data stored in the `World`
+        .insert_resource(Message {string: "Welcome to Bevy!"})
+        // Systems run every pass of the game loop and perform logic
+        .add_system(print_message_system.system())
```
djeedai, line 56: Is `.system()` obsolete or not? I think the welcome page omits it.

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,104 @@
...
 
-## `MinimalPlugins`
-
-## `DefaultPlugins`
+This empowers Bevy developers to modularly "build their own engine" using official, third party, and custom {{rust_type(type="trait" crate="bevy_app" name="Plugin" plural=true)}}. Bevy intentionally blurs the lines between engine developers and app developers.
 
 ## Writing your own plugins
 
+Plugins are collections of code that modify {{rust_type(type="struct" crate="bevy" mod = "app" name="App" no_mod = "true")}}.
+Any code in a plugin could be directly substituted directly on the base {{rust_type(type="struct" crate="bevy" mod = "app" name="App" no_mod = "true")}}.
```
djeedai, line 15:
```suggestion
Any code in a plugin could be directly substituted on the base {{rust_type(type="struct" crate="bevy" mod = "app" name="App" no_mod = "true")}}.
```

content/learn/book/welcome/plugins/_index.md
```diff
@@ -1,16 +1,104 @@
...
+
+```rust
+App::new().add_plugins(DefaultPlugins)
+```
+
+Take a look at the [source](https://github.com/bevyengine/bevy/blob/latest/crates/bevy_internal/src/default_plugins.rs) to see a full list of what's included.
+
+If you're looking to structure your Bevy app in an unusual way and don't want to use most of the functionality provided by the engine, you can choose to use  Bevy's {{rust_type(type="struct" crate="bevy" name="MinimalPlugins")}} instead.
+
+We can click through to the [source]((https://github.com/bevyengine/bevy/blob/latest/crates/bevy_internal/src/default_plugins.rs)) for the `impl PluginGroup for MinimalPlugins` to see that this adds {{rust_type(type="struct" crate="bevy_core" name="CorePlugin")}} and {{rust_type(type="struct" crate="bevy_app" name="ScheduleRunnerPlugin")}}.
```
djeedai, line 74:
```suggestion
We can click through the [source]((https://github.com/bevyengine/bevy/blob/latest/crates/bevy_internal/src/default_plugins.rs)) for the `impl PluginGroup for MinimalPlugins` to see that this adds {{rust_type(type="struct" crate="bevy_core" name="CorePlugin")}} and {{rust_type(type="struct" crate="bevy_app" name="ScheduleRunnerPlugin")}}.
```

content/learn/book/welcome/apps/_index.md
alice-i-cecile: It's obsolete; my bad.

content/learn/book/welcome/apps/_index.md
alice-i-cecile: Fixed! Should not be mut, should have a name!

content/learn/book/welcome/_index.md
```diff
@@ -1,9 +1,49 @@
...
 sort_by = "weight"
 template = "book-section.html"
 page_template = "book-section.html"
 +++
 
-TODO: Briefly discuss what Bevy is.
+Welcome to the Bevy game engine! A bevy may be a group of birds, but our engine is:
+
+- **Rust-first:** all the safety, tooling and power that come with a modern systems language
+- **Completely free:** [free-of-charge](https://github.com/sponsors/cart), [open source](https://github.com/bevyengine/bevy/blob/main/LICENSE), and hackable with an architecture designed for modularity
```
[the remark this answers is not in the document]
alice-i-cecile: I like that.

content/learn/book/welcome/_index.md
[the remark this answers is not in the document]
alice-i-cecile: Linking to the Cheatbook!

content/learn/book/welcome/apps/_index.md
```diff
@@ -0,0 +1,70 @@
...
+
+```rust
+use bevy::prelude::*;
+
+fn main() {
+    App::new()
+        // Plugins are App code that was written elsewhere,
+        // imported as a single unit for organization and clarity
+        .add_plugins(MinimalPlugins)
+        // Resources are global singleton data stored in the `World`
```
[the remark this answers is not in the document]
alice-i-cecile: Briefly done above :)

content/learn/book/welcome/setup/_index.md
```diff
@@ -95,6 +97,81 @@ Now run `cargo run` again. The Bevy dependencies should start building. This wil
...
+  ```toml
+  [dependencies]
+  bevy = { version = "0.5.0", features = ["dynamic"] }
+  ```
+
+  NOTE: Remember to revert this before releasing your game! Otherwise you will need to include `libbevy_dylib` alongside your game if you want it to run. If you remove the `dynamic` feature, your game executable can run standalone.
+
+For the fastest iterative compile times, we recommend the following configuration as well. You need to follow all three steps for any effect:
+
+1. **LLD linker**: The Rust compiler spends a lot of time in the "link" step. LLD is _much faster_ at linking than the default Rust linker. To install LLD, find your OS below and run the given command:
```
IceSentry, line 144: Maybe also suggest using `mold` for people using linux

content/learn/book/welcome/setup/_index.md
```diff
@@ -95,6 +97,81 @@ Now run `cargo run` again. The Bevy dependencies should start building. This wil
...
+   * **Arch**: `sudo pacman -S lld`
+   * **Windows**: Ensure you have the latest [cargo-binutils](https://github.com/rust-embedded/cargo-binutils)
+
+      ```sh
+      cargo install -f cargo-binutils
+      rustup component add llvm-tools-preview
+      ```
+
+   * **MacOS**: Modern LLD does not yet support MacOS, but we can use zld instead: `brew install michaeleisel/zld/zld`
+2. **Nightly Rust Compiler**: This gives access to the latest performance improvements and "unstable" optimizations
```
IceSentry, line 155: I think this step should clearly be marked as optional or at least be more obvious that this is purely for compilation speed and that bevy doesn't require nightly to work. Bevy works without any issues on stable rust so I wouldn't want anyone to think they _need_ nightly.

content/learn/book/welcome/apps/_index.md
```diff
@@ -0,0 +1,70 @@
...
+    println!("Hello, Bevy!")
+}
+```
+
+## What makes an App?
+
+So, what sort of data does our  {{rust_type(type="struct" crate="bevy" mod = "app" name="App" no_mod = "true")}} really store?
+Looking at the docs linked, we find three fields: `world`, `schedule` and `runner`.
+The `world` field stores all of our game's data, the `schedule` holds the systems that operate on this data (and the order in which they do so) and the `runner` interprets the schedule to control the broad execution strategy.
+You can read more about these by exploring the reference documentation linked just above.
```
IceSentry, line 34: It is not immediately obvious to me which documentation link this is referencing.

cart:
Merging this as I think its in a reasonable spot. But before publishing this, I'd still like to do a pass over the introduction, because as I mentioned 7.5 months ago (!!!!!! sorry sorry sorry !!!!!),  it is kind of like the front page of the website and I'd like to have firm control over how Bevy is introduced.

alice-i-cecile:
> But before publishing this, I'd still like to do a pass over the introduction, because as I mentioned 7.5 months ago (!!!!!! sorry sorry sorry !!!!!), it is kind of like the front page of the website and I'd like to have firm control over how Bevy is introduced.

Agreed! We'll do a proper pass before publication.

Issue #11, opened 2020-08-18
Change getting started to use git instead of cargo


IceSentry:
Currently, since the project is moving really fast I've seen a few issues that are fixed by using the git version of bevy instead of using cargo. I assume this will stabilize and won't be an issue, but for now I think it would be good to at least add a note to mention this in the installation guide.

Issue #10, opened 2020-08-14
CI is broken for pull request validation


cart:
Its trying to deploy the website during validation and (very very fortunately) failing because it doesnt have permission.

We should add another configuration specifically for PR validation.

guimcaballero:
It looks like [shalzz/zola-deploy-action](https://github.com/shalzz/zola-deploy-action) has a way to do this listed on the Readme:

> This example will build and deploy on master branch to gh-pages branch. Additionally will build only on pull requests.

This is what we're looking for right?

cart:
Closed by #24

PR #182, opened 2021-06-29
ECS chapter for new book

alice-i-cecile:
If you'd like to read (or review) this PR, you'll need to:

1. Install [zola](https://www.getzola.org/).
2. Clone this repo.
3. Use `zola serve --open` from that directory.

**To do:**
- [ ] Use  a more direct approach to introducing ECS concepts.
- [ ] Reduce the wall-of-text / wall-of-code alternation.
- [ ] Move some of the gorier details into API docs.
- [ ] Consider cutting `WorldCell` from this chapter.


content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+
+Bevy's scheduler is remarkably powerful: it uses the information about data access defined in our system parameters to automatically run systems in parallel.
+By default, every system in a stage runs in parallel with every other system in that stage (as long as threads exist to take the work): the only rule is that systems which have the ability to *write* to a particular piece of data (such as a resource) cannot be run at the same time as other systems which read or write to that same data.
+In Rust, this corresponds to mutable access, as declared by the use of `Query<&mut MyComponent>` or `ResMut<MyResource>`.
+
+On the next page, we'll create a simple "game" using the ECS so you can see how this all fits together.
+
+## Rust tips and tricks for Bevy's ECS
+
+If you're new to Rust, there are a few standard tricks and idioms that are used by Bevy's ECS that are worth learning about.
```
Nilirad, line 40: > there are a few standard tricks and idioms that are used by Bevy's ECS that are worth learning about.

I found a couple of potential problems in this sentence:

1. The word _tricks_ is a bit too colloquial and lacks strong meaning. Using _patterns_ instead seems more appropriate.
2. The phrase "that are used by Bevy's ECS" while maybe true, it's probably not what you actually meant. Did you mean something like "that you can adopt to use Bevy's ECS at it's fullest" instead?

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+Feel free to come back to this section later as you learn more about Bevy!
+
+### Tuple structs
+
+When working with tuple structs like `struct Score(u64)` or `struct Position(u8, u8)`, `.0`, `.1` and so on are used to access fields, counting up from 0 in the order they were defined.
```
Nilirad, line 45: It's not clear where the definition of the tuple struct ends and where the field access syntax starts. Separating the two concepts in two separate sentences may be the easiest solution.

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+
+`let foo.0 = bar` can seem opaque when you first read it, but in Rust it just means that you're assigning `bar` to the first field of the tuple struct `foo`.
```
Nilirad, line 47: I would avoid _foo_, _bar_ and similar meaningless identifiers. Names that reflect the role of a thing are much more better IMO. Example:

```
my_tuple.0 = new_value
```

Note that I removed `let` as it should be incorrect.

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+   1. This is commonly done when accessing fields and methods: calling `.title` on `Res<WindowDescriptor>` will grab the title field of `WindowDescriptor`, not the wrapping `Res<WindowDescriptor>` as the `Res` type has no field named `title`.
+3. When your types cannot be automatically dereferenced, you will need to manually dereference out of the outer smart pointer using `*`.
+   1. This is commonly needed when assigning a value to your data (rather than just one of its fields), or when using `match`.
+
+### Initializing structs with some default values
+
+Bevy often requires defining complex structs when creating entities with bundles or configuring the app with resources.
+In many cases, we want to use *some* default fields, but not all of them.
+As Rust does not have default arguments, we instead use [struct update syntax](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) in combination with the [`default()`](https://doc.rust-lang.org/std/default/trait.Default.html) method from the `Default` trait to allow us to quickly set some values while leaving others unchanged.
```
Nilirad, line 64:
```suggestion
As Rust does not have default arguments, we instead use [struct update syntax (SUS)](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) in combination with the [`default()`](https://doc.rust-lang.org/std/default/trait.Default.html) method from the `Default` trait to allow us to quickly set some values while leaving others unchanged.
```

It was stronger than me lol. Please don't accept it 😝

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+struct Attack(u32);
+
+let my_enemy = EnemyBundle {
+	attack: 10,
+	defense: 5,
+	// All other fields are initialized from their default value
+	// setting enemy: Enemy and life: Life(100)
+	..Default::default()
+}
+```
```
Nilirad, line 105: The example should just focus on the Struct Update Syntax, not bleed into Bevy features. Here the concept of _bundles_ is added, unnecessarily increasing complexity (especially since bundles are not introduced yet). A pure Rust example would be much more well understood.

Basically, just show the definition of a struct with many fields (not too many or the code explodes in length; 4 fields are enough). Say that the struct implements `Default` (don't show the full impl though, it's too much code and it's irrelevant), and then create an instance of the struct by filling a couple of fields and then use `..Default::default()`

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Type aliases don't play nice
+
+Bey's ECS uses Rust's type system to dispatch data to our systems as requested.
```
Nilirad, line 109:
```suggestion
Bevy's ECS uses Rust's type system to dispatch data to our systems as requested.
```

Typo correction.

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+
+```rust
+use std::any::type_name;
+
+struct Life(u32);
+struct Health(u32);
+
+// Our components have different types now!
+assert!(type_name(Life(42)) != type_name(Health(42));
+```
```
Nilirad, line 140: The use of type ids instead of type names feels more appropriate to me, since that is what really distinguished two different types.

Nilirad (changes requested):
I reviewed the chapter introduction. Mostly suggestions and a couple of change request for typos and other additions/corrections. Great work, anyway 🚀

content/learn/book/welcome/installation/_index.md
```diff
@@ -0,0 +1,118 @@
...
+### Install Rust
+
+Install Rust by following the [Rust Getting Started Guide](https://www.rust-lang.org/learn/get-started).
+
+Once this is done, you should have the ```rustc``` compiler and the ```cargo``` build system installed in your path.
+
+### Install OS dependencies
+
+* [Linux](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md)
+* Windows: Make sure to install [VS2019 build tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
```
MinerSebas, line 23: Now that we use `shaderc` instead of `bevy-glsl-to-spirv` you could also build for `x86_64-pc-windows-gnu` which doesn't require the `VS2019 build tools` unlike `x86_64-pc-windows-msvc`

Nilirad:
@MinerSebas I think this PR involves more files than it should, probably because @alice-i-cecile included the commit "Book structure" (252325d)

MinerSebas:
I think the problem is more that this targets the `master` branch, and not the `new-book` branch which already includes 252325d

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -0,0 +1,252 @@
...
+	// Each field of this bundle corresponds to a type of component 
+	// that will be inserted into the final entity
+	// The field names don't matter: only the types are retained in our ECS storage
+    life: Life,
+    attack: Attack,
+    defense: Defense,
+}
+```
+
+Now, let's get to work by spawning a player entity and an enemy entity for them to fight!
```
Nilirad, line 44: I think that this section could end here. After this point there is a lot of talk involving systems, that should not belong in the "entities have components" section. I think we shouldn't create the whole game in one single section. We can instead introduce each concept as soon as it's needed during the development.

First, I would remove the player input from the example (basically it becomes a zero player game), and then have the next sections as the following:
1. _Commands._ Introduce commands, like in the already existing section, then use the acquired knowledge to spawn the player and the enemy in a startup system.
2. _Queries._ Introduce queries. Show an example of how a query looks in the context of a system parameter, then link to the API docs for the full usage. Then add the update systems. At this point, the `NextAttack` resource doesn't exist yet, so make the player relentlessly attack the enemy until their hp goes to 0. 
3. _Resources._ Introduce resources. Add the `NextAttack` resource, and make the poor enemy defend themselves.

For each section, after enriching the example, I would continue, when appropriate, by talking about less common use cases of the concept, connecting with previous topics and so on.

I left out the dedicated section for systems. I don't know what we should do with that: systems pervade everything, and making an example starting with systems will inevitably bleed into other areas, like commands, queries and resources. On the other hand, a prior discussion of systems is needed to explain their role in relation to the `World` (the fact that they can read and mutate data) and to the `Schedule` (they can be set to run once or for each iteration of the game loop). Therefore, I think that introducing systems before the example could be the way to go, without obviously mentioning queries, commands or resources.

Nilirad (commented):
I found some critical issues about how some concepts are introduced, affecting the structure of the whole chapter. I would prefer to get this sorted out before reviewing the rest.

I also have some mixed feeling about adding such complex examples throughout each section, as they dilute the explanation making it very easy to lose the thread, which is time consuming and frustrating, especially to newcomers.

I suggest instead to adopt a simple example and develop it section after section.

Nilirad:
> 
> 
> I think the problem is more that this targets the `master` branch, and not the `new-book` branch which already includes [252325d](https://github.com/bevyengine/bevy-website/commit/252325d64d2c108e4d30db2a6c73c07d3b3003a9)

So changing the base branch would fix this?

content/learn/book/ecs/_index.md
ickk: Since `default` is the *only* required method of the `Default` trait, this seems needlessly wordy. My suggestion would be something along the lines of:

> ".. we instead use struct update syntax in combination with [`Default::default()`](https://doc.rust-lang.org/std/default/trait.Default.html), which allows us to succinctly specify only the values we are interested in."

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+`Life`, `Attack` and `Defense` will almost always be added to our entities at the same time, so let's create a **bundle** (collection of components) to make them easier to work with.
+
+```rust
+#[derive(Bundle)]
+struct CombatBundle {
+	// Each field of this bundle corresponds to a type of component 
+	// that will be inserted into the final entity
+	// The field names don't matter: only the types are retained in our ECS storage
```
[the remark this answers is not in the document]
ickk: or alternatively, reworded to clarify

> "The field names are just used when instantiating the bundle, only the types are stored in the ECS"

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+We can use this to specialize behavior between entities: only performing work on them in our systems if they have the correct combination of components.
+You don't want to apply gravity to entities without a position in your world, and you're only interested in using the UI layout algorithm to control the layout of UI entities!
+
+When we want to go beyond this tabular data storage, we can use **resources**: global singletons which store data in monolithic blobs.
+You might use resources to store one-off bits of state like the game's score, use it to interface with other libraries, or store secondary data structures like indexes to augment your use of entity-component data.
+Just like with components on entities, resources are accessed by type and you can only have one resource of each Rust type.
+
+In order to actually perform logic on all of this data, we must use systems, which are automatically run by our scheduler.
+**Systems** are Rust functions which request specific data from the world, as declared in their system parameters (function arguments): generally resources and entities that have a particular combination of components using queries.
```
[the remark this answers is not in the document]
ickk: - I agree that "using queries" is weird here. However, I think that whole elaboration of *what* data systems might request could be left out, since..

- I think this explanation doesn't really give the reader an idea of what the *purpose* of a system is. 

  As a reader, why do I care about Systems? What does this feature let me do?
  
  My suggestion would be to mention the purpose a bit more: perhaps something like:
  
  > **Systems** are Rust functions that request data from the World, in order to perform tasks - often a system is a procedure that specifies how entities may interact. They prescribe all of the rules and behaviours of the World.

MinerSebas:
> 
> 
> > I think the problem is more that this targets the `master` branch, and not the `new-book` branch which already includes [252325d](https://github.com/bevyengine/bevy-website/commit/252325d64d2c108e4d30db2a6c73c07d3b3003a9)
> 
> So changing the base branch would fix this?

Yes it should. I leave the Github documentation here, because this feature is pretty hidden: https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

alice-i-cecile:
Awesome feedback! Let's get the branch fixed up first...

alice-i-cecile:
> I also have some mixed feeling about adding such complex examples throughout each section, as they dilute the explanation making it very easy to lose the thread, which is time consuming and frustrating, especially to newcomers.

Yeah, I'm not sure how I feel about complex examples. The entities-components example definitely needs to be cut in favor of something simpler. At the same time, I want to have complete, runnable snippets that actually demonstrate the features.

There's a lot of ground to cover, and just talking about things in the abstract is not going to do a lot of good for beginners. I also very often find myself trying to figure out "oh what's the exact syntax / snippet for this again...", and looking at resources like this to find a complete working example I can hack on.

I'm not a fan of the running example, since it's hard to avoid shoe-horning features in, and the book is not always going to be read in a linear order. Originally I'd proposed end-of-chapter examples, but those were cut due to perceived effort required (per @cart's request). I think simplify + link to existing Bevy examples for more depth might be the way to go?

alice-i-cecile:
> I think that this section could end here. After this point there is a lot of talk involving systems, that should not belong in the "entities have components" section. I think we shouldn't create the whole game in one single section. We can instead introduce each concept as soon as it's needed during the development.
>
> First, I would remove the player input from the example (basically it becomes a zero player game), and then have the next sections as the following:
> 
> Commands. Introduce commands, like in the already existing section, then use the acquired knowledge to spawn the player and the enemy in a startup system.
> Queries. Introduce queries. Show an example of how a query looks in the context of a system parameter, then link to the API docs for the full usage. Then add the update systems. At this point, the NextAttack resource doesn't exist yet, so make the player relentlessly attack the enemy until their hp goes to 0.
> Resources. Introduce resources. Add the NextAttack resource, and make the poor enemy defend themselves.
> For each section, after enriching the example, I would continue, when appropriate, by talking about less common use cases of the concept, connecting with previous topics and so on.
> 
> I left out the dedicated section for systems. I don't know what we should do with that: systems pervade everything, and making an example starting with systems will inevitably bleed into other areas, like commands, queries and resources. On the other hand, a prior discussion of systems is needed to explain their role in relation to the World (the fact that they can read and mutate data) and to the Schedule (they can be set to run once or for each iteration of the game loop). Therefore, I think that introducing systems before the example could be the way to go, without obviously mentioning queries, commands or resources.

- @Nilirad

First off, this particular example needs to be *massively* reduced and simplified. I'll do that ASAP and then we can see how it looks.

IMO the proposed style of running example is *much* better handled by the quick start guides: that's the learning style they're designed for. The book itself is intended to be more comprehensive and academic, getting into the how and why on a deeper level and serving as a reference to refer back to later.

The entangled nature of many of these topics is rough for strictly-in-order topics: I'd prefer to lean more heavily on the chapter introduction to mention topics and preserve the value of each page as more in-depth reference material. We can't introduce entities without a way to spawn them, and talking about commands immediately is worthless for example. Instead, I'd like to lean on the high level introduction to briefly define concepts and not worry aggressively about out-of-order concepts.

We'll do our best on that of course, but "beginner who reads the entire book cover-to-cover in order" is not our only (or likely most common) audience.

content/learn/book/ecs/_index.md
```diff
@@ -0,0 +1,140 @@
...
+* the primary key of this database is the [`Entity`](https://docs.rs/bevy/latest/bevy/ecs/entity/struct.Entity.html) identifier, which can use to look up specific entities using [`Query::get(my_entity)`](https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.Query.html#method.get)
+
+Of course, this database is not very well-normalized: not all entities will have every component!
+We can use this to specialize behavior between entities: only performing work on them in our systems if they have the correct combination of components.
+You don't want to apply gravity to entities without a position in your world, and you're only interested in using the UI layout algorithm to control the layout of UI entities!
+
+When we want to go beyond this tabular data storage, we can use **resources**: global singletons which store data in monolithic blobs.
+You might use resources to store one-off bits of state like the game's score, use it to interface with other libraries, or store secondary data structures like indexes to augment your use of entity-component data.
+Just like with components on entities, resources are accessed by type and you can only have one resource of each Rust type.
```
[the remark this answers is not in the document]
alice-i-cecile: I think this whole line is a poor fit here. The point is to introduce the concepts, not flood the reader with details.

content/learn/book/ecs/_index.md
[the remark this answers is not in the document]
alice-i-cecile: Adapted your suggestion <3 I agree that there needed to be more *intent* present.

> Systems are Rust functions that request data from the `World` (such as resources or entities with particular components) in order to perform tasks.
> All of the rules and behaviours of our game are governed by systems.

content/learn/book/ecs/_index.md
alice-i-cecile: > If you're new to Rust, there are a few standard patterns and pitfalls that are particularly relevant to Bevy's ECS.

content/learn/book/ecs/_index.md
alice-i-cecile: > ".. we instead use struct update syntax in combination with Default::default(), which allows us to succinctly specify only the values we are interested in."

Stealing this bit :)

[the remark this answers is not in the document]
alice-i-cecile: I actually don't think this is grammatically correct. Explaining precisely *why* as a native English speaker is hard though 😅

content/learn/book/ecs/_index.md
alice-i-cecile: Agreed: this has been simplified and I left a note below to talk about its relevance to bundles.

content/learn/book/ecs/_index.md
alice-i-cecile: Absolutely. Changing.

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -0,0 +1,85 @@
...
+
+Due to the delayed effect of commands, and their relatively poor performance (they can only be executed one at a time in sequence), you should only use custom commands for tasks that truly need their world-altering powers.
+In many cases, an event plus an event-handling system will be faster, more ergonomic and easier to debug.
+
+## `NonSend` resources
+
+Non-send resources are resources that lack the `Send + Sync` trait bounds: they cannot be sent safely across threads.
+`NonSend` resources will typically be quite advanced: used for things like networking or interfacing with external libraries.
+They are accessed using `NonSend<R>` and `NonSendMut<R>` in a way that is directly analogous to `Res` and `ResMut`.
+The only difference is that they can only be accessed when we have exclusive `World` access, as accessing them can only be done on the main thread.
```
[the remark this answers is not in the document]
MinerSebas: For a concrete case, you can look at cart/bevy#10 and cart/bevy#11.
`#10` solved the Problem by making the system exclusive, but was replaced by `#11` which doesn't require the exclusive access to the world.

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -0,0 +1,252 @@
...
+On this page, we're going to create a simple little combat system, so lets start by defining some basic components.
+
+```rust
+// These are dataless "unit structs", which hold no data of their own
+// In Bevy, these are useful for distinguishing similar entities or toggling behavior
+// and are called "marker components"
+struct Player;
+struct Enemy;
+
+// These simple components wrap an i8 in a tuple struct
```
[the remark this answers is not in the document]
alice-i-cecile: Stating the obvious is the point here :) Many of our users are new to Rust (or even programming!), and practicing critical vocabulary (components, wrap, i8, tuple struct) through demonstrations is a nice way to learn.

content/learn/book/welcome/installation/_index.md
alice-i-cecile: This should be in a separate issue <3. That section was already merged; I just targeted the wrong branch.

content/learn/book/ecs/systems-queries/_index.md
```diff
@@ -0,0 +1,147 @@
...
+To get around this, we can use [`Query::single()`](https://docs.rs/bevy/0.5.0/bevy/ecs/system/struct.Query.html#method.single) and [`Query::single_mut()`](https://docs.rs/bevy/0.5.0/bevy/ecs/system/struct.Query.html#method.single_mut), depending on whether or not we need to mutate the returned data.
+
+Note that these functions return a `Result`: if you expect this could fail in real scenarios (in case the query does not contain exactly one entity), handle the result properly.
+Otherwise, just call `let (component_a, component_b) = query.single().unwrap()` to make use of the data quickly.
+
+## Looking up specific entities
+
+Each entity in our ECS data storages has a unique identifier, given by its [`Entity`](https://docs.rs/bevy/0.5.0/bevy/ecs/entity/struct.Entity.html), which defines the entity in terms of a `u32` [`id`](https://docs.rs/bevy/0.5.0/bevy/ecs/entity/struct.Entity.html#method.id) and a `u32` [`generation`](https://docs.rs/bevy/0.5.0/bevy/ecs/entity/struct.Entity.html#method.generation).
+We can fetch the `Entity` of each entity returned by our queries by including it in the first type parameter of `Query`:
```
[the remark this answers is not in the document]
alice-i-cecile: The semantics aren't quite adequate there: you can't put it in the filtering position. I'll clarify the wording though.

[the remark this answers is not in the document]
MinerSebas: Mention that its convention to put the `Entity` first.
```suggestion
We can fetch the `Entity` of each entity returned by our queries by including it as a type parameter of `Query` (By convention it is placed as the first parameter):
```

ickk:
> > > I think the problem is more that this targets the `master` branch, and not the `new-book` branch which already includes [252325d](https://github.com/bevyengine/bevy-website/commit/252325d64d2c108e4d30db2a6c73c07d3b3003a9)
> > 
> > 
> > So changing the base branch would fix this?
> 
> Yes it should. I leave the Github documentation here, because this feature is pretty hidden: https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

Out of curiosity, is this functionally the same as just a rebase on `new-book` follow by a force push?

MinerSebas:
> 
> 
> > > > I think the problem is more that this targets the `master` branch, and not the `new-book` branch which already includes [252325d](https://github.com/bevyengine/bevy-website/commit/252325d64d2c108e4d30db2a6c73c07d3b3003a9)
> > > 
> > > 
> > > So changing the base branch would fix this?
> > 
> > 
> > Yes it should. I leave the Github documentation here, because this feature is pretty hidden: https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request
> 
> Out of curiosity, is this functionally the same as just a rebase on `new-book` follow by a force push?

No, it wouldn't. Let me show it with some commit trees
Here Main represents the Main branch, Book the `new-book` branch with  252325d and PR this PR.

```
Main
    \
    Book
       \ 
         PR
```

Originally this PR tried to merge all changes into Main, but because Main didn't have 252325d  yet it was also shown in difff.

```
Main -------- Merge
    \         /
    Book     /
       \    /
         PR
```

The change was which branch was targeted. 
By targeting `new-book` the diff no longer shows changes that 252325d did, because they are already present in `new-book` 

```
Main
    \
    Book --- Merge
       \    /
         PR
```

But by the same token, rebasing onto `new-book` wouldn't have changed anything, because 1.) the target/base branch would have still been Main which doesn't have 252325d  and 2.) the rebase wouldnt have removed 252325d

Nilirad:
@alice-i-cecile 

> I want to have complete, runnable snippets that actually demonstrate the features.

The best solution that comes to my mind would be to have some sort of private playground. Basically the code shown in the book is the most relevant subset of the complete code example, with boilerplate completely stripped out. Then there's an "expand" button to immediately show the entire code, and a "play" button to run the code.
Obviously this would be a prohibitively high cost solution for us at this point, since we need to get the book up and running in a couple of months. As a simple alternative, I thought about showing the stripped code on the book and linking the full example to a gist. However I don't know if this is feasible since we want to get to import validated snippets, unless it's possible to hide lines during the import...
Anyway, things don't have to be perfect, I'm just concerned about having a screen completely covered by code to show a simple concept.

>IMO the proposed style of running example is much better handled by the quick start guides

Yeah, you convinced me, I also thought about quick start guides at a certain point. I think that the examples can be simplified a lot. To make the examples as practical and lightweight as possible, I thought about a couple of things:

- **Using `MinimalPlugins` or `ScheduleRunnerPlugin`.** Differently from `DefaultPlugins`, they give the ability to control the frequency of the game loop. This will make the user understand better what's going on.
- **Stripping out unnecessary features.** Excluding from the examples things like user input and assets will drastically reduce its cognitive load.
- **Simpler problem domain.** If the idea of the example is simple, it's implementation will also be simple. Less systems, less components, less entities.

alice-i-cecile:
> Using MinimalPlugins or ScheduleRunnerPlugin. Differently from DefaultPlugins, they give the ability to control the frequency of the game loop. This will make the user understand better what's going on.

This is a good call. I may also poke around and see if I can make a `CommandLinePlugins`, to create a custom runner designed for terminal games. This would be fantastic for examples, allowing us to gloss over the hard bits of making things work (and would be handy for quite a few games / applications as well).

Nilirad:
> I may also poke around and see if I can make a CommandLinePlugins, to create a custom runner designed for terminal games.

There is the `bevy_crossterm` plugin that does a similar job. It's outdated (supports Bevy 0.4) and maybe a bit too complex for our purposes, but it may be a good source of inspiration.

content/learn/book/welcome/installation/_index.md
MinerSebas: Created #183

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+```rust
+use std::any::type_name;
+
+// Our first component type
+struct Life(u32);
+
+// An attempt at a second component type, using a type alias
+type Health = Life;
+
+// Unfortunately, these two types share the same name :(
```
Nilirad, line 113:
```suggestion
// Unfortunately, these two types share the same id :(
```

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+	defense: 50,
+	..Default::default()
+}
+```
+
+This pattern is commonly used when working with bundles: groups of components typically used for entity initialization.
+
+### Type aliases don't play nice
+
+Bey's ECS uses Rust's type system to dispatch data to our systems as requested.
```
Nilirad, line 99:
```suggestion
Bevy's ECS uses Rust's type system to dispatch data to our systems as requested.
```

You forgot to commit this typo fix.

Nilirad (changes requested):
A couple of fixes in chapter intro.

Nilirad:
I thought about a potential chapter structure that could expose the topics in a more relaxed manner. I think that we can actually do that if we don't hurry about showing complete code examples too early. The secret is not strictly following the structure of the API, that, while being optimized for *usage*, it does not care about laying out concepts.

I sorted out the main ECS concepts in my mind (and on paper) and I came out with a peculiar structure, that while I don't expect it to be adopted as-is, can give some insight.

Basically, I subdivided the chapter in two cognitive sections, a theoric one and a practical one, with a soft boundary between the two. The first, theoric section, talks about the ecs world, where entities and components are stored inside. Then, in the second, mostly practical section, we talk about systems and the tools to govern them, with simple but representative examples.

- **The ECS world (static viewpoint of ECS).** The world defines what exists in a Bevy application. Let's see how it does that.
    - **Entities and components.** What are they? How do they relate to each other? Why is this kind of data organization so special? What does it have that a classical OOP approach doesn't?
    - **Resources.** What are they? Why do they exist when we already have entities? What's are the differences from an entity? When should they be preferred over an entity?
- **The ECS schedule (dynamic viewpoint of ECS).** The world is not static. The schedule makes stuff happen by reading from and writing data to the world, through systems.
    - **Systems and parameters.** What are systems needed for in an ECS application? How can we read from and write to the world? Using system parameters. What are they? What system parameters can we use? How do we set up a world? Startup systems. How do we make the world look alive? Normal/update systems (we may add a couple of simple examples to show how those system _looks_, without focusing the attention to the specific parameters that are not introduced yet). Now let's see how some common system parameters work.
    - **Commands.** Why do they exist? Here we start writing our first system, where we spawn and build our first entity (the tedious, component-wise way).
    - **Component bundles.** Introduce bundles. Show how it's easier to build an entity now. This *may* be included in the same section of commands, since it's not too much content and they are often used in startup systems to spawn entities.
    - **Resource fetching.** While resources have already been introduced, here they are shown in the context of system parameters. Also resource insertion during app setup should be shown, since in the theoric section code examples are very minimal if not absent.
    - **Queries, filtering and component change detection.** Queries are often used in normal/update systems to make stuff happen. The current text for the section(s) is already good, maybe some tweaks for integration would be needed at max. As there is much content here, splitting it in two or three sections may be a good idea (as it's already done).
    - **Generic systems and exclusive world access.** These two sections are a bit like an appendix, since they are not as much needed as the above, but they may still be good to have anyway. Actually, generic systems are not a proper Bevy ECS feature, but maybe it could be beneficial to show some examples, I can't know for sure.

[This resource](https://www.youtube.com/watch?v=aKLntZcp27M) may be a good source of inspiration for the OOP vs ECS approaches to game development, if you didn't see it already.

content/learn/book/ecs/filtering-queries/_index.md
```diff
@@ -0,0 +1,75 @@
...
+title = "Filtering queries"
+weight = 5
+template = "book-section.html"
+page_template = "book-section.html"
++++
+
+When components are fetched in queries, the data of *every* component in the first type argument passed will be fetched and made available to the system.
+However, this isn't always what you want!
+In many cases, you just want to filter the query based on the presence (or absence) of a component, and don't want to deal with unpacking data you're never going to use.
+This is particularly true when working with **marker components**: data-less structs designed to convey the identity or current state of an entity.
```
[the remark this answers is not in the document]
alice-i-cecile: Right; this is useful to discuss again here.

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -5,6 +5,68 @@ template = "book-section.html"
...
+struct Life(u8);
+struct Attack(u8);
+struct Defense(u8);
+
+// We can store arbitrary data in our components, as long as it has a 'static lifetime
```
Havvy, line 26: People get confused all the time that types without lifetimes are `'static`, so I fear this comment is going to confuse those people.

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+Feel free to come back to this section later as you learn more about Bevy!
+
+### Tuple structs
+
+Tuple structs, defined like `struct Score(u64);` or `struct Position(u8, u8);`, are structs with unnamed fields, accessed by the order in which they were defined.
+`.0`, `.1` and so on are used to access fields, counting up from 0 in the order they were declared.
+
+`my_position.0 = 42` can seem opaque when you first read it, but in Rust it just means that you're assigning `42` to the first field of the tuple struct `my_position` (by convention, the x coordinate).
+Using ordinary structs with named fields rather than tuple structs (even for simple components and resources!) can go a long way to improving the clarity of your code by providing more descriptive field names.
+In this example, `my_position.x = 42` immediately communicates intent, rather than relying on the end user to remember the convention used.
```
TypicalFork, line 51:
```suggestion
In this example, `my_position.x = 42` immediately communicates intent, rather than relying on future you to remember the convention used.
```
Minor nitpick: the end-user of a game usually doesn't look at the code.

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+
+### Smart pointers
+
+There are a few points to be aware of when working with "smart pointers" like components and resources:
+
+1. The value returned in a system when you ask for `Res<Score>` is not a plain `Score`, it is a `Res<Score>`. This is a resource that stores the score, similar to an ordinary `&` pointer, but with some special behavior (like change detection).
```
TypicalFork, line 57:
```suggestion
1. The value passed to a system when you ask for `Res<Score>` is not a plain `Score`: it is a `Res<Score>`. This is a resource that stores the score, similar to an ordinary `&` pointer, but with some special behavior (like change detection).
```

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+2. Because `Res` and other smart pointers implement the `Deref` trait, Rust will attempt to automatically dereference it to match required types when it can.
+   1. This is commonly done when accessing fields and methods: calling `.title` on `Res<WindowDescriptor>` will grab the title field of `WindowDescriptor`, not the wrapping `Res<WindowDescriptor>` as the `Res` type has no field named `title`.
+3. When your types cannot be automatically dereferenced, you will need to manually dereference out of the outer smart pointer using `*`.
```
TypicalFork, line 60:
```suggestion
3. When your types cannot be automatically dereferenced, you will need to manually dereference the outer smart pointer (i.e. `Res`) using `*`.
```
A small clarification to make it very clear.

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+   1. This is commonly needed when assigning a value to your data (rather than just one of its fields), or when using `match`.
```
TypicalFork, line 61:
```suggestion
   1. This is commonly needed when assigning a value to your data (rather than just one of its fields), or using `match`.
```

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+	defense: 50,
+	..Default::default()
+}
+```
+
+This pattern is commonly used when working with bundles: groups of components typically used for entity initialization.
+
+### Type aliases don't play nice
+
+Bevy's ECS uses Rust's type system to dispatch data to our systems as requested.
```
TypicalFork, line 99:
```suggestion
Bevy's ECS uses Rust's type system to dispatch data to our systems as required.
```

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+This is very convenient for safety and ergonomic reasons, but means that only one resource of each Rust type can exist at once, and only one component of each type can be stored on each entity.
```
TypicalFork, line 100:
```suggestion
This is very convenient for safety and ergonomic reasons but means that no two resources can have the same Rust type, and entities can, at most, only have one instance of each type of component.
```
In my opinion, this is clearer, but I might be wrong.

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+
+"One of each Rust type" has somewhat surprising consequences for new Rust users though: [type aliases](https://doc.rust-lang.org/reference/items/type-aliases.html) will not result in unique types, and should generally not be used to define resource or component types.
```
TypicalFork, line 102:
```suggestion
No two resources having the same Rust type has somewhat surprising consequences for new Rust users: [type aliases](https://doc.rust-lang.org/reference/items/type-aliases.html) will not result in unique types, and should generally not be used to define resource or component types.
```

TypicalFork (changes requested):
I've left some feedback on the introduction. Some of it is quite minor and potentially subjective.

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,125 @@ template = "book-section.html"
...
+
+Of course, this database is not very well-normalized: not all entities will have every component!
+We can use this fact to specialize behavior between entities: systems only perform work on entities with the correct combination of components.
+You don't want to apply gravity to entities without a position in your world, and you're only interested in using the UI layout algorithm to control the layout of UI entities!
+
+When we want to go beyond this tabular data storage, we can use **resources**: global singletons which store data in monolithic blobs.
+You might use resources to store one-off bits of state like the game's score, use it to interface with other libraries, or store secondary data structures like indexes to augment your use of entity-component data.
+
+In order to actually manipulate all of this data in interesting ways, we must use systems.
+**Systems** are Rust functions which request specific data from the world, as declared in their system parameters (function parameters): generally resources and entities that have a particular combination of components using queries.
```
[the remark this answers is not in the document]
TypicalFork: I tried to make it clearer but I may have lost some information.

content/learn/book/ecs/_index.md
asleeponmykeyboard:
```suggestion
In this example, `my_position.x = 42` immediately communicates intent, rather than relying on future developers to remember the convention used.
```
I think it should be 'future developers' because more than 1 person may be working on the code.

alice-i-cecile:
So https://github.com/bevyengine/bevy-website/pull/182/commits/4ed2ca5bcc7b8d17699f5cfcd7e84cd38ee3cbd1 adds a more involved example to the start of the chapter. There are two broad areas I want feedback on:

**1. Is it correct to include a high-level, minimal explanation example at this point in the book?**

My core argument here is "everything is too interwoven: there is no single linear order that works for learning the ECS". Components and entities are unusable without basic commands, but you can't get into how commands *actually* work without a huge amount of background. As I'd like to prioritize in-depth explanations of the details, to make the book more useful to convey subtleties and tricks, I want to keep e.g. "commands 101" and "custom commands" on the same page, even though the latter will not be useful until you're an advanced user.

By throwing an interwoven example in (and eventually, pointing the users to the quickstart guides) to supplement the high-level explanation, we can give the reader a vocabulary and basic gloss of the architecture used so they can follow along on their first read of the rest of the sub-chapters.

**2. How is this particular example? Should it be replaced by something else? Concrete suggestions are useful.**

I'm not thrilled with several things:
1. The player / enemy entity access pattern is not very pretty, although it *does* demonstrate the `query.get` API which is often seriously overlooked by newcomers.
2. The example won't compile until someone (probably me...) adds an API for https://github.com/bevyengine/bevy/issues/2276 and it's merged in. This can be solved with marker components in place of a side enum, but then you get conflicting queries and have to do the dumb `(With<Player>, Without<Enemy>>)` thing.
3. The complexity is a higher than I'd *really* prefer for a first real example. Some trickiness with run criteria, change detection, random access to component data, fancy match statements, system ordering...
4. We don't make use of the window created by `DefaultPlugins`. This could be solved by a better plugin set. I decided not to do input handling or graphics to manage complexity but it's not a very fun or impressive game at all. Maybe we can swallow the complexity here by swapping to a much simpler logic game?

mirkoRainer (commented):
> **1. Is it correct to include a high-level, minimal explanation example at this point in the book?**
I don't see an issue with this. New users may gloss over it, but intermediate users might read the example to understand "what's to come". Agree with you intent here.

> **2. How is this particular example? Should it be replaced by something else? Concrete suggestions are useful.**
> 
> I'm not thrilled with several things:
> 
> 1. The player / enemy entity access pattern is not very pretty, although it _does_ demonstrate the `query.get` API which is often seriously overlooked by newcomers.
Can confirm. Newcomer that learned after asking in Discord.

> 2. The example won't compile until someone (probably me...) adds an API for [Query::get_multiple for accessing multiple entities simultaneously from the same query bevy#2276](https://github.com/bevyengine/bevy/issues/2276) and it's merged in. This can be solved with marker components in place of a side enum, but then you get conflicting queries and have to do the dumb `(With<Player>, Without<Enemy>>)` thing.
Is this intended to release with 0.6? The code **definitely** needs to compile so may need another pass just before release to ensure all the APIs exist, etc. 
I don't quite understand the dumb thing explained here.

> 3. The complexity is a higher than I'd _really_ prefer for a first real example. Some trickiness with run criteria, change detection, random access to component data, fancy match statements, system ordering...
Perhaps there should be an expectation set here? Are there more in-depth articles/sections that could be linked out to regarding these topics? If so, I'd say providing the link in the examples would good enough for advance users wanting to understand more in-depth. But also new users can just ignore the links to get the game running.

> 4. We don't make use of the window created by `DefaultPlugins`. This could be solved by a better plugin set. I decided not to do input handling or graphics to manage complexity but it's not a very fun or impressive game at all. Maybe we can swallow the complexity here by swapping to a much simpler logic game?
On the note of fun. This game is not engaging, because there is not input from the user.
I haven't looked into this myself, but perhaps the user could be prompted with input to "roll" for the Player stats which are then randomly generated within a range. Likewise, the Enemy could be randomly generated as well. This may add unnecessary complexity, but it would definitely be a more engaging example.

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -5,6 +5,68 @@ template = "book-section.html"
...
+struct Name(String);
+```
+
+`Life`, `Attack` and `Defense` will almost always be added to our entities at the same time, so let's create a **bundle** (collection of components) to make them easier to work with.
+
+```rust
+#[derive(Bundle)]
+struct CombatBundle {
+	// Each field of this bundle corresponds to a type of component
```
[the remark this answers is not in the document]
alice-i-cecile: Lack of nice cargo format :( This will be resolved once we migrate these code blocks into their own, compiling .rs files.

Nilirad:
> Is it correct to include a high-level, minimal explanation example at this point in the book?

Yes, that's the point of an introduction. The only things that could be changed are the following. I think the last paragraph of the ECS overview can be entirely omitted. Even the second-last paragraph can be removed, except the first sentence, that can be merged to the previous paragraph. Those are not needed for a high-level understanding. One is an advanced API and the other is merely a technical constraint.



## About the ECS Example

> How is this particular example? Should it be replaced by something else? Concrete suggestions are useful.

These are my thoughts on how it could be improved and what may look off:

1. I suggest changing system function names to include the `_system` suffix, as convention in the Bevy source code and previous documentation.
2. Declare resources and components earlier, before the main function. In this way, readers don't get confused too much when they read the main function. I think that systems can still come after `main` though.
3. The `Side` component needs a comment showing its role inside the game.
4. I think entity indexing through `CombatantEntities` can be avoided. We can still talk about `Query::{get, get_mut}` in the queries section. Query filtering based on `Side` should do the job.
5. I don't think introducing bundles here has a high benefit/cost ratio. Inserting components one by one is not too much tedious (it is actually shorter in this case), and we can still introduce bundles later. Actually the real power of bundles come from the struct update syntax, where it can significantly reduce the number of lines. But not here, where it just increase them instead.
6. `p_` and `e_` prefixes for "player" and "enemy" in `combat` makes everything a bit difficult to understand despite making things more compact.
7. `check_victory` system is not immediately understandable. At first sight it even seems that the character whose life goes to 0 wins.
8. The lack of a definitive game over condition looks a bit underwhelming. Despite the `if` statement in `combat`, systems continue to run, and `Victory::Concluded` can give a wrong value before settling to the correct one.

### Alternative example

I still think it is difficult to simplify the combat example under a non overwhelming level. Probably some useful `bevy_ecs` features are still not present or the domain is more complex than we thought (probably both are factors). So I propose a simpler example, that I prepared below, featuring a single entity and no resources. The problem domain is an entity falling by some sort of "gravity" towards y = 0, and receiving fall damage. I made two versions, where the simpler one may be around 100 LOC if we add more comments. If a resource is *really* needed, i thought about an achievement that gets unlocked if a fall doesn't kill the entity but it leaves it with 1 `Life` point or less.

<details>
<summary>More complex version</summary>

```rust
use bevy::app::AppExit;
use bevy::log::LogPlugin;
use bevy::prelude::*;

// This component defines our entity's life.
struct Life(f32);

// This component is used to mark if our entity is currently airborne.
struct Falling {
    // The higher the initial height of falling, the higher the damage.
    height: f32,
}

fn main() {
    App::new()
        .add_plugins(MinimalPlugins)
        .add_plugin(LogPlugin)
        .add_startup_system(spawn_player_system)
        .add_system_set(
            SystemSet::new()
                .label("logic")
                .with_system(gravity_system.label("gravity"))
                .with_system(fall_damage_system.after("gravity")),
        )
        .add_system_set(
            SystemSet::new()
                .label("report")
                .after("logic")
                .with_system(report_life_system)
                .with_system(report_height_system),
        )
        .add_system(game_over_system.after("report"))
        .run();
}

// This system spawns the player at a fairly high elevation.
fn spawn_player_system(mut commands: Commands) {
    const INITIAL_HEIGHT: f32 = 15.0;

    commands
        .spawn()
        .insert(Life(20.0))
        .insert(Transform::from_translation(Vec3::new(
            0.0,
            INITIAL_HEIGHT,
            0.0,
        )))
        .insert(Falling {
            height: INITIAL_HEIGHT,
        });
}

// This system pulls down the entity towards the ground (at y = 0), at a constant velocity,
// only while it's falling.
fn gravity_system(mut query: Query<&mut Transform, With<Falling>>) {
    // We use query.single since we are sure the query will match only one entity here.
    if let Ok(mut transform) = query.single_mut() {
        transform.translation.y = (transform.translation.y - 1.0).max(0.0);
    }
}

fn fall_damage_system(
    mut commands: Commands,
    mut query: Query<(Entity, &mut Life, &Falling, &mut Transform), With<Falling>>,
) {
    if let Ok((entity, mut life, falling, mut transform)) = query.single_mut() {
        // Our entity has touched the ground
        if transform.translation.y <= 0.0 {
            transform.translation.y = 0.0;
            commands.entity(entity).remove::<Falling>();

            let damage = (falling.height - 3.0).max(0.0);
            life.0 = (life.0 - damage).max(0.0);
            info!("Damage: {}", damage);
        }
    }
}

fn report_life_system(query: Query<&Life, Changed<Life>>) {
    if let Ok(life) = query.single() {
        info!("Life: {}", life.0);
    }
}

fn report_height_system(query: Query<&Transform, Changed<Transform>>) {
    if let Ok(transform) = query.single() {
        info!("Height: {}", transform.translation.y);
    }
}

fn game_over_system(query: Query<&Life, Changed<Life>>, mut exit: EventWriter<AppExit>) {
    if let Ok(life) = query.single() {
        if life.0 <= 0.0 {
            info!("GAME OVER!");
            exit.send(AppExit);
        }
    }
}
```

</details>

<details>
<summary>Simpler version (only reports damage dealt)</summary>

```rust
use bevy::app::AppExit;
use bevy::log::LogPlugin;
use bevy::prelude::*;

// This component defines our entity's life.
struct Life(f32);

// This component is used to mark if our entity is currently airborne.
struct Falling {
    // The higher the initial height of falling, the higher the damage.
    height: f32,
}

fn main() {
    App::new()
        .add_plugins(MinimalPlugins)
        .add_plugin(LogPlugin)
        .add_startup_system(spawn_player_system)
        .add_system(gravity_system.label("gravity"))
        .add_system(fall_damage_system.after("gravity"))
        .run();
}

// This system spawns the player at a fairly high elevation.
fn spawn_player_system(mut commands: Commands) {
    const INITIAL_HEIGHT: f32 = 15.0;

    commands
        .spawn()
        .insert(Life(20.0))
        .insert(Transform::from_translation(Vec3::new(
            0.0,
            INITIAL_HEIGHT,
            0.0,
        )))
        .insert(Falling {
            height: INITIAL_HEIGHT,
        });
}

// This system pulls down the entity towards the ground (at y = 0), at a constant velocity,
// only while it's falling.
fn gravity_system(mut query: Query<&mut Transform, With<Falling>>) {
    // We use query.single since we are sure the query will match only one entity here.
    if let Ok(mut transform) = query.single_mut() {
        transform.translation.y = (transform.translation.y - 1.0).max(0.0);
    }
}

fn fall_damage_system(
    mut commands: Commands,
    mut query: Query<(Entity, &mut Life, &Falling, &mut Transform), With<Falling>>,
    mut exit: EventWriter<AppExit>,
) {
    if let Ok((entity, mut life, falling, mut transform)) = query.single_mut() {
        // Our entity has touched the ground
        if transform.translation.y <= 0.0 {
            transform.translation.y = 0.0;
            commands.entity(entity).remove::<Falling>();

            let damage = (falling.height - 3.0).max(0.0);
            life.0 = (life.0 - damage).max(0.0);
            info!("Damage: {}", damage);
            exit.send(AppExit);
        }
    }
}
```

</details>

alice-i-cecile:
I agree with your assessment: the combat example is not feasible to adequately simplify.

I'm going to replace it with a lightly tweaked version of your simpler example. I'll expand on the comments, and added a second entity. `.single()` is a lovely API, but it's not representative of typical ECS code and should not be the first thing we introduce.

content/learn/book/ecs/systems/_index.md
```diff
@@ -0,0 +1,42 @@
...
+
+In many cases, you don't *want* your systems to run constantly: instead, you may only want to have them run a single time at the beginning to perform some setup.
+To do this, use a **startup system**.
+
+Startup systems run exactly once, before any ordinary systems, and can add them using [`AppBuilder::add_startup_system](https://docs.rs/bevy/latest/bevy/app/struct.AppBuilder.html#method.add_startup_system).
+
+Carefully controlling if and when systems run is one of the most important tools you have for managing the behavior of the game.
+Check out the pages on [system ordering](../../game-logic/system-ordering/_index.md), [run criteria](../../game-logic/run-criteria/_index.md) and [states](../../game-logic/states/_index.md) in the next chapter for more details.
+
+## Generic systems
```
alice-i-cecile, line 34: I couldn't find a good simple example of generic systems that I was happy with. All of the good use cases I've seen for them are particularly complex, and while I *could* invent something contrived I don't want to teach bad habits.

mirkoRainer: Complex systems are complex. 🤷‍♂️

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -1,14 +1,124 @@
...
+            SystemState::new(&mut world);
+
+// Use .get() if you only need read-access for less restrictive ownership constraints
+let (camera_transform_query, mut commands, mut app_exit_event_writer) = system_state.get_mut(&world);
+
+// You can then work with the extracted system parameters however you wish
+app_exit_event_writer.send(AppExit);
+```
+
+### Manually running systems
```
alice-i-cecile, line 65: If https://github.com/bevyengine/bevy/pull/2427 lands in time for 0.6, we should use that here instead.

It also yeets .system for the methods in question!

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -1,14 +1,138 @@
...
+
+### Accessing multiple parts of the `World` simultaneously
+
+When working with non-trivial exclusive `World` logic, you're likely to run into cases where you need mutable access to more than one part of the `World` at once.
+This tends to make the compiler quite unhappy, but by carefully partitioning our data access we can ensure that our code doesn't violate Rust's memory safety rules.
+No `unsafe` needed!
+
+Right now, there are two main tools to do so:
+
+1. [`World::cell`](https://docs.rs/bevy/latest/bevy/ecs/world/struct.World.html#method.cell): Like the [concept of the same name from the Rust standard library](https://doc.rust-lang.org/std/cell/), this enables interior mutability by disabling Rust's compile-time checks for aliased mutability and replacing them run-time checks.
```
alice-i-cecile, line 42: I am not at all confident in this explanation; someone please chime in if you understand / can explain the mechanics better.

IceSentry:
There's a few inconsistencies with using .system(). The first example doesn't use .system() but most other examples after that use .system().

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -1,14 +1,138 @@
...
+In various places, Bevy code works directly with the `World` data, gaining exclusive, blocking access to it and allowing for arbitrary mutations.
+This is very flexible, and can be essential for advanced use cases, but prevents any other systems from running simultaneously and can often be harder to reason about and maintain.
+As a result, you should use only exclusive `World` when you have to.
+You might be working with the `World` if:
 
-TODO: discuss and demonstrate custom commands
+- you're running an exclusive system to access data in unusually broad ways (such as for saving the game or handling networking)
+- you're writing a custom command, to execute logic at the end of the system
+- you're initializing a resource using the `FromWorld` trait
+- you're working with a `NonSend` resource that cannot be sent across threads
```
MinerSebas, line 16: Perhaps call this `NonSend` Values, to avoid confusion with regular systems that use `NonSend`Resources and are not exclusive.
```suggestion
- you're working with `NonSend` values that cannot be sent across threads
```

alice-i-cecile: They're resources in every other respect, and are stored with other resources.

That said, I like this change as there's no blockers to us implementing NonSend components too.

content/learn/book/ecs/queries/_index.md
```diff
@@ -0,0 +1,267 @@
...
+#[derive(Component)]
+struct Rotten;
+
+#[derive(Default)]
+struct FruitInventory{
+	map: HashMap<Fruit, u8>
+}
+
+// The second filtering type parameter of Query can be ommitted when no filter is used
+fn take_inventory_system(fruit_query: Query<&Fruit>, fruit_inventory: FruitInventory){
```
JunichiSugiura, line 162: What is `fruit_inventory` argument used for?

IceSentry (commented):
I'm about half way through the chapter and I haven't seen anything major. One thing I noticed is that some code blocks use tabs and other use spaces. While I don't want to start a space vs tabs debate, it currently leads to inconsistent formatting in the rendering of code blocks.

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -5,6 +5,302 @@ template = "book-section.html"
...
+        commands.entity(entity).remove(InCombat);
+    }
+}
+```
+
+## Bundles
+
+As you might guess, the one-at-a-time component insertion syntax can be both tedious and error-prone as your project grows.
+To get around this, Bevy abstracts these patterns using **bundles**: named and typed collections of components.
+These are implemented by adding the {rust_type(type="trait" crate="bevy_ecs" mod = "bundle" name="Bundle" no_mod = "true")}}  trait to a struct; turning each of its fields into a distinct component on your entity when they are inserted.
```
sseemayer, line 182: The tag should open with `{{` to make it a working link.

content/learn/book/ecs/queries/_index.md
```diff
@@ -0,0 +1,283 @@
...
+So what's the difference?
+
+By using `With` filters instead of requesting the data directly, we only check whether a component *has* that value, rather than fetching that component's value.
+As a result, we don't need to unpack our query into `(fruit, _delicious)` when iterating over it, and as a bonus, we can allow other systems to change the value of our components that are only include in our `With` filters in parallel to the systems being run!
+
+### `Or` Queries
+
+By default, query filters (just like query data requests) operate on a "and" basis: if you have a filter for `With<A>` and another filter for `With<B>`, only entities with both the `A` and `B` components will be fetched.
+We can change this behavior by using the {{rust_type(type="struct" crate="bevy_ecs" mod = "query" name="Or")}} type, nesting primitive query filters like {rust_type(type="struct" crate="bevy_ecs" mod = "query" name="With")}}, {{rust_type(type="struct" crate="bevy_ecs" mod = "query" name="Without")}} and {{rust_type(type="struct" crate="bevy_ecs" mod = "query" name="Changed")}} inside of it to return entities that meet any of the criteria inside.
```
sseemayer, line 199: the `With` type link is opening incorrectly, needs `{{`

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -1,14 +1,140 @@
...
+- you're using {{rust_mod(crate="bevy" mod="ecs")}} as a standalone crate
 
-TODO: discuss and demonstrate exclusive systems
+## Basic usage
 
-TOD: discuss and demonstrate `NonSend` resources
+Generally speaking, the API of working directly with the {{rust_type(type="struct" crate="bevy_ecs" name="World")}} mirrors those elsewhere that you might be familiar with.
+
+Like with {{rust_type(type="struct" crate="bevy" mod = "ecs/system" name="Commands" no_mod = "true")}}, you can call {rust_type(type="struct" crate="bevy_ecs" name="World" method = "spawn")}}, {rust_type(type="struct" crate="bevy_ecs" name="World" method = "spawn_batch")}} and {rust_type(type="struct" crate="bevy_ecs" name="World" method = "despawn")}} to add and remove entities, adding components to them with {{rust_type(type="struct" crate="bevy_ecs" name="World" method = "insert")}} and {{rust_type(type="struct" crate="bevy_ecs" name="World" method = "insert_bundle")}}.
+Resources are simply accessed with {rust_type(type="struct" crate="bevy_ecs" name="World" method = "get_resource")}} and the mutable equivalent {rust_type(type="struct" crate="bevy_ecs" name="World" method = "get_resource_mut")}}.
```
sseemayer, line 25: Two cases here.

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -1,14 +1,140 @@
...
+
+Like with queries, you can call {rust_type(type="struct" crate="bevy_ecs" name="World" method = "get")}} and {rust_type(type="struct" crate="bevy_ecs" name="World" method = "get_mut")}} to access component data of a particular sort on a given entity.
+If you want access to *all* of the data on an entity, use {rust_type(type="struct" crate="bevy_ecs" name="World" method = "get_entity")}} and {rust_type(type="struct" crate="bevy_ecs" name="World" method = "get_entity_mut")}}, along with various {rust_type(type="struct" crate="bevy_ecs" name="World" method = "EntityRef)}} methods to operate on the returned object.
```
sseemayer, line 28: three here.

content/learn/book/ecs/exclusive-world-access/_index.md
```diff
@@ -1,14 +1,140 @@
...
+
+You can create new queries using {rust_type(type="struct" crate="bevy_ecs" name="World" method = "query")}} and {rust_type(type="struct" crate="bevy_ecs" name="World" method = "query_filtered")}}.
```
sseemayer, line 30: two here.

content/learn/book/ecs/detecting-changes/_index.md
```diff
@@ -0,0 +1,130 @@
...
+## Removal detection
+
+We can watch for the removal of components too, although the mechanisms are quite different.
+{{rust_type(type="struct" crate="bevy_ecs" mod = "system" name="RemovedComponents" no_mod = "true")}} is a system parameter, returning an iterator of all entity identifiers from whom components of that type were removed during the last frame.
+This includes entities with that component who were despawned.
+There are three important caveats here:
+
+1. While{{rust_type(type="struct" crate="bevy_ecs" mod = "query" name="Added" no_mod = "true")}} and {{rust_type(type="struct" crate="bevy_ecs" mod = "query" name="Changed" no_mod = "true")}} operate on the basis of "since the last time this system ran", removal detection only works for the last frame. This distinction is very important if your system does not run every frame due to states or run criteria.
+2. {{rust_type(type="struct" crate="bevy_ecs" mod = "system" name="RemovedComponents" no_mod = "true")}} does not return the *value* of the component that was removed, due to performance concerns. If you need this, you will need to include your own change-detecting system to cache the values before they are removed.
```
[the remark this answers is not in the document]
alice-i-cecile: This appears to be an issue with our short-codes; it appears above when starting a paragraph with `Added`. 

Search "are query filters, appearing in the second type parameter".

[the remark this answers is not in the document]
alice-i-cecile: As shown in https://github.com/bevyengine/bevy-website/pull/182#issuecomment-936311931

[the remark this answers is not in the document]
alice-i-cecile: Resolving for now as out of scope, see #219.

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -5,6 +5,302 @@ template = "book-section.html"
...
+        .insert(Allegiance::Hostile)
+        .insert(Name("Goofus".to_string()));
+}
+```
+
+### Adding and removing components
+
+Once an entity is spawned, you can use commands to add and remove components from them dynamically.
+
+```rust
```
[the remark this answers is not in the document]
alice-i-cecile: See #219. I think we should fix this in a separate PR.

alice-i-cecile:
Alright, this is ready for final review! I'd prefer to merge this in ASAP if possible, and then make further additions or corrections in separate PRS: the length of both this thread and the PR is already excessive.

colepoirier:
Small style note, if I run zola serve for the latest ecs book chapter pr the indentation looks to be double what it should be  http://127.0.0.1:1111/learn/book/ecs/queries/

mirkoRainer:
> Small style note, if I run zola serve for the latest ecs book chapter pr the indentation looks to be double what it should be http://127.0.0.1:1111/learn/book/ecs/queries/

This looks to be a tabs vs spaces thing.
Places where spaces are, the indentation looks fine:
image

But where there are tabs:
image

This should fix it if you want:
https://github.com/alice-i-cecile/bevy-website/pull/2

alice-i-cecile:
Merged, thanks!

content/learn/book/ecs/queries/_index.md
```diff
@@ -0,0 +1,288 @@
++++
+title = "Fetching data with queries"
+weight = 2
+template = "book-section.html"
+page_template = "book-section.html"
++++
+
+Once we have data stored on our entities in the form of components, we need to be able to get the data back out in a principled way.
```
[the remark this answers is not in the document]
alice-i-cecile: > I'd like to see the "introduce one additional concept textually, provide concrete example, expand with one additional concept, repeat", pattern applied more regularly. I'm seeing a pattern of "introduce a ton of concepts all at once" then "provide a bunch of code".

I agree, this is a problem. I'll do a full pass on the chapter with fresh eyes and fix this up.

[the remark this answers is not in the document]
cart: Awesome. Thanks!

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,342 @@ template = "book-section.html"
...
+You will first encounter this when spawning and despawning entities: we have no way of knowing precisely which other components our entities might have, and so we are forced to wait until we can ensure that we can safely write to *all* component data at once.
+
+## ECS by example
+
+Before we dive into the details of each of these features, let's take a quick look at a simple but fun whack-a-mole game that you can run and play.
+Unsurprisingly, the different parts of the ECS tend to be closely linked: components are not very useful without a way to spawn entities and systems that run our logic are very dull if we can't discuss the data they can access.
+The details of each part are more easily grasped if you have a basic sense of the whole.
+
+```rust
+use bevy::prelude::*;
```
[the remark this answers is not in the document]
alice-i-cecile: I agree, this example is way too complex for a new user.

content/learn/book/ecs/_index.md
```diff
@@ -6,4 +6,342 @@ template = "book-section.html"
...
+
+Of course, this database is [very ragged](https://www.transdatasolutions.com/what-is-ragged-data/): not all entities will have every component!
+We can use this fact to specialize behavior between entities: systems only perform work on entities with the correct combination of components.
+You don't want to apply gravity to entities without a position in your world, and you're only interested in using the UI layout algorithm to control the layout of UI entities!
+
+When we want to go beyond this tabular data storage, we can use **resources**: global singletons which store data, each in their own monolithic blob.
+You might use resources to interface with other libraries, store unique bits of state like the game's score, or store secondary data structures like indexes to augment your use of entity-component data.
+
+In order to manipulate and act on this data, we must use systems.
+**Systems** are Rust functions that request specific data, such as resources and entities, from the {{rust_type(type="struct" crate="bevy_ecs" name="World")}}. They define a query in their parameters (arguments) that selects data with a particular combination of components.
```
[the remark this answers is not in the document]
alice-i-cecile: I like the bullet points a lot. I'm going to steal from this.

KDecay:
Enabled the anchor links for the ECS section in https://github.com/alice-i-cecile/bevy-website/pull/4.

alice-i-cecile:
Closing this out in favor of the small, split PRs linked above. Working with all of the changes in a single PR is incredibly unwieldy as both an author and reviewer.

C-BJ:
Any plans to localize Bevy Book?

alice-i-cecile:
@C-BJ personally yes, but once it's significantly more stable. I don't think it makes sense until 1.0.

doup:
Maybe https://github.com/bevyengine/bevy-website/issues/265 can be updated listing the PRs on the description?

```
- [ ] https://github.com/bevyengine/bevy-website/pull/290
- [ ] https://github.com/bevyengine/bevy-website/pull/291
- [ ] https://github.com/bevyengine/bevy-website/pull/292
- [ ] https://github.com/bevyengine/bevy-website/pull/293
- [ ] https://github.com/bevyengine/bevy-website/pull/294
- [ ] https://github.com/bevyengine/bevy-website/pull/295
- [ ] https://github.com/bevyengine/bevy-website/pull/296
- [ ] https://github.com/bevyengine/bevy-website/pull/297
```

ickk:
> Maybe #265 description can be updated to list the PRs?

As an alternative, I'd advocate for setting up a project board instead of trying to use a tracking issue that needs to be manually kept in sync

content/learn/book/ecs/entities-components/_index.md
```diff
@@ -3,8 +3,293 @@ title = "Entities have components"
...
+        // The component will be inserted at the end of the current stage
+        commands.entity(entity).insert(InCombat);
+    }
+}
+
+// Now to undo our hard work
+fn end_combat_system(query: Query<Entity, (With<Combatant>, With<InCombat>>, mut commands: Commands){
+    for entity in query.iter(){
+        // The component will be removed at the end of the current stage
+        commands.entity(entity).remove(InCombat);
```
ebiggs, line 171: should this be remove::<InCombat>() ?

PR #183, opened 2021-06-30
Minor fixes in the new Book

MinerSebas:
Due to #182 accidentally targeting the `Master` Branch, I reviewed some changes that were already pushed to the `new-book` branch.

This PR applies the suggestions of those Reviews.

This also adds the changes from #167 and #169.
These changes would be otherwise lost as 252325d thinks the original File was deleted, instead of moved.


content/learn/book/welcome/installation/_index.md
```diff
@@ -20,7 +20,7 @@ Once this is done, you should have the ```rustc``` compiler and the ```cargo```
 ### Install OS dependencies
 
 * [Linux](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md)
-* Windows: Make sure to install [VS2019 build tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
+* Windows: If you use the `msvc` toolchain make sure to install [VS2019 build tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
```
alice-i-cecile, line 23: I'm not convinced this advice is adequately clear in either form. Newcomers (particularly common on Windows) have no idea what a toolchain is, or how to tell if they're using it.

VS2019 build tools also come with a dozen checkboxes to tick: it's not clear from the description that you need the C++ (if I recall correctly) installation and nothing else.

MinerSebas: Alternatively we could remove the Windows entry as should already be covered by the Rust install Instructions.
This is shown to me here https://www.rust-lang.org/tools/install
```
It looks like you’re running Windows. 
To start using Rust, download the installer, then run the program and follow the onscreen instructions. 
You may need to install the Visual Studio C++ Build tools when prompted to do so.
```

And when installing Rust, the Installer detects the missing dependency and (tries to) install the gnu toolchain instead.

alice-i-cecile:
Closing due to staleness; this chapter is about to get reworked pretty aggressively.

PR #191, opened 2021-07-21
fix code highlighting after zola 0.14

mockersf:
- move the config option to the correct section
- fix the background color

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -7,10 +7,6 @@ title = "Bevy Engine"
 # Whether to automatically compile all Sass files in the sass directory
 compile_sass = true
 
-# Whether to do syntax highlighting
-# Theme can be customised by setting the `highlight_theme` variable to a theme supported by Zola
-highlight_code = true
-
 # Whether to build a search index to be used later on by a JavaScript library
 build_search_index = true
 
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

cart:
Merged asap because some highlighting is better than no highlighting. Just got my local zola updated. And now I want to tweak this a bit :smile: 

Heres a quick / rough comparison:

image

1. right: old
2. top left: new (with new css tweaks to revert the background to the old color)
3. bottom left: current main with new background color

My preference is (1), (2), (3). I think (3)'s background sticks out too much. (2) has regrettably lost a lot of contrast compared to (1). I'd like the colors to pop more if possible (like the do in (1)).

I suspect i'll be messing around with this for awhile :)

cart:
Here is (2) again with the background colors properly synced up:
image

cart:
Note how bland it looks now :sob:

cart:
Good news! The new release has lots of configuration options. We can now build fully custom css themes: https://www.getzola.org/documentation/content/syntax-highlighting/

cart:
(which i will absolutely do)

mockersf:
> (which i will absolutely do)

That... doesn't surprise me at all from you 👍

PR #203, opened 2021-08-09
add test with md included as a doc comment

mockersf:
alternative to #148 and #150 

* validate inline code block as a doc comment of a rust crates
* hide extra lines needed for setup of doc comments


mockersf:
will be included in the next book version as #281

PR #214, opened 2021-10-20
[Merged by Bors] - Shallow clone bevy assets

NiklasEi:
Currently, this saves around 20% of data (150kB) for every website deploy execution. But I would expect the gain to increase with the growing history in the assets repository. Just feels like less of a waste of resources :slightly_smiling_face:


IceSentry (approved):

alice-i-cecile:
> Provide an argument of -- depth 1 to the git clone command to copy only the latest revision of a repo

https://www.perforce.com/blog/vcs/git-beyond-basics-using-shallow-clones

PR #216, opened 2021-10-24
Render bevy error codes

NiklasEi:
This PR generates a list of error codes from the markdown files in the bevy repository. They are rendered at `/errors`.

- [x] wait for bevyengine/bevy#2994


NiklasEi (commented):
~I will try pulling common parts of the two generation crates (assets/errors) into some kind of utility.~ (Edit: this doesn't really seem worth it at the moment)

Just wanted to have the draft PR open already so we have a place to discuss where/how to show the errors on the website in general.

alice-i-cecile:
Seems like a sensible idea. I like this.

mockersf:
for the website, which documents the released version, this will need to pull the `latest` branch of Bevy

NiklasEi:
> for the website, which documents the released version, this will need to pull the `latest` branch of Bevy

Good point. The codes now come from `latest`. To still be able to merge this before the `0.6` release, I added some general information on the page and link to the descriptions for Bevy main. If it's fine that the page is pretty empty until the release, we could merge this before `0.6`.

templates/learn.html
```diff
@@ -55,5 +55,18 @@
       </div>
     </div>
   </a>
+  <a class="card" href="/errors">
+    <div class="card-image">
+      <img src="/assets/bevy_icon_dark.svg" class="centered-card-image" alt="Bevy logo" />
+    </div>
+    <div class="card-text">
+      <div class="card-title">
+        Bevy Error index
```
mockersf, line 64:
```suggestion
        Bevy Error Index
```

mockersf:
It would be nice to have "Bevy Assets" and "Bevy Error Index" each have their own custom logo, but that can be done later. Something like the Bevy logo and a file for assets, and the logo with a warning sign for errors?

mockersf (approved):

cart:
Yeah I'm down to merge but I think we should remove Bevy Error Index from the `/learn` page until we have useful content here and we have a unique logo for the error index (i dont take issue with reusing the bevy logo as much as I do with the fact that two identical images back to back cuts down on perceived quality).

templates/learn.html
```diff
@@ -55,5 +55,18 @@
       </div>
     </div>
   </a>
+  <a class="card" href="/errors">
```
cart, line 58: I'm thinking errors should probably live under the `/learn` url

cart: `/learn/errors`

PR #221, opened 2021-11-12
Add fix for missing libudev package

timClicks:
I encountered a build error when playing around with Bevy tonight. This fixed the problem for me :)


DJMcNab:
We should point to the Linux docs document in the main repository for this, I think?

alice-i-cecile:
Closing this out in favor of pointing at the Linux docs.

PR #233, opened 2021-12-31
Update _index.md

godsmith99x:
Adds a note on necessary package dependencies for Fedora 35 users.

### DIFF
--- a/content/learn/book/getting-started/_index.md
+++ b/content/learn/book/getting-started/_index.md
@@ -35,6 +35,10 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
     ```sh
     cargo run --example breakout
     ```
+    ---
+    **Note**
+    Fedora 35 users will need to install the alsa-lib-devel and rust-libudev-devel packages in order to compile the breakout example.
+    ---
 
 ### Add Bevy as a Dependency
 


rparrett:
Totally agree that dependencies should be mentioned prior to telling folks to run the examples. They are mentioned on the [next page](https://bevyengine.org/learn/book/getting-started/setup/), which links to [linux_deendencies.md](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md).

If I recall correctly, though, this isn't an issue in the new welcome chapter of the book being worked on [here](https://github.com/bevyengine/bevy-website/pull/176).

rparrett:
It looks like `rust-libudev-devel` isn't mentioned there though, so a PR on the bevy repo adding that would probably be welcome!

cart:
Yeah I dont think duplicating platform-specific dependencies in the welcome chapter is worth the maintenance cost (or the bloat ... there are a lot of platform specific notes). The linux_dependencies.md file is where this type of content should live. Definitely happy to revisit how / where we link to that doc though. Maybe sync up with @alice-i-cecile on the "new bevy book" currently in development?

godsmith99x:
Cool. That all makes sense. Thanks for the review!

PR #246, opened 2022-01-08
Mention Rust 2021 first in the migration guide

nakedible:
The transition to Rust 2021 is easily missed from the news announcement, and will lead to confusing errors.

The actual explanation is simply copied from the news article, as it was very well explained there.


rparrett:
Looks good to me

alice-i-cecile (approved):

KDecay (approved):

PR #251, opened 2022-01-09
Add implied missing link.

nebkor:
It really looked like a note to do something had been left in as text, so hopefully this is on the right track :)


alice-i-cecile (approved):

alice-i-cecile:
Awesome, thank you!

PR #252, opened 2022-01-09
build more examples in webgl2 and use latest tag

mockersf:
build more examples in webgl2 and use latest tag


rparrett:
All the new examples are working over here!

PR #253, opened 2022-01-09
fix crevice example field name

payload:
I think field names of the rust struct directly match with the field names of the shader struct.

### DIFF
--- a/content/news/2022-01-08-bevy-0.6/index.md
+++ b/content/news/2022-01-08-bevy-0.6/index.md
@@ -488,7 +488,7 @@ Coupling an `AsStd140` derive with our new `UniformVec<T>` type makes it easy to
 ```rust
 // WGSL shader
 struct Mesh {
-    model: mat4x4<f32>;
+    transform: mat4x4<f32>;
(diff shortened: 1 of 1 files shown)


content/news/2022-01-08-bevy-0.6/index.md
```diff
@@ -488,7 +488,7 @@ Coupling an `AsStd140` derive with our new `UniformVec<T>` type makes it easy to
 ```rust
 // WGSL shader
 struct Mesh {
-    model: mat4x4<f32>;
+    transform: mat4x4<f32>;
```
payload, line 491: Unfold above to see the transform field of the rust struct.

alice-i-cecile (approved):

IceSentry:
The current Mesh struct still uses model
|https://github.com/bevyengine/bevy/blob/6bc5c609867908345f0b427cbd2725082883b31c/crates/bevy_pbr/src/render/mesh_struct.wgsl#L2

alice-i-cecile:
Closing this out then :)

PR #258, opened 2022-01-10
Remove outdated paragraph (v0.6 fix)

jch-13:
The `system()` call has been removed from the paragraph explaining it, but the whole sentence does not make sense anymore with `v0.6`.


alice-i-cecile:
Thanks, that's a great catch!

PR #259, opened 2022-01-10
Typo Fix

mirkoRainer:
Small fix to type.
disjoint -> disjointed


jakobhellermann:
https://english.stackexchange.com/a/122074 since "disjoint" here refers to "having no resources/components in common" I think it is the correct form

mirkoRainer:
🤔 
TIL. Thank you. Will close.

PR #263, opened 2022-01-11
Fix complexity description for iter_combinations

idanarye:
The newspost said it was super-exponential, but simple analysis shows that it should only be exponential.

The number of combinations is exponential at the size of the combination, so unless visiting each combination is super-exponential - the overall complexity should also be exponential. The bulk logic of visiting each combination is done in [`QueryCombinationIter::fetch_next_aliased_unchecked`](https://github.com/bevyengine/bevy/blob/v0.6.0/crates/bevy_ecs/src/query/iter.rs#L227), which seems polynomial. Not even exponential - definitely not super-exponential.

### DIFF
--- a/content/news/2022-01-08-bevy-0.6/index.md
+++ b/content/news/2022-01-08-bevy-0.6/index.md
@@ -704,7 +704,7 @@ fn system(query: Query<&Player>) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
Thanks for the analysis and the fix <3

alice-i-cecile:
See: https://stackoverflow.com/questions/39558086/how-to-prove-binomial-coefficient-is-asymptotic-big-theta-of-two-to-the-power-n

PR #266, opened 2022-01-11
Correct audio block on the landing page

NiklasEi:
* `AudioOutput` is an internal, non-send resource; the user-facing resource to play audio on is `Audio`
* We can load multiple audio encodings, so I removed the `mp3` to make it more general
* Stray whitespace that my IDE removes on its own :bowtie:

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -130,8 +130,8 @@
                 <h2 class="feature-title">Sound</h2>
                 Load audio files and play them on demand
                 <ul class="feature-sublist">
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
Yep: this is the right paradigm to use now. See [the example](https://github.com/bevyengine/bevy/blob/main/examples/audio/audio.rs).

alice-i-cecile (approved):

mockersf (approved):

PR #272, opened 2022-01-13
Remove Rust doc shortcodes from new book

### LINKED ISSUE #224: Replacement for shortcode docs links
## Problem definition

1. Short codes are error-prone to write, with a huge number of matching delimiters.
2. Short codes are very hard to read, due to their verbose syntax, arbitrary ordering and lack of support on Github and in IDEs. e.g. `{rust_type(type="struct" crate="bevy" mod = "ecs/system" name="EntityCommands" no_mod = "true")}}`.
3. Short codes are a strange DSL that new contributors to the book must memorize.
4. Short codes force authors to look up the fully qualified path each type.
5. Short codes require authors to explicitly state whether a given import is a `struct`, `enum`, `trait` or so on.

## Requirements

1. Human-legible form is easy to write and read.
  2. Acceptable level: `struct{{bevy::ecs::system::EntityCommands}}`
2. Machine-legible form is standard zola-compatible HTML that links to the appropriate docs.rs page.
3. Docs links must support custom formatting.
4. Human-legible form is committed and reviewed.

### Bonus points

1. Hot reloading still works.
2.  Implicit imports: automatically assume Bevy prelude: `struct{{EntityCommands}}`
3. Type elision: `{{bevy::ecs::system::EntityCommands}}`
4. Explicit imports: list `use bevy::ecs::system::EntityCommands` at the top of the file, which is processed out, allowing the author to write `struct{{EntityCommands}}`.

## Possible strategies

1. Fork Zola to add support for this.
    1. Probably more immediately elegant.
    2. Significantly more maintenance burden unless merged upstream.
    3. Should only involve writing our own [tera](https://tera.netlify.app/docs/) functions.
    4. Best we can do is likely `{{rs(name="App"}}` due to tera limitations :(
    4. Difficulties integrating with our CI.
2. Add a Rust preprocessor which modifies the `.md` files provided.
    1. These are stored in the `build` directory and are not commited.
    2. Begin with simple instructions to run the preprocessor before `zola serve`.
    3. Eventually, wrap this in its own minimal CLI to support hot reloading and one-command builds.
3. Use [reference style links](https://www.markdownguide.org/basic-syntax/#reference-style-links) instead of short-codes.
    1. Works out of the box!
    2. Very legible.
    2. No special formatting out of the box. Can this be added?
    3. Requires looking up the full URL, and modifying it to either use `latest` or doing a find-and-replace each Bevy version.

alice-i-cecile:
# Problem
- Shortcodes are awful to both read and write.
- Fixes #224.

# Solution
- [x] Remove short codes
- [x] Replace short codes with reference-style links
- [x] Remove shortcodes from Welcome chapter


NiklasEi:
This would be perfect to combine with a link check in CI as we have it on the main repo.
https://github.com/bevyengine/bevy/blob/main/.github/workflows/ci.yml#L167

alice-i-cecile:
Yep. I ended up having to remove a few links here because they were dead :(

content/learn/book/migration-guides/0.5-0.6/_index.md
```diff
@@ -29,9 +29,9 @@ members = [ "my_crate1", "my_crate2" ]
...
 
-All functions of {{rust_type(type="struct" crate="bevy_app" mod="" version="0.5.0" name="AppBuilder" no_mod=true)}} were merged into {{rust_type(type="struct" crate="bevy_app" mod="" version="0.6.0" name="App" no_mod=true)}}.
+All functions of [`AppBuilder`] were merged into [`App`].
 
-In practice this means that you start constructing an {{rust_type(type="struct" crate="bevy_app" mod="" version="0.6.0" name="App" no_mod=true)}} by calling {{rust_type(type="struct" crate="bevy_app" mod="" version="0.6.0" name="App" no_mod=true method="new")}} instead of {{rust_type(type="struct" crate="bevy_app" mod="" version="0.5.0" name="App" no_mod=true method="build")}} and {{rust_type(type="trait" crate="bevy_app" mod="" version="0.5.0" name="Plugin" no_mod=true method="build")}} takes a {{rust_type(type="struct" crate="bevy_app" mod="" version="0.6.0" name="App" no_mod=true)}} instead of a {{rust_type(type="struct" crate="bevy_app" mod="" version="0.5.0" name="AppBuilder" no_mod=true)}}
+In practice this means that you start constructing an [`App`] by calling `App::new()` instead of `App::build` and `Plugin::build` takes a [`App`] instead of a [`AppBuilder`].
```
cart, line 34: These functions used to be links. Can you re-add those? (and do a pass to make sure we haven't dropped links elsewhere? I'm seeing this happen in a number of places.)

content/learn/book/migration-guides/0.5-0.6/_index.md
```diff
@@ -61,9 +61,12 @@ impl Plugin for SomePlugin {
 }
 ```
 
+[`AppBuilder`]: https://docs.rs/bevy/0.5.0/bevy/app/struct.AppBuilder.html
+[`App`]: https://docs.rs/bevy/latest/bevy/app/struct.App.html
```
IceSentry, line 65: This should probably point to 0.6 otherwise it might not point to the right thing once 0.6 isn't the latest

IceSentry, line 65: This should probably point to 0.6 otherwise it might not point to the right thing once 0.6 isn't the latest

content/learn/book/migration-guides/0.5-0.6/_index.md
```diff
@@ -82,9 +85,11 @@ In order to use foreign types as components, wrap them using the newtype pattern
 struct Cooldown(std::time::Duration);
 ```
 
+[`Component`]: https://docs.rs/bevy/latest/bevy/ecs/component/trait.Component.html
```
IceSentry, line 88: Same thing here about pointing to 0.6

IceSentry (approved):
I had a quick look around locally by clicking on a bunch of random links and everything seems to work. 

The ease of reading the raw text is so much nicer!

cart (commented):
Ok this is more compelling than I expected.

Things I like:
* We still get special formatting (I didn't know you could use code blocks as link aliases)
* We can reuse link declarations
* Legibility of prose goes way up
* Easier to run deadlink checks

Thinks I dislike:
* Less ergonomic for single-use links, especially if we did ergo improvements for short codes (default to no_mod, shorter short-code name, full paths as variable input). The proposed approach requires repeating the symbol name, repeating the crate name, specifying the url prefix (`https://docs.rs/`), and adding the `.html` suffix. 
* We're throwing away our ability to theme based on symbol type (ex: match struct and trait colors to our syntax highlighting theme) and to add more complicated html (ex: show docs on hover, show full path on hover, etc). Not the biggest deal in the world atm because we don't do either of those things, but it does feel like a loss.

However since we've agreed that our ideal code ref situation would require a preprocessor (which in turn, would resolve the issues above), I think I'm cool with moving in this direction in the interim. Moving to a preprocessor would require rewriting shortcodes anyway.

content/news/2020-08-10-introducing-bevy/index.md
```diff
@@ -895,7 +920,7 @@ This will load new versions of assets whenever their files have changed.
 
 #### Adding New Asset Types
 
-To add a new asset type, implement the {{rust_type(type="trait" crate="bevy_asset" version="0.1.0" name="AssetLoader" no_mod=true)}} trait. This tells Bevy what file formats to look for and how to translate the file bytes into the given asset type.
+To add a new asset type, implement the `AssetLoader` trait. This tells Bevy what file formats to look for and how to translate the file bytes into the given asset type.
```
alice-i-cecile, line 923: This trait does not exist in the public API of Bevy 0.1. As a result, I cannot replace this link.

content/news/2020-08-10-introducing-bevy/index.md
```diff
@@ -924,13 +949,13 @@ We plan on extending the audio system with more control and features in the futu
 
 ## Render Graph
 
-All render logic is built on top of Bevy's {{rust_type(type="struct" crate="bevy_render" version="0.1.0" mod="render_graph" name="RenderGraph" no_mod=true)}}. The Render Graph is a way to encode atomic units of render logic. For example, you might create graph nodes for a 2D pass, UI pass, cameras, texture copies, swap chains, etc. Connecting a node to another node indicates that there is a dependency of some kind between them. By encoding render logic this way, the Bevy renderer is able to analyze dependencies and render the graph in parallel. It also has the benefit of encouraging developers to write modular render logic.
+All render logic is built on top of Bevy's `RenderGraph`. The Render Graph is a way to encode atomic units of render logic. For example, you might create graph nodes for a 2D pass, UI pass, cameras, texture copies, swap chains, etc. Connecting a node to another node indicates that there is a dependency of some kind between them. By encoding render logic this way, the Bevy renderer is able to analyze dependencies and render the graph in parallel. It also has the benefit of encouraging developers to write modular render logic.
```
alice-i-cecile, line 952: `RenderGraph` is not publicly exposed in 0.1, so I cannot relink this.

content/news/2020-08-10-introducing-bevy/index.md
```diff
@@ -924,13 +949,13 @@ We plan on extending the audio system with more control and features in the futu
...
 
 Bevy includes a number of nodes by default: `CameraNode`, `PassNode`, `RenderResourcesNode`, `SharedBuffersNode`, `TextureCopyNode`, `WindowSwapChainNode`, and `WindowTextureNode`. It also provides subgraphs for 2d rendering, 3d rendering, and UI rendering. But you are welcome to create your own nodes, your own graphs, or extend the included graphs!
 
 ### [Data Driven Shaders](https://github.com/bevyengine/bevy/blob/1d68094f59b01e14f44ed7db8907dbd011b59973/examples/shader/shader_custom_material.rs)
 
-Components and Assets can derive the {{rust_type(type="trait" crate="bevy_render" version="0.1.0" mod="renderer" name="RenderResources" no_mod=true)}} trait, which enables them to be directly copied to GPU resources and used as shader uniforms.
+Components and Assets can derive the `RenderResources` trait, which enables them to be directly copied to GPU resources and used as shader uniforms.
```
alice-i-cecile, line 958: Also not public.

alice-i-cecile:
@cart links have been re-added now! There's three links from the 0.1 announcement that are unfixable: they aren't part of the public API and those links were always dead. IMO we should just leave them unlinked at this point.

content/news/2020-08-10-introducing-bevy/index.md
cart: https://docs.rs/bevy_asset/0.1.0/bevy_asset/trait.AssetLoader.html

content/news/2020-08-10-introducing-bevy/index.md
cart: https://docs.rs/bevy_render/0.1.0/bevy_render/render_graph/struct.RenderGraph.html

cart:
Those types are all exported in the public api of `0.1.0`. I provided links in each comment.

alice-i-cecile:
Thanks @cart; I missed them because they weren't in the `bevy` crate itself. This should be ready-to-merge now.

cart:
I think the "fix missing Bevy 0.1 links" commit doesn't actually do that.

alice-i-cecile:
Oops; my linter went beserk. Let me try again...

cart:
LOL

cart:
No bors here

PR #276, opened 2022-01-20
Skip readme when generating error codes

NiklasEi:
After the readme was added, it started showing up on the website: https://bevyengine.org/learn/errors/#readme

### DIFF
--- a/generate-errors/src/lib.rs
+++ b/generate-errors/src/lib.rs
@@ -1,3 +1,4 @@
+use regex::Regex;
 use serde::Serialize;
 use std::fs::read_to_string;
 use std::{fs, io, path::PathBuf, str::FromStr};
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

generate-errors/src/lib.rs
```diff
@@ -34,17 +34,11 @@ impl From<&ErrorCode> for FrontMatterErrorCode {
...
-        // Todo: after the 0.6 release, remove this if statement
-        // For now we will allow this to be able to point to the `latest` branch (0.5)
-        // which does not yet include error codes
-        return Ok(());
-    }
     assert!(dir.is_dir(), "The path to the errors is not a directory");
     for entry in fs::read_dir(dir)? {
         let entry = entry?;
         let path = entry.path();
-        if path.file_name().unwrap() == ".git" || path.file_name().unwrap() == ".github" {
```
alice-i-cecile, line 47: Are there any other .git / .github files in this directory that we should be worried about still?

rparrett (approved):

mockersf:
it would be more future proof to check that the file name is of the expected pattern (`B[0-9]{4}\.md` I think)

mockersf (approved):

generate-errors/src/lib.rs
NiklasEi: Not inside the `errors` directory.

PR #281, opened 2022-02-15
[Merged by Bors] - Code validation and format checking

### LINKED ISSUE #83: Bevy book code should be compiled against new releases
There have been a number of cases where the Bevy book falls out of sync from the most recent release and stays that way for a while, confusing new users that come on board right after an exciting new release.

To that end, this code should be fully compiled with Bevy as a dependency, and then its examples checked against the release candidate before the release goes live. Copy-pasting the code into test scripts is tedious, so it would be nice if this occurred more automatically.

I know mdbook supports this smoothly (because I support in [Understanding Bevy](https://github.com/alice-i-cecile/understanding-bevy)); I'm not sure that you can do the same in Zola. As an added bonus, swapping to mdbook (#47) lets end users run the example code directly if they clone this repo.

### LINKED ISSUE #279: Enforce formatting of code examples in Bevy book
It can be quite challenging to maintain an appropriately consistent style when writing code snippets for examples. Indentation errors, offset parentheses and so on reduce the perception of quality to new readers, and distract from the learning experience.

Fortunately, `cargo fmt` is very good. We should ensure that submitted code passes `cargo fmt`. This is very closely related to #83, and we should ensure that the solution we select for that can be extended to fit this. 

Ideally, this process is also easy for authors to use. We probably won't be able to use `cargo fmt` directly, as the Rust files are embedded inside markdown, but a script that applies the formatting to all code blocks and can be run on command would be massively appreciated.

KDecay:
# Objective

Migrates #203
Closes #83

# Solution

- [x] Add code validation
- [x] ~~Add code format checking~~ Removed to unblock the code validation changes

# Problems

During the implementation of the code validation and the format checking I encountered some problems which I would appreciate guidance or help with.

## Need to be addressed

1. ~~Zola doesn't highlight the code blocks using the rust syntax if you add more than just the language to the code block attributes. Example `rust` highlights the code correctly, `rust,no_run` doesn't. So currently the code highlighting is broken.~~ Zola is order dependent for attributes (as noted by @mockersf). So changing the attribute order from `rust,no_run` to `no_run,rust` works just fine.
1. Running `cargo fmt` doesn't format the doc examples. Adding the `rustfmt.toml` option `format_code_in_doc_comments = true` enables this, but requires nightly. In my tests it also only works if you add the code examples directly without using `#[doc = include_str!("_index.md")]`. So currently the format checking is not working.

## Nice to have

1. The `lib.rs` file of the `code-validation` crate has to be manually updated each time a new `_index.md` file is added or removed. It might be a good idea to dive deeper and see if we can implement a solution where this work is done automatically. For now this should be more than fine though.


.github/workflows/ci.yml
```diff
@@ -30,3 +30,37 @@ jobs:
...
+     steps:
+       - uses: actions/checkout@v2
+
+       - uses: actions-rs/toolchain@v1
+         with:
+           toolchain: nightly
+           components: rustfmt, clippy
+           override: true
+
+       - name: Install alsa and udev
```
alice-i-cecile, line 62: I don't think we should need these deps for cargo fmt.

code-validation/rustfmt.toml
```diff
@@ -0,0 +1,3 @@
+use_field_init_shorthand = true
+newline_style = "Unix"
+format_code_in_doc_comments = true
```
alice-i-cecile, line 3: This is *very* nice. Once we get this working, we should transfer this to the main repo; this has been a frustration of mine.

code-validation/src/lib.rs
```diff
@@ -0,0 +1,170 @@
+//! This crate is used to validate the rust code of the `bevy` website.
+//!
+//! It is currently used to validate the rust code of the offical `bevy` book.
+//! The modules represents the folder structure of the website.
+
+mod learn {
```
alice-i-cecile, line 6: Ouch. I don't love how manual this process is.

alice-i-cecile:
> Zola doesn't highlight the code blocks using the rust syntax if you add more than just the language to the code block attributes. Example rust highlights the code correctly, rust,no_run doesn't. So currently the code highlighting is broken.

I'm not seeing an [existing issue](https://github.com/getzola/zola/issues) for this, but the author has been very responsive in the past. Let me make one and see what we can do.

mockersf:
if I remember correctly, Zola is order dependant and it should work with `no_run,rust`

KDecay:
That did the trick. Thank you!

This is important to know for everyone that creates and/or edits code examples. Do we have a place to note this down?

alice-i-cecile:
> Do we have a place to note this down?

IMO we should have a `CONTRIBUTING.md` for this repo. Can you start one in this PR and note that down?

alice-i-cecile:
So, the question is: do we want to merge this as is, or block on a code formatting solution. IMO code formatting is the much-less important problem here; and I'd prefer to get this rolling faster to improve book quality and contributor experience.

code-validation/src/lib.rs
alice-i-cecile: Yeah. We can poke at fixing this later IMO.

alice-i-cecile (approved):

.github/workflows/ci.yml
```diff
@@ -30,3 +30,34 @@ jobs:
...
+         with:
+           toolchain: stable
+           components: rustfmt, clippy
+           override: true
+
+       - name: Install alsa and udev
+         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
+
+       - name: Build & run tests
+         run: cd code-validation && cargo test
```
alice-i-cecile, line 49: Can you add directions on how to do this (and the code formatting step, if we can get it working) to CONTRIBUTING.md?

code-validation/src/lib.rs
KDecay: Agreed. This is already a huge improvement to not having code validation at all.

mockersf:
> 2. Running `cargo fmt` doesn't format the doc examples. Adding the `rustfmt.toml` option `format_code_in_doc_comments = true` enables this, but requires nightly. In my tests it also only works if you add the code examples directly without using `#[doc = include_str!("_index.md")]`. So currently the format checking is not working.

Would it work if you expand the macro then run the formatting?

KDecay:
> So, the question is: do we want to merge this as is, or block on a code formatting solution. IMO code formatting is the much-less important problem here; and I'd prefer to get this rolling faster to improve book quality and contributor experience.

Agreed. I sadly couldn't find anything online about the `format_code_in_doc_comments` setting that verified my findings. Having CI check formatting without ever failing is not the worst thing ever, but if we want to merge this we could also remove the CI task and add it once we find a solution.

> Would it work if you expand the macro then run the formatting?

Do you mean expanding the macro so that you get `#[doc = "the actual string of the file"]`? If that's the case then it didn't work for me. Seems like `#[doc = "..."] ` is not respected by `cargo fmt` which seems weird, but `format_code_in_doc_comments` is an unstable feature so who knows.

alice-i-cecile:
@KDecay IMO we should split out the validation and formatting checks in order to unblock this.

I really really want this in place before I (or others) attempt to write more content for the book.

KDecay:
Definitely agreed. I'll update the CI workflow so it only adds the code validation.

alice-i-cecile:
This no longer closes #279.

KDecay:
Updated the description accordingly.

alice-i-cecile (approved):

IceSentry:
So, just to make sure I understand this correctly. Let's say I want to have a small code block that only shows a few lines that are added to a block that was defined earlier, I would now need to have the entire valid program in the code block and hide all the lines I don't want to show?

KDecay:
@IceSentry Yup. This way we ensure that the code shown to the user is valid. You can also use the `ignore` tag, but that would defeat the purpose of this change.

IceSentry:
Right, makes sense, where can I get more information on how the hide-line parameter works?

IceSentry (approved):

alice-i-cecile:
It looks like it's an extension of [the rust-doc feature](https://doc.rust-lang.org/stable/rustdoc/documentation-tests.html#hiding-portions-of-the-example), but I can't immediately see where the hide-line parameter can be found.

KDecay:
@IceSentry https://www.getzola.org/documentation/content/syntax-highlighting/#annotations

PR #285, opened 2022-03-10
[Merged by Bors] - enable CI on bors branches

mockersf:
bors is timeouting because CI is not triggering on its branches (`trying` and `staging`)
- enable CI on bors branches

bors needs the job name in its config
- change the workflow name to the job name

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,5 +1,5 @@
 status = [
-    "CI",
+    "build",
 ]
 
 use_squash_merge = true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,10 @@ name: CI
 
 on:
   pull_request:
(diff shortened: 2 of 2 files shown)


cart:
Cool cool. I'll manually merge this after normal CI builds pass.

mockersf:
this pr doesn't enable CI on `master` branch which is bad practice according to bors, but I think it should be fine here

alice-i-cecile:
bors try

mockersf:
> Cool cool. I'll manually merge this after normal CI builds pass.

it should be possible to `bors r+` it, I think bors will use the CI setup from this PR

mockersf:
> bors r+

Best guess to why it's waiting is that Bors doesn't want to merge a PR it's currently trying :D

mockersf:
bors cancel

PR #286, opened 2022-03-11
Check for dead links in CI

### LINKED ISSUE #177: Check for dead links in CI
Just as in the main Bevy repo, this would be helpful functionality.

It's particularly nice when combined with /latest/ links to docs.rs, which will let us know if something has been moved and prompt a refresh of that section of the book.

alice-i-cecile:
Fixes #177.


mockersf:
not sure this check is a good idea, given how unstable it is even on bevy repo. It would also block from pushing a new section and links to it in the same commit

alice-i-cecile:
> not sure this check is a good idea, given how unstable it is even on bevy repo. It would also block from pushing a new section and links to it in the same PR

Yeah, I think I agree. Let's cut it, and we can reconsider it later.

alice-i-cecile:
Actually, closing this. We should figure out the strategy in #177, and then create a fresh PR to match that.

PR #289, opened 2022-03-14
Basic contributing directions

alice-i-cecile:
We should explain how to use the tooling of the website.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,6 +108,9 @@ Check that this works locally using:
 1. `cd code-validation`
 2. `cargo build`
 
+This website also hosts examples, assets and error codes for the Bevy engine.
+These are created by the scripts in the various `generate-*` folders, which scrape data sources from other repos and are run daily.
+
 ## Contributor's style guide
 
 When writing and reviewing learning material for the Bevy Book and Quick Start Games, please try to follow these guidelines:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,7 +100,7 @@ This path should prioritize:
 
 ## Setup
 
-This website is built using [Zola](https://www.getzola.org/): install it and use `zola serve --open` to examine the website as it would appear.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -98,6 +98,16 @@ This path should prioritize:
    5. Contributing to Bevy itself.
 6. Linking to further reading: official examples, `docs.rs` and (very sparingly) source code links.
 
+## Setup
+
+This website is built using [Zola](https://www.getzola.org/): install it and use `zola serve --open` to examine the website as it would appear.
```
IceSentry, line 103: is `--open` required? I always only used `zola serve`

IceSentry (approved):

CONTRIBUTING.md
alice-i-cecile: No, it's not. It's a bit more comfortable IMO, but eh, I'll cut it.

alice-i-cecile:
```suggestion
This website is built using [Zola](https://www.getzola.org/): install it and use `zola serve` to examine the website as it would appear.
```

CONTRIBUTING.md
```diff
@@ -98,6 +98,16 @@ This path should prioritize:
    5. Contributing to Bevy itself.
 6. Linking to further reading: official examples, `docs.rs` and (very sparingly) source code links.
 
+## Setup
```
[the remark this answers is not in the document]
alice-i-cecile: I could use more feedback and suggestions on this: I'm not very familiar with it.

I'm also not convinced it belongs here. In a lot of ways, this feels like it belongs in the READMEs of the generate-* folders?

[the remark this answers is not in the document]
IceSentry: Well, it's kinda part of the setup to run the site, so it should probably be in both the README and in here.

The only thing I would make sure to specify is to run the scripts directly in the folders because it won't work correctly if you run it from the top level folder.

IceSentry:
I believe this could be closed in favor of #372

PR #291, opened 2022-03-14
Commands section for ECS chapter

alice-i-cecile:
From #182.

# Status

- [ ] Review and revise


cart:
Closing as we will very likely start new PRs (and use this as a reference) when we pick up the new book effort again.

PR #292, opened 2022-03-14
Change detection section for ECS Chapter

alice-i-cecile:
Extracted from #182.

# Status

- [ ] Revisit and revise.


cart:
Closing as we will very likely start new PRs (and use this as a reference) when we pick up the new book effort again.

PR #293, opened 2022-03-14
ECS magic section for ECS chapter

alice-i-cecile:
Extracted from #182.

# Status

- [ ] Discuss framing: critical that new users don't feel compelled to understand this
- [ ] Revisit and revise


james7132:
Tempted to say we should have a separate section for "engine internals" akin to the Rustonomicon that explains the structure of internals and the type system magic that is employed. It's not needed for most users in for practical purposes. We can still point people who are genuinely curious there, but it shouldn't be right there on the sidebar for an intro to the engine.

alice-i-cecile:
> Tempted to say we should have a separate section for "engine internals" akin to the Rustonomicon that explains the structure of internals and the type system magic that is employed. It's not needed for most users in for practical purposes. We can still point people who are genuinely curious there, but it shouldn't be right there on the sidebar for an intro to the engine.

Yep, this is one of my big questions about this section. It's clearly valuable (I've pointed half-a-dozen grateful users to it already), but it's not at all clear that it belongs as part of the ECS chapter, or even the Bevy book.

doup:
As a newcomer, I got a similar sensation while reading the Commands section: https://github.com/bevyengine/bevy-website/pull/291 . Apart of the "engine internal", maybe some of these can be moved to a "how to…" section?

alice-i-cecile:
Closing this out. It's way too hard to keep up-to-date and of questionable value to readers at this level.

That said, expect a blog post on the topic from me in the future.

PR #301, opened 2022-03-16
Point to main branch of bevy for code validation

alice-i-cecile:
This allows us to verify that our code *will* work on 0.7, even if it currently doesn't on 0.6.

Obviously this should be revised shortly after each crates.io release.

### DIFF
--- a/code-validation/Cargo.toml
+++ b/code-validation/Cargo.toml
@@ -4,4 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-bevy = "0.6.0"
+# This should point to `main` when we are preparing for the next release,
+# but should be fixed at the latest release once it is published on crates.io
+# bevy = 0.6
+bevy = {git = "https://github.com/bevyengine/bevy"}


code-validation/Cargo.toml
```diff
@@ -4,4 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-bevy = "0.6.0"
+# This should point to `main` when we are preparing for the next release,
+# but should be fixed at the latest release once it is published on crates.io
+# bevy = 0.6
```
KDecay, line 9:
```suggestion
# bevy = "0.7"
```

code-validation/Cargo.toml
```diff
@@ -4,4 +4,7 @@ version = "0.1.0"
...
+bevy = {git = "https://github.com/bevyengine/bevy"}
```
KDecay, line 10:
```suggestion
bevy = { git = "https://github.com/bevyengine/bevy" }
```

KDecay (changes requested):
Makes sense to me! I changed the bevy version to "0.7" because that's the version we would have to insert on the next revision.

KDecay (approved):

Issue #307, opened 2022-03-30
Improve layout


doup:
Hi! I've seen the following layout related issues:

- https://github.com/bevyengine/bevy-website/issues/135
- https://github.com/bevyengine/bevy-website/issues/213
- https://github.com/bevyengine/bevy-website/issues/188

I think all of them can be improved at the same time if the layout is revisited. I've been looking around and I think [Docusaurus](https://docusaurus.io/docs) solution is quite good. It would allow:

- Header elements in desktop that hide into on the hamburger menu on mobile
- Call to actions (e.g. donate, get started), which stay visible in the header on mobile (until a breakpoint)
- Documentation menu that plays nicely with the main menu

Here is what I'm referring to:

https://user-images.githubusercontent.com/188612/160906129-5a71f62d-04ca-4070-aa31-ec68423473b5.mp4

- This is how the menu looks like on the `docs` section
- It clashes with the main menu, so it has two "levels" on the menu: docs menu & main menu
  - This could be applied only for the docs section
- When clicking on the hamburger the documentation menu shows
- But, the main menu can also be accessed with the `<-- Back to main menu` option

I could give it a try. Any thoughts?

alice-i-cecile:
I like that style, and I think this is a good design to learn from :) I agree that these should be tackled at once.

cart:
> It clashes with the main menu, so it has two "levels" on the menu: docs menu & main menu 

"Multiple hierarchies of menus" does bury top level navigation info in a way im not ecstatic about, but we do need to make a compromise somewhere and this is the best I've seen so far. I also like that this buys us space for doc-specific functionality like search buttons (and space for the bevy logo).

I think if we make tapping the bevy logo link to the hoe page,  that might help improve the discoverability of the "buried" content.

In short: im on board.

IceSentry:
Are you saying that you want to integrate Docusaurus or that you want to implement a solution that looks like it?

cart:
I read this as "implementing a solution that looks like it" / thats what I was agreeing to.

doup:
Yeah, sorry if I wasn't clear, I meant I want to copy their UX solution for the doble-menu problem.

doup:
Hi!

[...]

PR #314, opened 2022-04-01
[Merged by Bors] - Organise SCSS files by folder

doup:
- In preparation for #307 I would like to reorganize the SASS files into folders.
- The idea is to have folders by CSS areas: html elements, components (a.k.a. "repeating visual patterns") & pages. Root files are reserved for "core" SASS files: entry file, vars, functions, mixins… This should make the styling files easier to navigate.
- I don't like it, but I had to add a couple of `!important` on `.container` class, as it previously was `@import` order dependant and the reorder broke some styling. **If this is a big issue I can put some time to make a proper fix.**


cart:
Yeah we shouldn't be using `!important` (and I think its worth waiting for a fix there), but I do like the general direction here.

doup:
OK, fixed. Actually it was quite easy… 😇

IceSentry (approved):

PR #315, opened 2022-04-01
Improve header navigation

doup:
Builds on top of: #314 
Fixes: #307


CooCooCaCha:
I'm encountering a strange issue when I use Chrome DevTools or Firefox Devtools. When the screen is narrow the page appears to extend past the width of the screen. I think it might be due to an issue with position: fixed.

bevy_header_issue

CooCooCaCha:
I just confirmed it happens on an actual mobile device too.

sass/_functions.scss
```diff
@@ -0,0 +1,11 @@
+@function calc-rem($size) {
```
CooCooCaCha, line 1: Curious what the purpose of this function is. Why not just use `rem` units?

sass/_functions.scss
```diff
@@ -0,0 +1,11 @@
...
+    @return ($size / $size-body) * 1rem;
+}
+
+@function darken-color($color) {
```
CooCooCaCha, line 5: Why saturate and darken vs just darken?

IceSentry (commented):
At a quick glance the code looks good, I'll need to test it though, 

It's a lot of changes and I wonder if some of that could be done in smaller PRs. I guess seeing the diff after #314 will help simplify this PR.

IceSentry:
When I scroll down to the end the menu goes up which doesn't look like it's intended. I also confirm that I see the same thing that @CooCooCaCha mentioned where the page doesn't scale properly on mobile.

sass/_functions.scss
doup: I feel more natural to use `px`, matter of preference I suppose. The ideal though would be to use a limited set of size tokens (a "palette"). I _feel_ that it'll be easier to gather the set of sizes being used and decide on the "palette" if these are expressed on `px`.

sass/_functions.scss
doup: Sincerely, I don't remember exactly. I've copied few of these helpers from other projects. The idea I think was that darken/lighten are more pleasant when are combined with saturate/desaturate, but probably is super questionable. 😇 

Here, same as with font-sizes, the ideal would be to just have a set of color tokens and choose from there, instead of using these helpers.

Maybe there is a designer in the community willing to prepare a design system… 😅

IceSentry:
I'm not sure why, but with you latest commits I have a permanent vertical scrollbar that shows up.

doup:
I've fixed the issue with the header, the thing was that some pages overflowed horizontally and that broke the header. To be future proof I've limited the header width to `100vw` and changed to `overflow-x: hidden` on the content of the pages with menu (docs/assets/examples). But the real fix has been to tweak several pages to remove any source of horizontal overflow, notably:

- Make sponsors section responsive
- Tweak code blocks
- Improve assets grid

doup:
@IceSentry can you try again? I've found an explicity `overflow-y: scroll` on `html`, that was probably the culprit.

doup:
Oh, btw, **on the menu being cut**. Does this issue happen when you're at the **top of the page**, or **when you reach the bottom**?

- If it's **at the top**, then it might be because your browser doesn't support `position: sticky`… in that case I'll need to find another solution (or revert to the menu at being static).
- If it's **when reaching the bottom**, that's a limitation of `position: sticky`; that's the best I've achieved while allowing to scroll on the menu when the window height is small (e.g. in `assets` there is a long menu, if the window height is small it allows to scroll on it).

IceSentry:
It's when reaching the bottom of the page.

As for the vertical scroll bar, it's still there for me.

sass/_functions.scss
CooCooCaCha: I tend to prefer px too, however, I also think it's generally a good idea to stick with the patterns that are established in a codebase unless you want to start a conversation about px vs rem.

Eventually I think it would be a good idea to create a theme file with arrays of preset values. So instead of using `16rem` you'd use variables like `font-sizes-0` and `font-sizes-1`. That way values stay standardized across the codebase, except for special cases.

sass/_functions.scss
CooCooCaCha: Gotcha, I was more worried about the name since it's easy to confused `darken()` with `darken-color()`. Perhaps a more descriptive name?

IceSentry:
The vertical scrollbar seems to be fixed now.

As for the menu in the book, could you just make it sticky like the header?

sass/_functions.scss
doup: Fair enough. On the smaller PRs I'm using `rems` instead of `calc-rem`. I'll change this branch also to use `rem`. I'll leave this comment open as a reminder.

sass/_functions.scss
doup: The name certainly is not very inspired. 😓  In other projects I'm using the Dart implementation which uses the new Sass modules, Zola uses the older `libsass` (which rises some questions about how that will be maintained, but that's another issue) and had to change the implementation… This is cleaner, but unsupported:

```
@use 'sass:color';

@function darken($color) {
  @return color.scale($color, $lightness: -5%, $saturation: 5%);
}
```

Anyway, to the point, I might remove these and use some color variables instead.

sass/_functions.scss
doup: I've removed the `calc-rem` usage from the components, I've left it on `html` though, I think that's acceptable.

sass/_functions.scss
doup: I've removed the color functions. The solutions here should to use a palette… but let's post-pone that. ^_^

doup:
@IceSentry @CooCooCaCha with the intention to make this PR smaller I've been merging few improvements in other PRs. This one is still big but I don't think it can be squeezed much more.

Whenever you can, can you please review it again?
Thanks!

IceSentry:
Alright, there's still the book menu that gets hidden by the nav bar when you scroll all the way down, but it's not a super big deal. Other than that I haven't noticed anything wrong visually. I'll do a final pass on the code before approving.

IceSentry (approved):
LGTM

doup:
I think that there isn't much to do with the menu, it's how `position: sticky` behaves. See for example Procreates manual: https://procreate.art/handbook/procreate/interface-gestures/ once you reach the footer it happens the same.

You mentioned `position: absolute` in another message, I can give it a try, but I've the impression that getting the `left` position might be tricky/ugly…

Btw, I noticed I forgot to add the "section title" part on the header… **I need to add it back.** 😓 
This 👇 
image

doup:
OK, fixed:
image

IceSentry:
Oh, good catch with the section title, I was so focused on everything else when testing it I didn't even realize it was gone 😅

cart:
I dig the changes functionally. Just a few changes to address:

1. The text sections now "bleed out" of the bounds of their corresponding images and also feel "off center" now.
2. The text on the donate button feels "too high" now
3. I'd like the "Features" text to be slightly bigger and bolder
4. I think the navigation text should all be a bit bolder.

image

doup:
I've fixed all the issues you pointed out @cart.

For the first one (the feature page layout) I've just made it similar to what's now on production, and I've also improved responsive. It's not very good, but at least it doesn't break on mobile/tablet… The homepage needs some love, but that's for another PR.

cart:
Love the changes. The potential for "feature image background aspect ratio" changes when scaling the width down is slightly regrettable, but the improved use of space during scaling feels like a worthwhile trade to me. 

Great work!

PR #318, opened 2022-04-03
[Merged by Bors] - Switch to using .path instead of .permalink. This makes links relative.

CooCooCaCha:
When testing Vercel integration it was discovered that some links (like the news feed) would link to the production website instead of the preview site. This is because all permalinks in zola use the `base_url` from the config file, which is hardcoded.

You can also check this locally by running `zola serve` and using the url `localhost:1111` in your browser. Clicking on a news post will redirect you to `127.0.0.1:1111` because `127.0.0.1` is the `base_url` in development.

This PR switches from using `.permalink` to `.path` which makes these links relative.

### DIFF
--- a/templates/book-macros.html
+++ b/templates/book-macros.html
@@ -1,7 +1,7 @@
 {% macro book_nav_section(section_path, section_number) %}
 {% set section = get_section(path=section_path) %}
 <li class="book-nav-section">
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

alice-i-cecile:
Makes sense, and there's no harm in fixing this.

IceSentry (approved):

PR #319, opened 2022-04-04
[Merged by Bors] - Revised the title of news

C-BJ:


alice-i-cecile (approved):
I think this is the right call, and much clearer. It looks like this will fix the tab title on https://bevyengine.org/news/. 

This brings us into consistency with the actual display on the website.

doup (approved):

PR #321, opened 2022-04-04
[Merged by Bors] - Organize template files into folders

doup:
Organize template files into folders:
- `layouts`
- `macros`

### DIFF
--- a/templates/anchor-link.html
+++ b/templates/anchor-link.html
@@ -2,8 +2,8 @@
 -------------
 DO NOT REMOVE
 -------------
-This template is used to generate the header anchor links in docs/news/…
-This is configured by the fron-matter `insert_anchor_links` on markdown files.
+This template is used to generate the header anchor links (`#`) in docs/news/…
+This is configured by the `insert_anchor_links` front-matter option located on the markdown files.
 See documentation here: https://www.getzola.org/documentation/content/section/#front-matter
 #}
 <a class="anchor-link" href="#{{id}}">#</a>


templates/anchor-link.html
```diff
@@ -1 +0,0 @@
-<a class="anchor-link" href="#{{id}}">#</a>
```
alice-i-cecile, line 1: Do we no longer need this? It appears to have been simply deleted.

alice-i-cecile (commented):
I like this :) Just one quick question about anchor-link.

templates/anchor-link.html
IceSentry: I think it was never used as a template. Every link seems to have been inlined

IceSentry: Actually, I'm not sure that's true, for some reason with this PR the anchor-links in the book are not using the `#` symbol like everywhere else, but this PR doesn't touch that.

IceSentry: Alright, found it, it's not super obvious, but when using `insert_anchor_links` inside the zola header thing in the markdown file it will insert that template so it is in fact needed, it's just not used anywhere directly so it's a bit weird. 

@doup I would revert deleting that file and add a simple comment explaining that just so nobody deletes it again in the future.

doup: Reverted, and sorry. ^_^U

Btw, I couldn't make Teras white-space removal work with the comments, so the HTML begins just after the comment and I've removed the end-of-line.

IceSentry: What's "Teras white-space removal"?

doup:
I'm not sure about that, I don't see the criteria clear. As I see it, at root, all but `anchor-link` are page templates, no? I have the impression we would have to move all of them to `pages`… which could be good, clearer.

The downside is that we'll need to update the front-matter of the markdown files.

IceSentry:
Right, in my head things like book-section were just a "nested template" of book, but looking at it again it's indeed all specific pages.

Instead of separating it by pages, I'd separate it based on the location. For example, we could have a `/learn` folder that contains the learn page but also all the book/examples/asset stuff.

So, something like this:

```markdown
* templates
  * layouts
  * macros
  * shortcode
  * pages
    * learn
      * learn
      * book
      * book-section
      * example
      * examples
      * assets
    * news
      * news
      * news-page
    * community (since it's a single file it doesn't need a folder)
    * index
```

IceSentry:
At this point it doesn't matter that much. I'll approve it as is, but that's what I had in mind.

IceSentry (approved):

templates/anchor-link.html
doup: I'm refering to this: https://tera.netlify.app/docs/#whitespace-control
And there are users that have make it work: https://github.com/Keats/tera/issues/641

You can add a dash `-` on the closing comment tag and in theory it should remove the line-break between the comment & the link:

```
{# some comment -#}
<a>Link</a>
``` 

I was mentioning this because this piece of HTML gets added on the headers, and the line-break adds a space between the header text and the `#`…

Actually, I'll add line breaks as it'll make the template more readable, and that extra space it's not noticeable…

Hahaha, I sometimes get lost in this small details… 😅

templates/anchor-link.html
```diff
@@ -1 +1,9 @@
...
\ No newline at end of file
+{#
+-------------
+DO NOT REMOVE
+-------------
+This template is used to generate the header anchor links in docs/news/…
+This is configured by the fron-matter `insert_anchor_links` on markdown files.
```
IceSentry, line 6:
```suggestion
This is configured by the front-matter `insert_anchor_links` on markdown files.
```

doup: Fixed!

Issue #313, opened 2022-04-01
Host dev branch using vercel


IceSentry:
We want to be able to host a dev version of the website to make collaboration easier. The main issue stopping us is that github pages can only host a single branch.

Fortunately, there's another way. We could use [vercel](https://vercel.com)/ to host multiple branches for free and it's super easy to setup.

CooCooCaCha:
A few notes from testing the vercel <-> github integration on my fork.

- Vercel will automatically detect branch updates and deploy that branch to a unique (and ugly) url.
- Vercel will automatically comment on PRs with the generated url. I believe Vercel can be configured to not submit this comment.
- If a PR is updated, Vercel will redeploy and edit the original comment if needed.
- These urls will not be indexed by search engines. This is done by using the `x-robots-tag: noindex` HTTP header (https://vercel.com/support/articles/are-vercel-preview-deployment-indexed-by-search-engines#x-robots-tag-header).

Here's a screenshot of what the above looks like:
![vercel_github_screenshot](https://user-images.githubusercontent.com/850228/161401757-0b6578fc-7656-4280-ae7d-8bda4d94cd88.png)

alice-i-cecile:
Cart and I have been chatting with Vercel's sales team: they'd be open to offering us the equivalent of their pro plan if we're willing to credit them for their hosting. The guidelines are [here](https://vercel.com/support/articles/can-vercel-sponsor-my-open-source-project), but we're discussing to see if we can compromise on something more consistent with the other sponsorships.

alice-i-cecile:
So, for acknowledgement, we need:

1. A Infrastructure Sponsors section of our Sponsors section.
2. A link in the footer with the Vercel logo and OSS link (as described above), with the phrase "Website powered by Vercel".
3. A section in the README of the `bevy-website` repo for sponsors (with all of our Bronze+ tier sponsors).

This is ready for a test PR enabling Vercel integration + the acknowledgements, which we can then send to their sales team for approval.

cart:
> 2. A link in the footer with the Vercel logo and OSS link (as described above), with the phrase "Website powered by Vercel".

I think we should try pushing back on this, given that the website isn't really "powered" by it. We're using it as a dev tool to improve our PR efficiency, not as a technology stack for the website (or as a host for the production website).

alice-i-cecile:
[...]

PR #325, opened 2022-04-05
Vercel Build Script

CooCooCaCha:
This PR adds a custom build script so that vercel correctly compiles bevy assets before running `zola build`.

Use this by changing your vercel build command to `./vercel_build.sh` in your vercel settings
Screenshot 2022-04-05 175027
.


alice-i-cecile:
Part of #313.

IceSentry:
So, is that just a vercel feature? Like they look for any `./vercel_build.sh` and run it?

CooCooCaCha:
> So, is that just a vercel feature? Like they look for any `./vercel_build.sh` and run it?

You can actually just change the build command in your settings.

IceSentry:
Oh, that's pretty nice then.

mockersf:
is a custom script also needed to build examples? https://bevyengine.org/examples/

CooCooCaCha:
> is a custom script also needed to build examples? https://bevyengine.org/examples/

Good catch, I didn't know that existed. Working on it.

CooCooCaCha:
Updated the script to compile bevy examples. You can check it here https://bevy-website-hd0mg2wxd-wrjstuff.vercel.app/examples

Compiling examples also bumped up the build time to 9min 20s.

IceSentry:
Since it's for dev branches and quick iterations, could we skip the examples? It's a bit long especially if most of the dev work won't be on that page.

CooCooCaCha:
> Since it's for dev branches and quick iterations, could we skip the examples? It's a bit long especially if most of the dev work won't be on that page.

Not a bad idea. I'll leave the decision up to someone higher up the chain.

For the record the base website takes about 10s to build. With assets 1min 12s. With examples 9min 20s.  

Something we could do in the future is move examples to a separate repo and pre-build assets and examples. That way building the website (for prod and dev) would be super fast.

alice-i-cecile:
> Not a bad idea. I'll leave the decision up to someone higher up the chain.

I'd prefer to skip examples: they're rarely going to be relevant to this repo and the added friction is high enough that I'd prefer to just skip them.

> Something we could do in the future is move examples to a separate repo and pre-build assets and examples. That way building the website (for prod and dev) would be super fast.

Agreed here. It would also be *very* interesting to have web-available examples for each `bevyengine/bevy` PR to allow reviewers to quickly test out relevant functionality.

mockersf:
Can the example be built conditionally if the relevant css/html files are modified?

Or can the page be built without building the wasm, including a dummy content? The example pages are currently in dire need of a better style and more presence on the site

CooCooCaCha:
> Can the example be built conditionally if the relevant css/html files are modified?
> 
> Or can the page be built without building the wasm, including a dummy content? The example pages are currently in dire need of a better style and more presence on the site

I'm still figuring out how all that works but perhaps we could add a flag to the examples build script that skips the wasm step? Theoretically, the pages would still work but the examples themselves would appear blank.

In a followup PR we could do something more elegant like replace the examples with placeholders.

CooCooCaCha:
Ok I updated the script to disable WASM compilation only for vercel builds. 
This brings the compilation time to around 1min 30s. 

When WASM is disabled it just shows a blank space on the page as can be seen here: https://bevy-website-55e99xe2w-wrjstuff.vercel.app/examples

In the future we could potentially use git to see if any WASM files have changed and conditionally set the SKIP_WASM flag.

IceSentry (approved):

alice-i-cecile (approved):

alice-i-cecile:
We got ghosted on some of our questions, so I'm going to close this out and we'll go with a competitor...

cart:
In fairness I was the one that ghosted them right after they called out that we can deploy without their "overly permissioned" github integration by using their CLI. There is a chance we can still make this work, although idk if we could make the user experience great without writing additional automation to "close the loop" by somehow getting the preview url from the CLI and having a bot leave it in a comment on the relevant pr.

alice-i-cecile:
Ah, fair enough. All the same, I think I'd rather poke at Netlify's version of this feature rather than sinking more time in here.

PR #326, opened 2022-04-06
Improve feature sections responsive behavior

### LINKED ISSUE #308: Compress Feature page layout

doup:
- Improve features responsive behavior. Now it'll change the layout based on the width and orientation.
- Group classes under "feature" component
- Remove "reversed" classes in favor of `:nth-child(even)` selector

This is another preparation PR for #315 
Fixes #308

https://user-images.githubusercontent.com/188612/161916805-b31ab80b-ef63-4ab5-89bc-7f71a99b6a95.mp4

### DIFF
--- a/sass/components/_feature.scss
+++ b/sass/components/_feature.scss
@@ -4,13 +4,17 @@
     gap: 16px;
     width: 100%;
     margin: 0 auto;
-    max-width: 500px;
+    max-width: 450px;
     grid-template-areas:
-        'text'
         'image'
+        'text'
     ;
 
-    @media #{$bp-feature-horizontal} {
+    * {
+        box-sizing: border-box;
+    }
+
+    @media #{$bp-tablet-portrait-up} {
         max-width: 1000px;
         grid-template-columns: 1fr 1fr;
         grid-template-areas: 'text image';
@@ -18,8 +22,12 @@
     }
 
     &:nth-child(even) {
-        @media #{$bp-feature-horizontal} {
+        @media #{$bp-tablet-portrait-up} {
             grid-template-areas: 'image text';
+
+            .feature__image-wrapper {
+                justify-content: flex-end;
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


rparrett:
I was initially confused about the "BEM naming," but it seems logical.

Tested in MacOS Chrome/Firefox and iOS Chrome/Safari and looks good to me.

IceSentry:
I like it, it fixed the issue when I use a bigger default font-size!

sass/components/_feature.scss
```diff
@@ -0,0 +1,71 @@
+.feature {
+    display: grid;
```
IceSentry, line 2: I completely forgot about using `display: grid` when I was trying to fix it. It's perfect for this.

IceSentry (approved):
LGTM

doup:
@rparrett for a static site like this, it gives you poor man's encapsulation for the CSS (less chances to pollute & be polluted by other CSS classes; less specificity issues…) and the HTML is kind-of self documenting. The downside is its verbosity… I like that though. ^_^

cart:
I do like that this is better at using available space, but I'd prefer it if we could maintain the aspect ratios of the image containers:

image

Not a huge fan of the "squashed" look.

cart:
Or at least, I'd like to do a side by side comparison where it uses available space _and_ maintains aspect ratios. Its possible that maintaining aspect ratios _and_ using available space will feel "too big".

cart:
I just noticed that the blender "whats new" feature overview is very similar to ours, and they manage to maintain aspect ratios, but they use a "wider" one than we currently do:
https://www.blender.org/download/releases/3-1/

doup:
This one is tricky, making it fixed aspect-ratio in some cases (e.g. `2D Renderer`) makes it look too big. In Blender they are using the landscape aspect ratio for all the images; I'm not sure if it applies in this case. So instead, I've changed how the background color behaves, for mobile, instead of taking all the width it shrinks to the image.

I don't know, I'm still not convinced. Maybe is better to leave it as it was. 😓 

https://user-images.githubusercontent.com/188612/162588290-76b84bee-eceb-4d55-b62e-50090ecb09f0.mp4

SIGSTACKFAULT (approved):
Looks good but after a quick rebase it looks the same as main? Did this get rewritten and merged somewhere else?

doup:
I've compared with `main` and doesn't seem to have a lot of conflicts… in any case, I've toyed a little bit with it and I'm still not convinced. The images appear with different sizes in some breakpoints… apart from other issues.

I might give it a try again (starting fresh from `main`), but for now I think I'll just close this PR. If anyone wants to give it a try, the interesting file is `sass/components/_feature.scss`.

PR #290, opened 2022-03-14
[Merged by Bors] - Introduction for ECS chapter

alice-i-cecile:
Extracted from #182.

# Status

- [x] Simplify example
- [x] Avoid database analogy
- [x] Edit and revise for clarity and quality
- [x] Fix dead links
- [x] Refactor Breakout example on Bevy repo
- [x] Update text to sync closely with Breakout example

PR #327, opened 2022-04-07
[Merged by Bors] - Improve side menu in Book, Assets & Example

doup:
- In preparation for #315, add new `tree-menu` CSS component
- Update Books, Assets & Example to use this new component
  - In Books only the active book section is automatically opened, other sections with sub-sections start closed
  - Assets/Examples are always open, but are collapsible

https://user-images.githubusercontent.com/188612/162277066-bf9ab366-79ce-402e-af02-691ed2c4a279.mp4

https://user-images.githubusercontent.com/188612/162277084-cbbd3a9a-d436-4388-a7df-8e34d77433f0.mp4

https://user-images.githubusercontent.com/188612/162277125-924d7f84-b66a-47fb-ae08-c771a6a80a5a.mp4

### DIFF
--- a/sass/_mixins.scss
+++ b/sass/_mixins.scss
@@ -1,0 +1,5 @@
+@mixin flex-center() {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -28,6 +28,10 @@ $syntax-theme-background-hover: #414247;
 $default-image-background-color: #1b1b1b;
 $card-hover-background: #2f3033;
 
+// Animation
+$duration-fast: 100ms;
+$duration: 250ms;
+
 // General
 $content-top-margin: 30px;
 $default-padding: 15px;
--- a/sass/components/_book-nav.scss
+++ b/sass/components/_book-nav.scss
@@ -1,7 +1,7 @@
 $inactive-nav-color: #999999;
 $book-nav-width: 210px;
 
-@media screen and (max-width: 900px) {
+@media screen and (max-width: 980px) {
     #show-book-nav:not(:checked) + .book-nav {
         transform: translateX(calc(0px - #{$book-nav-width}));
     }
@@ -16,10 +16,12 @@ $book-nav-width: 210px;
 }
 
 .book-nav {
+    box-sizing: border-box;
     margin-top: $default-padding;
     flex: 0 0 $book-nav-width;
     user-select: none;
     transition: transform 0.3s;
+    padding-left: 8px;
 }
 
 .book-nav-sections {
--- a/sass/components/_tree-menu.scss
+++ b/sass/components/_tree-menu.scss
@@ -1,0 +1,103 @@
+.tree-menu {
+    $hover-color: rgba($color-white, 0.1);
+    $item-height: 32px;
+    $border-radius: 4px;
+
+    margin: 0;
+    padding: 0;
+    position: relative;
+
+    // TODO: REMOVE THIS SNIPPET ONCE THE HEADER IS REFACTORED
+    // vvv
+    z-index: 2;
+
+    * {
+        box-sizing: border-box;
+    }
+    // ^^^
+
+    &__state {
+        display: none;
+    }
+
+    &__item {
+        list-style: none;
+
+        > .tree-menu {
+            display: none;
+            padding-left: 16px;
+        }
+    }
+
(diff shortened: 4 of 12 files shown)


cart:
Oooh this is pretty slick. I'll give it a run.

cart:
Some assorted thoughts:

* I was skeptical at first, but removing the section and subsection numbers feels like a win. It buys more room in the sidebar and reduces the general feeling of "clutter". And these sections in practice are "unstable". You cant really safely refer to "section 2.4" anywhere because theres a possibility that a new section or subsection will be inserted in front of it.
* Triple-nested sections should be relatively uncommon, but I'd appreciate it if we tested that arbitrary nesting still works
* I sort of miss the "deselected items get out of your way via reduced lightness" design, rather than using blue text and box highlights to denote selected / hovered items. It adds a level of dimensionality that I liked. That being said, the new approach is much more standard and probably more legible to some people. Before merging I'll probably play with applying the "old" approach, just to see how I feel about it in the new context.
* I like that we're now using the whole area of a section as the "hit box" rather than the text itself. The old behavior was a nasty little paper cut. 
* I think I'm on board for the addition of casing to sections. I initially opted for lowercase to emphasize the "get out of your way" design (and to add a bit of character), but I think casing makes it feel more professional + gives us the ability to encode more semantic information, when needed.
* What is the license and provenance of the chevron svg?

New vs Old Image (for a side by side comparison):
image

BlackPhlox:
Since we're redesigning the side menu. We already have a dedicated _column_ for the side menu, wouldn't it be a good idea to have the side menu fixed when having desktop width? Since there is no fixed burger menu when scrolling down for desktop means that there is no quick access to the side menu, and users have to be at the top.

doup:
Now I see that maybe I over-stepped on some decisions, sorry for that. In any case any of the points you mention should be easy to revert (although, looks like it won't be necessary 😅). Dimming of the non-selected items (as it's now in prod), sounds like a nice UX win; let me know if you prefer me to look into it.

I did the chevron myself in Figma; it's actually in the same "file" as the chart in #290. Same for the hamburger and the "X" icons used in #315. I gave @alice-i-cecile the ownership of that file… but now that I think it, probably you should be the owner. ^_^

Oh, btw, the Tera macro is recursive so nested menus work as well. I just copied the "Getting Started" folder inside "FAQ" and it just works:

image

@BlackPhlox yup, the sticky menu is done in #315. I've actually made this PR because of that PR, on mobile the docs menu looks a little bit off: small text, with small hit bboxes. Once I sync with this PR the mobile menu should look muuuuch better.

doup:
Oh, btw, full disclaimer, the idea/design is a copy from: https://docusaurus.io/docs
The code is implemented from scratch though (and without JS).

alice-i-cecile:
> I was skeptical at first, but removing the section and subsection numbers feels like a win. It buys more room in the sidebar and reduces the general feeling of "clutter". And these sections in practice are "unstable". You cant really safely refer to "section 2.4" anywhere because theres a possibility that a new section or subsection will be inserted in front of it.

Agreed.

> Triple-nested sections should be relatively uncommon, but I'd appreciate it if we tested that arbitrary nesting still works

I should be able to work around this in the book, but it can't hurt if it works.


> I sort of miss the "deselected items get out of your way via reduced lightness" design, rather than using blue text and box highlights to denote selected / hovered items. It adds a level of dimensionality that I liked

I liked this too.

> I think I'm on board for the addition of casing to sections. I initially opted for lowercase to emphasize the "get out of your way" design (and to add a bit of character), but I think casing makes it feel more professional + gives us the ability to encode more semantic information, when needed

Yeah, I think that I'm fine either way here.

>  I gave @alice-i-cecile the ownership of that file… but now that I think it, probably you should be the owner. ^_^

Ideally we'd have a Bevy organization, or a shared `Bevy` user account :)

cart:
> I gave @alice-i-cecile the ownership of that file… but now that I think it, probably you should be the owner. ^_^

For now, if you can agree to give me full ownership / usage rights (or alternatively ... just license it permissively), I'd appreciate it. Until we get a legal Bevy org spun up, Id prefer to be the single steward of "legal assets" to simplify the transition, when it comes.

doup:
I agree to give full ownership/usage rights of `static/assets/icon-chevron-down.svg` to @cart.

cart (approved):
Brilliant! I'm happy with the dimming. I think this is good to go

PR #329, opened 2022-04-08
Tweak error-prone Bevy fast-compiles advice

alice-i-cecile:
# Problem

Currently, the Bevy book encourages fussing with complex fast compilation setups.

Fast compile advice is very cool, but comes with a *large* numbers of caveats and often breaks in surprising ways: https://github.com/bevy-cheatbook/bevy-cheatbook/issues/135

It should not be introduced to beginners, even if it may be valuable to advanced users later.

## Solution

- remove advice to set up fast compiles

This advice can and should be added back later, in a more appropriate location with carefully vetted caveats.

### DIFF
--- a/content/learn/book/getting-started/_index.md
+++ b/content/learn/book/getting-started/_index.md
@@ -41,6 +41,7 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
 
 Bevy is [available as a library on crates.io](https://crates.io/crates/bevy).
 
+
 Add the bevy crate to your project's Cargo.toml like this:
 
 ```toml
@@ -51,3 +52,16 @@ bevy = "0.6" # make sure this is the latest version
 This is the current `bevy` crate version:
 
 <a href="https://crates.io/crates/bevy"><img src="https://img.shields.io/crates/v/bevy.svg" style="height: 1.7rem;"/></a>
+
+**_NOTE:_** Bevy is currently being updated at a rapid pace. Taking a dependency on the git repo instead of the cargo crate will allow you to receive the latest updates as fast as possible. *However*, **there are often breaking changes made to APIs and behavior**. This means that it will be important to keep up with the latest developments with bevy. **This is not recommended for people who are just getting started with bevy.**
+```toml
+[dependencies]
+bevy = { git = "https://github.com/bevyengine/bevy" }
+```
+
+In general it's a good idea to lock in to a specific commit hash, which gives you control over when you take updates. You can find the [latest commit hash here](https://github.com/bevyengine/bevy/commits/main) (to the right of each commit).
+
+```toml
+[dependencies]
+bevy = { git = "https://github.com/bevyengine/bevy", rev = "25f62f7250a0d750068dc32533b9433f7985af98" }
+```
--- a/content/learn/book/getting-started/_index.md
+++ b/content/learn/book/getting-started/_index.md
@@ -41,7 +41,6 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
 
 Bevy is [available as a library on crates.io](https://crates.io/crates/bevy).
 
-
 Add the bevy crate to your project's Cargo.toml like this:
 
 ```toml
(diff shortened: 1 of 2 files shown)


cart:
The main reasons I chose to include it early:
* Iteration times are critical for gamedev. Slow compile times == bad impressions. We dont want to lose potential users early because they hate waiting around for 8 seconds (when they could be waiting < 2).
* Doing the fast compiles setup early (ideally before the first compile)  means they dont need to "fully recompile" a second time, which could also leave bad impressions

I do agree that things should _easily_ just work out of the box. I think ideally we recommend platform specific setups that are "approximately guaranteed to work". I do agree that the current recommendations make it too possible for users to "fail".

cart:
We could also force compilation failures with solid error messages for platforms where dynamic linking wont work (ex: wasm).

alice-i-cecile:
I agree with those positions, but have two concerns:

1. "Time to fun" is too high. There's already a frustrating amount of yak-shaving, especially for users who need to install a large number of depedencies.
2. The existing setups are too unreliable. I would much prefer to be able to link to e.g. the bevy_ci_template that we built for the jam with all of the little steps set up for each platform.

> We could also force compilation failures with solid error messages for platforms where dynamic linking wont work (ex: wasm).

I really like this.

alice-i-cecile:
I've removed the commit about depending on `main`: that's much less controversial and should be its own PR.

cart:
Maybe theres a middle ground solution: only introduce the `--dynamic` feature in the beginning. Only encourage its use on platforms that support it out of the box (and link to a separate long-form fast compiles document for platforms with more complicated setups).

alice-i-cecile:
> Maybe theres a middle ground solution: only introduce the --dynamic feature in the beginning. Only encourage its use on platforms that support it out of the box (and link to a separate long-form fast compiles document for platforms with more complicated setups).

Yeah, I think this may be the way. Overall this deserves a comprehensive plan; I'll chew on what I think this should look like.

superdump:
> The main reasons I chose to include it early:
> 
> * Iteration times are critical for gamedev. Slow compile times == bad impressions. We dont want to lose potential users early because they hate waiting around for 8 seconds (when they could be waiting < 2).
> * Doing the fast compiles setup early (ideally before the first compile)  means they dont need to "fully recompile" a second time, which could also leave bad impressions
> 
> I do agree that things should _easily_ just work out of the box. I think ideally we recommend platform specific setups that are "approximately guaranteed to work". I do agree that the current recommendations make it too possible for users to "fail".

If fast compile has not been up front like this then I would likely not have looked them up and then wasted a bunch of time for a long time. I feel like it’s good that it’s there in the getting started so you get a good environment setup. If it’s a stumbling block then we could provide a helper script or something to get going?

inodentry:
I agree with @cart and @superdump that configuring for fast compiles is important and we should keep this section.

I think the real issue is that this section recommends some controversial/experimental things, that lead users to run into issues if they follow the advice blindly without fully understanding what they are doing.

1. Dynamic linking is great ... but we currently do not mention that it has issues on many platforms. It only works if you are iterating on a native desktop build, not if you are targeting WASM or mobile. IIRC it also has issues with Windows and MacOS (ive seen many ppl complain)? As in, it only reliably and easily works on Linux? This stuff should be noted.

2. Nightly compiler. This should be removed. Rust nightly has had regressions before, and the wins of experimental options like `-Zshare-generics` are pretty minor. This stuff falls into the category of "experimental tinkering for advanced users" IMO, not something we should recommend to all beginners.

The linker advice is great. Fast linkers like LLD, mold (speaking of, should we recommend mold instead of lld on linux?), zld, make a huge difference, and don't really break anything.

In all, I love that we are helping beginners get fast compiles. I also agree that we should be more careful with what exactly we recommend, so that users can get most benefit, without "over-optimizing" by enabling controversial things that may cause them issues for little extra gain.

speakk:
One option is to have the more advanced and error prone optimization steps moved to the end of the tutorial. This way time-to-fun is much faster, and those steps being optional is much more clear by the fact that they are at the end.

cart:
Closing as this PR doesn't have any actual changes / has stalled. Feel free to open a new PR that makes any of the changes discussed.

PR #331, opened 2022-04-09
[Merged by Bors] - Remove advice to depend on `main`

alice-i-cecile:
Relying directly on main breaks the ability to use vital ecosystem plugins, and exposes users to new bugs. It is no longer net useful due to the swap to train releases (and improved quality of the engine and ecosystem).

### DIFF
--- a/content/learn/book/getting-started/_index.md
+++ b/content/learn/book/getting-started/_index.md
@@ -41,7 +41,6 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
 
 Bevy is [available as a library on crates.io](https://crates.io/crates/bevy).
 
-
 Add the bevy crate to your project's Cargo.toml like this:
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
@bevyengine/docs-team, I'd like to merge this before 0.7 launches.

cart (approved):
I'm on board for this (including merging it as-is asap for the 0.7 release), but it does feel like a shame to lose these docs entirely. Ideally we just move these somewhere else.

alice-i-cecile:
> but it does feel like a shame to lose these docs entirely. Ideally we just move these somewhere else.

The Cheatbook's guide on this is *really* good, and very prominent. Down the road I want to absorb that guide instead, as part of a chapter on "advanced productivity tips": https://bevy-cheatbook.github.io/setup/bevy-git.html

PR #334, opened 2022-04-13
[Merged by Bors] - news: release 0.7.0

cart:
This is a draft of the Bevy 0.7.0 blog post.

Note that it is still _very_ incomplete. Sections are missing, as are some screenshots, videos, and "real numbers". 

I've included my "commit files" to help illustrate progress + organization. I've sorted roughly by category and then indented commits that have been covered. I dont intend to cover every commit, but its a rough indicator of progression.


alice-i-cecile:
How can we tell which commits you've already made a stub / writeup for? Do you remove them from https://github.com/bevyengine/bevy-website/pull/334/files#diff-d0be4ee95a3c5ed139808ec8389aef9130dd40a6582c32262d1db033910f5447 as you work?

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+* GLTF animation importing
+* Unlimited* point lights in a scene
+* Improved clustered forward rendering: dynamic/adaptive clustering, faster and more accurate cluster assignment
+* Compressed texture support (KTX2 / DDS / .basis): load more textures in a scene, faster
+* Compute shader / pipeline specialization: Bevy's flexible shader system was ported to compute shaders, enabling hot reloading, shader defs, and shader imports
+* Render to texture: cameras can now be configured to render to a texture instead of a window
+* Flexible user-customizable mesh vertex layouts in shaders
+* ECS improvements: Order systems using their names, Query::get_many, use conflicting parameters in systems via ParamSets, WorldQuery derives
+* More audio control: pause, volume, speed, and looping
+* Hot reloading for "plugin-provided / built-in assets"
```
alice-i-cecile, line 33: I think it's worth calling out that shader hot-reloading now works.

cart:
> How can we tell which commits you've already made a stub / writeup for? Do you remove them from categorized_commits.md as you work?

I think I already answered this in the PR description.

content/news/2022-04-12-bevy-0.7/index.md
cart: Yeah that was the motivating use case / worth calling out explicitly. We will also touch on that in the "in-shader import path" coverage, as those features are intertwined.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+        animation_player.play(animations.dance.clone());
+    }
+}
+```
+
+[`AnimationPlayer`] can also be used to animate arbitrary [`Transform`] components, not just skeletons!
+
+This critical feature has been a long time coming, but we wanted to build it in a way that meshed nicely with the [new Bevy renderer](/news/bevy-0-6/#the-new-bevy-renderer) and didn't just "hack things in". This builds on our new [User-Customizable Mesh Vertex Layouts](LINKME), [Shader Imports](/news/bevy-0-6/#shader-imports), and [Material](/news/bevy-0-6/#materials) systems, which ensures that this logic is flexible and reusable, even with non-standard meshes and custom render pipelines.
+
+And we're just getting started! Multi-track animation blending and higher level animation state management should arrive in the very near future. Now is a great time to start contributing animation features to Bevy. We've smashed through most of the foundational technical hurdles and what remains is largely high level api design choices!
```
james7132, line 68: It'd be great to refer/link to these RFCs directly. Hint hint nudge nudge wink wink.

james7132 (changes requested):

superdump:
I definitely think the benefits of compressed texture support in terms of load times and RAM usage warrant a section in the article.

superdump:
Oh, is the list of highlights at the top the list of sections that will be in the article?

I wonder if a before and after of performance since 0.6.1 would be good too. There should be a significant difference. I'll test in bistro and check.

alice-i-cecile:
> I wonder if a before and after of performance since 0.6.1 would be good too. There should be a significant difference. I'll test in bistro and check.

I would love to see both this and `ecs_bench_suite`. I don't expect serious changes on the latter, but it's worth checking if we can.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+* Render to texture: cameras can now be configured to render to a texture instead of a window
+* Flexible user-customizable mesh vertex layouts in shaders
+* ECS improvements: Order systems using their names, Query::get_many, use conflicting parameters in systems via ParamSets, WorldQuery derives
+* More audio control: pause, volume, speed, and looping
+* Hot reloading for "plugin-provided / built-in assets"
+* Power usage options to enable only updating Bevy Apps when input occurs 
+
+<!-- more -->
+
+## Skeletal Animation
```
alice-i-cecile, line 38: I think this section needs to mention that `gltf` animations work seamlessly: it's super cool and not obvious.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+    let a_transform = query.get_mut(entities.a).unwrap();
+}
+```
+
+However, to respect Rust's mutability rules, we need to disallow apis that might produce "aliased mutability". Seasoned Bevy users will probably recognize this Rust borrow checker error:
+
+```rust
+fn system(mut query: Query<&mut Transform>, entities: Res<SomeEntities>) {
+    let a_transform = query.get_mut(entities.a).unwrap();
+    // This line files to compile because `query` is already mutably borrowed above
```
nxsy, line 227: Typo: "files" -> "fails"

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+    let b_transform = query.get_mut(entities.b).unwrap();
+}
+```
+
+_You_ know Entity A and Entity B are different entities at runtime. But Rust's borrow checker has no way to know that at compile time! I'm sure you can imagine game development scenarios that would benefit from having mutable access to multiple components at the same time. This borrow checker restriction was a common pain point and the workarounds were ... not fun (using scopes to ensure conflicting accesses are dropped, copying data, re-querying things, etc).
+
+Fortunately, **Bevy 0.7** introduces a brand new set of apis to save the day!
+
+```rust
+fn system(mut query: Query<&mut Transform>, entities: Res<SomeEntities>) {
```
DJMcNab, line 237:
```suggestion
fn use_get_many(mut query: Query<&mut Transform>, entities: Res<SomeEntities>) {
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+fn give_sword(mut query: Query<&mut Items>) { 
+    for mut items in query.iter_mut() {
+        // No more .0!
+        items.push(Item::new("Flaming Poisoning Raging Sword of Doom"));
+    }
+}
+```
+
+Astute `std` doc readers might notice that the Rust team [recommends only using Deref/DerefMut for smart pointers, to avoid confusion](https://doc.rust-lang.org/std/ops/trait.Deref.html). Components like `Items` _are not_ smart pointers. We personally don't care. This pattern works, it is already widely used in the Rust ecosystem, and Good UX Comes First.
```
DJMcNab, line 332:
```suggestion
Astute `std` doc readers might notice that the Rust team [recommends only using `Deref`/`DerefMut` for smart pointers, to avoid confusion](https://doc.rust-lang.org/std/ops/trait.Deref.html). Components like `Items` _are not_ smart pointers. We choose to ignore this advice, as this pattern works, is already widely used in the Rust ecosystem, and Good UX Comes First.
```

Mostly, I really don't like:
> We personally don't care.

For a number of reasons.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+Sometimes when building Bevy Apps you might find yourself repeating the same sets of components over and over in your queries:
+
+```rust
+fn move_players(mut players: Query<(&mut Name, &mut Transform, &mut Item)>) {
+    for (name, transform, item) in players.iter_mut() {
+    }
+}
+
+fn despawn_players(mut players: Query<(Entity, &mut Name, &mut Transform, &mut Item)>) {
```
DJMcNab, line 346: I'm not a huge fan of this example, since for despawning, it seems unlikely that you'd need full access to all of these components, and certainly not for the same reason as in `move_players`

Unfortunately, I'm not sure of any better examples. Maybe this is almost always unidiomatic, but we provide it for completeness?

DJMcNab (commented):
I've done a first pass, and I'm broadly happy.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+<div class="release-feature-authors">authors: Rob Swain (@superdump), @robtfm, @dataphract, @cart</div>
+
+With the upper limit of 256 point lights removed, the only limit on lights is what the hardware can support and bottlenecks in our algorithms. To increase the number of lights, we made a number of optimizations to our clustering algorithms.
+
+* **Dynamic Light Clusters**: By default cluster x/y slices are now dynamically configured based on the lights in the scene, which can significantly increase performance in some scenes. The clustering behavior is now also user-configurable as FixedZ (the new default dynamic x/y behavior), custom fixed x/y/z slice values, single-cluster, and "no clustering".
+* **Light Frustum Change Detection**: We now use Bevy ECS's change detection feature to only recalculate the view frustum of lights that have changed.
+* **Cluster Assignment Optimizations**: The cluster assignment data access patterns and data structures received a variety of tweaks that improved performance.
+
+// TODO: Show many lights video here
```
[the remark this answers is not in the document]
inodentry: Markdown doesn't have a comment syntax. The closest thing you can get is "abusing link anchor syntax" as `[]:#(comment)`, if you want to have some markdown that will not output anything to the rendered files. Using HTML comments as `<!-- comment -->` means they still get passed through / outputted to the resulting HTML files, though the browser would, ofc, treat them as comments and ignore them.

In this case, none of that should matter, as these are TODO notes on a WIP document. If anything, using `//` is nice, because it is *not* treated as a comment, meaning that people reviewing the WIP PR using a rendered HTML view can see them and read them, which is useful.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+In **Bevy 0.7**, systems are now automatically labeled with a `SystemTypeIdLabel` tied to their type. This enables much more ergonomic and clear system ordering:
+
+```rust
+app
+  .add_system(update_velocity)
+  .add_system(movement.after(update_velocity))
+```
+
+The Bevy ECS labeling system is powerful and there are still legitimate use cases for custom labels (such as labeling multiple systems with the same label). But most common use cases can take advantage of the ergonomic auto-labeling functionality.
```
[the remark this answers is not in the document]
inodentry: Agreed, it feels backwards. It presents internal implementation details *first*, and only gets to why users should care later.

I would propose starting with the user-facing implications. Something like: "In bevy 0.7, you no longer have to come up with a custom label every time you need to order a system relative to another system; for simple cases you can just use the system's function, just like when you added the system!"

Then, if you really want to get into the implementation details (tbh I'm not sure such level of detail is even appropriate for a release blog post), you can elaborate, after the code example: "... and here is how this works. We have a `SystemTypeIdLabel` type ... ..."

In general, I think this sort of mindset should be applied everywhere. The news posts would read much better if they start by presenting the user-facing changes and motivations, and then maybe dive deeper into how it was done / implementation details (but not too much). Doing it the other way around feels backwards, as a reader. You are telling me a bunch of technical stuff, and I don't even know why I should care yet, until later.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+However this can feel repetitive when you're doing it for tens or hundreds of entities. We added a way to make this even easier, without needing to resort to macros:
+
+```rust
+commands.spawn_bundle(SpriteBundle {
+    texture: some_texture,
+    ..default()
+})
+```
+
+This is equivalent in functionality to `..Default::default()`, it's just more compressed. And you can still use the longer form if you prefer. The `default()` function is included in Bevy's prelude by default, so you don't need to manually import it. Ergonomics for the win!
```
inodentry, line 203:
```suggestion
This is equivalent in functionality to `..Default::default()`, it's just more compressed. And you can still use the longer form if you prefer. The `default()` function is included in Bevy's prelude, so you don't need to manually import it. Ergonomics for the win!
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+Bevy ECS solves a hard problem: providing easy and fast access to data in parallel while still respecting Rust's strict mutability and ownership rules. Since our first release, we've supported efficiently accessing specific entities in our ECS Queries:
+
+```rust
+struct SomeEntities {
+    a: Entity,
+    b: Entity,
+}
+
+fn system(mut query: Query<&mut Transform>, entities: Res<SomeEntities>) {
```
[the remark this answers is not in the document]
inodentry: This is super pedantic and IMO hurts readability.

I don't think a general name like `system` is inappropriate for a code example that just showcases some syntax and does nothing in particular.

(i have the same response to your other similar suggestions below)

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+* Skeletal animation and mesh skinning
+* GLTF animation importing
+* Unlimited* point lights in a scene
+* Improved clustered forward rendering: dynamic/adaptive clustering, faster and more accurate cluster assignment
+* Compressed texture support (KTX2 / DDS / .basis): load more textures in a scene, faster
+* Compute shader / pipeline specialization: Bevy's flexible shader system was ported to compute shaders, enabling hot reloading, shader defs, and shader imports
+* Render to texture: cameras can now be configured to render to a texture instead of a window
+* Flexible user-customizable mesh vertex layouts in shaders
+* ECS improvements: Order systems using their names, Query::get_many, use conflicting parameters in systems via ParamSets, WorldQuery derives
```
ItsDoot, line 31: > WorldQuery derives

Without bevyengine/bevy#4125, these are very annoying to use since you can't make a system function public, because the generated impl makes use of private types.

content/news/2022-04-12-bevy-0.7/index.md
inodentry: Yea, this is needlessly inflammatory.

Also, if you consider me as part of the "bevy org", you are not speaking for me, as i personally do care and i am against (ab)using deref in this way.

content/news/2022-04-12-bevy-0.7/index.md
inodentry: I have a real-world example where this is useful. It's not "unidiomatic".

Say, you have a "player controller" that is represented as multiple components, like

```rust
/// Player controller configuration/properties
struct PlayerControlParams {
    gravity_up: f32,
    gravity_down: f32,
    acceleration_start: f32,
    acceleration_end: f32,
    // ...
}
/// Current player movement state
enum PlayerMoveState {
    Standing,
    Walking,
    Jumping,
    Falling,
}
/// Desired movement, as per joystick or other input
struct PlayerWantsMovement(Vec2);
```

These are separate components, because some systems (like the "joystick input" or whatever) might only need specific ones. But systems that are part of the player controller would typically need all of them.

```rust
#[derive(WorldQuery)]
struct PlayerControlQuery<'w> {
    transform: &'w mut Transform,
    state: &'w mut PlayerMoveState,
    input: &'w PlayerWantsMovement,
    params: &'w PlayerControlParams,
}

fn player_gravity(q: Query<PlayerControlQuery>) { /* ... */ }
fn player_walk(q: Query<PlayerControlQuery>) { /* ... */ }
fn input_gamepad(q: Query<&mut PlayerWantsMovement>, /* ... */) { /* ... */ }
fn input_keyboard(q: Query<&mut PlayerWantsMovement>, /* ... */) { /* ... */ }
```

Now, this is probably *too much* to include as an example in the release blog post. I just wanted to show a real use case for this. Feel free to use it as inspiration and come up with something more concise.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+
+Skeletal animations can now be played, paused, scrubbed, looped, reversed, and speed controlled using the new [`AnimationPlayer`] component and [`AnimationClip`] asset:
+
+```rust
+#[derive(Component)]
+struct Animations {
+    dance: Handle<AnimationClip>,
+}
+
+fn system(mut query: Query<(&Animations, &mut AnimationPlayer)>) {
```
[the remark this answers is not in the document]
alice-i-cecile: Yeah, I agree with DJ here (sorry Ida!). It's better to model good habits, and the added context of a name can help the reader clue in faster to what the example is supposed to do.

content/news/2022-04-12-bevy-0.7/index.md
alice-i-cecile: https://github.com/bevyengine/rfcs/pull/51: Composition

https://github.com/bevyengine/rfcs/pull/49: Primitives

content/news/2022-04-12-bevy-0.7/index.md
alice-i-cecile, line 165:
```suggestion
The Bevy ECS labeling system is powerful and there are still legitimate use cases for custom labels (such as labeling multiple systems with the same label and exporting a stable public API as a plugin author). But most common use cases can take advantage of the ergonomic auto-labeling functionality.
```

content/news/2022-04-12-bevy-0.7/index.md
alice-i-cecile: Agreed: I think we should r+ that change for 0.7 to make sure this feature functions as intended.

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
DJMcNab: Yeah. 
The version for describing the useless system was attempting to be a joke, but I do think that `fn system` is not a good pattern, since system names should always be verbs.

content/news/2022-04-12-bevy-0.7/index.md
cart: I'm on board for this. I left one comment that agrees with a documentation change. If we resolve that before release, I'm down to squeeze it in.

content/news/2022-04-12-bevy-0.7/index.md
cart: I figured the `## GLTF Animation Importing` section directly following `## Skeletal Animation` that calls this out explicitly / in detail was enough. I agree its super cool and I think it deserves its own section.

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Solid points. I wanted to make it clear that its a system, but modeling good behavior is more important. I'm convinced!

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Yeah I did "not actual comments" on purpose to make them appear in the rendered document.

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: I agree that the inclusion of SystemTypeIdLabel is dubious the focus on "auto-labeling" as the mechanism of the feature distracts from the point.

But I do think the general "problem -> solution" format is a great way to illustrate progress because it tells a consistent story. I think covering the "this is what manual labeling looked like" first is important.

If you show the solution first, the user's first thoughts when they see it are no longer guaranteed to be "wow what a great improvement over the status quo". If they aren't familiar with the previous state of things, they might just think "yeah that tracks logically". 

Imagine this was a rendering feature that makes things much prettier. I think it generally makes sense to show the "older / uglier" thing first. That way we've primed the reader to discover things that are "better" in the new picture. The reverse order results in the reader looking for things that are "worse", which is a very different story. And not providing a contrasting image at all is an even bigger miss.

Definitely not saying either of you were actually suggesting removing the "before" picture here (or reordering it). Just providing rationale in case you were.

I'll try to reframe this to remove the focus on auto-labeling as the "after" picture and remove SystemTypeIdLabel. But I'll leave the introduction setting up the "before" picture (including discussion of labels) unless someone wants to push back on that.

content/news/2022-04-12-bevy-0.7/index.md
cart: Solid points all around. I'm happy to own this decision personally if that sits better with you all (I just default to "we" when discussing features and design choices generally, even when I've fully designed and implemented something). And toning down the language is a very positive improvement here that better reflects the tone we should be projecting.

content/news/2022-04-12-bevy-0.7/index.md
cart: Cool yeah I can definitely make this a bit less arbitrary.

content/news/2022-04-12-bevy-0.7/index.md
cart: Yeah that seems worth including!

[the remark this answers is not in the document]
cart: I do also agree with @inodentry that we should open with a something like "In bevy 0.7, you no longer have to come up with a custom label every time you need to order a system relative to another system; for simple cases you can just use the system's function, just like when you added the system!"

content/news/2022-04-12-bevy-0.7/index.md
cart: (I just committed @DJMcNab's suggestion, but ill leave this open until we make a call on the "we" thing)

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
DJMcNab: Definitely not wanting to remove the before picture.
I'll be honest here, I did my first pass review quite quickly this morning, and completely skimmed over the before picture; that is, this comment is the first time I've even noticed it was there.

content/news/2022-04-12-bevy-0.7/index.md
alice-i-cecile: I think "as a project, we've chosen to..." is good language here. Not everyone agrees with the call, but in the context of Bevy as a project / organization, we've made a decision.

content/news/2022-04-12-bevy-0.7/index.md
alice-i-cecile: As the person who suggested doc hidden I'm on board with that tweak. Left a comment on the PR with the appropriate suggestion to make it easy.

alice-i-cecile, line 31:
```suggestion
* ECS improvements: Order systems using their names, Query::get_many_mut, use conflicting parameters in systems via ParamSets, WorldQuery derives
```

The mutable one is the tricky / important one :)

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,383 @@
...
+    for (animations, mut animation_player) in query.iter_mut() {
+        animation_player.play(animations.dance.clone());
+    }
+}
+```
+
+[`AnimationPlayer`] can also be used to animate arbitrary [`Transform`] components, not just skeletons!
+
+This critical feature has been a long time coming, but we wanted to build it in a way that meshed nicely with the [new Bevy renderer](/news/bevy-0-6/#the-new-bevy-renderer) and didn't just "hack things in". This builds on our new [User-Customizable Mesh Vertex Layouts](LINKME), [Shader Imports](/news/bevy-0-6/#shader-imports), and [Material](/news/bevy-0-6/#materials) systems, which ensures that this logic is flexible and reusable, even with non-standard meshes and custom render pipelines.
```
alice-i-cecile, line 66: Looks like the link should be to https://dev-docs.bevyengine.org/bevy/render/render_resource/struct.VertexBufferLayout.html ?

cart: I intended this to link to the part of the blog post that covers this (which hasn't been written yet).

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+        // slow down playback
+        sink.set_speed(0.5);
+    }
+}
+```
+
+You can also now loop audio playback:
+
+```rust
+audio.play_in_loop(music);
```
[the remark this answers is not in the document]
IceSentry: The with_volume isn't required, but it shows the builder pattern

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+For those who don't know, Bevy is a refreshingly simple data-driven game engine built in Rust. You can check out [Quick Start Guide](/learn/book/getting-started/) to get started. Bevy is also free and open source forever! You can grab the full [source code](https://github.com/bevyengine/bevy) on GitHub. Check out [Bevy Assets](https://bevyengine.org/assets) for a collection of community-developed plugins, games, and learning resources.
+
+To update an existing Bevy App or Plugin to **Bevy 0.7**, check out our [0.6 to 0.7 Migration Guide](/learn/book/migration-guides/0.6-0.7/).
+
+As always, there are a _ton_ of new features, bug fixes, and quality of life tweaks in this release, but here are some of the highlights: 
+
+* Skeletal animation and mesh skinning
+* GLTF animation importing
+* Unlimited* point lights in a scene
+* Improved clustered forward rendering: dynamic/adaptive clustering, faster and more accurate cluster assignment
```
superdump, line 26: This is only completely true when the faster cluster light assignment PR is merged. Otherwise it’s just as accurate.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+
+## Light Clustering Features and Optimizations
+
+<div class="release-feature-authors">authors: Rob Swain (@superdump), @robtfm, @dataphract, @cart</div>
+
+With the upper limit of 256 point lights removed, the only limit on lights is what the hardware can support and bottlenecks in our algorithms. To increase the number of lights, we made a number of optimizations to our clustering algorithms.
+
+* **Dynamic Light Clusters**: By default cluster x/y slices are now dynamically configured based on the lights in the scene, which can significantly increase performance in some scenes. The clustering behavior is now also user-configurable as FixedZ (the new default dynamic x/y behavior), custom fixed x/y/z slice values, single-cluster, and "no clustering".
+* **Light Frustum Change Detection**: We now use Bevy ECS's change detection feature to only recalculate the view frustum of lights that have changed.
```
superdump, line 118: This feels minor to me. It’s not a heavy calculation.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+fn system(mut query: Query<(&Animations, &mut AnimationPlayer)>) {
+    for (animations, mut animation_player) in query.iter_mut() {
+        animation_player.play(animations.dance.clone());
+    }
+}
+```
+
+[`AnimationPlayer`] can also be used to animate arbitrary [`Transform`] components, not just skeletons!
+
+This critical feature has been a long time coming, but we wanted to build it in a way that meshed nicely with the [new Bevy renderer](/news/bevy-0-6/#the-new-bevy-renderer) and didn't just "hack things in". This builds on our new [User-Customizable Mesh Vertex Layouts](LINKME), [Shader Imports](/news/bevy-0-6/#shader-imports), and [Material](/news/bevy-0-6/#materials) systems, which ensures that this logic is flexible and reusable, even with non-standard meshes and custom render pipelines.
```
superdump, line 65:
```suggestion
This critical feature has been a long time coming, but we wanted to build it in a way that meshed nicely with the [new Bevy renderer](/news/bevy-0-6/#the-new-bevy-renderer) and not just "hack things in". This builds on our new [User-Customizable Mesh Vertex Layouts](LINKME), [Shader Imports](/news/bevy-0-6/#shader-imports), and [Material](/news/bevy-0-6/#materials) systems, which ensures that this logic is flexible and reusable, even with non-standard meshes and custom render pipelines.
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+
+In **Bevy 0.7**, we added the ability to automatically "upgrade" to using unbounded storage buffers for Clustered Forward Rendering on platforms that support them, enabling unlimited* point lights. There is an asterisk there because in practice this is limited by memory and hardware constraints. 
+
+## Light Clustering Features and Optimizations
+
+<div class="release-feature-authors">authors: Rob Swain (@superdump), @robtfm, @dataphract, @cart</div>
+
+With the upper limit of 256 point lights removed, the only limit on lights is what the hardware can support and bottlenecks in our algorithms. To increase the number of lights, we made a number of optimizations to our clustering algorithms.
+
+* **Dynamic Light Clusters**: By default cluster x/y slices are now dynamically configured based on the lights in the scene, which can significantly increase performance in some scenes. The clustering behavior is now also user-configurable as FixedZ (the new default dynamic x/y behavior), custom fixed x/y/z slice values, single-cluster, and "no clustering".
```
superdump, line 117:
```suggestion
* **Dynamic Light Clusters**
  * By default cluster x/y slices are now dynamically configured based on the lights in the scene, which can significantly increase performance in some scenes.
  * The clustering behavior is now also user-configurable as FixedZ (the new default dynamic x/y behavior, fixing the number of z slices), custom fixed x/y/z slice values, single-cluster, and "no clustering", giving you control when you know a certain cluster configuration will perform even better.
  * Additionally, in 0.6 the visible volume that is covered by all the clusters basically matched the full visible volume of the view frustum. This meant that if all the point lights were in the foreground, all the clusters beyond the lights were wasted space. In 0.7, it is possible to limit the far bound to be closer than the camera far bound, which means the lights can be spread across more clusters, which can significantly increase rendering performance.
```

content/news/2022-04-12-bevy-0.7/index.md
superdump: Oh, the light view frustum! For some reason when I read this the first time, I thought it meant the camera view frustum.

superdump: I can't resolve this so I'll just delete it.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+* **Light Frustum Change Detection**: We now use Bevy ECS's change detection feature to only recalculate the view frustum of lights that have changed.
+* **Cluster Assignment Optimizations**: The cluster assignment data access patterns and data structures received a variety of tweaks that improved performance.
+
+// TODO: Show many lights video here
```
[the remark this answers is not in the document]
superdump: Let me know if you want the original file.

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+// TODO: Add iterative sphere refinement here if it is merged
```
superdump, line 122: It isn't yet, but the PR does fix a bug with orthographic projections, though we haven't yet been able to isolate what part of the PR fixes it as we had implemented the new algorithm and then fixed problems we saw along the way, thinking they were caused by the new algorithm, but I guess maybe it works better now? :)

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+
+The benefits don't stop there either - because the textures are compressed and can be used by the GPU in that format, reading from them uses less memory bandwidth, which can bring performance benefits. The Bistro scene gains about 10% in frame rate from using compressed textures.
+
+![bistro compressed](bistro_compressed.png)
+
+
+Another benefit is that mipmaps are supported, which makes for smoother, less noisy textures. Bevy currently doesn't have automatic support for generating mipmaps for "normal" textures, so using compressed textures is a nice way to have mipmaps now, even though we don't support them for standard textures yet!
+
+In summary, Bevy now supports loading compressed textures from `.dds`, `.ktx2`, and `.basis` files. This includes support for the standard ASTC, BCn, and ETC2 formats, as well as "universal" formats like ETC1S and UASTC that can be transcoded to formats supported by specific systems at runtime. The GLTF loader was also extended to support loading these formats.
```
superdump, line 158:
```suggestion
Another benefit is that mipmaps are supported, which makes for smoother, less noisy textures. Bevy currently doesn't have support for automatically generating mipmaps for uncompressed textures, so using compressed textures is a nice way to have mipmaps now!

In summary, Bevy now supports loading compressed textures from `.dds`, `.ktx2`, and `.basis` files. This includes support for the standard ASTC, BCn, and ETC2 formats, as well as 'universal' formats like ETC1S and UASTC that can be transcoded to the standard formats supported by specific systems at runtime. The glTF loader was also extended to support loading these formats.
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+}
+```
+
+## Flexible Mesh Vertex Layouts
+
+<div class="release-feature-authors">authors: @cart, @parasyte</div>
+
+In **Bevy 0.7**, it is now easy to make shaders support any Mesh vertex layout and arbitrary vertex attributes. Bevy's "shader pipeline specialization" system was extended to support "specializing on mesh vertex layouts".
+
+For most Bevy users, this means that [Materials](/news/bevy-0-6/#materials), including the built in [`StandardMaterial`] and custom shader materials now support arbitrary Meshes automatically, provided those Meshes have the vertex attributes required by the material shaders.
```
superdump, line 200:
```suggestion
For most Bevy users, this means that [Materials](/news/bevy-0-6/#materials), including the built in [`StandardMaterial`] and custom shader materials now support arbitrary Meshes automatically, provided those Meshes have the vertex attributes required by the material shaders. It also means that if your Mesh is missing any attribute required by its material, rendering can fail gracefully.
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+    let a_transform = query.get_mut(entities.a).unwrap();
+}
+```
+
+However, to respect Rust's mutability rules, we need to disallow apis that might produce "aliased mutability". Seasoned Bevy users will probably recognize this Rust borrow checker error:
+
+```rust
+fn system(mut query: Query<&mut Transform>, entities: Res<SomeEntities>) {
+    let a_transform = query.get_mut(entities.a).unwrap();
+    // This line files to compile because `query` is already mutably borrowed above
```
superdump, line 336:
```suggestion
    // This line fails to compile because `query` is already mutably borrowed above
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+
+fn give_sword(mut query: Query<&mut Items>) { 
+    for mut items in query.iter_mut() {
+        // No more .0!
+        items.push(Item::new("Flaming Poisoning Raging Sword of Doom"));
+    }
+}
+```
+
+Astute `std` doc readers might notice that the Rust team [recommends only using `Deref`/`DerefMut` for smart pointers, to avoid confusion](https://doc.rust-lang.org/std/ops/trait.Deref.html). Components like `Items` _are not_ smart pointers. We choose to ignore this advice, as this pattern works, is already widely used in the Rust ecosystem, and Good UX Comes First.
```
superdump, line 441:
```suggestion
Astute `std` doc readers might notice that the Rust team [recommends only using `Deref`/`DerefMut` for smart pointers, to avoid confusion](https://doc.rust-lang.org/std/ops/trait.Deref.html). Components like `Items` are _not_ smart pointers. We choose to ignore this advice, as this pattern works, is already widely used in the Rust ecosystem, and Good UX Comes First.
```

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+    transform: &'w mut Transform,
+    items: &'w mut Items,
+}
+
+fn move_players(mut players: Query<PlayerQuery>) {
+    for player in players.iter_mut() {
+    }
+}
+
+fn despawn_players(mut players: Query<(Entity, PlayerQuery)>) {
```
superdump, line 476: It looks like the `PlayerQuery` can represent a commonly-queried set of components, and you can arbitrarily extend it with other components in the query. That is super cool!

alice-i-cecile:
I've added the coverage of doc improvements here: https://github.com/cart/bevy-website/pull/6 :) Feedback would be appreciated!

content/news/2022-04-12-bevy-0.7/index.md
```diff
@@ -0,0 +1,654 @@
...
+
+We plan to continue iterating on these APIs with even more functionality and usability improvements!
+
+[`AudioSink`]: http://docs.rs/bevy/0.7.0/bevy/audio/struct.AudioSink.html
+
+## EventLoop Power Saving Modes
+
+<div class="release-feature-authors">authors: @aevyrie</div>
+
+By default Bevy will run updates "as fast as it can" (limited by the monitors' refresh rate). This is great for most games, but some application types (such as GUI apps) need to prioritize CPU and GPU power usage.
```
superdump, line 620:
```suggestion
By default Bevy will run updates "as fast as it can" (limited by the screen's refresh rate). This is great for most games, but some application types (such as GUI apps) need to prioritize CPU and GPU power usage.
```

content/news/2022-04-12-bevy-0.7/index.md
cart: I've actually removed this section because it has become clear that this feature isn't ready for user consumption yet / only really works for internal bevy shaders atm.

content/news/2022-04-12-bevy-0.7/index.md
cart: merged!

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Added. Thanks!

content/news/2022-04-12-bevy-0.7/index.md
cart: I think the current phrasing is more correct. If we remove the first part of the "and" this becomes:

"This critical feature has been a long time coming, but we wanted to build it in a way that didn't just "hack things in"

The suggested "not" would look like this:

"This critical feature has been a long time coming, but we wanted to build it in a way that not just "hack things in"

content/news/2022-04-12-bevy-0.7/index.md
cart: Agreed. I dig it!

content/news/2022-04-12-bevy-0.7/index.md
cart: Added!

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Cool cool. I've changed how this is presented to make "auto labeling" conceptually an implementation detail, rather than the "feature" itself.

content/news/2022-04-12-bevy-0.7/index.md
cart: I drew inspiration from @inodentry's example and made this a bit less arbitrary.

content/news/2022-04-12-bevy-0.7/index.md
cart: This section has been added and I updated the link.

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Nice catch. Thanks!

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Yeah I think when the context is arbitrary (like places where you might put `foo`), `system` is a reasonable choice. I'd prefer to leave these cases as-is.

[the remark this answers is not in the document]
cart: I've resolved this (and similar suggestions), but feel free to continue the discussion if you feel inclined.

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
cart: Fixed! (i manually applied the suggestion because we don't use tabs for indentation in blog posts).

content/news/2022-04-12-bevy-0.7/index.md
cart: Nice catch. I removed the "accurate" statement. We can re-add it if we merge that pr.

content/news/2022-04-12-bevy-0.7/index.md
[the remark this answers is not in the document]
IceSentry: Interesting, I don't use tabs in my editor either, not sure how they ended up there. Doesn't matter as long as it's fixed 🎉

content/news/2022-04-12-bevy-0.7/index.md
superdump: My brain split at “wanted to” so then yours is “wanted to don’t just hack things in” and mine is “wanted to not just hack things in”. It feels like one of those rotating optical illusions where if you look at it one way it looks like it’s rotating anti-clockwise, and another clockwise.

content/news/2022-04-12-bevy-0.7/index.md
superdump: We merged the PR so now you can add it back. :)

content/news/2022-04-15-bevy-0.7/index.md
```diff
@@ -0,0 +1,1166 @@
...
+
+## Website Improvements 
+
+<div class="release-feature-authors">authors: @doup</div>
+
+The [Bevy Book](/learn/book) now has a much nicer pager widget that displays previous / next section names:
+
+![pager](pager.png)
+
+We also added an "improve this page" footer link to make it easier for Bevy Book readers to contributed changes.
```
inodentry, line 723:
```suggestion
We also added an "improve this page" footer link to make it easier for Bevy Book readers to contribute changes. 
```

content/news/2022-04-15-bevy-0.7/index.md
```diff
@@ -0,0 +1,1166 @@
...
+
+We plan to continue iterating on these APIs with even more functionality and usability improvements!
+
+[`AudioSink`]: http://docs.rs/bevy/0.7.0/bevy/audio/struct.AudioSink.html
+
+## Sprite Anchors
+
+<div class="release-feature-authors">authors: @mockersf</div>
+
+[`Sprite`] components can now define an [`Anchor`] point, which determines the "origin" of the sprite. Sprites still default to a "center" origin, but this is now configurable:
```
inodentry, line 637: I think it is also worth including the term/keyword "pivot point" somewhere in this description. A lot of people know it by that name and might be searching for that.

cart: good call!

PR #335, opened 2022-04-13
[Merged by Bors] - Make web examples more visible

IceSentry:
The web examples are currently hard to find if you don't already know they exist.

Make them visible in the learn section and the next steps of the book


alice-i-cecile (approved):
I think this is a nice improvement! The examples themselves need more polish, but they should be exposed.

PR #341, opened 2022-04-15
[Merged by Bors] - fix path to example bevymark and games

mockersf:
When building examples for wasm, example bevymark and games fail as it has changed path in the Bevy repo.

This changes the path to the correct one


alice-i-cecile (approved):

alice-i-cecile:
(Let's get the ball rolling on fixing stuff...)

PR #346, opened 2022-04-16
[Merged by Bors] - Add missing import to PresentMode section

rparrett:
`PresentMode` isn't in the prelude.


IceSentry (approved):
Good catch, I feel like it should be in the prelude though.

PR #347, opened 2022-04-16
[Merged by Bors] - Clarify the config api migration

IceSentry:
The 0.7 example of the migration guide for the config api had a few issues. It passed the wrong datatype and it wasn't cleared what the argument of the closer was.

### DIFF
--- a/content/learn/book/migration-guides/0.6-0.7/_index.md
+++ b/content/learn/book/migration-guides/0.6-0.7/_index.md
@@ -208,7 +208,7 @@ fn camera_system(cameras: Query<&Camera, With<FirstPassCamera>>) {
 // 0.6
 struct Config(u32);
 
-fn local_config(local: Local<Config>) {
+fn local_is_42(local: Local<Config>) {
     assert_eq!(*local.0, 42);
 }
 
@@ -219,18 +219,15 @@ fn main() {
 }
 
 // 0.7
-struct Config(u32);
-
-fn local_config(local: u32) -> impl FnMut(ResMut<Config>) {
-    move |mut val: ResMut<Config>| {
-        val.0 = local;
-
+fn local_is_42(local: u32) -> impl FnMut() {
+    // This closure will be the system that will be executed
+    move || {
         assert_eq!(val.0, 42);
     }
 }
 
 fn main() {
-    App::new().add_system(local_config(42)).run();
+    App::new().add_system(local_is_42(42)).run();
 }
 ```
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

rparrett (approved):

content/learn/book/migration-guides/0.6-0.7/_index.md
```diff
@@ -222,17 +222,15 @@ fn main() {
 struct Config(u32);
 
 fn local_config(local: u32) -> impl FnMut(ResMut<Config>) {
```
DJMcNab, line 224:
```suggestion
fn local_config(local: u32) -> impl FnMut() {
```

content/learn/book/migration-guides/0.6-0.7/_index.md
```diff
@@ -222,17 +222,15 @@ fn main() {
...
-    move |mut val| {
+    move |mut val: ResMut<Config>| {
```
DJMcNab, line 225:
```suggestion
    move || {
```

As a note, you shouldn't need to specify this type, since it was already specified in the return type.

content/learn/book/migration-guides/0.6-0.7/_index.md
```diff
@@ -222,17 +222,15 @@ fn main() {
...
         val.0 = local;
 
         assert_eq!(val.0, 42);
```
DJMcNab, line 228:
```suggestion
        assert_eq!(val.0, 42);
```

DJMcNab (approved):
I don't really like this example, but we can at least make the two examples do the same thing.

content/learn/book/migration-guides/0.6-0.7/_index.md
IceSentry: Yep, but I saw a video of someone reviewing the migration guide and they were confused about what that parameter was. That's the reason I added it, not because it's necessary.

DJMcNab: Hmm. It's still idiomatic not to duplicate writing it out.

IceSentry: Agreed, maybe I could add a comment. Something like

```suggestion
	// This closure will be the system that will be executed
    move || {
```

content/learn/book/migration-guides/0.6-0.7/_index.md
```diff
@@ -219,20 +219,15 @@ fn main() {
...
-struct Config(u32);
-
-fn local_config(local: u32) -> impl FnMut(ResMut<Config>) {
-    move |mut val| {
-        val.0 = local;
-
+fn local_is_42(local: u32) -> impl FnMut() {
+    // This closure will be the system that will be executed
+    move || {
         assert_eq!(val.0, 42);
```
rparrett, line 225:
```suggestion
        assert_eq!(local, 42);
```

Missed this one, I think.

PR #348, opened 2022-04-16
[Merged by Bors] - Add incorrect code block support

### LINKED ISSUE #343: Mark code that shouldn't compile or fails in runtime
The official Rust book has fancy Ferris showing whether a given snippet compiles or not [like so](https://doc.rust-lang.org/book/ch00-00-introduction.html#how-to-use-this-book). Rust documentation system has it too (no Ferris tho)

So I suggest following that and explicitly show that a piece of code won't compile or will give a runtime error.
e.g. in recent 0.7 blog post we have this https://bevyengine.org/news/bevy-0-7/#world-system-param chapter and it would be a bit better if the code snippet after "... &World will conflict with any mutable Query:" was made into reddish borders or had a corresponding label.

Jengamon:
This should resolve #343 by adding a shortcode that goes around code blocks that should be marked incorrect (`incorrect_code_block()`), highlights them in a dark red, and makes the left border red. Looks like this:
Screen Shot 2022-04-16 at 4 50 25 PM

from:

~~~markdown
Just keep in mind that `&World` will conflict with _any_ mutable Query:

{% incorrect_code_block() %}
```rust
fn invalid_system(world: &World, transforms: Query<&mut Transform>) {
}
```
{% end %}

In these cases, consider using our new [ParamSets](/news/bevy-0-7/#paramsets) to resolve the conflict:

```rust
fn valid_system(set: ParamSet<(&World, Query<&mut Transform>)>) {
}
```
~~~


Jengamon:
I would have loved to do something like
~~~markdown
```rust,{.incorrect}
.....
```
~~~

or something to that effect, but that breaks syntax highlighting. This was the best option that I found to get this feature working. It works by having a shortcode just add a `<div class="incorrect">` around the body that it's given, then I added some Sass in _html.scss to get all the colors working given a `div.incorrect`.

sass/elements/_html.scss
```diff
@@ -11,8 +12,21 @@ html, body {
...
+    background-color: #542326;
+    border-left: 10px solid red;
+    border-radius: 10px;
+
+    .z-code,
+    .z-code code {
+        background-color: #542326;
+    }
+}
+
```
doup, line 25: I think is better to put this snippet in `components/_syntax-theme.scss` which is where all the syntax highlighting  styles are. This file (`html.scss`) should be only for `html` element tweaks (although it also touches `body` right now, this is fixed in #315).

doup (changes requested):
The solution looks good. 👍 

The only thing, I think the styles should go in another file. Check the comment.

doup:
Btw, maybe an issue can be opened in Zola to allow adding extra classes into code blocks? They already have some annotations, maybe it's not difficult for them to add this option.

See: https://www.getzola.org/documentation/content/syntax-highlighting/#annotations

sass/elements/_html.scss
Jengamon: ok, I've moved it~

doup:
Hi! Now that I see it again, I think the PR should contain only the actual changes, not the code formatting changes. But, that's my opinion, I don't know what's the Bevy policy here.

cc @alice-i-cecile

Jengamon:
Ok, I've undone all the formatting for _html.scss, but _syntax-theme.scss didn't have a consistent space formatting (the properties are sometimes 1 space, sometimes 4.) I could undo the selector formatting.

doup (approved):
OK, makes sense for syntax-theme file. 👌

alice-i-cecile:
Yeah, we should try to fix and enforce formatting issues in a seperate PR if possible.

alice-i-cecile (approved):

PR #363, opened 2022-05-06
[Merged by Bors] - Remove Bevy Merch link

cart:
I just pulled my Bevy sweatshirt out of the dryer and the logo has started peeling. Its only been through a wash/dry cycle 3 times now. That is unacceptable. This is the second known case of quality issues with Spring's products (a month ago someone else hit the same issue) and I'm guessing more will start to surface as the products age.

I'm pressed for time atm because I've got a flight to catch tomorrow morning and a wedding celebration all weekend, but I no longer feel comfortable listing this on the home page so I wanted to be proactive here.

When I get back I'll engage with Spring and see if they can make this right (and hopefully resolve the quality issues), but I will almost certainly be looking for alternatives and offering refunds from my own pocket (if Spring won't take responsibility here).

What a bummer.

### DIFF
--- a/templates/layouts/base.html
+++ b/templates/layouts/base.html
@@ -81,7 +81,6 @@
                 {{header_macros::header_item(name="News", current_path=current_path)}}
                 {{header_macros::header_item(name="Community", current_path=current_path)}}
                 {{header_macros::header_item(name="Assets", current_path=current_path)}}
-                <a class="header-item" href="https://merch.bevyengine.org">Merch</a>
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
Oof. That's unfortunate. I think we should remove this ASAP, but I'll leave some time for others to weigh in.

alice-i-cecile (approved):

cart:
Nah I want this merged now. Better to mitigate damage asap and discuss after the fact.

alice-i-cecile:
Sounds good. That was my instinct, but I realized that you'd just merge it ASAP yourself if you felt this way.

PR #369, opened 2022-05-11
[Merged by Bors] - Indeterminate assets loading

doup:
If an asset has compression enabled the loading feedback break miserably. This change fixes that situation by introducing an "indeterminate" loading state. It switches to that state when the stream has loaded more data that it supposed to do.

I haven't tested this with a server with gzip enabled; if someone is willing to properly test it with a server go on. Otherwise I think the changes are simple enough to be just reviewed in code.

- [Commit with the actual change](https://github.com/bevyengine/bevy-website/pull/369/commits/30fc918e9182343630d1abad83c563927b8a76aa)
- [Commit with changes to the callback signatures, so all of them have the same "style" as `update`](https://github.com/bevyengine/bevy-website/pull/369/commits/4e8e976a83c64ee3854c477890ee68d08fa964fe)

> **⚠️ THE VIDEO BELOW IS TO SHOW THE EFFECT**
> **⚠️ IN REALITY THIS STATE WILL TRIGGER WHEN THE BAR IS AT 100%; NOT AT 75% AS IN THE VIDEO.**

https://user-images.githubusercontent.com/188612/167780112-6f735f02-64c3-4d2f-88a4-ea7a9845eef3.mp4


doup:
I've used the previous branch, that's why there are so many commits… the good ones are the last three.

cart:
Seems reasonable. Lets give it a try!

PR #376, opened 2022-05-28
Ignore `.vscode` folder

Weibye:
- Add `.vscode` folder to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 public
 **/.DS_Store
 .idea/
+.vscode/
 content/assets
 content/learn/errors
 content/examples


Nilirad:
I think you should rather ignore the folder on your system's .gitignore instead. Each editor will put its own folders, so this is more of a user concern rather than anything else.

See for example https://blog.martinhujer.cz/dont-put-idea-vscode-directories-to-projects-gitignore/

Weibye:
> I think you should rather ignore the folder on your system's .gitignore instead

huh, TIL. 
Well that makes sense, so I'm closing this.

PR #386, opened 2022-06-17
[Merged by Bors] - make customFetch global

IceSentry:
To try and reduce the likelihood of this breaking again we can simply register customFetch globally and use it directly from the window. It still relies on wasm-bindgen not changing a specific line, but at least it's one line instead of 5.

I tested it locally and it works.

Obviously doing things globally isn't ideal, but I renamed it to be a bit more explicit.

I tried simply overriding `fetch` itself globally, but it broke everything.

### DIFF
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -42,7 +42,7 @@ add_category()
         # Patch generated JS to allow to inject custom `fetch` with loading feedback.
         # See: https://github.com/bevyengine/bevy-website/pull/355
         sed -i.bak \
-          -e 's/getObject(arg0).fetch(/window.customFetch(/' \
+          -e 's/getObject(arg0).fetch(/window.loadingBarFetch(/' \
           ../../content/examples/$category_slug/$example_slug/$example.js
 
         echo "+++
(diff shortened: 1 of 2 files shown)


generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -42,11 +42,7 @@ add_category()
         # Patch generated JS to allow to inject custom `fetch` with loading feedback.
         # See: https://github.com/bevyengine/bevy-website/pull/355
         sed -i.bak \
-          -e 's/function init(input) {/function init(customFetch, input) { customFetch = customFetch || fetch;/' \
-          -e 's/input = fetch(/input = customFetch(/' \
```
doup, line 46: Be aware, I think there were to different calls to `fetch`:
- `input = fetch(` => for `.wasm` file?
- `getObject(arg0).fetch(` => for assets?

With this change-set I think only the assets loading progress will show. If you slow down the connection via "network" panel you should be able to notice it.

IceSentry: Ah, good catch, yeah it didn't show up locally but I did notice a small time without a loading bar.

doup (approved):

ickk (approved):

ickk:
> I tried simply overriding fetch itself globally, but it broke everything.

yeah overwriting window.fetch itself is not a good idea, because if anything else on the page needs to use fetch it'll get hijacked by our code and start generating loading bars. 

What this PR does seems pretty reasonable for this case, where the example pages only have a single canvas on the page, but won't generalise to multiple canvases per page (something to look out for other future use cases of embedded bevy canvases)

doup:
Probably this should be raised to `wasm-bindgen` to see if there is a way for a long term solution.

PR #391, opened 2022-07-06
[Merged by Bors] - Specify file on `main.rs` mention

AngelOnFira:
This is a small change, but in walking through the Bevy book with some students I'm teaching this summer, they misunderstood this and thought it should be added to the main function instead.

It's definitely more to reduce confusion and ambiguity for newer Rust developers and might be a bit annoyingly verbose for others. Not sure.


alice-i-cecile (approved):
Agreed!

PR #355, opened 2022-04-18
[Merged by Bors] - Improve examples loading feedback

doup:
This is a proposal for #338.

- Builds on top of the code by @ickk using `TransformStream`.
- Adds loading feedback on examples.
- The solution is hacky, it monkey-patches `fetch` in the example template.

https://user-images.githubusercontent.com/188612/163875627-b11bf330-0fb8-4991-a710-e12e8657fe07.mp4

Co-Authored-By: ickk <17050131+ickk@users.noreply.github.com>

PR #392, opened 2022-07-08
[Merged by Bors] - Add Examples to nav and add "intro" text

cart:
I think we've ironed out enough of the kinks to justify making the examples page more prominent. We've been sitting on a lot of value here for awhile :smile: 

For posterity: big thanks to @mockersf for putting together this feature in #225 and @doup for adding loading feedback in #355 and a bunch of other visual improvements.

### DIFF
--- a/sass/pages/_assets.scss
+++ b/sass/pages/_assets.scss
@@ -1,4 +1,9 @@
 .assets {
+    .assets-intro {
+        font-size: 1.2rem;
+        margin-bottom: 20px;
+    }
+
     .asset-section {
         font-size: 2.4rem;
         margin: 0 0 20px;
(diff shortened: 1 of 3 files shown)


IceSentry (approved):
Not sure if we want to keep it in the learn section if it's already in the header, but other than that I tried it locally and it looks good.

alice-i-cecile:
I'm on board with this choice :)

mockersf (approved):

cart:
> Not sure if we want to keep it in the learn section if it's already in the header, but other than that I tried it locally and it looks good.

I think we probably do as we don't want people to accidentally miss the examples when looking for learning material.

Issue #231, opened 2021-12-30
Bevy Assets Cards Layout Update


BlackPhlox:
## (WIP)
Part of #159's tracking issues :
- Prettier small cards in list view
- Assets can have one primary tag and several secondary tags, can filter on tag

Requires metadata specified in #230

[A page view per asset prototype](https://www.figma.com/proto/BAWOVsxMB8JrQnfpGaIAIW/Bevy-Store?node-id=2%3A1&scaling=min-zoom&page-id=0%3A1) referenced in #159 from [discord](https://discord.com/channels/691052431525675048/691052431974465548/839581143476338758) by giusdp

[Design prototypes of the small cards in bevy-assets list view](https://www.figma.com/file/CZ1svP4l70eMuPYXIOK9d3/bevy_design_sheet?node-id=610%3A38):

![bevy-website](https://user-images.githubusercontent.com/25123512/147786531-80622d82-dd37-46e8-9cb2-0625aadf71f4.png)

ickk:
With the help & input from @alterae, @BlackPhlox, @cart, @DJMcNab, @mockersf, @neo97x, @TheRawMeatball, @Weasy666, we iterated on this design for asset cards: <https://jsfiddle.net/tqnr5a6v/>
![image](https://user-images.githubusercontent.com/17050131/148511406-45d114a7-2575-4f5b-9910-5c81b06bc7a7.png)

I think the big useful feature that is still missing is a download counter.

Would love to hear thoughts on this.

DJMcNab:
I think if we're making 'warning' judgements on licenses (which to be clear, I think we probably should), then I think we would probably do well to have a ⚠ icon next to them with a tooltip. That is, something clearer than orange highlighting, which is another kind of highlighting, and could therefore be misinterpreted as positive (and is potentially colourblind unfriendly).

My suggestion for a way to do the 'larger set' of highlighted blue ones was anything which would be accepted by bevy's `deny.toml` (ideally automatically checked based on that and the latest crates.io release; we should also check for licenses changing and require that to be manually reviewed).

I am slightly biased towards only highlighting MIT/Apache as blue makes sense, as a nudge towards it.; but I think doing it for any permissive license makes sense.

ickk:
I thought about the license thing a bit; I think from both a UX perspective and from a 'what opinion does bevy want to have', and I think just having blue solely for dual mit/apache makes the most sense.

1. Users don't need to think about any additional license complexities at all (if they choose 'the blue one').
[...]

PR #394, opened 2022-07-18
[Merged by Bors] - asset cards with more metadata

IceSentry:
## Objective

Update the asset cards to support more metadata. The main goal is to be able to show the supported bevy version. Everything else is secondary right now.

## Solution

Support more metadata by using the styling from https://github.com/bevyengine/bevy-website/issues/231.

I'm missing a few things, like different colors for bevy versions, but that's because I don't know how to do that with zola templates. I also ignored the author name and crate version for now because it doesn't apply to every crate and it can be iterated on in the future.

image

Supported metadata:
- `bevy_versions`: array of supported version
- `licenses`: array of licenses

The metadata currently doesn't exist, but the code is written in such a way to not break if it doesn't exist. I created #397 to generate the required metadata.

A lot of the values were just found by me playing around with values and finding something that looked good. I'm absolutely opened to changing them. To be clear, a lot of what I have right now are just values that ended up working, but there's probably a better way to do most of those things.

## Remaining questions

- ~~What to do with assets without images.~~
  - Solved: I added a gradient to make it a bit less empty.
- How to handle different color based on the value of the version tag by using zola templates

## Notes

Until we add the required metadata, the cards will have a lot of empty space at the bottom, I already have a prototype ready to add the missing metadata, so that should be fixed pretty fast.


IceSentry:
@TimJentzsch since I can't figure out how to request a review, hopefully this ping will do the job.

sass/components/_card.scss
```diff
@@ -85,4 +85,4 @@
     margin-left: auto;
     margin-right: auto;
     align-items: center;
-}
+}
```
mockersf, line 88:
```suggestion
}

```

mockersf:
How it looks for me (Safari and Firefox):
Screenshot 2022-07-18 at 12 06 49

* The icons seems to not be aligned with the centre of the tags
* The license icon is too close to the license tag
* If there are no version, the license is on the left

There is also an issue when there are too many bevy versions, this would hide the licenses. That could be solved later as it's unlikely to be an issue today

For the remaining questions, I would like this to move without solving them:
* Assets without an image: let's keep them empty for now, and take the time to find a good placeholder
* Colors for versions: colors are... too complicated. This info is needed now, even without colors

TimJentzsch:
> - The icons seems to not be aligned with the centre of the tags
> - The license icon is too close to the license tag
> - If there are no version, the license is on the left

We should be able to fix all of that with :sparkles: flexbox :sparkles: 

The main ingredients:

- `display: flex`: This is where the fun begins!
- `justify-content: space-between`: The first item goes to the start, the second item goes to the end.
- Wrap the version and license each in a `div`. Only the content of the `div` is conditional, that way you keep the license to the right.
- `align-items: center`: Aligns the icons at the center of the tags.
- `gap: 5px`: Space between icon and tag name.
- `flex-flow: row-wrap`: If there are too many bevy versions, the license wraps in the next row.

That should be everything we need. In a couple of hours I might also have time to do a mockup with vanilla HTML and CSS if needed.

afonsolage:
Since the new layout always expects an image (like a thumbnail), maybe add a [default bevy image](https://bevyengine.org/processed_images/4e424e8bfdfe1d4800.png) when there is no image on metadata?

image

The current layout adapts when there is no image.

mockersf:
> Since the new layout always expects an image (like a thumbnail), maybe add a [default bevy image](https://bevyengine.org/processed_images/4e424e8bfdfe1d4800.png) when there is no image on metadata?

Yes, we'll need a placeholder, but not the Bevy logo

TimJentzsch:
> The current layout adapts when there is no image.

In my opinion that makes more sense than having a big gray box that is not used

IceSentry:
For the image thing, I was mostly questioning what to do right now, but yes in the future we should either make images mandatory or at the very least some sort of generated placeholder. I believe once cards are more clearly used to show case images we'll see more people submitting images. Could be worth explicitly calling out in the blog post.

IceSentry:
For the bevy and license logo, that was me playing around with it and breaking it after making my original PR it was super late though so I stopped and went to bed. I'll try to fix it later today.

doup:
It's looking good!

For the image adding some light background image (maybe some grey cogs?) + the name of the asset on top should look good enough. I think that removing it, with this vertical layout, is going to look ugly as the title/text will be misaligned between cards.

I have a couple of minor nits on margins/padding/alignment. On top of what `@mockersf` already mentioned in https://github.com/bevyengine/bevy-website/pull/394#issuecomment-1187016606, I would say that the margins/padding would look better if they were more consistent:
- Reduce bottom gap of the card (on the current online version on both assets/examples cards the same gap is used for both axis). You probably have some `margin-bottom`; just removing it should work as the parent `grid` container is responsible for the gap.
- Increase left padding on asset title to match description
- Increase bottom padding for de card

image

Also, on your screenshot it looks like the text is not vertically centered on the "pills":

image

IceSentry:
Alright, I rewrote most of the styling, but this time I used css grid and it made everything way easier. I tried to apply most of the feedback that I saw here.

I also added a gradient for the assets without images just so it's less empty

image

@doup sorry I didn't see your feedback before committing, but I think I fixed all of that

sass/pages/_assets.scss
```diff
@@ -90,22 +90,34 @@
...
                 background-color: rgb(110, 110, 247);
             }
         }
-
-        a {
-            text-decoration: none;
-            color: $link-color;
-        }
     }
 }
```
TimJentzsch: Nit: Can we add a trailing linebreak here? GitHub complains about that

templates/macros/assets.html
```diff
@@ -1,56 +1,94 @@
...
-<div class="card" id="{{ post.title | slugify }}">
-    <a href="{{ post.extra.link }}">
-        {% if post.extra.image %}
-        <div class="card-image card-image-default">
+<div id="{{ post.title | slugify }}">
+    <a class="asset-card" href="{{ post.extra.link }}">
+        <div class="asset-card__banner">
+            <div class="asset-card__title">{{post.title}}</div>
+            {% if post.extra.image %}
             <img src="{{ image_macros::resize_image(path=post.extra.image, width=340, height=340) }}" />
```
TimJentzsch, line 31: This seems to be missing an `alt` attribute`

templates/macros/assets.html
```diff
@@ -1,56 +1,94 @@
...
+                {% for version in post.extra.bevy_versions %}
+                <span class="asset-card__tag">{{version}}</span>
+                {% endfor %}
+            </div>
+            {% endif %}
+            {% if post.extra.licenses %}
+            <div class="asset-card__tag-list asset-card__licenses">
+                {{ assets_macros::license_logo_svg() }}
+                {% for license in post.extra.licenses %}
+                <span class="asset-card__tag">{{license}}</span>
```
TimJentzsch, line 50: Similar here, except that `alt` attributes don't work on `svg`s.
From what I've read, the best way to do this is to add a `title` element inside the `svg`: [Blog post](https://css-tricks.com/accessible-svgs/#aa-2-inline-svg), [W3 Reference](https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements)

templates/macros/assets.html
```diff
@@ -1,56 +1,94 @@
...
+</li>
 {% endmacro %}
+
+{% macro assets_menu(root, prefix) %}
+<ul class="tree-menu">
+    {% for s in root.subsections %}
+    {{ self::assets_menu_row(prefix=prefix, section_path=s)}}
+    {% endfor %}
+</ul>
+{% endmacro %}
```
TimJentzsch, line 94: Nit: Trailing line break is missing here

sass/pages/_assets.scss
IceSentry: Is it possible you didn't refresh? My latest commit should have already fixed that one

templates/macros/assets.html
```diff
@@ -1,56 +1,94 @@
...
+    <div class="{{label_class}}">
+        <a class="tree-menu__link" href="#{{section.title | slugify}}">{{ section.title }}</a>
         {% if section.subsections %}
-            <ul class="tree-menu">
-                {% for s in section.subsections %}
-                    {{ self::assets_menu_row(prefix=prefix, section_path=s) }}
-                {% endfor %}
-            </ul>
+        <label class="tree-menu__toggle" for="{{id}}">
+            <img class="tree-menu__chevron" src="/assets/icon-chevron-down.svg">
```
TimJentzsch, line 74: This is missing an `alt` attribute and should maybe have a `title` attribute to indicate that you can toggle something here (if I understand this correctly)

TimJentzsch (changes requested):
The screenshots you sent look good to me!

I did a review run on accessibility issues and found a few, would be nice to fix those before merging :)

templates/macros/assets.html
IceSentry: I want to convert the svg to an image asset like the bevy logo anyway so that will work

templates/macros/assets.html
```diff
@@ -1,56 +1,94 @@
...
-                {{ post.description | striptags | truncate(length=100) | safe }}
+        <div class="asset-card__description">
+            {{ post.description | striptags | truncate(length=120) | safe }}
+        </div>
+        <div class="asset-card__tags">
+            {% if post.extra.bevy_versions %}
+            <div class="asset-card__tag-list asset-card__bevy-versions">
+                <img src="/assets/bevy_icon_dark.svg" alt="Small bevy icon">
+                {% for version in post.extra.bevy_versions %}
+                <span class="asset-card__tag">{{version}}</span>
```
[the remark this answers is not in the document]
IceSentry: I literally had that exact "Supported bevy versions" alt text earlier, I have no idea why I changed it lol and yeah, using a title is a good call.

templates/macros/assets.html
```diff
@@ -37,7 +24,7 @@
         <div class="asset-card__tags">
             {% if post.extra.bevy_versions %}
             <div class="asset-card__tag-list asset-card__bevy-versions">
-                <img src="/assets/bevy_icon_dark.svg" alt="Small bevy icon">
+                <img src="/assets/bevy_icon_dark.svg" title="Supported bevy versions">
```
TimJentzsch, line 27: Nit: we need both the `alt` and `title`, as the `alt` is also rendered when the `img` can't be loaded.

(Sorry, I should have made that clearer in my comment)

templates/macros/assets.html
```diff
@@ -71,7 +58,8 @@
         <a class="tree-menu__link" href="#{{section.title | slugify}}">{{ section.title }}</a>
         {% if section.subsections %}
         <label class="tree-menu__toggle" for="{{id}}">
-            <img class="tree-menu__chevron" src="/assets/icon-chevron-down.svg">
+            <img class="tree-menu__chevron" src="/assets/icon-chevron-down.svg"
+                title="Toggle {{section.title}} section">
```
TimJentzsch, line 62: Same here, we also need the `alt` attribute.

templates/macros/assets.html
```diff
@@ -21,14 +8,14 @@
 </svg>
 {% endmacro license_logo_svg %}
 
-
 {% macro card(post) %}
 <div id="{{ post.title | slugify }}">
     <a class="asset-card" href="{{ post.extra.link }}">
```
TimJentzsch, line 13: The [MDN docs for `a`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) state:

> Content within each `<a>` **should** indicate the link's destination.

Such that users with visual impairments/using screen-readers can tell where they are going.
I assume that this is not the case here, because we are wrapping a `div` element.

However, I could not find any guidelines to make this accessible. If anyone knows anything, please let me know!

Maybe it's fine though because of the card title that's deeper in the hierarchy?

TimJentzsch (changes requested):

templates/macros/assets.html
TimJentzsch: I think they mean that the content of the `<a>` should indicate where the link leads to: [Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#strong_link_text)

TimJentzsch (approved):
Awesome, looks good to me!

sass/pages/_assets.scss
```diff
@@ -18,29 +18,109 @@
...
+            background-color: rgba(0, 0, 0, 0.5);
+        }
+
+        &__description {
+            grid-area: desc;
+            display: flex;
+
+            padding: 0.3rem;
+            font-size: 1rem;
+        }
```
doup, line 77: I've tried with a little bit of extra padding on the title/description (0.6rem ~ 8px), what do you think?

```suggestion
        &__title {
            position: absolute;
            width: 100%;
            bottom: 0;

            padding: 0.6rem;
            font-weight: bold;
            font-size: 1.2rem;
            background-color: rgba(0, 0, 0, 0.5);
        }

        &__description {
            grid-area: desc;
            display: flex;

            padding: 0.6rem;
            font-size: 1rem;
        }
```

image

sass/pages/_assets.scss
```diff
@@ -18,29 +18,109 @@
         margin-bottom: 52px;
     }
 
-    .card {
-        margin: 0;
-        height: 8rem;
-    }
+    .asset-card {
```
doup, line 21: I would move `asset-card` into it's own file `sass/components/_asset-card.scss`.
Pages folder (`pages/*`) is kind-of legacy; where both BEM and non-BEM gets mixed.

IceSentry: Sure, I'm still new to BEM so wasn't sure how file structure should work so I just updated the existing file. I'll do that tomorrow.

sass/pages/_assets.scss
```diff
@@ -18,29 +18,109 @@
...
+                background-color: $default-image-background-color;
+            }
+        }
+
+        &__title {
+            position: absolute;
+            width: 100%;
+            bottom: 0;
+
+            padding: 0.6rem;
```
doup, line 65: This shrinks the vertical padding for the title, this way it doesn't look so "airy"… and I stop. 😇 
Anyway, if you think `0.6rem` is too much please reduce as you wish… it's just that I find `0.3rem` too tight.

```suggestion
            padding: 0.3rem 0.6rem;
```

image

sass/pages/_assets.scss
doup: Sure, go for it. `0.6` was a suggestion.

doup (approved):

sass/pages/_assets.scss
```diff
@@ -18,29 +18,109 @@
...
+            background-color: rgba(0, 0, 0, 0.5);
+        }
+
+        &__description {
+            grid-area: desc;
+            display: flex;
+
+            padding: 0.6rem;
+            font-size: 1rem;
+        }
```
doup, line 77: I've noticed that this doesn't handle extra long text correctly.

**Option A**: use 4 lines ellipsis ([the mixin uses this snippet](https://css-tricks.com/almanac/properties/l/line-clamp/)).

```suggestion
        &__description {
            @include text-ellipsis(4);

            grid-area: desc;

            padding: 0.6rem;
            font-size: 1rem;
        }
```

**Option B**, see comment:

sass/pages/_assets.scss
```diff
@@ -18,29 +18,109 @@
...
-        margin: 0;
-        height: 8rem;
-    }
+    .asset-card {
+        display: grid;
+        grid-template-areas:
+            "banner"
+            "desc"
+            "tags";
+        grid-template-rows: 12rem 5.5rem 2rem;
```
doup, line 27: **Option B**, for comment: https://github.com/bevyengine/bevy-website/pull/394/files#r924157060

```suggestion
        grid-template-rows: 12rem minmax(5.5rem, 1fr) 2rem;
        height: 100%;
```

- `minmax` to allow the description grow, but keep a minimum height
- `height` for the other cards, so that they grow to match the one with the long description

image

sass/pages/_assets.scss
IceSentry: 😍 I love it, I that was the missing part.

sass/pages/_assets.scss
ickk: I'm biased against growing the cards vertically; when we were designing the initial mockups we had a lot of discussion about the amount of text that would be appropriate for cards, and we landed on the ratios and everything that we did based on a relatively little amount of text. 

Descriptions should be short and to the point. Longer exposition about a particular asset belongs on it's asset page (whenever we build that feature) or on the crates.io/repo/webpage linked to imo

doup: This was just an option I was giving him, related comment: https://github.com/bevyengine/bevy-website/pull/394/files#r924157060
He used **"Option A"** in the end, which is just "ellipsis at the fourth line": https://github.com/bevyengine/bevy-website/pull/394/commits/1a93da13e2cdbe0d8d8b2ac4949dba917e76f775

sass/pages/_assets.scss
doup: Sorry, I didn't realize, I saw it a after you removed it again on the comment. 🤦

alice-i-cecile:
@IceSentry this needs merge conflict resolution, but IMO we should ship this and iterate.

IceSentry:
I'd rather shipping the metadata loading first. That way the cards will have actual data to show. Right they will just look weirdly empty

alice-i-cecile:
Sounds good to me. I'll mark this as Blocked so we don't forget.

IceSentry:
Well, technically it isn't blocked, it could easily be merged without any issue, it would just look a bit empty at the bottom since there would be no license or bevy versions

alice-i-cecile:
Technically, but I'd be against it :p

IceSentry:
The other PR is ready it just needs more reviews. it's just that most people don't look at `bevy-website` 😦

sass/components/_asset-card.scss
```diff
@@ -0,0 +1,107 @@
...
+        background-color: $card-hover-background;
+        border: 2px solid #6b6b6b;
+    }
+
+    &__banner {
+        grid-area: banner;
+        display: block;
+        position: relative;
+
+        background: linear-gradient(45deg,
```
cart, line 26: Totally a matter of opinion, but I'd like to propose a more subtle / aligned gradient: `linear-gradient(0deg, #37393e 0%, #1b1b1b 100%)`

image

The current one draws the eye a bit too much and feels a bit aggressive / like a bright light is shining on something very metallic / reflective. And notably with that analogy the light should behave differently for each tile based on its position relative to your eye, but it doesn't. Which also contributes to my (personal) impression that it feels "off":

image

sass/components/_asset-card.scss
```diff
@@ -0,0 +1,107 @@
+.asset-card {
+    display: grid;
+    grid-template-areas:
+        "banner"
+        "desc"
+        "tags";
+    grid-template-rows: 12rem 5.5rem 2rem;
```
cart, line 7: I think we should probably rein in the height of the text section a bit (cut down on the character limit if necessary, maybe reduce the font size). The cards have _just_ a bit too much height right now. Most descriptions are short and this results in _a lot_ of wasted space. I'm down to do that in a separate PR though.

cart (commented):
I'm loving these changes!

sass/components/_asset-card.scss
IceSentry: Good idea, I like it, a lot more subtle while still being nicer than a solid color.

sass/components/_asset-card.scss
IceSentry: I'll see what I can do. I picked the height based on the longer descriptions but it was mostly just playing with some values. I'm down with reducing the size a bit

IceSentry: I reduced the padding a bit and limited it to 3 lines now. This is how it looks.
image

I tried going to 2 lines max but it looked weird.

rparrett:
I think that we may have lost lazy-loading of the asset images in the rebase.

templates/macros/assets.html
```diff
@@ -1,56 +1,93 @@
...
-    <a href="{{ post.extra.link }}">
-        {% if post.extra.image %}
-        <div class="card-image card-image-default">
-            <img src="{{ image_macros::resize_image(path=post.extra.image, width=340, height=340) }}" loading="lazy" />
+<div id="{{ post.title | slugify }}">
+    <a class="asset-card" href="{{ post.extra.link }}">
+        <div class="asset-card__banner">
+            <div class="asset-card__title">{{post.title}}</div>
+            {% if post.extra.image %}
+            <img src="{{ image_macros::resize_image(path=post.extra.image, width=340, height=340) }}"
```
rparrett, line 21:
```suggestion
            <img src="{{ image_macros::resize_image(path=post.extra.image, width=370, height=370) }}"
```

sass/components/_asset-card.scss
cart: I dig it! One more small tweak: I do think adding back a bit more padding is "worth it".

templates/macros/assets.html
```diff
@@ -1,11 +1,16 @@
...
-    <path d="M554.9 154.5c-17.62-35.25-68.12-35.38-85.87 0c-87 174.3-84.1 165.9-84.1 181.5c0 44.13 57.25 80 128 80s127.1-35.88 127.1-80C639.1 319.9 641.4 327.3 554.9 154.5zM439.1 320l71.96-144l72.17 144H439.1zM256 336c0-16.12 1.375-8.75-85.12-181.5c-17.62-35.25-68.12-35.38-85.87 0c-87 174.3-84.1 165.9-84.1 181.5c0 44.13 57.25 80 127.1 80S256 380.1 256 336zM127.9 176L200.1 320H55.96L127.9 176zM495.1 448h-143.1V153.3C375.5 143 393.1 121.8 398.4 96h113.6c17.67 0 31.1-14.33 31.1-32s-14.33-32-31.1-32h-128.4c-14.62-19.38-37.5-32-63.62-32S270.1 12.62 256.4 32H128C110.3 32 96 46.33 96 64S110.3 96 127.1 96h113.6c5.25 25.75 22.87 47 46.37 57.25V448H144c-26.51 0-48.01 21.49-48.01 48c0 8.836 7.165 16 16 16h416c8.836 0 16-7.164 16-16C544 469.5 522.5 448 495.1 448z"/>
+{% macro init_svg() %}
+<svg style="display: none" version="2.0">
+    <defs>
+        <symbol id="asset-card__scale-balanced-svg" viewBox="0 0 640 512">
+            <!--! Font Awesome Pro 6.1.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2022 Fonticons, Inc. -->
```
cart, line 7: Hmm this is explicitly calling itself out as "font awesome pro". We shouldn't be using any icons from there (and I dont think this one is part of pro).

cart: https://fontawesome.com/license/free

IceSentry: That's the exact svg with the same comment I get when I copy it from there https://fontawesome.com/icons/scale-balanced?s=solid

I'm not sure why it says it's under the pro license

sass/components/_asset-card.scss
IceSentry: I increased the padding a bit and made the area for the tags slightly larger so it could have a more consistent padding.

image

cart: Lovely!

templates/macros/assets.html
cart: If you download the "free" package, the svgs have the correct license embedded in them. I just pushed that svg content.

IceSentry: Interesting, it's a bit weird that they have the wrong license on the site for the exact same svg. That's good to know though so I downloaded it for the future.

PR #397, opened 2022-07-19
[Merged by Bors] - Generate extra metadata from external sources

IceSentry:
## Objective

The new asset cards (#394) need more metadata for license and bevy compatible versions.

## Solution

- For assets with a link to github or gitlab:
  - Get the Cargo.toml file using the apis, parse it and read the bevy version and license fields.
  - This is done on a best effort basis. It's not always possible to have this information. For example, when the Cargo.toml is not at the top level. 
- For assets with crates.io links
  - Use the crates.io dbdump to get the required data.
- If the metadata is entered manually directly in bevy_assets it will be used instead.

The github client relies on a GITHUB_TOKEN env var to exist in CI or locally. It should be already there since we use github actions. The rate limit in github actions is 1000 per hour https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions so we should be fine for a little while.

The gitlab client implementation ignores the token because we have so few crates that it doesn't matter.

## Notes

Querying the crates.io dump is pretty slow and the deps-tree required is pretty big. I think we should consider eventually rolling our own way to do that, but it works well right now.

This is going to be pretty slow to execute in the job once it is merged, but it's pretty easy to parallelize all the generate-* jobs so I did it in #401 .

### DIFF
--- a/generate-assets/.gitignore
+++ b/generate-assets/.gitignore
@@ -1,2 +1,4 @@
 target
 assets
+data/
+.env
--- a/generate-assets/src/lib.rs
+++ b/generate-assets/src/lib.rs
@@ -1,5 +1,14 @@
+use anyhow::{bail, Context};
+use cratesio_dbdump_csvtab::CratesIODumpLoader;
+use github_client::GithubClient;
+use gitlab_client::GitlabClient;
 use serde::Deserialize;
-use std::{fs, io, path::PathBuf, str::FromStr};
+use std::{fs, path::PathBuf, str::FromStr};
+
+pub mod github_client;
+pub mod gitlab_client;
+
+type CratesIoDb = cratesio_dbdump_csvtab::rusqlite::Connection;
 
 #[derive(Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
@@ -9,10 +18,37 @@ pub struct Asset {
     pub description: String,
     pub order: Option<usize>,
     pub image: Option<String>,
+    pub licenses: Option<Vec<String>>,
+    pub bevy_versions: Option<Vec<String>>,
 
     // this field is not read from the toml file
     #[serde(skip)]
     pub original_path: Option<PathBuf>,
+}
+
+impl Asset {
+    /// Parses a license string separated with OR into a Vec<String>
+    fn set_license(&mut self, license: Option<String>) {
+        if self.licenses.is_some() {
+            return;
+        }
+        if let Some(license) = license {
+            let licenses = license
+                .split(" OR ")
+                .map(|x| x.trim().to_string())
+                .collect();
+            self.licenses = Some(licenses);
+        }
+    }
+
+    fn set_bevy_version(&mut self, version: Option<String>) {
+        if self.bevy_versions.is_some() {
+            return;
+        }
+        if let Some(version) = version {
+            self.bevy_versions = Some(vec![version]);
+        }
+    }
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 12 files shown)


IceSentry:
I added support for querying the crates db, but I'm not sure how it will work in CI. Right now it downloads it whenever we generate. I made sure to not run it on valide though.

BlackPhlox:
I implemented ci caching for crates db, so it only fetches the zip after 24 hours, which fits the timestamp for when the zip file gets updated by new data: https://github.com/BlackPhlox/bevy-website/blob/8bbe21286035b60547e6ba630241896c8fb1741f/.github/workflows/ci.yml#L16-L24

BlackPhlox:
> Also, surprisingly, this is way slower than the network call to github. I would have assumed, since it's a local db it would be pretty quick. I guess there's just too many crates.

The current bottleneck that takes the longest for the local db is [reversed dependencies lookup](https://github.com/alyti/cratesio-dbdump-lookup/blob/0ab1c88acab98323af56824e0abeb369542574eb/src/lib.rs#L277).

IceSentry:
Ah, nice, I'll add the ci caching.

> The current bottleneck that takes the longest for the local db is [reversed dependencies lookup](https://github.com/alyti/cratesio-dbdump-lookup/blob/0ab1c88acab98323af56824e0abeb369542574eb/src/lib.rs#L277).

Oh right, completely forgot to update that call to use the same approach as you did which doesn't do a reverse lookup I think?

BlackPhlox:
I think it properly a good idea to have a notice about dbdump size and instructions if users want to run the generate-assets locally, something like https://github.com/IceSentry/bevy-website/pull/3

BlackPhlox:
The caching should also have to be added to `.github/workflows/deploy.yml`, though I don't know if they are going to share the cache, ideally, they should.

BlackPhlox:
That's because `get_crate` uses `rev_dependency` 😅 `get_crate` combines all query to get the full crate and its reverse dependencies as you get on [crates.io](https://crates.io/crates/bevy_config_cam/0.3.0/dependencies). Also, I think my naming of `rev_dependency` is wrong, I should have just called it `dependency_lookup` or the like.

generate-assets/src/lib.rs
```diff
@@ -9,12 +18,36 @@ pub struct Asset {
...
 }
 
+impl Asset {
+    /// Parses a license string separated with OR into a Vec<String>
+    fn set_license(&mut self, license: Option<String>) {
+        if self.licenses.is_some() {
+            return;
+        }
+        if let Some(license) = license {
+            let licenses = license.split("OR").map(|x| x.trim().to_string()).collect();
```
ickk, line 36: should this be `" OR "` to avoid splitting on licenses that happen to contain the characters?

IceSentry: Yeah, that's probably a good idea
```suggestion
            let licenses = license.split(" OR ").map(|x| x.trim().to_string()).collect();
```

generate-assets/.gitignore
```diff
@@ -1,2 +1,4 @@
 target
 assets
+data/
+.env
```
alice-i-cecile, line 4:
```suggestion
.env

```

alice-i-cecile (approved):
Code is clear and the general strategy seems sensible. I'm fine to start moving forward on this.

mockersf (approved):
Code looks good but I didn't test it.

I think we need @cart  👍  for this PR because of the token

IceSentry:
The `GITHUB_TOKEN` secret is supposed to be configured by default in any github actions job and for now I don't use any gitlab token so I don't think it needs any manual step.

mockersf:
no manual steps, just approval @cart expressed his preference to be aware of token use

IceSentry:
Oh, makes sense, it's definitely going to use that token a lot more now.

cart (approved):
Yup this looks good to me!

Something we should look out for: if we get rate limited on our github token, that might block our other uses of it. If CI stuff starts randomly failing after merging this, we'll probably need to revert the github integration part until we can sort something out.

PR #400, opened 2022-07-23
[Merged by Bors] - add .vscode to gitignore

IceSentry:
I have a few extensions causing me problem in this repo and I like disabling them in a vscode/settings.json

Since main bevy already has that ignore I feel it's ok to also add it here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ content/learn/errors
 content/examples
 static/assets/examples
 static/processed_images
+.vscode/


alice-i-cecile (approved):
Merging as trivial.

bors r+

Nilirad:
I don't like this. It's better that the PR author instead ignores editor specific files on their end instead of requesting a change in every project they contribute to.

TimJentzsch:
You can do this by setting a global gitignore file: <https://stackoverflow.com/a/7335487>

Nilirad:
I recognize though that it could be beneficial for the project to ignore common files, for user friendliness.

alice-i-cecile:
I decided to opt for user friendliness and consistency with the main repo here :) I think your solution is more correct, but I want to make sure novices can participate with a minimum amount of yak shaving.

PR #401, opened 2022-07-24
[Merged by Bors] - Parallel actions in CI and add a cache

IceSentry:
## Objective

CI takes a long time because each generate job is done sequentially and there's no caching.
Fix that.

## Solution

Create one job per generate-* and wait until they are done to build the website.

- Each job uploads artifacts
- The final jobs download all the artifacts and use them to build the website
- Each job also caches their respective build artifacts to speed up build time
  - For the wasm-examples it also caches the wasm files. The cache is invalidated once the bevy version tagged latest changes

This is all done only for the CI workflow. Eventually this could also be done with the deploy workflow by using something like https://docs.github.com/en/actions/using-workflows/reusing-workflows.

This is what the jobs look like now. The timing were done with a pre-existing cache.

vivaldi_TOqEly8nzH

## Notes

The longest job by far is still generating examples right now, but generating the assets metadata is about to get a lot longer. I believe we can create multiple jobs to parallelize the examples a bit more. Probably one per category. This can be done in a future PR.

I changed how the assets path are generated because it was causing issues when trying to zip them before the upload,


alice-i-cecile (approved):

IceSentry:
Notes for future me, consider using this to update deploy job https://docs.github.com/en/actions/using-workflows/reusing-workflows

alice-i-cecile:
bors retry

mockersf:
job name change should be done in https://github.com/bevyengine/bevy-website/blob/master/.github/bors.toml too

mockersf (approved):

PR #402, opened 2022-07-24
Add example WASM optimization

nicopap:
- Add the standard code size optimization flags to `Cargo.toml`
- Use wasm-opt from binaryen

This reduces the generated wasm file sizes from `12M` to `5.9M`.

**DISCLAIMER**:

This increases **massively** deployment times. On my very
beefy machine, I get 30 minutes to generate all examples
on a cold build.

### DIFF
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -19,19 +19,19 @@ sed -i.bak 's/asset_folder: "assets"/asset_folder: "\/assets\/examples\/"/' crat
 
 export CARGO_TARGET_DIR="target"
 
-if [ "$1" = --ci ] ; then
+if [ "$1" = --optimize-wasm ] ; then
+    # enable optimizations
+    cargo_profile="wasm-release"
+    cargo_target_dir="wasm-release"
+    # Optimize for size
+    wasm_opt_flag="-Oz"
+else
     # disable optimizations
     cargo_profile="dev"
     cargo_target_dir="debug"
     # Do not optimize on --ci check,
     # wasm-opt takes A LOT of time to run
     wasm_opt_flag="-O0"
-else
-    # enable optimizations
-    cargo_profile="wasm-release"
-    cargo_target_dir="wasm-release"
-    # Optimize for size
-    wasm_opt_flag="-Oz"
 fi
 
 examples_dir="../../content/examples"
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -6,7 +6,6 @@
 git init bevy
 cd bevy
 git remote add origin https://github.com/bevyengine/bevy
-git restore Cargo.toml
 git pull --depth=1 origin latest
 
 # remove markdown files from assets so that they don't get picked up by Zola
@@ -21,18 +20,16 @@ sed -i.bak 's/asset_folder: "assets"/asset_folder: "\/assets\/examples\/"/' crat
 export CARGO_TARGET_DIR="target"
 
 if [ "$1" = --ci ] ; then
+    # disable optimizations
+    cargo_profile="dev"
+    cargo_target_dir="debug"
     # Do not optimize on --ci check,
     # wasm-opt takes A LOT of time to run
     wasm_opt_flag="-O0"
 else
-    echo "
-
-[profile.release]
-opt-level = \"z\"
-lto = \"fat\"
-codegen-units = 1
-" >> Cargo.toml
-
+    # enable optimizations
+    cargo_profile="wasm-release"
+    cargo_target_dir="wasm-release"
     # Optimize for size
     wasm_opt_flag="-Oz"
 fi
@@ -61,10 +58,10 @@ add_category()
         out_dir=$examples_dir/$category_slug/$example_slug
         mkdir $out_dir
         cp examples/$category_path/$code_filename $out_dir
-        cargo build --release --target wasm32-unknown-unknown --example $example
+        cargo build --profile $cargo_profile --target wasm32-unknown-unknown --example $example
 
         wasm-bindgen --out-dir $out_dir \
-            --no-typescript --target web target/wasm32-unknown-unknown/release/examples/$example.wasm
+            --no-typescript --target web target/wasm32-unknown-unknown/$cargo_target_dir/examples/$example.wasm
 
         wasm-opt $wasm_opt_flag \
             --output "$out_dir/${example}_bg.wasm.optimized" \
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -1,8 +1,12 @@
 #!/bin/sh
 
+# Print all executed lines
+# set -x
+
 git init bevy
 cd bevy
 git remote add origin https://github.com/bevyengine/bevy
+git restore Cargo.toml
 git pull --depth=1 origin latest
 
 # remove markdown files from assets so that they don't get picked up by Zola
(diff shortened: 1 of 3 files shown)


generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -14,6 +18,26 @@ sed -i.bak 's/canvas: None,/canvas: Some("#bevy".to_string()),/' crates/bevy_win
 # setting the asset folder root to the root url of this domain
 sed -i.bak 's/asset_folder: "assets"/asset_folder: "\/assets\/examples\/"/' crates/bevy_asset/src/lib.rs
 
+export CARGO_TARGET_DIR="target"
+
+if [ "$1" = --ci ] ; then
+    # Do not optimize on --ci check,
+    # wasm-opt takes A LOT of time to run
+    wasm_opt_flag="-O0"
```
nicopap, line 26: Note: I set the flag to `-O0`, because this allows validating the `wasm-opt` command in CI, reducing divergence between ci and deploy, and avoiding nasty surprises when deploying for the first time

generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -14,6 +18,26 @@ sed -i.bak 's/canvas: None,/canvas: Some("#bevy".to_string()),/' crates/bevy_win
...
+else
+    echo "
+
+[profile.release]
+opt-level = \"z\"
+lto = \"fat\"
+codegen-units = 1
+" >> Cargo.toml
```
mockersf, line 34: I think this could be added to the bevy repo using another profile name, see https://github.com/bevyengine/bevy/pull/5443#issuecomment-1193285708

nicopap, line 34: I was wondering whether we should just add those lines to the main `Cargo.toml` in the https://github.com/bevyengine/bevy repo. I opted to add them here because it's easier :P

generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/sh
 
+# Print all executed lines
+# set -x
+
 git init bevy
 cd bevy
 git remote add origin https://github.com/bevyengine/bevy
+git restore Cargo.toml
```
mockersf, line 9: why is this needed?

nicopap: Because of the `echo >> Cargo.toml` :D I'll be glad to add it to the main repo `Cargo.toml`

nicopap:
So relying on the new profile is a bit tricky, since this syncs to the latest release (with `git pull origin latest`). The CI checks for this PR will work, but until https://github.com/bevyengine/bevy/pull/5443 is merged and 0.8 released, deploy will fail.

I tested it locally with my branch, and it works, but "works on my machine" is never the most encouraging.

generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -14,6 +17,24 @@ sed -i.bak 's/canvas: None,/canvas: Some("#bevy".to_string()),/' crates/bevy_win
 # setting the asset folder root to the root url of this domain
 sed -i.bak 's/asset_folder: "assets"/asset_folder: "\/assets\/examples\/"/' crates/bevy_asset/src/lib.rs
 
+export CARGO_TARGET_DIR="target"
+
+if [ "$1" = --ci ] ; then
```
IceSentry, line 22: Shouldn't this be the other way around? Optimize when in CI, but don't optimize when ran locally? Unless this is more meant as a deploy vs not deploy

IceSentry:
Right now this is going to take a very long time because it builds all examples sequentially. To make this faster, what could be done is to build each example category in parallel. I tried doing that by using a job matrix and each category was an entry in the matrix array. That way you can pass the category to the sh script and it only compiles that category. 

One issue with that is that it needs to fully compile bevy for each job independently, but I'm not sure if it's worth it to try and build it separately and upload build artifact to each job. It's wasteful, but time wise the end result are probably going to be similar.

Another thing to consider is that my approach assumes categories are passed as arguments to the script, this is nice for ci, but is less nice when running it locally, so you would need to take this into account with your ci flag.

IceSentry:
To be clear, the parallelization I'm suggesting aren't required for this PR and could be done later.

generate-wasm-examples/generate_wasm_examples.sh
nicopap: Correct, I use `--ci` here for CI vs deploy

nicopap:
Wouldn't the shell job system be enough, add `&` after the calls to `add_category`? Not that I am suggestion we do it in this PR.

generate-wasm-examples/generate_wasm_examples.sh
IceSentry: Ah right, so ci would be the name of the workflow and not just the general concept of CI, makes sense. I'd still suggest changing it to deploy since it feels a bit clearer to me, but I'm fine with it either way. Maybe ci-job or ci-workflow?

IceSentry:
I don't think that would work because you need to wait for the completion of those jobs before going to the next, but also, since this used for compiling they would just be fighting for resources. Using a matrix would make it easier to parallelize because you don't need special setup to detect the task completion and they wouldn't share processing resources.

IceSentry:
With that said, since it would only be slow when doing a deploy I think it's not that big of an issue. Especially if we combine it with the things I did in #401

generate-wasm-examples/generate_wasm_examples.sh
nicopap: Good thinking. I'll go with something more explicit though, like `--optimize-wasm`

generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -14,6 +17,24 @@ sed -i.bak 's/canvas: None,/canvas: Some("#bevy".to_string()),/' crates/bevy_win
 # setting the asset folder root to the root url of this domain
 sed -i.bak 's/asset_folder: "assets"/asset_folder: "\/assets\/examples\/"/' crates/bevy_asset/src/lib.rs
 
+export CARGO_TARGET_DIR="target"
```
IceSentry, line 20: Why is this an export?

IceSentry: Actually, is this used at all?

IceSentry (approved):
I'm not sure what's up with CARGO_TARGET_DIR, but other than that LGTM

generate-wasm-examples/generate_wasm_examples.sh
nicopap: It's actually the default.

This shouldn't  affect the caller's environment (it only affects programs called within the script, just tried it with with a `test.sh`, content of which was `export FOO="bar"`).

The default value of `CARGO_TARGET_DIR` is indeed `"target"`, on my machine I set it to something else, but it breaks this script. So I set the value to the default one here, so that when `cargo` is called in this script, it uses a consistent target dir.

IceSentry: I guess it doesn't break anything to leave it there, but it feels a bit weird since in CI it should already be the default. I guess it makes it less likely to break unexpectedly.

nicopap:
Note from `@rparrett`'s comment on https://github.com/bevyengine/bevy/pull/5443 it is possible to build the stress test with `-O3` separately from the rest of the examples. Should allow better optimization where it matters.

NiklasEi:
I was about to open another PR doing the (almost) same thing using the wasm-opt action ([see the diff](https://github.com/bevyengine/bevy-website/compare/master...NiklasEi:bevy-website:optimize_web_examples)). Maybe using the action here could save us a little bit of time compared to the `apt-get install`?

cart:
I do like how simple @NiklasEi's impl is. How do the outputs compare? Any tradeoffs?

NiklasEi:
The install times that I mentioned before should not really matter.

One difference is the order in which things are done. Here every single example is first built and then optimized. With the action, all builds will complete first. Only then does the optimizing start. That would be a shorter feedback cycle in case something goes wrong with building.

Another difference might be the binaryen version. I don't know how fast new versions are available for apt, but the action can be behind by a few months now and then. It only takes a minute, but [the version in the action needs to be updated manually](https://github.com/NiklasEi/wasm-opt-action/blob/main/Deployment.md).

Apart from that the same `wasm-opt` configurations (=> outputs) should be possible with both approaches. I am happy about every line of bash though, that can be replaced by some workflow configuration.

nicopap:
The advantage of the script-based approach is that it serves as a ready-to-use example for potential users. But I'd rather see Niklas' change be merged than have to update this PR...

cart:
Perhaps also relevant: I think we should seriously consider moving off of bash for these things to ensure this can run on all platforms easily. https://github.com/bevyengine/bevy-website/pull/542#issuecomment-1463046585

Another benefit to using a local script is that you can test the results locally. If we move to actions then opt only happens in CI.

I think my take is: we should port this to cargo/xshell and do the optimization locally (although binaryen needs manual setup for the system so it should probably be optional).

PR #404, opened 2022-07-25
[Merged by Bors] - add intro text to assets page

IceSentry:
## Objective

The assets page has no text describing what it actually is, it should be made a bit clearer. 

## Solution
I used the description from the bevy_assets README to clarify it a bit more.

vivaldi_znFCzIkiSO

## Notes

I'm sure some people would like a clear disclaimer related to licenses. I'm open to the idea, but this could be done in a separate PR and shouldn't block this PR.

### DIFF
--- a/templates/assets.html
+++ b/templates/assets.html
@@ -13,6 +13,11 @@
 
 {% block page_content %}
     <div class="assets">
+        <div class="assets-intro">
+            A collection of Bevy assets, plugins, learning resources, and apps made by the community. If you would like to
+            share what you're working on, <a href="https://github.com/bevyengine/bevy-assets">submit a PR</a>!
+        </div>
+
         {% for subsection in section.subsections %}
         {% set section = get_section(path=subsection) %}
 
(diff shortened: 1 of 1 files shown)


templates/assets.html
```diff
@@ -13,6 +13,11 @@
 
 {% block page_content %}
     <div class="assets">
+        <div class="assets-intro">
+            A collection of Bevy assets, plugins, learning resources, and apps made by the community. If you would like to
```
Nilirad, line 17:
```suggestion
            A collection of third-party Bevy assets, plugins, learning resources, and apps submitted by community members. If you would like to
```

IceSentry: I tried to not change the wording from the README to be as not controversial as possible. I don't mind the "third party", but I'm not sure what's wrong with "made by the community".

Jazarro: Agreed, 'made by the community' rolls off the tongue much more easily. In addition, the next sentence already contains the word 'submit', so now you're using the same word twice in short succession.

Nilirad (approved):
Looks good on browser. I added a suggestion about using “pull request” instead of the PR acronym.

Nilirad (approved):

cart (approved):
Good call!

PR #405, opened 2022-07-26
[Merged by Bors] - Add utility to write Zola rustdoc `hide_lines`

### LINKED ISSUE #373: Hidden code in examples is not hidden when display on Zola
In Rust doc tests, you can use the `# ` line prefix to hide lines. This prevents them from showing up in your helpful minimal explanations, and allows you to avoid repetitive boilerplate or tests.

In #295, I tried to do the same thing, as our examples our currently compiled. However, this does not work: 
image

Fixing this is critical for the new book release: we either need to enable hiding in some form or figure out an alternate approach to compiling examples.

doup:
This PR adds an utility app that automatically writes [`hide_lines` Zola annotations](https://www.getzola.org/documentation/content/syntax-highlighting/#annotations) on book contents.

Fixes #373


IceSentry:
I really like this, it will make writing code blocks way easier. The code looks fine to me, but I think you should add either a README or at least a top level doc comment that explains what it actually does. It took me a while to figure out exactly what it was doing and I'm already very familiar with the problem.

write-rustdoc-hide-lines/src/code_block_definition.rs
```diff
@@ -0,0 +1,252 @@
...
+        Some(CodeBlockDefinition {
+            annotations,
+            hide_lines_idx,
+        })
+    }
+
+    pub fn get_hidden_ranges(&self) -> Option<&HiddenRanges> {
+        self.hide_lines_idx.map(|idx| match &self.annotations[idx] {
+            Annotation::HideLines(ranges) => ranges,
+            Annotation::Other(_) => panic!("Unreachable?"),
```
IceSentry, line 108:
```suggestion
            Annotation::Other(_) => unreachable!(),
```

write-rustdoc-hide-lines/src/formatter.rs
```diff
@@ -0,0 +1,84 @@
...
+    let result = visit_dir_md_files(dir, &|entry| {
+        println!("{:?}", entry.path());
+
+        let file = File::open(entry.path())?;
+        let file_len = file.metadata().unwrap().len().try_into().unwrap();
+        let mut contents = String::with_capacity(file_len);
+        let mut is_inside_rust_code_block = false;
+        let mut rust_block: Vec<String> = vec![];
+
+        for line in io::BufReader::new(file).lines() {
```
IceSentry, line 24: I think this loop could be extracted to a function that way this could have a doc test explaining what this whole app does with an actual example

write-rustdoc-hide-lines/write_rustdoc_hide_lines.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+cargo run --release -- ../content/learn/book
```
IceSentry, line 2: I wonder if we should just run this in CI. I assume right now the intended workflow is to let authors run it manually

write-rustdoc-hide-lines/README.md
```diff
@@ -0,0 +1,15 @@
+# Write `rustdoc` `hide_lines` annotations
+
+This utility iterates over the Markdown files on the given folder.
+Searches `rust` code blocks and ensures that the [`hide_lines` Zola annotation](https://www.getzola.org/documentation/content/syntax-highlighting/#annotations) match the code block "hidden" lines (the ones starting with `#`).
```
Nilirad, line 4: I would prefer a reference-style link instead, to not break the flow of the sentence.

doup: You mean something like this, no? 👉  https://github.com/bevyengine/bevy-website/pull/405/commits/b1ded3b9f46d6fcaff48763422ac815fa3d4ab71

write-rustdoc-hide-lines/src/formatter.rs
doup: Thanks for the suggestion, I'll look into it.

write-rustdoc-hide-lines/write_rustdoc_hide_lines.sh
doup: Yes, for now the idea is to run it manually… but probably makes sense to do it on CI. Anyway, I haven't looked into it as I don't know almost nothing about setting up GitHub CI (and didn't want to research). :-)

IceSentry: No worries, that can be done in a separate PR, I can take care of the CI stuff.

write-rustdoc-hide-lines/src/formatter.rs
```diff
@@ -0,0 +1,84 @@
...
+
+            if is_inside_rust_code_block && is_code_block_open {
+                panic!("Nested '```rust' code block not allowed");
+            } else if is_code_block_open {
+                is_inside_rust_code_block = true;
+            }
+
+            // Skip the line, save it as is
+            if !is_inside_rust_code_block {
+                contents.push_str(&format!("{}\n", &line));
```
IceSentry, line 37: According to clippy this should avoid allocating
```suggestion
				writeln!(&mut contents, "{}", &line)?;
```

write-rustdoc-hide-lines/src/formatter.rs
```diff
@@ -0,0 +1,84 @@
...
+                            definition.set_hidden_ranges(real_hidden_ranges);
+                        }
+                    }
+                }
+
+                // Rewrite code block Zola annotations
+                rust_block[0] = definition.into_string();
+
+                // Write code block
+                contents.push_str(&format!("{}\n", &rust_block.join("\n")));
```
IceSentry, line 66:
```suggestion
                writeln!(&mut contents, "{}", &rust_block.join("\n"))?;
```

write-rustdoc-hide-lines/src/formatter.rs
IceSentry, line 24: I tried doing what I suggested, but since this is parsing a file using doc comments it was pretty much impossible to write a doc test with a doc comment inside the doc test, but essentially I wanted this to be extracted so we could have a test that looked like this:
```rust
#[cfg(test)]
mod tests {
    use indoc::indoc;

    #[test]
    fn format_file_test() {
        let file = indoc! {r#"
            ```rust
            # test
            # test 2
            fn not_hidden() {

            }
            # test 3
            ```
        "#}
        .as_bytes();

        let contents = super::format_file(std::io::BufReader::new(file), file.len());

        assert_eq!(
            contents.unwrap(),
            indoc! {r#"
                ```rust,hide_lines=1-2 6
                # test
                # test 2
                fn not_hidden() {

                }
                # test 3
                ```
            "#}
        );
    }
}
```

IceSentry: If you just extract the loop to a function called `format_file` that takes a `std::io::BufRead` you can add that test directly

write-rustdoc-hide-lines/README.md
Nilirad: Yes. If you want, you can also get rid of the index and use the anchor text as reference.

doup: Done!

write-rustdoc-hide-lines/src/formatter.rs
doup: Oh, I just realized I was on an older Rust version. These weren't reported in `1.60`. I'll try to fix this, although I'm not sure what to do with the error type, as `writeln!` is not compatible with `std::io::Error`… I suppose I need to change `visit_dir_md_files` to `dyn Error` or something like that? (like when `main` returns `Result`)

IceSentry: Ah, right, I added `anyhow` on my branch for that

doup: I've used `anyhow` in the end… and, I've found a couple of articles/videos on error handling that I'll check later. ^_^

write-rustdoc-hide-lines/src/utils.rs
```diff
@@ -0,0 +1,25 @@
+use anyhow::Result;
+use std::{
+    ffi::OsStr,
+    fs::{self, DirEntry},
+    path::Path,
+};
+
+pub fn visit_dir_md_files(dir: &Path, cb: &dyn Fn(&DirEntry) -> Result<()>) -> Result<()> {
```
IceSentry, line 8: I don't think this really need it's own file, could just be declared directly in the formatter file.

write-rustdoc-hide-lines/src/formatter.rs
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+
+        // Rewrite file
+        fs::write(entry.path(), contents)?;
+
+        Ok(())
+    });
+
+    match result {
```
IceSentry, line 81: This function should probably just return the result and let main handle the logging.

IceSentry (approved):
I added a few more comments, but I'll give my approve right now since it works well and it's simple enough that the small nitpicks aren't a big deal.

write-rustdoc-hide-lines/src/formatter.rs
doup: Changed in 7bbd5cf.

write-rustdoc-hide-lines/src/utils.rs
doup: Changed in 7bbd5cf.

write-rustdoc-hide-lines/src/main.rs
```diff
@@ -0,0 +1,15 @@
...
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let folder = args.get(1);
+
+    if let Some(folder) = folder {
+        let path = Path::new(&folder);
+        formatter::run(path);
+    } else {
+        panic!("Please provide a folder.");
+    }
```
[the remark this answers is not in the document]
doup: Much better 👍 
Changed in 7bbd5cf.

write-rustdoc-hide-lines/src/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+    if dir.is_dir() {
```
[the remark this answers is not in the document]
doup: Changed in 7bbd5cf.

write-rustdoc-hide-lines/src/formatter.rs
doup: I've refactored it to a `format_file` function that accepts a lines iterator (see b0606b2).
And I've added tests for: add, modify & remove annotations (see 8b2993b).

But… I've a doubt now, I noticed that the formatter modifies all line endings. For example:

image

image

I think it's not bad behaviour but I'm not sure if it's good either. 😅 
It's doing more than it should, as in theory this should only change the Markdown `rust` code blocks.

cc @alice-i-cecile

doup: For now, I'm going to assume that that's the good behavior. I've updated the tests accordingly, see c94e00f.

But, if you think we should respect the line endings then we'll need to change the approach, maybe we can get the code block ranges using regex and be more surgical… ¯\_(ツ)_/¯

For reference, right only two files are affected by this:

- `/ecs/detecting-changes/_index.md`
- `/ecs/ecs-magic/_index.md`

A third file (`/ecs/entities-components/_index.md`) gets modified, but only the `rust` code blocks.

doup: Resolved, see the other comment for more.

write-rustdoc-hide-lines/src/code_block_definition.rs
```diff
@@ -0,0 +1,252 @@
...
+                    .collect(),
+            )
+        } else {
+            Annotation::Other(String::from(text))
+        }
+    }
+}
+
+impl Annotation {
+    fn into_string(&self) -> String {
```
doup, line 47: Btw @IceSentry, any suggestion for this? I don't understand why it's OK to use the `&self` signature for `CodeBlockDefinition::into_string` though… By clippy, both should be consuming `self` (or that's what I understand): https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention

image

doup: Fixed in 6244271.

doup:
In case you didn't see, most of the issues are fixed but I realized that it's modifying all lines endings, see: https://github.com/bevyengine/bevy-website/pull/405#discussion_r933167679

write-rustdoc-hide-lines/src/formatter.rs
alice-i-cecile: Yeah, I won't block on this; that's the right behavior anyways.

alice-i-cecile (approved):

write-rustdoc-hide-lines/src/formatter.rs
IceSentry: Yeah, we should already enforce consistent line endings anyway so I don't think it's a big issue

alice-i-cecile:
bors try

doup:
@alice-i-cecile the build error was because the examples have changed and the `generate_wasm_examples.sh` is not in sync. [This is fixed in `master`](https://github.com/bevyengine/bevy-website/commit/99eb4d2dcee109a734da7e173ff84fe2768f9c6e#diff-72650f297809b15f8948308241f94e225d0fc47cf931f457b653ada9d3ba49f5). I've applied the same fix (b2c11e8).

cart (approved):
Very nice improvement! I agree that it makes sense to integrate this into CI, but that can definitely be a new PR.

doup:
The failure looks unrelated to this PR. I'm not sure how to proceed though. :-/

alice-i-cecile:
bors try

alice-i-cecile:
Okay, looks like we need to update the book code to main!

doup:
@alice-i-cecile do I need to do something from my part?

alice-i-cecile:
Nope; there's no problem with this PR. However, someone (which could be you), needs to update the example code to use 0.9 syntax.

doup:
Synced with latest changes in `new-book`.

PR #411, opened 2022-07-30
[Merged by Bors] - fix possible typo (drives instead of derives)

raffimolero:


alice-i-cecile:
Good catch :)

bors r+

PR #412, opened 2022-07-30
[Merged by Bors] - Add missing Text::from_section and query::QuerySingleError to 0.7-0.8 migration guide

afonsolage:
While upgrading my game to `0.8` I noticed those two was missing on migration guide. Both migration guide text I copied from their PRs, so feel free to suggest any text changes.

### DIFF
--- a/content/learn/book/migration-guides/0.7-0.8/_index.md
+++ b/content/learn/book/migration-guides/0.7-0.8/_index.md
@@ -538,3 +538,14 @@ bevy = { version = "0.8", default-features = false, features = [
     "bevy_scene",
 ] }
 ```
+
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

IceSentry:
The from_section wasn't added because it's not a breaking change, but I'm fine with adding it since I believe it needs the visibility.

IceSentry (approved):

afonsolage:
> The text from_section wasn't added because it's not a breaking change, but I'm fine with adding it since I believe it needs the visibility.

I had to change this in order to my code compile, so it's a breaking change, since there is no `with_section` anymore

IceSentry:
Oh, you're right it was renamed. I thought it was just added next to it, not renamed.

PR #413, opened 2022-07-30
[Merged by Bors] - Typo in a type

CleanCut:
For the example to work, `a` needs to be the same type in both places. I arbitrarily chose `f32`.


alice-i-cecile (approved):
bors r+

PR #417, opened 2022-07-30
[Merged by Bors] - Add an outline on focused elements for accessibility

### LINKED ISSUE #395: Add visual indicator for links on `:focus` event
Making a website usable with keyboard navigation is one of the corner stones of accessibility.

To enable this goal, it's important to know where you currently are while navigating through the website. The Bevy website currently provides no such indications (e.g. on links and the cards), making keyboard navigation almost impossible.

The `:focus` CSS pseudo-class indicates that the element is focused (e.g. by the keyboard).
This is usually indicated by a colored outline around the object.

[Relevant MDN Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#focusable_element_must_have_focus_styling)

How to reproduce:

- Make sure the website is focused
- Use tab and shift+tab to navigate between the intractable elements. You might see the link preview of your browser changing, but can't visually distinguish the element that's currently focused.

(Tested on Firefox and Chromium)

TimJentzsch:
Fixes #395.

For users that depend on keyboard navigation, it is crucial to have a visual indication of which element is currently focused.

This PR adds a bright blue outline to the focused element:

- Header links:
    
    Focus outline on the header links

- Header buttons:

    Focus outline on the header buttons

- Cards:

    image

### DIFF
--- a/sass/_utils.scss
+++ b/sass/_utils.scss
@@ -3,7 +3,7 @@
 }
 
 // Visual outline on focused elements, for accessibility
-*:focus {
+*:focus-visible {
     border-radius: 5px;
     outline: $focus-outline;
 }
--- a/sass/_utils.scss
+++ b/sass/_utils.scss
@@ -2,8 +2,9 @@
     display: none;
 }
 
-a:focus {
-    outline: none;
+// Visual outline on focused elements, for accessibility
+*:focus {
+    outline: $focus-outline;
 }
 
 .section {
--- a/sass/_utils.scss
+++ b/sass/_utils.scss
@@ -4,6 +4,7 @@
 
 // Visual outline on focused elements, for accessibility
 *:focus {
+    border-radius: 5px;
     outline: $focus-outline;
 }
 
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -59,3 +59,6 @@ $card-list-gap: 15px;
         --header-height: 72px;
     }
 }
+
+// Accessibility
+$focus-outline: solid #b1d9ff 3px;
--- a/sass/components/_main-menu.scss
+++ b/sass/components/_main-menu.scss
@@ -42,6 +42,7 @@
 
         // Instead we have the outline on the text inside the link
         &:focus span {
+            border-radius: 5px;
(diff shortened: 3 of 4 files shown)


alice-i-cecile:
Can we round the outline on the header links a bit? Those corners are pretty rough.

TimJentzsch:
> Can we round the outline on the header links a bit? Those corners are pretty rough.

Every website that I've ever seen has the hard edges for the focus outlines, but it's probably just because the amount of people that will see them is relatively low.

It looks like this now:
Rounded edges on focus outlines

alice-i-cecile:
I like that!

alice-i-cecile (approved):

sass/_utils.scss
```diff
@@ -2,8 +2,10 @@
     display: none;
 }
 
-a:focus {
-    outline: none;
+// Visual outline on focused elements, for accessibility
+*:focus {
+    border-radius: 5px;
+    outline: $focus-outline;
```
plof27, line 8: This is all quite good, but using the `:focus` pseudo-class results in the focus styling being applied whenever the mouse is held down, in addition to when the keyboard focuses an element. This is likely unintentional, and looks a bit odd on most elements (although I kinda like it on the cards).

I'd recommend using `:focus-visible` instead, which is applied only when the keyboard focuses an element (but ignores focus events from other devices, mostly). Further reading [here](https://css-tricks.com/the-focus-visible-trick/) and [here](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible).

plof27 (changes requested):
One small styling nitpick, but otherwise this looks good!

sass/_utils.scss
TimJentzsch: Neat, I didn't know about this!

plof27 (approved):
Looks great!

IceSentry (approved):
For some reason I can't change the focus locally with my keyboard., but the code looks good.

Nilirad:
If it had a little bit of transparency wouldn't have been bad IMO.

PR #419, opened 2022-07-30
[Merged by Bors] - Suggest to add bevy using cargo add

### LINKED ISSUE #356: Suggest that users use `cargo add` command to add Bevy
[Merged PR](https://github.com/rust-lang/cargo/pull/10472) for the feature.

This spares users from having to reason about the `Cargo.toml` directly, and makes sure they're always using the latest release.

We should probably wait a bit to make sure that this lands.

TimJentzsch:
Closes #356.

With this PR, we will suggest to add bevy as dependency using `cargo add bevy`, which is the easiest option now.

I still left in instructions on how to manually add it in `Cargo.toml` (e.g. if someone is on an older Rust version that doesn't have `cargo add`), but we could also take that out.

:warning: In the original issue there were some concerns regarding typo squatting, so those should probably be discussed before merging this.

### DIFF
--- a/content/learn/book/getting-started/_index.md
+++ b/content/learn/book/getting-started/_index.md
@@ -41,13 +41,17 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
 
 Bevy is [available as a library on crates.io](https://crates.io/crates/bevy).
 
-Add the bevy crate to your project's Cargo.toml like this:
+The easiest way to add it to your project is to use `cargo add`:
+
+```cli
+$ cargo add bevy
+```
+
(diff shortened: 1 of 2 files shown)


mockersf (approved):
I don't think typo squatting should be a concern at all

alice-i-cecile (approved):
Agreed with Francois, this isn't any more of a concern.

IceSentry (approved):

IceSentry:
For the record, I spoke with the community members that had concerns with `cargo add` and they had no complaints related to bevy suggesting using it.

PR #422, opened 2022-07-31
Prevent distortion of images in blog cards

### LINKED ISSUE #420: Blog post image previews are distorted on mobile
In the News section. This is most comically noticeable in the 0.8 release notes.
Screenshot_20220730-183202

TimJentzsch:
Fixes #420.

On mobile devices, the images were previously distorted. This PR enforces that the images always keep their original aspect ratio.

TBH it looks a bit weird, but probably better than before:
Preview of the changes. The images now preserve their aspect ratio, but only take up a low portion of the card height.

### DIFF
--- a/sass/components/_card.scss
+++ b/sass/components/_card.scss
@@ -34,9 +34,17 @@
 .centered-card-image {
     display: block;
     align-self: center;
-    height: 85%;
-    max-width: 85%;
+    height: auto;
+    width: auto;
+    max-width: 100%;
+    max-height: 100%;
     border-radius: $border-radius;
+
+
+    @media #{$bp-tablet-landscape-up} {
+        max-height: 85%;
+        max-width: 85%;
+    }
 }
 
 .card-text {


nicopap (approved):

alice-i-cecile (approved):
I'm on board, even if it does look a bit weird. Waiting for more opinions before merging.

IceSentry (approved):
I think I would have preferred if it just cropped the sides, but the images with text would look even weirder and I can't think of a better alternative. Maybe just add a bit of padding to the sides. It will make the image even smaller but the lack of padding doesn't help in my opinion.

cart:
Yeah I personally think cropping is worth experimenting with. The weirdness of scaling down that far is probably greater than the weirdness of cutting things off. Generally these images have "centered" composition / wouldn't suffer much from losing things in the periphery. Worth experimenting though. We should also experiment with adjusting the card ratios to give the images more room. That might make the method in this pr look a little better.

doup:
`object-fit: cover` on `.centered-card-image` looks good for images, but ugly when there is text (e.g. Bevy logo).
**Btw, looks like these cards should be using `min-height` instead of `height`…**

image

TimJentzsch:
Yea, currently a lot is fixed height and `overflow: hidden`, so mobile users miss part of the context.

Alternative designs for the images I can think of:

- Hide the images on mobile (more boring, but more room for actual content)
- Put the images with ~80% width above the text (images can be bigger and content has full width, but you have to scroll a lot more)
- Make the images only next to the title and subtitle and the description can be below them (aspect ratio for images better, but we'd need different designs for mobile and desktop)

cart:
Maybe we could hide the "preview" text on mobile and use that extra room to preserve aspect ratio? That text isn't particularly important.

alexfriesen:
How about showing the text under the image?
You could set an break-point and switch to `flex-direction: column`. 
It could look like this: 
chrome-capture-2022-8-26 (2)

cart:
I think "text under image" is probably the way to go here. If anyone wants to pick that up, feel free!

doup:
I have a solution for this, although it'll take some time before I prepare a PR. There are other screens that use this `.card` class in somewhat different ways and I need to review those.

Anyway, issues that will be fixed:
- Fully responsive
- Image aspect ratio is respected in all the breakpoints
- 3 lines ellipsis via CSS instead of Zola, this way the ellipsed line is filled until the end
- When the title is multi-line it handles it gracefully (visible on phone-landscape)

🎥 
- First element is the old version
- Second one is the new

https://user-images.githubusercontent.com/188612/222958921-20cae62d-b614-4dc5-9b1f-48033f9eb013.mp4

Issue #361, opened 2022-04-26
Convert assets GIFs to a smaller format


doup:
Assets page GIFs take around 16mb (at the time of writing this) and unfortunately don't get optimized by Zola `resize_image` (see issue #339 & PR #360). Ideally these should be resized to fit in `340x340` and converted to a better image format (e.g. webp, gifv, mpeg…)

Ideally this should be automated.
- `ffmepg` could be used as it supports gif => webp.
- or, maybe there is some crate that supports this conversion and can be integrated with the assets generation code

alice-i-cecile:
Other suggestions that came up: 
- only animation on hover
- pagination

nicopap:
A quick search shows that there are 5 assets that weight more than 1M:
```
3.0M Apps/Games/Kataster.gif
1.2M Apps/Games/bevy_combat.gif
4.2M Assets/2D/bevy_ecs_ldtk.gif
2.7M Assets/3D/bevy-hikari.gif
1.5M Assets/Helpers/bevy_match3.gif
```

<details>
<summary>Click to expand list of all assets of more than 50K</summary>
<pre>
52K ./Apps/Games/libracity.png
52K ./Apps/Games/One-Click-Ninja.png
52K ./Apps/Games/punchy.png
52K ./Assets/3D/bevy_mod_picking.png
60K ./Apps/Games/typey-birb.png
60K ./Assets/2D/bevy_life.png
68K ./Apps/Games/bounce-up.png
68K ./Apps/Games/build_a_better_buddy.png
68K ./Assets/Helpers/bevy_plot.png
68K ./Assets/Helpers/bevy_proto.png
72K ./Apps/Games/not-snake.png
96K ./Apps/Tools/nodus.gif
104K ./Assets/Networking/ggrs_logo.png
124K ./Apps/Games/unflock.png
124K ./Assets/Shapes/bevy_smud.png
152K ./Assets/2D/bevy_tileset.gif
164K ./Learning/minesweeper-tutorial.png
172K ./Apps/Games/nbody_moon_creator.gif
172K ./Learning/Learn-Bevy-Ecs-By-Ripping-Off.png
176K ./Apps/Games/cheaters-never-win.png
216K ./Apps/Games/super-kaizen-overloaded.png
256K ./Apps/Games/bevy-tetris-yat.gif
276K ./Apps/Games/limbo_pass.png
328K ./Assets/Animation/bevy_easings.gif
384K ./Assets/Animation/bevy_tweening.gif
[...]

PR #424, opened 2022-07-31
[Merged by Bors] - Enable lazy loading for card images

TimJentzsch:
This is a first step to improve the performance on the asset page (#361).

The card images are now lazily loaded, meaning that images are only loaded once they are in the viewport of the user.


nicopap (approved):

nicopap:
For web rookies like me:
- https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes

alice-i-cecile (approved):

mockersf (approved):

PR #426, opened 2022-07-31
[Merged by Bors] - add tracy upgrade to migration guide

hymm:
Users using tracy need to upgrade their tracy version.


alice-i-cecile (commented):
bors r+

PR #428, opened 2022-08-01
[Merged by Bors] - add "before migrating" section to migration guide

IceSentry:
## Objective

It's not uncommon for people to forget to update their rust versions. 

## Solution

Add a short warning to remind people to run `rustup update` and a short description as to why

image

## Notes

For now this is only on the latest migration guide, but I already have plans to make migration guide a dedicated section on the website and use a template that could include this warning on every page.


alice-i-cecile:
I'm on board but you're missing a word in the title :)

IceSentry:
How about this title?

alice-i-cecile (approved):

rparrett (approved):
Looks good to me

PR #429, opened 2022-08-02
[Merged by Bors] - automatically resize examples to canvas parent

mockersf:
## Objective

Fixes https://github.com/bevyengine/bevy/issues/4963
Issue is not in Bevy but on CSS of the page to resize the canvas

## Solution

Use the new option from Bevy 0.8 to resize canvas to parent


alice-i-cecile (approved):

IceSentry:
I think the css should stay there. When loading the example is larger and gets resized once loaded 
image

IceSentry (approved):

PR #434, opened 2022-08-10
[Merged by Bors] - Typo in getting-started/resources

Pand9:
`delta` is correct because it's in the snippet below and I've checked that it works.


TimJentzsch (approved):
Notably, [`delta_seconds` also exists](https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html#method.delta_seconds), but I agree that it makes sense to change it to `delta`, because it's used in the code snippet.

Weibye (approved):

IceSentry (approved):

mockersf (approved):

PR #441, opened 2022-08-12
[Merged by Bors] - Remove unused ROADMAP.md

TimJentzsch:
This file has not been changed since 2 years, so I don't think we need it.
If the entry is still relevant, it should be moved to an issue instead.


Nilirad (approved):
I agree, it's also outdated data.

mockersf (approved):

PR #443, opened 2022-08-13
Add validation for image existence, image filesize, dimensions, and format

rparrett:
Throwing this out here to start a discussion.

## Should we do something like this?

While we're storing images in a git repo and handling submissions via pull requests, I think it makes sense to enforce at least file size limits to reduce repo bloat.

Validating that the image exists and is not corrupt also seems like a no-brainer to me.

## What should the limits be? Should we go further and validate image aspect ratio, or enforce a specific size?

I'm not sure that we need to validate dimensions at all. The current optimal image size is an odd portrait aspect ratio, but a promising redesign to the asset cards would make that slightly larger and landscape. So enforcing aspect ratio at this moment seems a bit silly.

I tossed in 1000x1000 as a bounding box as more of a sanity-check. Images might be well under the file size limit but still wildly non-optimally sized.

### DIFF
--- a/generate-assets/Cargo.lock
+++ b/generate-assets/Cargo.lock
@@ -3,6 +3,18 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "adler32"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
+
+[[package]]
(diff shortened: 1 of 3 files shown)


IceSentry:
I don't think we need to enforce specific dimensions. At least not for now. A size limit is good enough in my opinion. If people want to better match the dimension of the asset card they can either submit a transparent background or make sure they have a landscape image. The images aren't currently fixed size in the new asset cards anyway, so it wouldn't make sense to enforce this. 

As for the implementation, I think the `validate()` function should return a `Result` instead of a `bool`.

rparrett:
I agree that the error handling in the validator could use a bit of polish.

 It seems that this thing is designed to show all validation errors for all assets and then ultimately fail if there's at least one validation error. Maybe `validate` should return a `Vec<Result>` and they should all get collected at the end?

Either way, I'd prefer that happen separately before or after this.

IceSentry:
Yeah, doing it in another PR makes sense, it's just that people don't look at this repo as often so I tried pushing for it a bit lol

rparrett:
Will re-implement this on top of #446 later.

PR #446, opened 2022-08-18
[Merged by Bors] - Improved error handling in asset validator

rparrett:
Some minor improvements to this tool both for contributors who may be working with it and for end-users making PRs to `bevy-assets`

- Show a friendlier error when failing to specify the `bevy-assets` path as a cli argument
- Show an error if the user specifies the `bevy-assets` path incorrectly
- Show errors that occur during `parse_assets`
- Check that images exist
- Group errors by asset

I felt a bit awkward working around the `AssetValidator` trait, and I'm not super strong with rust error handling / `anyhow` so any feedback would be appreciated.

Now outputs something like

```
(a bunch of metadata spam)

UnFlock
  DescriptionWithFormatting

taipo
  DescriptionTooLong
  ImageInvalidLink

Error: 2 asset(s) are invalid.
```

The change to `lib.rs` is much smaller than it looks. It consists of:
- removing a level of indentation by exiting early
- allowing `fs::read_dir` to potentially fail rather than hiding it behind `is_dir`
- removing the unwrap from `toml::from_str` in favor of `?`

### DIFF
--- a/generate-assets/src/bin/validate.rs
+++ b/generate-assets/src/bin/validate.rs
@@ -62,8 +62,7 @@ impl AssetValidator for Section {
     fn validate(&self) -> Vec<Result<(), AssetError>> {
         self.content
             .iter()
-            .map(|content| content.validate())
-            .flatten()
+            .flat_map(|content| content.validate())
             .collect()
     }
 }
--- a/generate-assets/src/bin/validate.rs
+++ b/generate-assets/src/bin/validate.rs
@@ -1,45 +1,54 @@
+use anyhow::{anyhow, Context, Result};
 use regex::Regex;
 
 use generate_assets::*;
 
-fn main() -> Result<(), ()> {
-    let asset_dir = std::env::args().nth(1).unwrap();
-    if let Ok(asset_root_section) = parse_assets(&asset_dir, None, None, None) {
-        if asset_root_section.validate() {
-            Ok(())
-        } else {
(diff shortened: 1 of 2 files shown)


IceSentry:
I like the direction where this is going. I'll review it tomorrow, but at a glance it looked good.

mockersf (approved):

generate-assets/src/bin/validate.rs
```diff
@@ -1,45 +1,75 @@
...
+    if errors.is_empty() {
+        return Ok(());
     }
+
+    eprintln!();
+    for error in &errors {
+        eprintln!("{}", error);
+    }
+
+    Err(anyhow!("{} asset(s) are invalid.", errors.len()))
```
IceSentry, line 31: you can use the bail! macro from anyhow for this
```suggestion
    bail!("{} asset(s) are invalid.", errors.len())
```

IceSentry (approved):
LGTM

generate-assets/src/bin/validate.rs
IceSentry: It's an anyhow macro, so clippy wouldn't know about it. I'm honestly not sure how idiomatic that is. I just saw the Err(anyhow and thought of the macro. I don't mind either way honestly it was just going through my head,

rparrett:
rebased

PR #454, opened 2022-08-30
[Merged by Bors] - Update MacOS requirements

0xD3n:
To run rust on MacOS you require Xcode or it will give the Following error: 
xcrun: error: invalid active developer path

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -22,7 +22,7 @@ Once this is done, you should have the ```rustc``` compiler and the ```cargo``` 
 ### Install OS dependencies
 * [Linux](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md)
 * Windows: Make sure to install [VS2019 build tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
-* MacOS: No dependencies here
+* MacOS: xcode-select --install or [Xcode](https://apps.apple.com/de/app/xcode/id497799835?mt=12)
 
 ### Code Editor / IDE
 
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -22,7 +22,7 @@ Once this is done, you should have the ```rustc``` compiler and the ```cargo``` 
 ### Install OS dependencies
 * [Linux](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md)
 * Windows: Make sure to install [VS2019 build tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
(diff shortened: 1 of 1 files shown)


content/learn/book/getting-started/setup/_index.md
```diff
@@ -22,7 +22,7 @@ Once this is done, you should have the ```rustc``` compiler and the ```cargo```
 ### Install OS dependencies
 * [Linux](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md)
 * Windows: Make sure to install [VS2019 build tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
-* MacOS: No dependencies here
+* MacOS: xcode-select --install or [Xcode](https://apps.apple.com/de/app/xcode/id497799835?mt=12)
```
rparrett, line 25:
```suggestion
* MacOS: Install the Xcode command line tools with `xcode-select --install` or the [Xcode app](https://apps.apple.com/en/app/xcode/id497799835)
```

The command would look nicer in an inline code block and we should probably link to the English app store.

rparrett:
I think this is a dependency of rust rather than bevy, but this could be pretty confusing for new users so adding it makes sense to me.

DJMcNab:
> I think this is a dependency of rust rather than bevy, but this could be pretty confusing for new users so adding it makes sense to me.

I was about to send the same message with the opposite conclusion. However, the linked getting started page doesn't list any macos steps on my machine, so I can't verify if it correctly gives these instructions. If it does, my stance would be that we can't solve people ignoring given steps by adding more steps.

content/learn/book/getting-started/setup/_index.md
0xD3n: Okay this makes this look better thank you.

0xD3n:
I was just using bevy and I just formatted my Mac. when I started using it again I got this error and after research you need Xcode this could be a new thing.

cart:
I think short term it makes sense to include this. But if the rust guide is ever updated, we should remove this.

DJMcNab:
I think we need to work out what needs xcode. Based on the source of the getting started guide (`view-source:https://www.rust-lang.org/learn/get-started`), there aren't any special instructions for macOS.

rparrett:
Xcode itself isn't really required, but rust needs a linker and such, I assume, and installing the "xcode command line tools" is the equivalent of installing a "build-essentials" type package.

The reason to mention Xcode at all as an app is that some people might already have that without even knowing what the xcode command line tools are.

rparrett:
I don't have a machine that I can nuke to test this stuff with, unfortunately. `rustup-init.sh` doesn't mess with this stuff as far as I can tell, but maybe rustup itself does?

DJMcNab:
Yes, but I trust Rust to have a sane installation strategy, so if their instructions don't tell you to install xcode, then you don't need to install xcode.

The logical conclusion, trusting that @0xD3n didn't explicitely say no to rustup installing xcode automatically, is that one of our dependencies has a dependency on something xcode provides.

rparrett:
Pinging @mockersf for mac expertise.

PR #463, opened 2022-09-26
add font-swap

alexfriesen:
This PR might improve render performance of the website a little (even though it's already super fast). 
`font-display: swap;` will display the fallback font until the requested font is loaded and swap them. So the browser wont block rendering the text.

### DIFF
--- a/sass/fonts/_firamono.scss
+++ b/sass/fonts/_firamono.scss
@@ -3,6 +3,7 @@
      font-family: 'Fira Mono';
      font-style: normal;
      font-weight: 400;
+     font-display: swap;
      src: url('../fonts/fira-mono-v8-latin-regular.eot'); /* IE9 Compat Modes */
      src: local('Fira Mono Regular'), local('FiraMono-Regular'),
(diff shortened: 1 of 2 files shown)


DJMcNab:
Is it worth just using native fonts, which might be more best practise.

See e.g. https://markdotto.com/2018/02/07/github-system-fonts/

alexfriesen:
@DJMcNab I think the current Font looks really good and adds a lot of character to the website. It would be a loss to just go back to the okay looking native fonts. Your suggestion might be worth a discussion when redesigning the page. 

In this PR I'm trying to make a tiny optimisation 😅

cart:
Hmm showing a visual "re-render" after the actual font loads might not be worth the time we save here, especially given that the site already loads pretty fast. I haven't used this setting before, but it seems like loading in the real font will result in some "relayouting", which I personally normally associate with "jank". Anyone else have thoughts on this / experience with this setting?

cart:
And yeah @DJMcNab I don't think using native fonts is worth giving up additional "brand identity". I don't want the site to look different on different OS-es / configs and I like the additional creative expression.

cart:
Closing for now as I'm pretty sure the current approach is preferable

Issue #448, opened 2022-08-18
Asset card image containers should have a constant aspect ratio


rparrett:
Currently, the cards change in width with the screen size.

This means that there is no image aspect ratio that will "look good" on all screen sizes. Even if all images were authored with a consistent aspect ratio, they would be letterboxed on some screens. The aspect ratio also ends up pretty extreme on large mobile displays in portrait mode.

It would be great to rework the cards so that the aspect ratio of the image container is consistent across screen sizes.

See also: [discord conversation](https://discord.com/channels/691052431525675048/695741366520512563/1009903697494556755)

oCaioOliveira:
Hello @rparrett, I've been new here and I'm trying to contribute. This issue still open? Running the clone of your repository looks the problem are solved, can you confirm this?

rparrett:
Cool!

Here's a screen recording that hopefully clears up what I attempted to describe in the issue text.

https://user-images.githubusercontent.com/200550/203440984-f2741046-3118-4311-b6b1-72dbe8f3b9b1.mp4

Right now, if someone submits their project to Bevy Assets, they always have to compromise when they choose the aspect ratio for their cover image because it is likely to be different on various screen sizes.

rparrett:
Maybe the least clear part of the issue text is saying that the "image" should have a constant aspect ratio. They're not getting squished or anything, but the container's aspect ratio is changing so the images are displayed with inconsistent letterboxing.

<img width="315" alt="image" src="https://user-images.githubusercontent.com/200550/203441918-49b987b7-cb2c-45fe-a84b-3d361a1629ac.png">

<img width="374" alt="image" src="https://user-images.githubusercontent.com/200550/203442000-a25fcf60-810f-4268-9e7e-d4b1831c524a.png">

oCaioOliveira:
Ok, thanks for the extra description @rparrett, I'll work on it.

oCaioOliveira:
I worked on it with @ErickMVdO, and this is the result @rparrett. Is it what you needed?
The PR it's open.
[Video demonstration](https://youtu.be/8TEz34Y7ylY).

PR #464, opened 2022-09-30
[Merged by Bors] - Add additional image validation

rparrett:
Updating this old code on top of the recent improvements to the validator.

Some previous discussion in #443
Some recent motivation in [#253](https://github.com/bevyengine/bevy-assets/pull/253)

## Objective

- Shorten the feedback loop for users submitting their projects to `bevy_assets`
- Help maintainers enforce restrictions related to image size
- ~~Improve the overall look and feel of the assets page by enforcing a reasonable aspect ratio~~
- Improve the performance of the assets page and reduce the repository size by enforcing a maximum file size
- Make validation errors output a little more clear for users

## Example Output

Note: Some of these validations are no longer present in this PR.

```
Fish Fight: Punchy
  Image dimensions must not exceed 1000x1000 px.
  Image aspect ratio must be between 1 and 2.

taipo
  Description must be at most 100 chars in length.
  Image file not found.

Bevy Combat
  Image file must be at most 1000000 bytes.

Typey Birb
  Image aspect ratio must be between 1 and 2.

Cheaters Never Win
  Image dimensions must not exceed 1000x1000 px.
  Image aspect ratio must be between 1 and 2.

Error: 28 asset(s) are invalid.
```

## Discussion Points

This is currently a draft because many existing images do not currently meet these restrictions.

For file size restrictions, we could manually fix up the files.

But for image aspect ratio, I think we would need to find a way to "grandfather in" the old images. I contemplated a static list of grandfathered assets, but that would prevent validations from working if users attempt to update those in the future.

Also, see https://github.com/bevyengine/bevy-website/issues/448 where we decided to use a single aspect ratio on the website. If that were fixed, we might want to rework the some of the validations like

- Enforce a fixed image size rather than a maximum
- Remove the separate aspect ratio check all-together

The current file size limit in the PR seems clearly too large. I think something like 250kb would be more appropriate, but that would require a lot more work on existing assets.

### DIFF
--- a/generate-assets/Cargo.lock
+++ b/generate-assets/Cargo.lock
@@ -47,12 +47,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
-name = "bit_field"
-version = "0.10.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dcb6dd1c2376d2e096796e234a70e17e94cc2d5d54ff8ce42b28cef1d0d359a4"
-
-[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -84,12 +78,6 @@ name = "bumpalo"
 version = "3.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d261e256854913907f67ed06efbc3338dfe6179796deefc1ff763fc1aee5535"
-
-[[package]]
-name = "bytemuck"
-version = "1.12.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aaa3a8d9a1ca92e282c96a32d6511b695d7d994d1d102ba85d279f9b2756947f"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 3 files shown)


IceSentry:
I believe this should be updated to only check for file size restrictions and ignore aspect ratio. With the current card layout, using the wrong aspect ratio might not look great, but it isn't an issue like size is.

rparrett:
Okay, that is done.

```
bevy_ecs_ldtk
  Image file size 4201746 exceeds maximum 2097152 bytes.

Kataster
  Image file size 3046854 exceeds maximum 2097152 bytes.

Error: 2 asset(s) are invalid.
```

IceSentry:
> Unfortunately, image is now choking on a webp 

Oh, 😦 that is unfortunate, well, I guess we'll be able to add it back in the future

For the CI failure, https://github.com/bevyengine/bevy-website/pull/520 this should fix it, it's not caused by this PR

rparrett:
> Oh, 😦 that is unfortunate, well, I guess we'll be able to add it back in the future

Yeah, happy to do that later. I am pretty sure that the next release of `image` will fix the issue.

rparrett:
We will need to do something about the images for `kataster` and `bevy_ecs_ldtk` or I assume the next person to submit something to `bevy-assets` will get stuck with that problem.

IceSentry (approved):

alice-i-cecile (approved):

alice-i-cecile:
Fingers crossed this should work now.

bors r+

rparrett:
> error: usage of sparse registries requires `-Z sparse-registry`

well that's new and exciting

mockersf (approved):

mockersf:
> > error: usage of sparse registries requires `-Z sparse-registry`
> 
> well that's new and exciting

I'm going to go with cargo bug on the server side 🤷

PR #469, opened 2022-10-25
[Merged by Bors] - migration guide and release note generator

IceSentry:
## Objective

Generate migration guide and release note automatically.

## Solution

The general steps are pretty simple. 

- Get all the commits on main 
- Get all PRs merged by bors and make sure they were closed after the target date
	- For the migration guide, make sure the PR has the `C-Breaking-Change` label
- For each commit, find the corresponding PR by using the title of the PR
	- We do this because github doesn't return a commit hash with the PR and some of the metadata we want needs a commit hash
- For each commit, get the list of contributors from the graphql api. For whatever reason, github doesn't give this information through the rest api
	- This is pretty slow because it requires a network call for each commit unfortunately. There's probably some ways to speed this up, but it does the job and takes about 3 minutes on my machine. To avoid being rate limited, I wait for a few seconds and then retry when it fails and it seems to work well.

## Notes

This doesn't necessarily need to live in the repo, but since I have the code and we'll probably reuse it in the future, might as well share it.

The markdown generation is a bit intense because the parser I used wouldn't let me simply write it back, so I needed to implement it myself.


rparrett:
I think the parsing could likely be simplified with [into_offset_iter](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/struct.Parser.html#method.into_offset_iter).

IceSentry:
I'm not sure how I could use it to simplify the parsing.

IceSentry:
Yeah, I just tried it, the code is a bit simpler, but the output generates a lot more errors when going through markdownlint. Since I already have the code I'll just leave it as is.

cart:
As a heads up, when generating the list of prs for 0.9, this PR shows up: https://github.com/bevyengine/bevy/pull/3223.
Notably, I deleted the ignore-reddit branch 7 days ago, which probably affected the metadata being used. This seems to imply that we might need to use a different date/time data source.

cart:
(there are quite a number of these actually)

IceSentry:
Oh yeah, I never filtered on the branch, didn't think about that. I used the issues api because PRs are issues and it worked, but there's actually a pulls api that let's me filter on the branch https://docs.github.com/en/rest/pulls/pulls I'll try to update it later tonight to use it instead.

cart:
Nice nice! Short term I've come up with a solution:
* Add `GithubIssuesResponse::closed_at`
* Extract the date from it
* Filter out anything outside of that date

Very easy tweaks to the existing infra! So I'm unblocked for this release. But using the branch is a more exact solution, so we should definitely use that if we can.

generate-migration-guide/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+}
+
+/// Generates the list of contributors and a list of all closed PRs sorted by area labels
+fn generate_release_note(date: &str, path: PathBuf) -> anyhow::Result<()> {
+    let prs = get_merged_prs(date, None)?;
+    let mut authors = HashSet::new();
+    let mut pr_map = HashMap::new();
+    let mut areas = HashMap::<String, Vec<i32>>::new();
+    for pr in &prs {
+        authors.insert(pr.user.login.clone());
```
cart, line 97: Another issue is that this excludes "pr collaborators" from the contributors list. We should find some way to include this list of people:
image

rparrett:
Another note: Need to manually filter out PRs that were cherry-picked for 0.8.1
```
- [bevy_pbr: Fix tangent and normal normalization][5666]
- [Sync up bevy_sprite and bevy_ui shader View struct][5531]
- [fix order of exit/close window systems][5558]
- [Fix View by adding missing fields present in ViewUniform][5512]
```

cart:
> (Or not, since this is technically "new stuff since the last blog post")

Yeah I'm cool with including these (at least for the contributors list) to properly credit contributors.

generate-migration-guide/src/main.rs
IceSentry: Yeah, I haven't figured out yet how to get that info. It's not on the issue or PR response from the rest api. I think most people that collaborated on PRs probably also have PRs where they were authors, but there's probably a couple that we missed anyway.

IceSentry: Wait, now that I think about it, I can probably just get all the commits on main, that way I don't need to filter anything other than the date since every commit contains the PR description and it's authors.

cart: Oooh yeah thats a good call.

cart: Hmm so you can list co-authors like this: `git log v0.8.0..origin/main --pretty=format:"%(trailers:key=Co-authored-by)"`, but that will only give you the emails and first / last names (generally), not github accounts (which is what we use).
So far I haven't been able to find any api that provides the co-author github user accounts.

cart: I've spent a good portion of time trying to build the `email->username` correlation, but I haven't found anything yet. The "repo contributors" api does have an email field in the schema, but it doesn't actually show up in the response.

There is a "release notes generator" which might have worked :https://docs.github.com/en/enterprise-server@3.5/rest/releases/releases#generate-release-notes-content-for-a-release

However it only includes PRs in the "merged" state, which doesn't work for our needs.

So far the _only_ way I've found to accurately generate this list is by scraping each page on the GitHub "compare branches" view:
https://github.com/bevyengine/bevy/compare/v0.8.0...main
Which I have done for past releases by manually copy/pasting each page's content and then extracting relevant info using regexes.

cart: I'm going to do that for this release as I can't afford to spend more time on this.

IceSentry: you can use `https://api.github.com/search/users?q=<email>` to find the user

alice-i-cecile (changes requested):
IMO we should scrape commits on main, like you said, and then extract emails from there. Once that's done this LGTM.

generate-migration-guide/src/main.rs
IceSentry: I now use the graphql api and this gives us a full list of contributors for each commit

cart:
I'm still seeing commits from previous releases in the outputs (and like the previous issue mentioned above, it seems like its because its using "last modified" instead of "closed at").

cart:
Likewise, there are a couple of stragglers on the "boundary" of 2022-11-12 because the date on its own doesn't match the exact time of the 0.9 release. If possible, if we can take a "commit range" as input (or even branch names like `v0.9.0..origin/main`) that would allow us to be perfect and only include the exact commits we need.

cart:
[correct_commits.txt](https://github.com/bevyengine/bevy-website/files/10738329/correct_commits.txt)
[generator_commits.txt](https://github.com/bevyengine/bevy-website/files/10738331/generator_commits.txt)

IceSentry:
> it seems like its because its using "last modified" instead of "closed at"

🤔 That's a bit strange, because it is supposed to be using the closed_at field. So it might be that github named it differently in the api.

> there are a couple of stragglers on the "boundary" of 2022-11-12

I completely forgot about that. I originally used a date because it was easier to figure out and less manipulation for me, but it's pretty much just used to retrieve the oldest commit on that date and everything derives from it. I'll just take in the actual commit sha instead. Maybe I could use a git tag, assuming we even have those for each version, but a commit sha should do the trick.

cart:
This api appears to do the trick:

```rust
    pub fn compare_commits(&self, from: &str, to: &str) -> anyhow::Result<()> {
        let request = self.get(&format!(
            "https://api.github.com/repos/bevyengine/bevy/compare/{from}...{to}"
        ));
        let response = request.call()?;
        let str = response.into_string().unwrap();
        println!("{str:#?}");
        Ok(())
    }
```

cart:
Where `from` is "v0.9.0" and `to` is "main"

cart:
Just returns a giant blob of json with every commit + some metadata. Only contains the first committer, so we'll still need to make calls to the other apis to get those.

Haven't plugged this in to anything yet / I've gotta hop out for a bit. So feel free to do that if you're interested.

IceSentry:
I need to go out for a bit too, but I'll do it later tonight.

cart:
Pushed my work so far. The number of "matching" PRs isn't right (~250 instead of the ~430 after the base commit date). Maybe the "title matching" algorithm needs changing? I've got to run again :)

cart:
Well that would explain it ... the api is only returning 250 commits. Maybe we need paging or something?

IceSentry:
Ah right, yeah, I had to support paging for the other api I used. Should be easy enough to reuse it. If you haven't done it yet I can do it in an hour.

IceSentry:
@cart I implemented the paging. Right now it finds 466 commits using from v0.9.0 to main. I haven't checked if this is all correct, but it's more than 250 lol

I also removed some dead code so the diff is a bit noisy

cart (approved):
Great work on this. Thanks for making our lives much easier!

PR #472, opened 2022-10-29
Updated instructions to get Mold linker working in MacOS

TGMcLean:
# Objective

- It'd be nice to have the Mold linker working in MacOS. These instructions update the installation instructions to make it so.

# Solution

- Updated instructions show how to use Brew / MacPorts to install Mold binaries; then how to update the `rustflags` annotations to ensure the linker binaries are found.


alice-i-cecile (approved):

alice-i-cecile:
Blocked momentarily until the linked PR is merged.

alice-i-cecile:
In light of the [proposed license changes]( https://www.reddit.com/r/rust/comments/yvqjmd/mold_linker_licensing_recent_strategy_change)  to  mold, and the uncertainty involved for what those might actually mean, I think this is irresponsible to recommend. Many of our users are attempting to make closed source games on small budgets, and recommending a project that pushes AGPL or "call for pricing" on them is unwise :(

TGMcLean:
Thanks, @alice-i-cecile.

PR #478, opened 2022-11-10
[Merged by Bors] - News: Release 0.9.0

cart:
This is an incomplete draft of the 0.9.0 release blog post. Feel free to leave comments / suggest changes / help me fill in gaps I missed!

NOTE: This is an _early_ draft and there are still a lot of missing pieces / placeholders. Please only review content that exists and point out missing pieces that don't already have placeholders. Anything with a `[4724]`-style markdown link is a placeholder for content that _might_  need to be included (but also might not be ... these placeholders still need some curation).

PR #479, opened 2022-11-10
[Merged by Bors] - 0.9 book fixes

alice-i-cecile:
Fixes breaking changes for 0.9.

Should be merged at the same time as #478.


hymm:
should the `query.iter()`'s be changed to `&query`?

alice-i-cecile:
> should the query.iter()'s be changed to &query?

IMO no: this is significantly harder for beginners to follow. Methods are easy, trait impls for &T are rough.

PR #485, opened 2022-11-14
[Merged by Bors] - updates getting started to mention 0.9.0

galkowskit:
It seems the rest was either already updated or the same as in 0.8.1. I thought it would be prudent to update the docs to mention the newest version.


alice-i-cecile (approved):
Oops! I missed this.

bors r+

PR #492, opened 2022-11-22
[Merged by Bors] - Add sponsor tier information

cart:
It makes sense to define what each tier means.


IceSentry (approved):
Are sponsors aware that this is visible/was it already public? I'm just curious if some sponsors prefer not having a money amount next to their name.

cart:
Yup these numbers are already public on my Github Sponsors profile.

Issue #423, opened 2022-07-31
Add instructions on how to set up Assets and Examples page locally


TimJentzsch:
When testing the website locally, I get a 404 error when trying to access the Assets and Examples pages.

I assume I have to run the `generate-assets/generate_assets.sh` script to do that.

This information should be accessible via the README file to make it easier for new contributors to set everything up.

oCaioOliveira:
Hello @TimJentzsch, @ErickMVdO and me oppened a [PR](https://github.com/bevyengine/bevy-website/pull/496) for this issue with a better description how to acces Assets.

PR #494, opened 2022-11-26
[Merged by Bors] - Card image containers with constant aspect ratio

ErickMVdO:
Changes in files: 
- _item-grid.scss: to justify content on different screens.
 `display: flex;`
 `flex-direction: row;`
 `flex-wrap: wrap;`
 `justify-content: space-evenly;`
- _asset-card.scss: to maintain a standard size on the cards.
`min-width: 280px;` 
 `max-width: min-content;`

Both in the path: sass/components/

Fixes https://github.com/bevyengine/bevy-website/issues/448

Co-authored-by: Caio <caiocesar31951910@hotmail.com>


rparrett:
Thanks for taking this on.

CSS isn't really a strength of mine, so forgive the slightly uninformed commentary. There may be some trade-offs here that I don't understand.

There seem to be a couple awkward points about this change that I am seeing.

1. When the last row of a section has fewer than 3 items, they are displayed with different spacing and are centered. I think that having these left-aligned would look nicer.

image

2. On mobile / displays where only a single column is shown, the cards no longer take up the entire width of the display, and are no longer aligned with the section headers.

image

ErickMVdO:
we believe there are only 3 ways to do this

1. increasing as much the height as width of the image
Screenshot from 2022-11-28 12-29-18

2. increasing only the width and stretching the image
Screenshot from 2022-11-28 12-29-49

3. increasing only the width without stretching the image (how it is currently done on the site)

in this last commit we chose to do it the first way

oCaioOliveira:
Hey @rparrett, @ErickMVdO and me made new changes, and when we start the contribution we had some difficulty for access the assets tab, so we contribute with addition on README.md, answering this [issue](https://github.com/bevyengine/bevy-website/issues/423) that we found. We opened the [PR](https://github.com/bevyengine/bevy-website/pull/496) about this, if you have access to review this I appreciate that.

rparrett:
The square aspect ratio doesn't seem ideal. It seems like this sort of thing should be possible. Here's an example from another engine's asset store: https://www.unrealengine.com/marketplace/en-US/store (see the showcase section)

@doup @IceSentry would you mind taking a look? I'm a bit out of my depth with modern css.

ErickMVdO:
but in this example from another engine's asset store, the images are cropped depending on screen size. Is this supposed to be normal behavior?
Screenshot from 2022-11-28 15-14-33
Screenshot from 2022-11-28 15-14-50

rparrett:
Sorry, that seems like the wrong example. I'm pretty sure that any cropping was discussed and vetoed.

rparrett:
Maybe this is a better example: https://www.construct.net/en/game-assets

IceSentry:
Can you explain the reasoning of those changes? I'm not sure I get _why_ this is necessary.

IceSentry:
I also don't understand what's the point of going back to using a flex layout. Css grid is designed exactly for grid-like layouts like this one.

rparrett:
This PR should link to #448 which explains the motivation.

IceSentry:
Right, in that case, I do agree that a fixed aspect ratio that isn't square is indeed ideal.

ErickMVdO:
actually I think I didn't understand very well, because I thought I should remove the margins of the images to keep them at a constant size. But now I think I get it, the images just need to be a constant size across all screens, right?

@IceSentry about the grid, I saw that it was wrong to keep the flex layout and I went back to the grid anyway

ErickMVdO:
in this last commit we tried to do based on the example of this [site](https://www.construct.net/en/game-assets), was that right?

sass/components/_asset-card.scss
```diff
@@ -2,12 +2,13 @@ $asset-card-padding: 0.4rem;
 
 .asset-card {
     display: grid;
+    max-width: 94vw;
+    aspect-ratio: auto 10/9;
```
rparrett, line 6: Is `auto 10/9` a valid value here?
https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio
It seems like it should be 10/9?
In chrome at least, it seems to be behaving just as `auto` which doesn't seem like what we want?

sass/components/_asset-card.scss
```diff
@@ -106,4 +107,25 @@ $asset-card-padding: 0.4rem;
...
+}
+
+@media (min-width: 285px) {
+    .asset-card{
+        max-width: 100vw !important;
+    }
+}
+@media (min-width: 315px) {
+    .asset-card{
+        aspect-ratio: auto 10/6 !important;
```
rparrett, line 119: Can we extract this into a variable and use the same value for every breakpoint?

rparrett (changes requested):

sass/components/_asset-card.scss
```diff
@@ -2,12 +2,13 @@ $asset-card-padding: 0.4rem;
 
 .asset-card {
     display: grid;
+    max-width: 94vw;
```
rparrett, line 5: Why is this needed? 94vw seems to be "approximately" the width of the viewport minus the container padding, but that seems a bit janky when the container padding is defined in pixels.

In my testing, things seem to work fine without any of the new `max-width: _vw;` lines.

On very very small screen sizes, the content can overflow, but at that point the screen size doesn't seem realistic for any common device.

sass/components/_asset-card.scss
rparrett: Ah, I see, this is the aspect ratio of the entire card rather than just the banner area.

rparrett:
I think the ideal behavior would be for `__description` and `__tags` to have a constant `rem`-based height, but for `__banner` to be `auto-fill`?

sass/components/_asset-card.scss
ErickMVdO: We had put 94vw just because of the galaxy fold, which has a width of 280px, but we will make the necessary changes

ErickMVdO:
we put an aspect ratio for each screen size, which one do you think would be best suited for all screens?

rparrett:
16x9 seems like a good starting point. That will be the aspect ratio of screenshots of bevy with its default window size.

ErickMVdO:
done 👍

rparrett:
Great! I think this is now generally behaving as described in #448, and the changes seem quite straightforward.

I am seeing one strange thing in chrome, firefox, and safari. In the following screenshot I made the banner background red for visibility:

image

There is a difference in how cards with an without images are rendered. The overall height of the card is smaller and the banner's background leaks through.

oCaioOliveira:
We'll check this @rparrett.

doup (approved):

doup:
A guess: the red background leakage is because `` is an inline element, try to set `` to `display: block`.

ErickMVdO:
@doup thas was it, thank you

rparrett (approved):
Thanks Erick and Caio.

This is definitely a bit of a compromise on mobile, but I think it's probably the right way to go. 

In the future, it may be possible to do two columns on most mobile displays by reworking the tags section.

PR #495, opened 2022-11-27
Encourage updating Rust

interfect:
This should help stop people from making mistakes like https://github.com/bevyengine/bevy/issues/6777.

Novice Rust programmers might see "latest stable release" and think that that means "what I installed last month when I started learning Rust". However, that's not likely to be the case.

Rust makes stable releases often enough that anyone who doesn't update it every couple weeks or so actually needs to update Rust in order to use Bevy.

This should make that clearer.

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -15,8 +15,10 @@ All Bevy app and engine code is written in Rust. This means that before we begin
 
 ### Installing Rust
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
Hmm. I think that this is unlikely to help users in practice: this will be read once, and then forgotten. I do think the MSRV approach discussed in that issue is much more direct.

hymm:
should this be closed now that https://github.com/bevyengine/bevy/pull/6852 has been merged?

PR #497, opened 2022-11-29
Ease novices into the context of Components

thadguidry:
It makes sense to give introduce Components from an ontological perspective as well as programming perspective.  Those coming from or have some small experience with other languages other than Rust might equate a Component to a supercharged "property".  Our examples such as "Name" component attached to Person, would be a property in other languages.  So, let's add additional context to help introduce the concept of Components.

### DIFF
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -9,7 +9,7 @@ insert_anchor_links = "right"
 
 All app logic in Bevy uses the Entity Component System paradigm, which is often shortened to ECS. ECS is a software pattern that involves breaking your program up into **Entities**, **Components**, and **Systems**. **Entities** are unique "things" that are assigned groups of **Components**, which are then processed using **Systems**.
 
-For example, one entity might have a `Position` and `Velocity` component, whereas another entity might have a `Position` and `UI` component. Systems are logic that runs on a specific set of component types. You might have a `movement` system that runs on all entities with a `Position` and `Velocity` component.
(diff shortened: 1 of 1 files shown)


content/learn/book/getting-started/ecs/_index.md
```diff
@@ -9,7 +9,7 @@ insert_anchor_links = "right"
 
 All app logic in Bevy uses the Entity Component System paradigm, which is often shortened to ECS. ECS is a software pattern that involves breaking your program up into **Entities**, **Components**, and **Systems**. **Entities** are unique "things" that are assigned groups of **Components**, which are then processed using **Systems**.
 
-For example, one entity might have a `Position` and `Velocity` component, whereas another entity might have a `Position` and `UI` component. Systems are logic that runs on a specific set of component types. You might have a `movement` system that runs on all entities with a `Position` and `Velocity` component.
+For example, one entity might have a `Position` and `Velocity` component, whereas another entity might have a `Position` and `UI` component. If it helps, you can think of Components as supercharged 'properties' that you can attach to Entities.  Components bring optimized cache and storage efficiency for reuse. **Systems** are logic that runs on a specific set of component types. You might have a `movement` system that runs on all entities with a `Position` and `Velocity` component.
```
cart, line 12: I generally push back against comparisons to other languages / concepts when first introducing ECS concepts because we should assume that people have no prior experience, and generally the analogy breaks down pretty quickly / might end up being unhelpful. Comparing components to "properties" could mean any number of things:
* Field-like getters/setters in languages like C#
* Argument-like "props" passed into React components
* Etc

"Supercharging"  implies it is exactly one of the many options, as you supercharge relative to a reference point.

I also think opening with the implementation detail "components bring optimized cache and storage efficiency for reuse" is too dense, too early (and slightly redundant with the next paragraph).

I'm on board for adding a sentence or two of extra clarity here, but it should be from "first principles" as much as possible.

thadguidry: You are probably right.  Let's close this.

PR #499, opened 2022-12-15
Fix window position on anchors navigation

### LINKED ISSUE #488: Content is cut off when navigating to anchors
This seems to be due to the height of the `position: fixed` header not being accounted for.

### Repro

Navigate here:

https://bevyengine.org/learn/book/migration-guides/0.8-0.9/#make-resource-trait-opt-in-requiring-derive-resource-v2

### What I see

image

### What I expect

image

ciro-c:
When navigating to anchors, the content of the page often gets cut off. With this modification any anchors in pages that renders from `base.html` will have the window position adjusted to it's corrected position.

Fixes #488 

Co-Authored-By: Victor Lima <limavictor49@gmail.com>

### DIFF
--- a/templates/layouts/base.html
+++ b/templates/layouts/base.html
@@ -110,4 +110,13 @@
             </main>
         </div>
     </body>
+    <script>
+        // Moves window position when navigating to anchors to prevent content cut
+        if (location.hash) {
+            scrollBy(0, -70)
+        };
+        window.onhashchange = () => { 
+            scrollBy(0, -70) 
+        };
+    </script>
 </html>


rparrett:
Thanks for looking into this. Fixing that issue will have a big impact.

I wasn't aware of this when writing that issue, but we have some css that's intended to fix this issue here: https://github.com/bevyengine/bevy-website/blob/8dfc51cc233a3e0ddeb6680318763c1572a5901d/sass/_utils.scss#L29 that doesn't seem to be working.

A css solution would likely be preferred.

I tried swapping that out for something based on https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors
```
:target::before {
    content: "";
    display: block;
    height: calc(var(--header-height) + 16px);
    margin: calc(0px - var(--header-height) -16px) 0 0;
}
```

And it seems to work, but it would be nice to understand why what we have now isn't working properly.

ciro-c:
@rparrett  When I was looking into this issue I found another issue with the same problem https://github.com/twbs/bootstrap/issues/1768, and in that discussion I found this link [Jump link positioning](https://nicolasgallagher.com/jump-links-and-viewport-positioning/demo/) that shows methods to solve this problem with css, but is not supported in all browsers. So not knowing the target of the page I decided to use js. I'll change the solution to css.

rparrett:
What I've found is that the CSS linked above does work properly, but only for the assets and examples pages, where it seems like we are manually constructing the anchor tags in a format that is inconsistent with how Zola builds its own anchor tags.

doup:
There is already a utility class for this: `.anchor-target`. See `assets.html` or this new PR: https://github.com/bevyengine/bevy-website/pull/502

doup:
I think I spoke too fast… 😅
The file to change could be `templates/anchor-link.html`, but it's a little bit hacky (and doesn't fully work, yet):

```
<a class="anchor-link" href="#{{id}}_">#</a><a class="anchor-target" id="{{id}}_"></a>
```

1. Note the added `_` to the href used in the `.anchor-target` `id`; the `#` now points to `{{id}}_`
2. Since `anchor-link` is not before the actual heading element (is inside), it doesn't count for the margin… so it's not working correctly.

This results in, the `Make Resource trait...` header is offseted, but not fully:
image


I think (🤞) the offset can be fixed by adding an extra CSS variable (e.g. `--extra-offset`), the idea is:
- By default `--extra-offset` is `0`
- Then, the styling for the news/learn content, defines `--extra-offset` for the headings… which I think would pick the inner `.anchor-target`.
  - This is needed because usually `.anchor-target` is being used outside the heading: `.anchor-target + h2`… but in `anchor-link.html` I'm putting it inside `h2 > .anchor-target` (so it doesn't respect the top margin).

doup:
It actually works… but it feels like a BIG hack. Probably this is something that could be reported upstream in Zola. If you go via this route, all heading should have their own `--extra-anchor-offset` defined.

`sass/_vars.scss`: 
```scss
:root {
    --extra-anchor-offset: 0px;
}
```

`sass/_utils.scss`:
```scss
.anchor-target {
    top: calc(0px - var(--header-height) - var(--extra-anchor-offset) - 8px);
}
```

`sass/pages/_book.scss`:
```scss
.book-content h3 {
  --extra-anchor-offset: 32px;
}
```

**Result:**
image

rparrett:
Throwing another potential solution into the ring: #503

PR #501, opened 2022-12-19
[Merged by Bors] - add a Community page with our members

mockersf:
add a new page using data from https://github.com/bevyengine/bevy-community
add a new link in https://bevyengine.org/community/, the new page will be available as https://bevyengine.org/community/people

see https://github.com/bevyengine/bevy-community/pull/1


templates/community.html
```diff
@@ -114,5 +114,24 @@
             </div>
         </div>
     </a>
+    <a class="card" href="/community/people/">
```
cart, line 117: Not something we need to do this PR, but I think we should consider merging `community` and `people` into one page. Community links at the top (smaller / more consolidated than the current list) directly followed by the people list.

It would give the people page more prominence / discoverability. And it would give more of an impression of "community driven development".

cart (commented):
I love this!

templates/people.html
```diff
@@ -0,0 +1,166 @@
...
+    {{assets_macros::assets_menu(prefix="page-menu", root=section)}}
+{% endblock %}
+
+{% block page_menu %}
+    {{assets_macros::assets_menu(prefix="page-menu", root=section)}}
+{% endblock %}
+
+{% block page_content %}
+
+<svg style="display: none" xmlns="http://www.w3.org/2000/svg" >
```
cart, line 16: One potential issue with the current "one person per category" approach is that these categories are sets that might intersect. This will be even more true once we add "subject matter expert" info.

Ex: I am a maintainer, a org member, a subject matter expert (rendering, ecs, assets, etc), and a community member. I am also the "project lead" which may or may be communicated on this page as a category.

The current categories are reasonable / not confusing (community members that join the org would just move into the org category, org members that become maintainers would move up). And I do like the hard lines between them. But as soon as we add SMEs, things start to break down I think.

Maybe the current categories are the right approach, but we then add the concept of "labels for SMEs (very similar to area labels on github) and other things like it? Then we could also sort org members by "SME-ness" to ensure that information is prioritized / discoverable. And in the future if we need more "labels" we could add them that way. Maintainers could also have SME tags.

I do think I'd like the fact that I am the project lead to be communicated (as this is a separate role with separate responsibilities). If we create a new "project lead" category above maintainer and move me up, then it would mean I would no longer be perceived as a "maintainer", despite also having that role. So maybe it should be a "label". But then its a little confusing that "maintainer" is the only role that isn't a "label". So maybe all "roles" should be "labels"?

I'd also like this page to be a minimal introduction to "how the bevy project is run". To start, a short description of maintainer, org member, and community member would be nice, which we could render below each category title. I'd like to write the initial descriptions for these, so maybe I'll just add that in a follow up PR.

When SMEs roll out (for those who are confused, you'll hear about these soon), I'd like the SME concept to be communicated here as well.

Maybe we blur the lines for the Bevy Org by making every "role" a "label" and creating some "implicit" categories via newlines/padding?

Ex: `Project Lead`, `Maintainer`, `SME-Rendering`, etc would be labels.

```
## The Bevy Org
[description of the bevy org, including "project lead", "maintainer", and "SME" label descriptions]

[(cart, project-lead-label, maintainer-label, sme-labels)]

[(mockersf, maintainer-label, sme-labels), (alice-i-cecile, maintainer-label, sme-labels), (robswain, maintainer-label, sme-labels)]

[(someone, sme-labels), (someone_else)]


## Community
```

cart: "Labels" / the problem they solve also isn't something we need to solve in this PR. Largely just using this PR as a platform to discuss the problem / hopefully reach consensus.

mockersf:
This is how this page would look with only me 
Screenshot 2022-12-20 at 01 00 23

templates/people.html
cart: Also, it might be better (for space-saving and prominence reasons) to merge the implicit "project lead" and "maintainer sections

sass/components/_card.scss
```diff
@@ -9,13 +9,36 @@
...
     border: 2px solid #383838;
+    position: relative;
 
     &:hover {
         background-color: $card-hover-background;
         border: 2px solid #6b6b6b;
     }
 }
 
+.card-profile-picture {
```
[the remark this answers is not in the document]
mockersf: `2.1em` feels too arbitrary, I'll settle on `2.0em` 👍

generate-community/src/bin/validate.rs
```diff
@@ -53,7 +54,7 @@ fn validate_node(node: &CommunityNode) -> Result<(), String> {
             };
 
             if let Some(bio) = &member.bio {
-                if bio.chars().count() > MAX_BIO_LENGTH {
+                if bio.graphemes(true).count() > MAX_BIO_LENGTH {
                     Err(format!("Bio is longer than the maximum allowed length of {}. It is currently {} characters long.", MAX_BIO_LENGTH, bio.len()))?;
```
TheRawMeatball, line 58: The error reporting still reports in byte length

cart: :facepalm:

PR #502, opened 2022-12-21
Fix errors page anchors offset so error headings are visible

doup:
Fixes the scroll position for the error links. Right now the error headings end up under the header (first screenshot). With this fix now it scrolls to the correct position. This uses the same solution as in [`assets.html`](https://github.com/bevyengine/bevy-website/blob/master/templates/assets.html#L26).

**Before**
image

**After**
image


doup:
#503 is a much better (and standard) solution that solves the same issue in several places.

PR #503, opened 2022-12-21
[Merged by Bors] - Fix header text being cut off when navigating to anchors

### LINKED ISSUE #488: Content is cut off when navigating to anchors

rparrett:
Fixes #488 

Closes #502 
Closes #499 

Using `scroll-padding-top`. I am just discovering this CSS property, but it seems to be just the thing we need.

https://caniuse.com/?search=scroll-padding-top
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top

> This allows the author to exclude regions of the scrollport that are obscured by other content (such as fixed-positioned toolbars or sidebars) or to put more breathing room between a targeted element and the edges of the scrollport.

### DIFF
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -54,9 +54,11 @@ $card-list-gap: 15px;
 // CSS Vars
 :root {
     --header-height: 60px;
+    scroll-padding-top: 76px;
 
     @media #{$bp-tablet-landscape-up} {
         --header-height: 72px;
+        scroll-padding-top: 88px;
     }
 }
 
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -54,11 +54,10 @@ $card-list-gap: 15px;
 // CSS Vars
 :root {
     --header-height: 60px;
-    scroll-padding-top: 76px;
+    --scroll-padding-top: calc(var(--header-height) + 16px);
 
     @media #{$bp-tablet-landscape-up} {
         --header-height: 72px;
-        scroll-padding-top: 88px;
     }
 }
 
--- a/sass/elements/_html.scss
+++ b/sass/elements/_html.scss
@@ -3,6 +3,7 @@ html {
     font-family: 'Fira Sans', sans-serif;
     font-size: calc-rem($size-body-mobile);
     background-color: $color-grey-900;
+    scroll-padding-top: 76px;
 
     * {
         box-sizing: border-box;
@@ -12,5 +13,6 @@ html {
 
     @media #{$bp-tablet-portrait-up} {
         font-size: calc-rem($size-body);
+        scroll-padding-top: 88px;
     }
 }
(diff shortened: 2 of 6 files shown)


doup:
This is the way, much better. Seing the _caniuse_ data this is a no-brainer.
At some point I learnt about this property… ^_^U

sass/_vars.scss
```diff
@@ -54,9 +54,11 @@ $card-list-gap: 15px;
 // CSS Vars
 :root {
     --header-height: 60px;
+    scroll-padding-top: 76px;
```
doup, line 57: I would define this as a CSS var, which is based on `--header-height` (no need to add second definition for _landscape-up_ breakpoint):

```suggestion
    --scroll-padding-top: calc(var(--header-height) + 16px);
```

sass/elements/_html.scss
```diff
@@ -3,6 +3,7 @@ html {
     font-family: 'Fira Sans', sans-serif;
     font-size: calc-rem($size-body-mobile);
     background-color: $color-grey-900;
+    scroll-padding-top: 76px;
```
doup, line 6: Then here just use the CSS var. And remove the `88px` definition for `tablet-landscape-up` (btw, you have a bug there as it would grow the scroll padding before the header height changes).

```suggestion
    scroll-padding-top: var(--scroll-padding-top);
```

doup (changes requested):
I haven't tested it locally, but this is the correct approach. I've added a couple of comments to reduce code duplication.

sass/_vars.scss
rparrett: Oops. The bits in `_vars.scss` didn't get cleaned up when I put the properties in the proper place in `_html.scss`.

Good idea to add a variable though. I'll fix that up.

sass/elements/_html.scss
rparrett: > btw, you have a bug there as it would grow the scroll padding before the header height changes

I'm not sure I totally understand. Was this resolved in the latest commit though?

doup (approved):
I've tested it and works perfectly!

sass/elements/_html.scss
doup: Yes! Latest changes look good. 👍 

About the "bug", I misread the code. You were using different breakpoints in `:root` and in `html` to change the `scroll-padding-top`; but since you were repeating the `scroll-padding-top` property probably there wasn't any issue. Anyway, with the CSS vars approach this is irrelevant.

cart (approved):
Perfect! This works for me too. Glad to have this working.

PR #504, opened 2022-12-22
[Merged by Bors] - Sync `new-book` with `master`

doup:
Sync `new-book` with latest changes in `master`. This should unblock #405.

There weren't really any major conflicts, the only thing I noticed is that in the `new-book` the `getting-started` section doesn't exist anymore (now is `welcome`). I've merged the `getting-started` files from `master` anyway and added a note in `ROADMAP.md` (which I presume you were using to track the new-book TODO).

**I haven't done a rebase, just merged `master` into `sync-with-master`**. Which will probably mess with the co-authored information… 😇 

cc @alice-i-cecile


doup:
OK, MD lints need to be fixed and the CI tasks need explicit dependency set:
image

alice-i-cecile:
Thank you so, so much for this. I'll merge it once CI is passing.

IceSentry:
For the fix for the missing link between markdownlint and build-webiste, I'd do it in a separate PR once this is merged. The fix should be pretty easy, but it would get buried here. 

For the CI failures, I'd suggest fixing all the markdownlint in a separate PR too, but before merging this one.

alice-i-cecile:
Yep, I think we should enable `markdownlint` on `main` in a separate PR, and then try this.

doup:
PR for `markdownlint` on `main`: https://github.com/bevyengine/bevy-website/pull/510

alice-i-cecile:
Merging #510 now; once that's in and this is rebased I'll merge this in ASAP.

doup:
OK, synced with the latest changes in `master`.

mockersf:
CI is failing because there are two jobs called `markdownlint`

doup:
I've removed the duplicate `markdownlint` and done few tweaks to the CI config ([see diff](https://github.com/bevyengine/bevy-website/compare/7f6d8d7..bab33a5)):
- Run `markdownlint` and `test-code` before `generate*` tasks
- Temporalily disable `test-code`. I've tried with `job.if`, but it disabled also the dependents. So for now I just commented out a couple of steps.

`test-code` is not passing right now, I've opted to "disable" it . There should be a follow up PR that fixes the issues and re-enables it.

This is how the new setup looks like, having `markdownlint`/`test-code` earlier it'll fail faster and we can skip the costly `generate-*` tasks:

image

doup:
Oh, come-on, now zola fails. 🙃

PR #508, opened 2023-01-09
[Merged by Bors] - fix extra scroll bar on book

hymm:
Book pages are showing a double scroll bar after https://github.com/bevyengine/bevy-website/pull/501.

image

`height: 100%` was added to `media-content` in that PR. As far as I can tell removing it doesn't change the appearance of the people page at all. `media-content` is used as container inside each car. But the card is at a fixed height of 12rem and has overflow hidden. So I wouldn't expect the height to be needed here.

### DIFF
--- a/sass/pages/_content.scss
+++ b/sass/pages/_content.scss
@@ -1,7 +1,6 @@
 $content-font-size: 1.22rem;
 .media-content {
     width: 100%;
-    height: 100%;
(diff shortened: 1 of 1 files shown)


cart (approved):
Much appreciated! I agree that theres a bit too much cross-context CSS overlap atm, but thats a problem for another day and this was a high priority quality issue.

PR #510, opened 2023-01-11
[Merged by Bors] - Add `markdownlint` in CI

doup:
- Add [super-linter](https://github.com/github/super-linter) with [markdownlint](https://github.com/DavidAnson/markdownlint) configuration
- Fix all markdownlint issues in the `content` folder


mockersf:
you should also add it to https://github.com/bevyengine/bevy-website/blob/master/.github/bors.toml

doup:
Btw, I had to change one of the headers in `content/news/2022-07-30-bevy-0.8/index.md`: 

```diff
+ Scripting and Modding Progress: Untyped ECS APIs
- Scripting / Modding Progress: Untyped ECS APIs
```

It looks like when generating the anchor to the header (`#foo-bar`) markdownlint/GitHub and Zola have a different algorithm. So the linter was detecting a couple of dead links on that file (which weren't). I opted to change the title as I think it's a good lint rule, even if it might fail with certain characters.

mockersf (approved):
didn't review everything but the CI config looks OK, and a few randomly picked toml changes are OK

alice-i-cecile (approved):
bors r+

PR #514, opened 2023-01-14
[Merged by Bors] - Community: Add Role tags and consolidate Bevy Organization

cart:
This adds Bevy Organization Role "tags" to the Bevy People/Community page (Project Lead, Maintainer, SME). These are displayed on each card. I've also added a short description of the Bevy Org and each role.
It also consolidates all org sections (Maintainers, Bevy Org) into a single "The Bevy Organization" section. 

You might notice we have a new role! We'll be announcing what that is and the people that will occupy it in the next day or so.

The Bevy People page now looks like this.

image

Note that this currently uses my fork of bevy-community. We should merge this first, then migrate bevy-community to the new format.

### DIFF
--- a/generate-community/generate_community.sh
+++ b/generate-community/generate_community.sh
@@ -1,9 +1,12 @@
 #!/bin/sh
 
-git init bevy-community
+git clone https://github.com/cart/bevy-community
 cd bevy-community
-git remote add origin https://github.com/bevyengine/bevy-community
-git pull --depth=1 origin main
+git checkout roles-and-unify
+# git init bevy-community
+# cd bevy-community
+# git remote add origin https://github.com/bevyengine/bevy-community
+# git pull --depth=1 origin main
 cd ..
 
 cargo run --bin generate -- bevy-community ../content/ community
--- a/generate-community/src/bin/generate.rs
+++ b/generate-community/src/bin/generate.rs
@@ -1,9 +1,11 @@
 use rand::{prelude::SliceRandom, thread_rng};
 use serde::Serialize;
 use std::{
+    collections::BTreeMap,
     fs::{self, File},
     io::{self, prelude::*},
-    path::Path,
+    path::{Path, PathBuf},
+    str::FromStr,
 };
 
 use generate_community::*;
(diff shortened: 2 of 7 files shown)


IceSentry:
The entire block with the text feels a bit dense, but I'm not entirely sure what to suggest to fix it. Also, the badges seem a bit tall and if someone is an SME in a few areas then it will become 2 lines and it might be too tall.

Personally, I'd keep the separate sections or at least keep maintainers separate and have a badge for org-members.

cart:
> The entire block with the text feels a bit dense, but I'm not entirely sure what to suggest to fix it.

Some ideas:

1. Make it all one paragraph so it takes up less vertical space
2. Don't have text here and just link to the new Bevy Organization doc. 

> Also, the badges seem a bit tall and if someone is an SME in a few areas then it will become 2 lines and it might be too tall.

Yeah sadly they get way less legible if we reduce the font size (one way to reduce height). And yeah going beyond three tags will wrap around, which will use even more vertical space.  We might be able to sort out a grid formatting approach that makes the cards (and the others in their row) as big as they need to be, while maintaining a smaller size for things without tags. The maintainers and I wracked our brains trying to solve the layout problem but couldn't come up with anything particularly satisfying. We largely chose to punt the "running out of space" problem to the future. I don't want to hold back the SME rollout on this.

But I'll try to poke around a bit to regain space where possible.
 
> Personally, I'd keep the separate sections or at least keep maintainers separate and have a badge for org-members.

I personally really like having a unified way of communicating org roles, and the implication that The Bevy Organization is "one thing" rather than a bunch of separate things. Using sections for specific roles makes it harder to allow people to occupy more than one role at once. And theres a reasonable chance we'll add more role types in the future.

IceSentry:
For the general size of the first section I think the issue, at least for me, is just the density. One option could be to use a few sub headers to break things up a bit. I would suggest having #Roles and #Members as sub-headers. Where roles would contain the explanation of the roles and members, well, the members. I think there should be at least some explanation here because not everyone is comfortable with reading an md file on github, especially when we start to be big enough that artist look into bevy and not just programmers.

Something like this: 
image

(This was just quickly thrown together in the dev console)

For the badges, I think just removing some vertical padding should be enough, my main issue is that it doesn't look like the same padding is used for the explanation and for the actual badge. I haven't looked at the css yet though, so maybe this isn't easy. I agree that this isn't a blocker at all.

As for separate sections for roles, that's up to you, I don't mind either way.

cart:
> For the badges, I think just removing some vertical padding should be enough, my main issue is that it doesn't look like the same padding is used for the explanation and for the actual badge. I haven't looked at the css yet though, so maybe this isn't easy. I agree that this isn't a blocker at all.

Weirdly the same padding css is used, but i literally can't make the vertical padding smaller within the card (even if I add `!important`. I have full control over horizontal padding though.

cart:
I really like the "roles and members" sections. Definitely helps break things up a bit!

IceSentry:
Also, I haven't looked how easy that would be, but adding alt-text to the roles badges with their description would probably be a good idea too. Could be done later though.

james7132:
Wonder if we could make the role descriptions both shorter and tooltips on the tags? That'd cut out the bulk of the text.

IceSentry:
I don't think they should be only tooltips. That would reduce the size yes, but that would also be less accessible.

IceSentry:
The height issue is caused by the line-height property of `.media-content`. That would be the second time this class is causing issue lol. Probably not worth touching it now, but worth looking into it later.

cart:
Line height worked! I reduced the vertical padding and shrunk the card height to account for the space gained.
I also added the sections mentioned above.
Given that we're announcing today, I'll leave it there. But we can definitely keep tweaking later.

PR #516, opened 2023-01-14
[Merged by Bors] - Use upstream bevy-community

cart:
Now that bevy-community is updated to the new format, we should use it!


cart:
Haha the bevy-community PR wasn't merged yet due to a mismatched validator source. This should work in a minute or so.

PR #518, opened 2023-01-15
[Merged by Bors] - Make the People page less confusing

inodentry:
When I first looked at this page, I found it confusing. I had to read the README in the `bevy-community` repo to understand how it all works and why there is a separate Community Members section.

My first thoughts were "why are there so many people with no Role tags under the bevy org section? what is their role in the org? seems like there are a bunch of random people listed here, and some are separated under the community heading and others lumped in with the project people?"

This PR helps make the page more understandable.
 
 - Add some text under "Roles", to describe the role of the org members without any "special role" tag
 - Add some text under "Community Members", to clarify the purpose of that section

### DIFF
--- a/templates/people.html
+++ b/templates/people.html
@@ -85,7 +85,7 @@
                     <div class="people-role-description-text">(Subject Matter Experts) have proven themselves to be experts in a given development area. They help guide Bevy's direction in their area and they are great people to reach out to if you have questions about a given area.</div>
                 </div>
                 <div class="people-role-description">
-                    <div class="people-role-description-text">The other members, without a special role, are part of the triage team. They can manage GitHub issues and pull requests.</div>
+                    <div class="people-role-description-text">Other members, without a special role, are part of the GitHub organization. They have the power to manage GitHub issues and pull requests.</div>
                 </div>
             </div>
         </p>
@@ -102,7 +102,7 @@
         {% if section.title == "Community Members" %}
         <div class="people-section-description media-content">
         <p>
-            People from the Bevy community, who are not part of the organization.
+            People in the Bevy community who are not part of the organization.
         </p>
         <p>
             Anyone can add themselves here! <a href="https://github.com/bevyengine/bevy-community">Submit a pull request</a>!
--- a/templates/people.html
+++ b/templates/people.html
@@ -84,6 +84,9 @@
                     <div class="people-role people-role-top-level people-role-sme">SMEs</div>
                     <div class="people-role-description-text">(Subject Matter Experts) have proven themselves to be experts in a given development area. They help guide Bevy's direction in their area and they are great people to reach out to if you have questions about a given area.</div>
                 </div>
+                <div class="people-role-description">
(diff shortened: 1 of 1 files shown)


templates/people.html
```diff
@@ -84,6 +84,9 @@ <h2>
                     <div class="people-role people-role-top-level people-role-sme">SMEs</div>
                     <div class="people-role-description-text">(Subject Matter Experts) have proven themselves to be experts in a given development area. They help guide Bevy's direction in their area and they are great people to reach out to if you have questions about a given area.</div>
                 </div>
+                <div class="people-role-description">
+                    <div class="people-role-description-text">The other members, without a special role, are part of the triage team. They can manage GitHub issues and pull requests.</div>
```
alice-i-cecile, line 88:
```suggestion
                    <div class="people-role-description-text">Oher members, without a special role, are part of the GitHub organization. They have the power to manage GitHub issues and pull requests.</div>
```

templates/people.html
```diff
@@ -95,6 +98,18 @@ <h2>
             Members<a class="anchor-link" href="#org-members">#</a>
         </h2>
         {% endif %}
+
+        {% if section.title == "Community Members" %}
+        <div class="people-section-description media-content">
+        <p>
+            People from the Bevy community, who are not part of the organization.
```
alice-i-cecile, line 105:
```suggestion
            Members of the Bevy community who are not part of the GitHub organization.
```

templates/people.html
```diff
@@ -95,6 +98,18 @@ <h2>
...
+        </p>
+        <p>
+            Anyone can add themselves here! <a href="https://github.com/bevyengine/bevy-community">Submit a pull request</a>!
```
alice-i-cecile, line 108: I really like this,

templates/people.html
inodentry: I specifically did not want to use the word "members" here, to avoid confusion with org members, who are listed under a heading "Members".

Also, everywhere else on the page, we call it "The Bevy organization", and it lists all the people in the GH org anyway, without calling it that. Calling it "the GitHub organization" here feels weird and confusing.

My goal with this blurb of text was to make this "Community" part of the page clearly distinguished from the "Org" part of the page above it. The two sections not being distinct enough was what caused my initial confusion and motivated me to make this PR.

templates/people.html
inodentry: Why do you prefer the term "GitHub organization" over "triage team"? I'm not against it, I just want to know your thoughts.

alice-i-cecile: The triage team is an artifact at this point: we should try to remove it if we can.

alice-i-cecile (approved):

alice-i-cecile:
@cart when you get a chance we could use an `A-Community`.

IceSentry (approved):

alice-i-cecile (commented):
bors r+

cart:
@alice-i-cecile good call. I added `A-People` as we've reached relative consensus that calling this feature "Bevy People" everywhere (including repo names) is the move.

PR #532, opened 2023-02-15
[Merged by Bors] - donate page backed by the Bevy people

mockersf:
have the donate button link to a page with people from the Bevy org with a sponsor link


templates/donate.html
```diff
@@ -0,0 +1,199 @@
...
+    {{assets_macros::assets_menu(prefix="page-menu", root=section)}}
+{% endblock %}
+
+{% block page_content %}
+
+<svg style="display: none" xmlns="http://www.w3.org/2000/svg" >
+    <!--! Font Awesome Pro 6.2.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2022 Fonticons, Inc. -->
+    <defs>
+        <symbol id="heart" viewbox="0 0 512 512">
+            <path d="M47.6 300.4L228.3 469.1c7.5 7 17.4 10.9 27.7 10.9s20.2-3.9 27.7-10.9L464.4 300.4c30.4-28.3 47.6-68 47.6-109.5v-5.8c0-69.9-50.5-129.5-119.4-141C347 36.5 300.6 51.4 268 84L256 96 244 84c-32.6-32.6-79-47.5-124.6-39.9C50.5 55.6 0 115.2 0 185.1v5.8c0 41.5 17.2 81.2 47.6 109.5z"/>
```
alice-i-cecile, line 20: These constants are very mysterious; can you add comments?

mockersf: they aren't that mysterious... it's the path for the "heart" from font awesome as SVG...

alice-i-cecile: I expected as much, but wasn't sure. I'm fine to leave it if you think it'll be obvious to folks with more web experience.

cart:
Just pushed some changes:
1. Remove the indentation / make use of space better
2. Added clarity + direction to the sponsorship process
3. Made the cards link to sponsorships if you click anywhere on the card (that isn't a link). I tried to do this without onclick, but couldn't get the formatting right with `<a>` links.

cart:
image

PR #534, opened 2023-02-19
Sponsorship Pledge

cart:
Bevy is a community-driven people-first project. We do not have a business model. We don't sell anything. For now, we don't even have a centralized legal organization. Instead, our work is fully supported by sponsorships from generous individuals and companies. But introducing money into the equation creates concerns about power dynamics and incentives. And some companies require contracts as a part of their sponsorship programs. Bevy leaders (Project Leads and Maintainers) _must serve the community's interests above all else_. Therefore leaders that accept any money within the context of Bevy must abide by the following pledge:

(see sponsorship-pledge.html for pledge contents)

The @bevyengine/maintainer-team has agreed to (and helped iterate on) this pledge. Maintainers, please create a new PR after this one is merged to sign this and add relevant disclosures. I'll be following up with a couple of changes to my entry as well (as I have an exciting announcement to make).

### DIFF
--- a/templates/sponsorship-pledge.html
+++ b/templates/sponsorship-pledge.html
@@ -1,0 +1,59 @@
+{% extends "layouts/base.html" %}
+
+{% block content %}
+<div class="media-content">
+
+<h1>Bevy's Sponsorship Model</h1>
+
+Bevy is a community-driven people-first project. We do not have a business model. We don't sell anything. For now, we don't even have a centralized legal organization. Instead, our work is fully supported by sponsorships from generous individuals and companies. But introducing money into the equation creates concerns about power dynamics and incentives. And some companies require contracts as a part of their sponsorship programs. Bevy leaders (Project Leads and Maintainers) <i>must serve the community's interests above all else</i>. Therefore leaders that accept any money within the context of Bevy must abide by the following pledge:
+
+<h2>The Pledge</h2>
+
+<ol>
+    <li>
+        <b>The needs of the <i>people</i> in the Bevy community come first</b>. Everything we do should be in the interest of the <i>individuals in the Bevy community</i>. Yes ... companies use (and support) Bevy. This is a good thing. But companies are <i>not</i> people. Companies <i>consist of people</i>. The voice of an individual in a company should be weighted exactly the same as every other voice in our community. No more. No less.
+    </li> 
+    <li>
+        <b>Money can <i>never</i> buy project direction</b>. Bevy's design direction must always be determined by community needs and (as much as possible) consensus. For example: a company can never directly pay us to adopt a feature that goes against the community's interests. Nor can they indirectly pull the strings via financial pressure, such as threat of un-sponsoring.
+    </li>
+    <li>
+        <b>Money can <i>never</i> buy project priority</b>. We must always do the work that we believe is best for the Bevy community at a given moment. We cannot be paid to bump a company's pet feature up in our priority list. Our time (within the context of what we commit to the Bevy community) is not for sale. If we are "full time" Bevy leaders, none of our time can be bought. If we are "part time Bevy leaders", only the remaining time can be bought. 
+    </li>
+    <li>
+        <b>Bevy leaders must be able to speak their mind</b>. We will never sign away our right to disclose our thoughts or opinions about entities we receive money from. This means no "non-disparagement agreements". "Non-disclosure agreements" are only ok within the context of protecting private information (that does not directly relate to Bevy). 
+    </li> 
+    <li>
+        <b>Bevy leaders must be open about who supports them (within the context of Bevy)</b>. No hidden entities behind the scenes. If we receive $2,400 a year or more from an entity or have a financial stake in one that relates to the Bevy project in any reasonable way, we must disclose it within 6 months of receiving the money or stake.
+    </li>
+    <li>
+        <b>Bevy leaders will hold each other accountable</b>. If any of us suspects a violation of the pledge, we will immediately seek to rectify it. First amongst ourselves if there is reasonable doubt, and then publicly if the situation demands it.
+    </li>
+    <li>
+        <b>The pledge and the Bevy Community come first ... always</b>. If we are confronted with the choice of upholding the pledge or losing the financial support of an entity, we will choose the pledge every time. The amount of money on the table has no bearing on this. We must be honest with ourselves and each other. If we can't say no to large sums of money with pledge-violating strings attached, we don't deserve to be Bevy leaders.
+    </li>
+</ol>
+
+<p>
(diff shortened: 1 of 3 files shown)


templates/sponsorship-pledge.html
```diff
@@ -0,0 +1,59 @@
...
+    </li> 
+    <li>
+        <b>Bevy leaders must be open about who supports them (within the context of Bevy)</b>. No hidden entities behind the scenes. If we receive $2,400 a year or more from an entity or have a financial stake in one that relates to the Bevy project in any reasonable way, we must disclose it within 6 months of receiving the money or stake.
+    </li>
+    <li>
+        <b>Bevy leaders will hold each other accountable</b>. If any of us suspects a violation of the pledge, we will immediately seek to rectify it. First amongst ourselves if there is reasonable doubt, and then publicly if the situation demands it.
+    </li>
+    <li>
+        <b>The pledge and the Bevy Community come first ... always</b>. If we are confronted with the choice of upholding the pledge or losing the financial support of an entity, we will choose the pledge every time. The amount of money on the table has no bearing on this. We must be honest with ourselves and each other. If we can't say no to large sums of money with pledge-violating strings attached, we don't deserve to be Bevy leaders.
```
alice-i-cecile, line 32:
```suggestion
        <b>The pledge and the Bevy Community always come first</b>. If we are confronted with the choice of upholding the pledge or losing the financial support of an entity, we will choose the pledge every time. The amount of money on the table has no bearing on this. We must be honest with ourselves and each other. If we can't say no to large sums of money with pledge-violating strings attached, we don't deserve to be Bevy leaders.
```

templates/sponsorship-pledge.html
```diff
@@ -0,0 +1,59 @@
...
+Also note that this only applies to Bevy leadership. We welcome and encourage companies to fund development efforts that improve Bevy in whatever way they want. However Bevy leadership <i>cannot be involved in any way that violates The Pledge</i>. Bevy leadership must always be the filter that protects the community from changes that go against their interests. Companies should do their best to collaborate with the Bevy community (and Bevy leaders) to ensure their changes are acceptable.
+</p>
+
+<h2>Pledge Signers</h2>
+<ul>
+    <li>
+        <a href="https://github.com/sponsors/cart">Carter Anderson (@cart)</a> 
+        <ul>
+            <li>Work Status: Full-time Project Lead / Maintainer</li>
+            <li>Sponsors (above $2,400 / year threshold): Foresight Mining Software Corporation ($2,000 / month), Embark ($1,000 / month), MeetKai ($1,000 / month), Encultured ($1,000 / month), Striked ($200 / month), storytold ($200 / month), Metabuild ($100 / month), Keygen ($100 / month), VPSServer ($100 / month), @joshuajbouw ($200 / month), @CleanCut ($200 / month), @rozgo ($100 / month)</li>
```
alice-i-cecile, line 54:
```suggestion
            <li>Sponsors (including all those  above $2,400 / year threshold): Foresight Mining Software Corporation ($2,000 / month), Embark ($1,000 / month), MeetKai ($1,000 / month), Encultured ($1,000 / month), Striked ($200 / month), storytold ($200 / month), Metabuild ($100 / month), Keygen ($100 / month), VPSServer ($100 / month), @joshuajbouw ($200 / month), @CleanCut ($200 / month), @rozgo ($100 / month)</li>
```

alice-i-cecile: There are $100 / month sponsors on here, that I think are probably worth including.

IceSentry, line 54: I'm not sure I understand the threshold means in this context. Every sponsor listed is below the threshold.

JulianRuiseco:
All for this!

templates/sponsorship-pledge.html
aevyrie: Per year, vs per month. :)

aevyrie: This is a pretty understandable point of confusion. Maybe the sponsor numbers should also be listed as per year?

IceSentry: Oooh, yeah, that makes sense. In that case, I'd suggest using the same unit in the list and the threshold. Also, maybe explain what the 2400/year number represents.

aevyrie:
Have you considered limiting the percent of leaders a company can sponsor directly?

I've seen this be a point of (perceived) contention in the rustlang project itself, though I don't know if it's a real issue. For example, if FooCorp decides they want to support Bevy, and begin to personally sponsor every person in a leadership position in the organization, that could certainly lead to the perception that FooCorp has a ton of soft power over the organization, regardless of what the pledge says. It may also lead to a situation where the people in the organization actually do have reliance on an outsized sponsor.

afonsolage:
What if a `Bevy Leader` wants to do some work to a company for financial needs for like 6 months, he/she can resign temporarily from his position as `Bevy Leader`, became a `normal` member do the work and get back to the `Bevy Leader` position after the work has been done?

I don't know if it's because I live in a third world country, but there are some cases where you have to do some work to raise money due to survival.

Also I think should be stated somewhere who judge if the pledge was violated or not, since there can be some edge cases like the up mentioned.

cart:
> Have you considered limiting the percent of leaders a company can sponsor directly?

The maintainers and I will discuss it. I definitely understand this being a concern. Given that a "legal bevy org" is roughly impending, idk how we would handle this case in that context, as each sponsor's contributions would be "spread out" across the devs.

> What if a Bevy Leader wants to do some work to a company for financial needs for like 6 months, he/she can resign temporarily from his position as Bevy Leader, became a normal member do the work and get back to the Bevy Leader position after the work has been done?

Yup this is definitely allowed. Its all about being open with the community. If you are going from Full Time to Part Time (or no time), you should tell your community and your sponsors (if you previously made a Full Time commitment to them). The Work Status field in the Signers section exists as a part of that transparency. We support basically any work schedule / time commitment, as long as you are open about it in the context of raising funds.

> Also I think should be stated somewhere who judge if the pledge was violated or not, since there can be some edge cases like the up mentioned.

In the document we've stated that for now (1) maintainers should hold each other accountable and (2) this operates on the honor system. As soon as we have a "legal bevy org" we can have a centralized enforcement policy that can actually "do" things (like revoke access to money). This is a "promise to the community", so if anyone in the community believes we are breaking that promise, they can and should bring that up to maintainers (and/or the community at large if that feels necessary).

templates/sponsorship-pledge.html
cart: I've switched to per month as that is the unit we use everywhere else / how sponsorships generally work.

cart:
https://github.com/bevyengine/bevy-website/pull/536

PR #538, opened 2023-02-21
[Merged by Bors] - Remove web-streams polyfills

ickk:
Firefox was lacking support for `pipeThrough` when this feature was added in April 2022. Full support was added shortly after in June [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeThrough#browser_compatibility). 

It's been a decent period of time since, and given we don't test against older browsers in any way I think it makes sense to remove these polyfills to improve page loading times.


IceSentry (approved):
I haven't tested the code, but the change makes sense to me

cart:
Seems reasonable. Lets merge and see if anyone yells :)

PR #542, opened 2023-02-21
Local build instructions

ickk:
While preparing #538 I needed to test the changes locally. Finding your way through the build process feels like *maze of twisty little passages, all alike*, which could be frustrating for new contributors. 

I roughly documented the process, but it could probably be generalised and made more concrete. We could even make a couple of simple scripts that automate setting up an environment similar to ci to ease the process.

### DIFF
--- a/local_build_instructions.md
+++ b/local_build_instructions.md
@@ -1,6 +1,6 @@
 # Windows local build instructions
 
-Building bevy-website locally is tedious, especially on windows. Here is one (brute force) method:
+Building bevy-website locally is tedious, especially on Windows. Here is one (brute force) method:
 
 1. [Install WSL](https://learn.microsoft.com/en-us/windows/wsl/install), with Ubuntu.
 
--- a/local_build_instructions.md
+++ b/local_build_instructions.md
@@ -1,4 +1,4 @@
-### Windows local build instructions ###
+# Windows local build instructions
 
 Building bevy-website locally is tedious, especially on windows. Here is one (brute force) method:
 
--- a/local_build_instructions.md
+++ b/local_build_instructions.md
@@ -1,0 +1,42 @@
+### Windows local build instructions ###
+
+Building bevy-website locally is tedious, especially on windows. Here is one (brute force) method:
+
+1. [Install WSL](https://learn.microsoft.com/en-us/windows/wsl/install), with Ubuntu.
+
+note: You could instead use [git bash](https://gitforwindows.org/), but some of the details might be different.
+
+2. Open a WSL bash shell
+3. [install rustup](https://rustup.rs/) `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
+4. Add the wasm target `rustup target add wasm32-unknown-unknown`
+5. Install build dependencies `sudo apt install build-essential pkg-config`
+6. Install wasm-bindgen tool `cargo install wasm-bindgen-cli`
+
+Now we have everything we need to start generating the site. We can roughly follow the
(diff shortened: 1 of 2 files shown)


local_build_instructions.md
```diff
@@ -0,0 +1,42 @@
+### Windows local build instructions ###
+
+Building bevy-website locally is tedious, especially on windows. Here is one (brute force) method:
```
alice-i-cecile, line 3:
```suggestion
Building bevy-website locally is tedious, especially on Windows. Here is one (brute force) method:
```

local_build_instructions.md
```diff
@@ -0,0 +1,42 @@
+### Windows local build instructions ###
```
alice-i-cecile, line 1:
```suggestion
# Windows local build instructions
```

alice-i-cecile:
This lines up with my experience building this on Windows.

IceSentry:
Personally, I prefer using git bash because it comes with git, so no additional install step and it uses your already existing installation of rust.

local_build_instructions.md
```diff
@@ -0,0 +1,42 @@
...
+2. Open a WSL bash shell
+3. [install rustup](https://rustup.rs/) `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
+4. Add the wasm target `rustup target add wasm32-unknown-unknown`
+5. Install build dependencies `sudo apt install build-essential pkg-config`
+6. Install wasm-bindgen tool `cargo install wasm-bindgen-cli`
+
+Now we have everything we need to start generating the site. We can roughly follow the
+outline of the CI build script `.github/workflows/ci.ympl`
+
+7. cd into the website repo. There are a bunch of directories starting with "generate-".
```
IceSentry, line 18: I would add that these steps are optional if you don't plan to work on those related pages.

IceSentry: Like, when I was working on the asset page, I never generated all the other pages.

IceSentry: Thinking about it, we should probably just provide a `generate.sh` at the root that does all those steps

mockersf:
I would prefer to discourage people from building parts they don't need as they are optional.

Another interesting option to explore could be to have the deploy process save everything generated to a branch that can be used when people need all the parts .

local_build_instructions.md
ickk: Agree with you on git-bash being more minimal.. But I'm also not going to install a whole rust environment to git-bash when I already use WSL. I think there's value in having both.

I was hoping we could either flesh this out to include the other platforms (which will be pretty similar to the bulk, but I can't test mac), or just provide an automation for setting up the environment.

ickk:
As for each thing being optional, I thought it would be pretty obvious for anyone wanting to contribute, but maybe some more explicit language would be good.

Mostly I just wanted to document what I did so next time I don't have to rediscover missing dependencies step by step. Feel free to change the wording, add stuff, whatever.

ickk:
> Another interesting option to explore could be to have the deploy process save everything generated to a branch or an artefact to download that can be used when people need all the parts.

This could be useful, does sound maybe a little convoluted. Even so, I think regular documentation is important for when a contributor needs to change something that impacts the generated assets themselves.

cart:
Maybe instead of going further down this path we should use cargo/xshell instead of bash for facilitating these builds (like we do for `cargo run -p ci`?

We should probably have a cargo workspace setup for this anyway.

generate_wasm_examples will be the biggest challenge. the `find` and `sed` usage specifically

PR #545, opened 2023-02-21
Add tonemapping example images for use in the book and rustdoc

superdump:


doup:
If these are going to be referenced from [the API reference](https://docs.rs/bevy/latest/bevy/) (not the website docs) then I think it's better to move it into a separate `assets-docs` folder so it's not removed in case there is a website redesign (or tooling is used to optimize the website `assets`, etc.).

If this approach is used, then I would also add a `assets-docs/README.md` to explain the purpose of this folder.

alice-i-cecile:
Seconding doup's suggestion. Once that's done this LGTM.

superdump:
Done.

alice-i-cecile (approved):

mockersf:
Do we have an idea on how to handle versioning if we notice a bug in how those images were generated and need to be updated, but still have to link to a before fix version in the corresponding docs?

doup:
Maybe just version the files like `foo.v0.webp`? Without adding a relation to any Bevy version.

mockersf:
https://github.com/rust-lang/rfcs/pull/3397
It also mentions a few alternatives that would work today

SIGSTACKFAULT (approved):

Selene-Amanita:
A bit late to the party:

This image should probably be moved there too: https://github.com/bevyengine/bevy/tree/main/assets/docs ?

(Created in https://github.com/bevyengine/bevy/pull/9061 )

PR #561, opened 2023-03-03
[Merged by Bors] - Add sidebar to news posts

doup:
This PR adds a sidebar using [Zolas TOC](https://www.getzola.org/documentation/content/table-of-contents/). In desktop first two levels are shown, in mobile just the top level headers.

**DESKTOP**
image

**MOBILE**
image

### DIFF
--- a/sass/components/_page-with-menu.scss
+++ b/sass/components/_page-with-menu.scss
@@ -27,11 +27,9 @@
     &__menu {
         @include scrollbar-v;
 
-        --sidebar-margin-top: 0px;
-
         position: sticky;
-        top: calc(var(--header-height) + #{$margin-top} + var(--sidebar-margin-top));
-        height: calc(100vh - var(--header-height) - #{$margin-top * 2} - var(--sidebar-margin-top));
+        top: calc(var(--header-height) + #{$margin-top});
+        height: calc(100vh - var(--header-height) - #{$margin-top * 2});
         overflow-y: auto;
         padding-bottom: $margin-top;
     }
@@ -39,16 +37,5 @@
     &__content {
         grid-area: content;
         overflow-x: hidden;
-    }
-
-    &--news {
-        @media #{$bp-tablet-landscape-up} {
-            grid-template-columns: 1fr $sidebar-width;
-            grid-template-areas: 'content menu';
-        }
-
-        .page-with-menu__menu {
-            --sidebar-margin-top: 32px;
-        }
     }
 }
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
Yeah this layout looks good enough for now.

doup:
I've changed the layout so it looks like the book (see PR description). For reference, this is what it looked like before the latest changes: https://github.com/bevyengine/bevy-website/pull/561/commits/3bc8d5879a39b1dff6299515f0f8272882f5a303 The issue is that aligning the menu with the image is tricky… as the text can be multi-line, we would need to refactor the layout.

image

Another option is to to put the menu on the right, but without margin adjustment. The order (menu-content) is not consistent with the book, but at least the spacing is consistent and looks while scrolling:

image

**The difference is subtle, there is more top margin on the first image; and while scrolling that margin keeps there for the sidebar.**

doup:
I've moved the sidebar to the right, I think it makes more sense for a blog post. The PR description is updated. To check the previous versions try these commits:
- V1: https://github.com/bevyengine/bevy-website/pull/561/commits/ba85ea50a85c4e126c8974c9b4bf8e4d6f75820b
- v2: https://github.com/bevyengine/bevy-website/pull/561/commits/3bc8d5879a39b1dff6299515f0f8272882f5a303
- V3 (as in the PR description): https://github.com/bevyengine/bevy-website/pull/561/commits/7f83e27597b80f940e65f996a4a585c943ce10d7

cart:
Lovely work! I think my one critique is that the content feels a bit too compressed now. Not much to do about that with the current margins (i don't think we should make the ToC any thinner than it is). Maybe we could fudge it by making the margins smaller (and the content area larger) on desktop / when there is room?

cart:
If we can't sort that out before release I don't think its a huge deal. _Very_ shippable as-is. Thank you so much for putting this together on such short notice!

cart:
Expanding the content area "well" will be hard I think, especially if we try to retain alignment with the headerbar ... which probably shouldn't change width according to the page you're on (that could be jarring). So we either expand the width of the content area / header bar globally (when possible) or we allow the content in blog posts to grow "beyond" the headerbar margins.

cart (approved):
In fact, lets merge this as-is to make sure it gets in. We can tweak spacing later if we find time / find spacing we prefer more.

doup:
We could change the width for bigger screens, so space is not wasted.

On the other hand, there is this typographic rule suggesting that text lines should be around 60-80 chars to improve readability (this can be compensated with bigger `line-height` on long lines…); by that rule it should even be narrower. :P

In any case it would useful for the Book if we end up adding the page TOC.

PR #563, opened 2023-03-04
[Merged by Bors] - Code in title: don't overwrite the font size

mockersf:
titles with code don't look pretty because of the changed font-size:

Screenshot 2023-03-04 at 01 03 24


cart (approved):
Muuuuuch appreciated. This was pretty nasty.

PR #564, opened 2023-03-04
[Merged by Bors] - 0.10 shader features

cart:


content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -686,6 +686,87 @@ app from the app to send it between the main thread and the rendering thread. Th
...
+
+```rust
+for (var i: u32 = 0u; i < #{MAX_DIRECTIONAL_LIGHTS}; i = i + 1u) {
+}
+```
+
+They can also be defined inline in shaders:
+
+```rust
+#define MAX_DIRECTIONAL_LIGHTS 10
```
LiamGallagher737, line 721: Here the 10 is not clear to me, is this a default value?

cart: Its just an arbitrary value chosen to illustrate what setting values look like. Its equivalent to `let max_directional_lights = 10` in rust.

LiamGallagher737: I get that. What I don't understand is how this fits in with passing constant values in from rust. If I passed in a value of 15 using `ShaderDefVal::Int("MAX_DIRECTIONAL_LIGHTS".to_string(), 15)` would that override the value defined in the shader?

cart: Ahh yeah the precedence order is worth clarifying. Good call!

content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -686,6 +686,87 @@ app from the app to send it between the main thread and the rendering thread. Th
 only valid to do after all the plugin build methods have been called, because any plugin may
 want to modify the rendering sub app.
 
+## ShaderDef Values
+
+<div class="release-feature-authors">authors: @mockersf</div>
+
+Bevy's shader processor now supports ShaderDefs with values. This allows developers to pass constant values into their shaders:
```
alice-i-cecile, line 693: ShaderDefs should probably have a link.

LiamGallagher737 (approved):

PR #566, opened 2023-03-04
[Merged by Bors] - 0.10 Reflection

cart:


alice-i-cecile (approved):
Good explanations!

cart:
Forgot to note that many of the examples (and the ReflectFromReflect opening blurb) are pretty much direct adaptations of @MrGVSV's pr descriptions. They were very helpful!

content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -739,6 +739,106 @@ query.iter().for_each(|mut component| {
...
+
+assert_eq!(*x.path::<u32>("data.1").unwrap(), 123);
+assert_eq!(*x.path::<u32>("some_tuple.0").unwrap(), 10);
+```
+
+## Pre-Parsed Reflection Paths
+
+<div class="release-feature-authors">authors: @MrGVSV</div>
+
+Reflection paths enable a lot interesting and dynamic editor scenarios, but they do have a downside: calling `path()` requires parsing strings every time. To solve this problem we added [`ParsedPath`], which enables pre-parsing paths and then reusing those results on each access:
```
LiamGallagher737, line 774:
```suggestion
Reflection paths enable a lot of interesting and dynamic editor scenarios, but they do have a downside: calling `path()` requires parsing strings every time. To solve this problem we added [`ParsedPath`], which enables pre-parsing paths and then reusing those results on each access:
```

content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -739,6 +739,106 @@ query.iter().for_each(|mut component| {
...
+  some_tuple: (10, 20),
+};
+
+assert_eq!(*x.path::<u32>("data.1").unwrap(), 123);
+assert_eq!(*x.path::<u32>("some_tuple.0").unwrap(), 10);
+```
+
+## Pre-Parsed Reflection Paths
+
+<div class="release-feature-authors">authors: @MrGVSV</div>
```
MrGVSV, line 772: This PR was just a rebase of @james7132's PR, so they should probably be listed as an author as well :)

content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -739,6 +739,106 @@ query.iter().for_each(|mut component| {
...
+let element = parsed_path.element::<usize>(&some_value);
+```
+
+Much more suitable for repeated access, such as doing the same lookup every frame!
+
+## `ReflectFromReflect`
+
+<div class="release-feature-authors">authors: @MrGVSV</div>
+
+When using Bevy's Rust reflection system, we sometimes end up in a scenario where we have a "dynamic reflect value" representing a certain type `MyType` (but under the hood, it is not that type). Such scenarios happen when we call `Reflect::clone_value`, use the reflection deserializers, or create the dynamic value ourselves. Unfortunately, we can't just call `MyType::from_reflect` as we do not have knowledge of the concrete `MyType` at runtime.
```
MrGVSV, line 787: Nit: I think this could be a little clearer. Maybe something like?
```suggestion
When using Bevy's Rust reflection system, we sometimes end up in a scenario where we have a "dynamic reflect value" representing a certain type `MyType` (even though under the hood, it isn't really that type). Such scenarios happen when we call `Reflect::clone_value`, use the reflection deserializers, or create the dynamic value ourselves. Unfortunately, we can't just call `MyType::from_reflect` as we do not have knowledge of the concrete `MyType` at runtime.
```

content/news/2023-03-04-bevy-0.10/index.md
alice-i-cecile:
```suggestion
<div class="release-feature-authors">authors: @MrGVSV, @james7132 </div>
```

LiamGallagher737 (approved):

PR #567, opened 2023-03-04
[Merged by Bors] - 0.10 AsBindGroup storage buffers, Shapes, and enum Visibility

cart:


content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -739,6 +739,68 @@ query.iter().for_each(|mut component| {
...
+
+![plane](plane.png)
+
+[`Plane`]: https://docs.rs/bevy/0.10.0/bevy/prelude/shape/struct.Plane.html
+
+## Configurable Visibility Component
+
+<div class="release-feature-authors">authors: @ickk</div>
+
+The [`Visibility`] component controls whether or not an [`Entity`] should be rendered. **Bevy 0.10** changes it from being an `is_visible: bool` to an enum with multiple modes:
```
alice-i-cecile, line 764:
```suggestion
The [`Visibility`] component controls whether or not an [`Entity`] should be rendered. **Bevy 0.10** reworked the type definition: rather having a single `is_visible: bool` field, we now use an enum with an additional mode:
```

content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -739,6 +739,68 @@ query.iter().for_each(|mut component| {
...
+
+```rust
+pub enum Visibility {
+  Hidden,    // unconditionally hidden
+  Visible,   // unconditionally visible
+  Inherited, // inherit visibility from parent
+}
+```
+
+The enum is much easier to understand! And in previous Bevy versions, "inherited visibility" and "hidden" were essentially the only two options. Now entities can opt to be visible, even if their parent is hidden!
```
alice-i-cecile, line 774:
```suggestion
Much easier to understand! In previous Bevy versions, "inherited visibility" and "hidden" were essentially the only two options. Now entities can opt to be visible, even if their parent is hidden!
```

PR #569, opened 2023-03-04
[Merged by Bors] - 0.10 Blog Post: Iterating through a World's Entities

james7132:
Might be a bit too deep in the weeds, can simplify the wording a bit if need be.


content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -839,6 +839,32 @@ assert!(concrete_value.is::<MyStruct>());
 [`Entity`]: https://docs.rs/bevy/0.10.0/bevy/ecs/entity/struct.Entity.html
 [`World`]: https://docs.rs/bevy/0.10.0/bevy/ecs/world/struct.World.html
 
+## Iterating through a World's Entities
+
+<div class="release-feature-authors">authors: @james7132</div>
+
+In **Bevy 0.9**, `World::iter_entities` allows users to get an iterator over all of the entities in the `World` in `Entity` form. In **Bevy 0.10**, this has been changed to be an iterator over `EntityRef`, which gives full read-only access to all of the entity's components instead of just getting it's ID. It's new implementation should also be significantly faster than fetching the `EntityRef` by hand (though note that a `Query` will still be faster if you know the exact components you're looking for). This gives users free reign to arbitrarily read any entity data from the World, and may see use in scripting language integrations and reflection-heavy work flows.
```
theotherphil, line 846:
```suggestion
In **Bevy 0.9**, `World::iter_entities` allows users to get an iterator over all of the entities in the `World` in `Entity` form. In **Bevy 0.10**, this has been changed to be an iterator over `EntityRef`, which gives full read-only access to all of the entity's components instead of just getting its ID. Its new implementation should also be significantly faster than fetching the `EntityRef` by hand (though note that a `Query` will still be faster if you know the exact components you're looking for). This gives users free reign to arbitrarily read any entity data from the World, and may see use in scripting language integrations and reflection-heavy work flows.
```

alice-i-cecile (approved):
I like this one: I see users asking about this sort of workflow pretty regularly.

cart (approved):

PR #578, opened 2023-03-04
[Merged by Bors] - 0.10 camera output modes

cart:


content/news/2023-03-04-bevy-0.10/index.md
```diff
@@ -1061,6 +1061,36 @@ Bevy's [`Plane`] shape can now be subdivided any number of times.
...
+
+// Configure the camera to skip writing to the final output texture
+// This can save a pass when there are multiple cameras, and can be useful for
+// some post processing situations
+camera.output_mode = CameraOutputMode::Skip;
+```
+
+_Most_ single-camera and multi-camera setups will not need to touch this setting at all. But if you need it, it will be waiting for you!
+
+MSAA requires an extra intermediate "multisampled" texture, which gets resolved to the "actual" unsampled texture. In some corner case multi-camera setups rendering to the same texture, this can create weird / inconsistent results based on whether or not MSAA is enabled or disabled. We've added a new `Camera::msaa_writeback` `bool` field which (when enabled) will write the current state of the unsampled texture to the intermediate MSAA texture (if a previous camera has already rendered to the target on a given frame). This ensures that the state is consistent regardless of MSAA configuration. This defaults to true, so you only need to think about this if you have a multi-camera setup and you _don't_ want MSAA writeback.
```
LiamGallagher737, line 1089:
```suggestion
MSAA requires an extra intermediate "multisampled" texture, which gets resolved to the "actual" unsampled texture. In some corner cases multi-camera setups render to the same texture, this can create weird / inconsistent results based on whether or not MSAA is enabled or disabled. We've added a new `Camera::msaa_writeback` `bool` field which (when enabled) will write the current state of the unsampled texture to the intermediate MSAA texture (if a previous camera has already rendered to the target on a given frame). This ensures that the state is consistent regardless of MSAA configuration. This defaults to true, so you only need to think about this if you have a multi-camera setup and you _don't_ want MSAA writeback.
```

cart: This fundamentally changes the meaning of what i was saying / is no longer expressing. I think adding "that" before "rendering" in the original sentence adds the clarity you're looking for.

cart: (and changing "rendering" to "render")

cart: Done!

PR #583, opened 2023-03-05
[Merged by Bors] - Update Ruins Video and Screenshot

coreh:


cart (approved):
Oooh that looks a lot better!

bors r+

PR #585, opened 2023-03-06
Update the Bevy Book to 0.10

cart:


content/learn/book/getting-started/ecs/_index.md
```diff
@@ -28,7 +28,7 @@ Bevy ECS is Bevy's implementation of the ECS pattern. Unlike other Rust ECS impl
 
     ```rs
     fn print_position_system(query: Query<&Transform>) {
-        for transform in query.iter() {
+        for transform in &query {
```
IceSentry, line 31: I believe this was kept intentionally because `.iter()` is more beginner friendly.

cart: Definitely worth discussing. But I still think we should still use `&`, given that this is the style we "recommend". We chose to update all of our examples to use this pattern / we encourage it in bevy internals.

Its also "more ergonomic" which will make code examples look cleaner.

PR #593, opened 2023-03-07
[Merged by Bors] - Remove a non-breaking change from the 0.10 migration guide

JoJoJet:
The type `UnsafeWorldCellEntityRef` was introduced after 0.9 and renamed before 0.10, so this was not a breaking change.


devil-ira (approved):

james7132 (approved):
bors r+

JoJoJet:
Hmm... I think the even bigger problem is that https://github.com/bevyengine/bevy/pull/6404 seems to have no migration guide, despite removing public methods from `World`.

PR #596, opened 2023-03-07
Add smooth scrolling behavior

Winterwolf365:
so that clicking trough different sections in the site is smoother

### DIFF
--- a/sass/elements/_html.scss
+++ b/sass/elements/_html.scss
@@ -5,6 +5,7 @@ html {
     font-size: calc-rem($size-body-mobile);
     background-color: $color-grey-900;
     scroll-padding-top: var(--scroll-padding-top);
+    scroll-behavior: smooth;
 
     * {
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
My experience with this is that this is a fairly controversial change. Pinging @doup @IceSentry for opinions.

rparrett:
I am not personally a fan of this due to its behavior when navigating to a specific anchor tag from an **outside** link or a bookmark.

I just tested this again with the PR, and (MacOS, Desktop, `zola serve`)

- Chrome animates a scroll to the content after the page loads.
- Firefox does... something, and ends up with content incorrectly positioned on the screen.
- Safari seems to try to animate a scroll, but ends up popping to the wrong content.

IceSentry:
Yeah, same thing for me, I don't personally like it and it seems a bit inconsistent in how it behaves

PR #600, opened 2023-03-08
on_enter and on_exit are not a thing

hazelsparrow:
these were [vetoed ](https://github.com/bevyengine/bevy/pull/7790#issuecomment-1442156182) in the original PR but the migration guide was never updated.


alice-i-cecile (approved):

alice-i-cecile:
Thanks!

PR #601, opened 2023-03-09
Various changes

Keats:
Some changes while looking at the repo after https://github.com/getzola/zola/pull/2116/commits/cd90fd0756f7286ac717f3ca6cdaebd54140317d

The biggest one is the change in the community building where it was using page for members for no discernable reasons. `load_data` is made for exactly that kind of use case. By creating pages, you end up with stuff like that https://bevyengine.org/community/donate/cart/ but [Zola should warn on that.](https://github.com/getzola/zola/issues/2135)

You could skip 90% of the code by just storing the members in a single TOML file for each category as an array of tables though.

Note: some of my changes might need me to release Zola 0.17.2 before it works.


generate-community/src/bin/generate.rs
```diff
@@ -1,65 +1,118 @@
...
+        group.shuffle(&mut rng);
+        org_members.extend(group);
+    }
 
-    donate.write(Path::new(&content_dir), Path::new(&content_sub_dir), 0)?;
+    // TODO: collect files to copy to static dir here
+
+    let org_members: Vec<_> = org_members
+        .into_iter()
+        .map(MemberOutput::from_member)
```
mockersf, line 84:
```suggestion
        .into_iter()
        .filter(|member| member.sponsor.is_some())
        .map(MemberOutput::from_member)
```
The page this is used on needs only the members with a sponsor link

templates/donate.html
```diff
@@ -66,144 +67,8 @@
...
-                    </div>
-                </div>
-            </div>
-
-            {% endfor %}
-        </div>
-        {% endif %}
+        {% set members = load_data(path="org_members.json") %}
 
+        {{ members_macro::members_grid(members=members) }}
```
mockersf, line 72:
```suggestion
        {{ members_macro::members_grid(members=members | filter(attribute="sponsor")) }}
```
Only members with a sponsor link should be displayed here

templates/people.html
```diff
@@ -47,208 +48,67 @@
...
             </div>
 
-            {% endfor %}
-        </div>
+            <a class="anchor-target" id="org-members"></a>
+            <h2>
+                Members<a class="anchor-link" href="#org-members">#</a>
+            </h2>
+        {% else %}
+            {% set members = load_data(path="community_members.json") %}
```
mockersf, line 100:
```suggestion
            {% set members = load_data(path="community_members.json", required=false) %}
```

templates/people.html
```diff
@@ -47,208 +48,67 @@
...
+                <p>
+                    People in the Bevy community who are not part of the organization.
+                </p>
+                <p>
+                    Anyone can add themselves here! <a href="https://github.com/bevyengine/bevy-community">Submit a pull request</a>!
+                </p>
+            </div>
         {% endif %}
 
+        {{ members_macro::members_grid(members=members) }}
```
mockersf, line 111:
```suggestion
        {% if members %}{{ members_macro::members_grid(members=members) }}{% endif %}
```

templates/people.html
```diff
@@ -47,208 +48,67 @@
...
-                        {% if post.extra.profile_picture %}
-                            <div class="card-profile-picture">
-                                {% if post.extra.profile_picture is starting_with("https://") %}
-                                    <img src="{{ post.extra.profile_picture }}" />
-                                {% else %}
-                                    <img src="{{ get_url(path=post.extra.profile_picture) }}" />
-                                {% endif %}
-                            </div>
-                        {% endif %}    
+            {% set members = load_data(path="org_members.json") %}
```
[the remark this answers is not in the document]
Keats: It is required in practice though? This page will be empty if we don't generate the JSON files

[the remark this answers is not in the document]
mockersf: to build the full website yup, but for people who only want to work on part of it, like the book, they don't need it and we want to avoid the need for them to run the script

content/faq/index.md
cart, line 1: This existed as an "actual page" because we want to be able to link to `bevyengine.org/faq` as a nice sharable / short link to the github discussions query

templates/people.html
```diff
@@ -47,208 +48,67 @@
     <div class="assets-intro media-content">
         The Bevy community! If you want to see yourself here, <a href="https://github.com/bevyengine/bevy-community">submit a pull request</a>!
     </div>
+
     {% for subsection in section.subsections %}
         {% set section = get_section(path=subsection) %}
-
-        <a class="anchor-target" id="{{ section.title | slugify }}"></a>
```
cart, line 53: The sidebar links stop working with this change

cart (commented):
Very on board for this general direction. Thanks for putting this together!

Keats:
I'll address all the comments next week, thanks for the comments!

alice-i-cecile:
Many of these changes have been incorporated elsewhere. Thanks!

PR #602, opened 2023-03-10
POC: Quick-Start Guide and "New Book"

akappel:
## Description
This is a proof-of-concept to solicit feedback. After reviewing "bevy/rfcs#23", I decided to do the following:

1. Rename the current book to "Quick-Start Guide" that lives in "Learn", above "The Bevy Book".
 2. Create an empty "The Book" placeholder in "Learn" for the deep-dive stuff.

It was a bit hacky; I basically copied and renamed files referencing "book" to "quick-start" and made changes in JS, CSS, templates, etc. to account for it. If the feedback is positive, I can go back and review better ways to accomplish some of the changes so it's not so copy-pasta-ish.

Any feedback is welcome. In the short-term, if this sounds actionable, I can create a PR for just item 1. That way it can be filled out/restructured by @alice-i-cecile with tutorial content. Then we can have a separate PR to add the new-and-improved book back in when it's ready.

## Screenshots
Screen Shot 2023-03-10 at 10 58 26 AM
Screen Shot 2023-03-10 at 11 00 26 AM
Screen Shot 2023-03-10 at 11 00 41 AM


alice-i-cecile:
This PR should target the new-book branch: we don't want to expose the messy WIP in such a public way.

akappel:
@alice-i-cecile for sure, I changed the target but, woof, that's a lot of conflicts I was hoping to avoid. I'll go back to the drawing board and chew on this more with the `new-book` branch in mind.

akappel:
One question: given the screenshots, are you overall pleased with the division between "Quick-Start Guide" and "The Bevy Book"? I can try and target that as the goal with any changes to `new-book`. It seems like it was already going down that path.

alice-i-cecile:
Yep, I'm on board with that division, from what I can see. Basically the quick start guide should have "welcome to Bevy" + "here's how you install it", and very little else.

ickk:
> Basically the quick start guide should have "welcome to Bevy" + "here's how you install it", and very little else.
@alice-i-cecile 

I think there's value to having a one or two pager that shows a very simple "hello-world" or "draw triangle" app structure.

It serves two purposes: 
1. As a "template" to start a new project, without relying on extra junk like github or cargo-template.
2. It is effectively a one-pager for marketing to potential new users; "Here's what using bevy looks like! it's cool, give it a try!".   Reading `new-book`, cheatbook, examples or API-docs can be too heavy when you just want to get a first impression.

That said, It should be more terse than the current book - providing only cursory explanations so that readers can follow along & understand the *intent*, without necessarily explaining any details.

alice-i-cecile:
> > Basically the quick start guide should have "welcome to Bevy" + "here's how you install it", and very little else.
> > @alice-i-cecile
> 
> I think there's value to having a one or two pager that shows a very simple "hello-world" or "draw triangle" app structure.
> 
> It serves two purposes:
> 
> 1. As a "template" to start a new project, without relying on extra junk like github or cargo-template.
> 2. It is effectively a one-pager for marketing to potential new users; "Here's what using bevy looks like! it's cool, give it a try!".   Reading `new-book`, cheatbook, examples or API-docs can be too heavy when you just want to get a first impression.
> 
> That said, It should be more terse than the current book - providing only cursory explanations so that readers can follow along & understand the _intent_, without necessarily explaining any details.

I'm sold: I think that makes sense. I'd like to get something very simple and graphical: I don't think that the println examples do a good job showcasing "we're an engine" for this purpose.

Issue #603, opened 2023-03-11
0.9 -> 0.10 migration errors


djeedai:
(reopened from bevyengine/bevy#8033)

Here's a list of migration errors with no corresponding items in the migration guide:

- The stageless migration really doesn't explain the basics of how to migrate from _e.g._ `add_system_to_stage()` to ... (?) There's not even a single example.
- `ReportExecutionOrderAmbiguities` (likely: `bevy::ecs::schedule::ScheduleBuildSettings::ambiguity_detection`)
- `ModifiesWindows` (?)
- `StageLabel` (?)
- `MonitorSelection` moved to inner of `WindowPosition`, `Window::monitor` field deleted
- `App::add_system_set()` removed, how to replace?
- `ResMut<State<X>>` doesn't build anymore; this needs the new `States` trait which is not mentioned in the migration guide
- `State::current()` (likely `state.0`). (`NextState` is mentioned but it's not super clear either)

(TODO : didn't finish my update)

Weibye:
Additionally: How exactly should one setup systems that previously ran with a fixed timestep?

denormative:
It took half a day to upgrade the relatively simple roguelike I was working on and there's no way I could have done it without finding these notes online: https://zenn.dev/hyoi/scraps/5ea4bc5c9f6436 And I've still got a weird bug where the action N only triggers after I've triggered the N+1's action, but I'm guessing that's likely because of needing use NextState instead of just pushing the new state on the stack like I used to.

Anyway, I get how this new stateless system is more flexible and such, but I think the upgrade documentation was way to minimal to be of use to people not following the conversions in the PRs/issues.

djeedai:
> Anyway, I get how this new stateless system is more flexible and such, but I think the upgrade documentation was way to minimal to be of use to people not following the conversions in the PRs/issues.

Yes I agree. I _feel_ reading some descriptions the potential stageless has, and I'm pretty sure it's much better than stages, but in practice being the number one feature of this release, a complete rework of a fundamental system, and not having a single migration example in the migration guide is quite frustrating.

@alice-i-cecile I hope we can fix that guide quickly before too many people struggle. I can try to help but given I genuinely don't understand yet how to fix most of the items listed I won't be able to make a PR for now. Is anyone with ECS insight able to jump on this please?

alice-i-cecile:
I'll do my best to tackle this in the next couple of days.

denormative:
[...]

PR #604, opened 2023-03-11
Improve 0.9 -> 0.10 scheduling migration advice

alice-i-cecile:
Fixes most of #603.

### DIFF
--- a/content/learn/book/migration-guides/0.9-0.10/_index.md
+++ b/content/learn/book/migration-guides/0.9-0.10/_index.md
@@ -98,7 +98,8 @@ app.configure_sets(
         AfterUpdate::CommandFlush,
         CoreSet::PostUpdate,
     ).chain()
-).add_system(apply_system_buffers.in_set(AfterUpdate::CommandFlush));
+).add_system(apply_system_buffers.in_base_set(AfterUpdate::CommandFlush));
+
 ```
 
 #### Label types
--- a/content/learn/book/migration-guides/0.9-0.10/_index.md
+++ b/content/learn/book/migration-guides/0.9-0.10/_index.md
@@ -44,7 +44,7 @@ Stages had two key elements: they ran one after another, and they applied comman
 
 The former can be replaced by system sets (unless you need branching or looping scheduling logic, in which case you should use a schedule), and the latter can be controlled manually via `apply_system_buffers`.
 
-To migrate from Bevy's built-in stages, we've provided the [`CoreSet`](https://docs.rs/bevy/latest/bevy/app/enum.CoreSet.html), [`StartupSet`](https://docs.rs/bevy/latest/bevy/app/enum.StartupSet.html) and [`RenderSet`](https://docs.rs/bevy/latest/bevy/render/enum.RenderSet.html) system sets.
+To migrate from Bevy's built-in stages, we've provided the [`CoreSet`](https://docs.rs/bevy/0.10/bevy/app/enum.CoreSet.html), [`StartupSet`](https://docs.rs/bevy/0.10/bevy/app/enum.StartupSet.html) and [`RenderSet`](https://docs.rs/bevy/latest/0.10/render/enum.RenderSet.html) system sets.
 Command flushes have already been added to these, but if you have added custom stages you may need to add your own if you were relying on that behavior.
 
 Before:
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
I've omitted the other smaller fixes, as a) they're unrelated and b) I'm not immediately sure how to solve them. Please feel free to open more PRs to fix those.

rparrett:
Just in case we want to address the custom stages part, I think these are equivalent.

```rust
// Bevy 0.9
#[derive(Debug, Hash, PartialEq, Eq, Clone, StageLabel)]
pub struct AfterUpdate;
app.add_stage_after(CoreStage::Update, AfterUpdate, SystemStage::parallel());

// Bevy 0.10
#[derive(Debug, Hash, PartialEq, Eq, Clone, SystemSet)]
#[system_set(base)]
pub struct AfterUpdate;
app.configure_set(
    AppSet::AfterUpdate
        .after(CoreSet::UpdateFlush)
        .before(CoreSet::PostUpdate),
);
```

One thing I wish the migration guide addressed more explicitly are scenarios where the user was using various core bevy stages because they need a command flush between systems.

We are "mostly" just saying (via code snippets) that `StartupStage::PostUpdate` is now `StartupSet::PostUpdate`. But this situation is so common that I think it would help to spell out in code how to use `apply_system_buffers` in that situation.

Malax:
Awesome to see better docs! Thanks for working on these! FWIW: My biggest hurdle migrating was the difference between `in_set` and `in_base_set`. Adding a system to a set via `in_set` would still add it to a base set. That caused 

```
thread 'main' panicked at '`"Update"` and `"Render"` have a `before`-`after` relationship (which may be transitive) but share systems.'
```

errors that I wasn't able to understand without resorting to the amazing folks on Discord. (Thread: https://discord.com/channels/691052431525675048/1083571454404214885/1083571454404214885)

Would be cool to see this addressed in the migration notes, but I am not sure how. Given that I don't understand the new scheduling fully, I'm not really qualified to make suggestions :)

content/learn/book/migration-guides/0.9-0.10/_index.md
```diff
@@ -38,9 +38,244 @@ As a result, the Minimum Supported Rust Version (MSRV) is "the latest stable rel
...
+}
+
+app.configure_sets(
+    (
+        CoreSet::UpdateFlush,
+        AfterUpdate::Parallel,
+        AfterUpdate::CommandFlush,
+        CoreSet::PostUpdate,
+    ).chain()
+).add_system(apply_system_buffers.in_set(AfterUpdate::CommandFlush));
```
rparrett, line 101:
```suggestion
).add_system(apply_system_buffers.in_base_set(AfterUpdate::CommandFlush));

```

rparrett:
Labels *could* be removed for brevity from the custom stages section since they are covered just below, if that's a concern.

alice-i-cecile:
@rparrett I've adapted and added your suggestion, thank you!

@Malax excellent edge case, but I'm unsure how to address it here. This is definitely one of the rougher edges in the whole design IMO.

djeedai:
Excellent, this is miles better than before, thanks a lot @alice-i-cecile!

Ouch on the subtle difference between `OnEnter`/`OnExit` and `OnUpdate`, that's a big footgun :(

djeedai (approved):

content/learn/book/migration-guides/0.9-0.10/_index.md
```diff
@@ -38,9 +38,245 @@ As a result, the Minimum Supported Rust Version (MSRV) is "the latest stable rel
...
+If you were relying on multiple `FixedTimestep` run criteria with distinct periods, you should swap to using timers, via the `on_timer(MY_PERIOD)` or `on_fixed_timer(MY_PERIOD)` run conditions.
+
+Before:
+
+```rust
+app.add_system_set(
+    SystemSet::new().with_run_criteria(FixedTimestep::step(0.5)).with_system(update_pathfinding)
+).add_system_set(
+    SystemSet::new().with_run_criteria(FixedTimestep::step(0.1)).with_system(apply_damage_over_time)
+);
```
rparrett, line 212:
```suggestion
app.add_system_set(
    SystemSet::new()
        .with_run_criteria(FixedTimestep::step(0.5))
        .with_system(update_pathfinding),
)
.add_system_set(
    SystemSet::new()
        .with_run_criteria(FixedTimestep::step(0.1))
        .with_system(apply_damage_over_time),
);
```

Just formatting

content/learn/book/migration-guides/0.9-0.10/_index.md
```diff
@@ -38,9 +38,245 @@ As a result, the Minimum Supported Rust Version (MSRV) is "the latest stable rel
...
+            FixedTimestep::step(0.5)
+        )
+        .with_system(fixed_update),
+);
+```
+
+After:
+
+```rust
+app.insert_resource(FixedTime::new_from_secs(0.5))
```
Weibye, line 194:
```suggestion
// This will affect the update frequency of fixed time for your entire app
app.insert_resource(FixedTime::new_from_secs(0.5))

```
Not sure if needed, or the actual phrasing of it.

Weibye (approved):
These are good changes and now is in a state where I would be able to migrate my apps. Everything I struggled with has been answered.

content/learn/book/migration-guides/0.9-0.10/_index.md
```diff
@@ -38,9 +38,245 @@ As a result, the Minimum Supported Rust Version (MSRV) is "the latest stable rel
...
+
+State transitions are now applied via the `apply_state_transitions` exclusive system, a copy of which is added [`CoreSet::StateTransitions`](https://docs.rs/bevy/latest/bevy/app/enum.CoreSet.html) when you call `App::add_state`. You can add more copies as needed, specific to the state being applied.
+
+`OnEnter` and `OnExit` systems now live in schedules, run on the `World` via the `apply_state_transitions` system.
+By contrast, `OnUpdate` is now a system set which is nested within `CoreSet::Update`.
+
+Before:
+
+```rust
+#[derive(Debug, Clone, Copy Eq, PartialEq, Hash)]
```
rparrett, line 251:
```suggestion
#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]

```

alice-i-cecile:
Thanks a ton for the reviews, merging now :)

PR #607, opened 2023-03-15
Adds GitHub repo link to header

akappel:
## Description
I find myself sometimes just wanting to get to the Bevy repo directly from the website, instead of through "Learn/Official Examples" or a google search. I guess I could bookmark it, but I figured having a button linking directly to it would be helpful for others.

## Implementation
Adds a GitHub mark to the header bar, linking out to `bevyengine/bevy`. I set the height to what I found for `.logo` and it seems to scale well. I also added some hover brightness changing to align it with the UX of the buttons next to it. I'm a CSS novice, so if there's better ways to do it, please let me know!

Screen Shot 2023-03-15 at 9 12 00 AM


james7132 (approved):

cart:
The `@media` query that controls the menu collapse logic needs to be updated to include the width of the logo I think. The github logo gets slightly cut off right before the menu collapses.
image

cart:
Fixed!

cart (approved):
We should definitely discuss removing the Getting Started link, but that can happen later

PR #610, opened 2023-03-16
Add redirects for migration guides

rparrett:
We moved our migration guides and left a bunch of broken links in our wake.

Zola's `aliases` can apparently help us with this.

This adds aliases for the old urls: `learn/book/migration-guides/0.8-0.9/` to the new migration guide pages and section.


alice-i-cecile (approved):
This appears to be the correct usage. Shame we don't have a live deployed preview to test it with...

mockersf (approved):

PR #581, opened 2023-03-05
Fix cards responsive issues

doup:
Changes:

- Adds a new `link-card` [BEM _component_](https://getbem.com/naming/). This superseed `card` and it's for this image+text combination used in Learn/News/Community.
- Update Learn/News/Community to use `link-card`
- Change Learn to use two columns (following #612)
- Update Learn/Community to load their data from a TOML file
- Extract common "card" styling to a mixin and use it in `asset-card`/`link-card`/`card`.
- Tweak `item-grid` class and use it for 1/2/variable columns. Now it's used almost everywhere.
    - This class controls the spacing between elements, so `margin-bottom` from `card` is not needed anymore (or the `community-card` fix)

For the result, see: https://github.com/bevyengine/bevy-website/pull/422#issuecomment-1455071771

Fixes #420 

## Follow-up tasks

- Remove `card` class and related and instead add: `person-card` & `example-card` following BEM convention.

PR #612, opened 2023-03-16
Use two columns for Learn section

rparrett:
This just slaps `.community-card` and `.community-grid` on the learn section.

We can't really change the cards globally because this probably won't work well / is not what we want for the news section.

But if anyone has some suggestions for reorganizing the styles I'd be happy to execute them (ping @doup).

Desktop:
localhost_1111_learn
Mobile:
localhost_1111_learn (1)

### DIFF
--- a/templates/learn.html
+++ b/templates/learn.html
@@ -1,8 +1,8 @@
 {% extends "layouts/base.html" %}
 
 {% block content %}
-<div class="padded-content">
-  <a class="card" href="/learn/book/introduction">
+<div class="padded-content community-grid">
+  <a class="card community-card" href="/learn/book/introduction">
     <div class="card-image">
       <img src="/assets/book.svg" class="centered-card-image" alt="Bevy book" />
     </div>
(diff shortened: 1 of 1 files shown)


doup:
I'm looking into this for #581 and I think the best is to add some extra [_modifiers_](https://getbem.com/naming/) to [`.item-grid`](https://github.com/bevyengine/bevy-website/blob/main/sass/components/_item-grid.scss). I would do:

- default behavior would be 1 column grid, this would be used in News
- `.item-grid--two-cols`: for Learn and Community
- `.item-grid--multi-cols`: the current `.item-grid` behavior get's moved here. This would be used in assets/examples/people. The name can be improved ^_^

```scss
.item-grid {
    display: grid;
    grid-template-columns: 1fr;
    gap: 16px;

    &--two-cols {
        @media #{$bp-tablet-portrait-up} {
            grid-template-columns: 1fr 1fr;
        }
    }

    &--multi-cols {
        grid-template-columns: repeat(auto-fill, minmax(270px, 1fr));
    }
}
```

To use it, you need to add both the "block" and the "modifier", like:

```html
<div class="item-grid item-grid--two-cols">
```

doup:
**Remember**, old `item-grid` usages need will need also `item-grid--multi-cols`:
```diff
- <div class="item-grid">
+ <div class="item-grid item-grid--multi-cols">
```

rparrett:
Thanks for pointing me to that PR. I'd rather try to help push that forward, so I'll close this out.

PR #625, opened 2023-04-30
Add mastodon verification to index.html

Gingeh:
Adds a `<link rel="me"` to the site so that the official account can be [verified](https://docs.joinmastodon.org/user/profile/#verification).

Screenshot of @bevy@mastodon.social with a green checkmark next to bevyengine.org


cart:
Thanks for putting this together! But I'm closing this in favor of #627 as I think that approach is preferable.

PR #626, opened 2023-04-30
Guide users to upgrade Rust if necessary when getting started

daniel-butler:
- Address this issue: https://github.com/bevyengine/bevy/issues/8519


mockersf:
Bevy uses the [`rust-version` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) to specify its minimum supported Rust version: https://github.com/bevyengine/bevy/blob/latest/Cargo.toml#L13

PR #627, opened 2023-04-30
Add rel="me" support to community links and enable for Mastodon

cart:
Gives us the option to label links as "rel=me" and enables it for our mastodon link.


cart:
Alternative to #625. I think this is preferable because it avoids duplicating the mastodon link (and allows us to use this for future community links if needed). Mastodon supports verification from non-root links, so this should work.

Gingeh:
This would also require the website field on the mastodon profile to be changed to `https://bevyengine.org/community/`

cart:
Hmmm I wasn't aware of that and it does appear to be the case. Looks like your approach is the way to go!

PR #628, opened 2023-05-04
Removing outdated note about Windows support for dynamic_linking

zendril:
I tested the dynamic linking on Windows 11 and it works. I also heard from others in the Discord chats that it seems to be a resolved issue. Thus a small PR to update the docs to reflect that.

Here is the relevant discussion with Alice in the General room on Discord where they call out that they believe it is fixed: https://discord.com/channels/691052431525675048/691052431974465548/1103445630237540362

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -145,7 +145,7 @@ You might think to simply develop in release mode instead, but we recommend agai
 
 Bevy can be built just fine using default configuration on stable Rust. However for maximally fast iterative compiles, we recommend the following configuration:
 
-* **Enable Bevy's Dynamic Linking Feature**: This is the most impactful compilation time decrease! If `bevy` is a dependency, you can compile the binary with the "dynamic" feature flag (enables dynamic linking). Note that right now, this doesn't work on Windows.
+* **Enable Bevy's Dynamic Linking Feature**: This is the most impactful compilation time decrease! If `bevy` is a dependency, you can compile the binary with the "dynamic" feature flag (enables dynamic linking).
 
   ```sh
   cargo run --features bevy/dynamic_linking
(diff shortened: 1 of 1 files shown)


IceSentry:
Using the dynamic_linking feature still doesn't work for me on windows 10. 

I still get the issues mentioned in that thread https://github.com/bevyengine/bevy/issues/1110

I'm surprised it works for anyone on windows honestly. Did you test it with default features disabled?

IceSentry:
To clarify a bit, it does work when enabling the optimization like mentioned in the linked thread, but it seems wrong to not mention to users that they need to do additional configurations to get it to work.

zendril:
Perhaps that is why. I did do the optimizations. 

Ok, so it seems like it isn't 'broken' completely (which is the way the previous text read to me).. seems like it needs a bit more clarification that it will only work on windows when enabling the optimizations?

content/learn/book/getting-started/setup/_index.md
```diff
@@ -145,7 +145,7 @@ You might think to simply develop in release mode instead, but we recommend agai
 
 Bevy can be built just fine using default configuration on stable Rust. However for maximally fast iterative compiles, we recommend the following configuration:
 
-* **Enable Bevy's Dynamic Linking Feature**: This is the most impactful compilation time decrease! If `bevy` is a dependency, you can compile the binary with the "dynamic" feature flag (enables dynamic linking). Note that right now, this doesn't work on Windows.
+* **Enable Bevy's Dynamic Linking Feature**: This is the most impactful compilation time decrease! If `bevy` is a dependency, you can compile the binary with the "dynamic" feature flag (enables dynamic linking).
```
paul-hansen, line 148: Suggestion for how to add a note about performance optimizations being required on Windows.
Also fixes reference to the old flag name "dynamic" which is now "dynamic_linking"
```suggestion
* **Enable Bevy's Dynamic Linking Feature**: This is the most impactful compilation time decrease! If `bevy` is a dependency, you can compile the binary with the "dynamic_linking" feature flag (enables dynamic linking). **Important!** On Windows you _must_ also enable the [perfomance optimizations](#compile-with-performance-optimizations) or you will get a [`too many exported symbols`](https://github.com/bevyengine/bevy/issues/1110#issuecomment-1312926923) error.
```

paul-hansen (commented):
Related issue that can be closed when this is merged: https://github.com/bevyengine/bevy-website/issues/131

cart (approved):

PR #640, opened 2023-05-19
Fix example canvas growing on resize

### LINKED ISSUE #432: Example canvas now grows on each resize
Introduced in #429

As [mentioned in the PR adding this feature to Bevy](https://github.com/bevyengine/bevy/pull/4726/#issuecomment-1126418215), using CSS for parent sizing in a way that relies on the child size creates a feedback loop, resulting in the height growing on each resize.

devil-ira:
Prevent the `fit_canvas_to_parent` setting from growing the canvas by locking the aspect ratio to that of the initial resolution of the bevy `Window`.

Fixes #432.

### DIFF
--- a/templates/example-webgpu.html
+++ b/templates/example-webgpu.html
@@ -54,6 +54,22 @@
     import init from 'https://wasm-examples.pages.dev/{{ category.title }}/{{ page.extra.technical_name }}/wasm_example.js'
 
     const canvasEl = document.getElementById('bevy');
+
+    let once = false;
+    const observer_callback = (_mutations, _observer) => {
+        if (!once) {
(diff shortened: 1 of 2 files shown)


mockersf:
Could you also do this fix on https://github.com/bevyengine/bevy-website/blob/main/templates/example-webgpu.html?

cart (approved):

mockersf:
> Done. Might deduplicate this later since we have a templating system already.

Definitely! After the 0.11 release. The WebGL2 example should become closer to the WebGPU page, list every example with `wasm = true`, use screenshots, ... Right now it would be cheating as the screenshots used for WebGPU are from the 0.11. And the script is not yet available on the latest branch anyway.

PR #643, opened 2023-05-25
Filter out unavoidable winit control flow exception in WebGPU examples

marlyx:
During startup winit is using an exception as a part of its control flow. This is exception is "normal" and not a error but will end up in the browsers error console like this:

image

This PR adds a simple exception handler that will silently drop this specific exception and propagate all other exceptions.

### DIFF
--- a/templates/example-webgpu.html
+++ b/templates/example-webgpu.html
@@ -106,6 +106,16 @@
         })
     }
     window.bevyLoadingBarFetch = loadingBarFetch;
-    init();
+
+    {#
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

JMS55 (approved):

mockersf:
could you also port this fix to https://github.com/bevyengine/bevy_github_ci_template?

marlyx:
> could you also port this fix to https://github.com/bevyengine/bevy_github_ci_template?

Sure thing. Will make a PR for that today

PR #649, opened 2023-06-20
Properly handle plurals and code blocks

mattfbacon:
For a type called `App` that is plural, it should look like this:

> `App`s

and not

> `Apps`

because the latter makes it sound like a singular type named "Apps".

### DIFF
--- a/content/learn/book/contributing/docs/_index.md
+++ b/content/learn/book/contributing/docs/_index.md
@@ -42,9 +42,6 @@ We made an extension to the markdown syntax that makes linking to Rust API docs 
 * Short Type: {{rust_type(type="struct", crate="std" mod="collections", name="HashMap", no_mod=true)}}
 
     ```{{/*rust_type(type="struct" crate="std" mod="collections" name="HashMap" no_mod=true)*/}}```
-* Plural Type: {{rust_type(type="struct" crate="std" mod="collections" name="HashMap" no_mod=true plural=true)}}
-
-    ```{{/*rust_type(type="struct" crate="std" mod="collections" name="HashMap" no_mod=true, plural=true)*/}}```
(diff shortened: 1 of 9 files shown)


mattfbacon:
I don't think that CI failure has anything to do with my change

cart:
I think the resulting styling here is too awkward / degrades perceived quality too much:

image

* The spacing between `App` and the `s` is too separate (due to the padding)
* The `s` takes on the "link styling" which feels out of place.

I'm on board for a change like this, but I think it should attempt to re-color the `s` to something more natural (maybe as the "normal text" white) without adjusting spacing.

mattfbacon:
I wouldn't want to rely on color alone to indicate important semantic differences for accessibility reasons. Ideally, only actual code should be inside a `<code>` tag.

For my initial solution I valued simple markup—what you'd get if you did it in Markdown, like ``[`App`s](https://example.com/struct.App.html)``.

> The s takes on the "link styling" which feels out of place.

To address this specifically, would taking the S out of the `<a>` tag work?

cart:
At that point I think we should just remove `plural = true` entirely in favor of tacking manual `s` on things. Thats ok with me.

mattfbacon:
I made this change. However, it seems that either Zola or Tera is injecting an unwanted space (or maybe newline?) between the shortcode and the `s` that is directly after it in the markup. Not sure how to suppress this.

cart (approved):
Turns out the newline in the shortcode was resulting in the extra space :)

mattfbacon:
Wow, that should be considered a bug in whatever relies on that. A string of characters that doesn't end in a newline is not a line after all

mattfbacon:
For example Neovim fixes the line if you even touch the file.

PR #657, opened 2023-06-28
News: Release 0.11

cart:
## How this works

(This contains sections as of https://github.com/bevyengine/bevy/commit/10f5c9206847ae01b8dc833c2680562e7bd46664)

For the Bevy 0.11 release blog post, I'm opening the doors to other people writing blog post sections. Specifically, if you worked on a feature in a substantial way and are interested in presenting it, you can now ask to claim a section by leaving a comment in this PR. If you claim a section, submit a pull request to the `release-0.11.0` branch in this repo. For the next week, we will be filling in sections (the release target is Friday July 7th or Saturday July 8th). Please don't claim a section if you don't plan on completing it within that timeline. Also don't claim a section if you weren't an active participant in the design and implementation of the change (unless you are a Maintainer or SME).

I will claim any unclaimed sections.

Try to match the style of previous release blog posts as much as possible.

1. Show, don't tell. Don't bombard people with information. Avoid large walls of text _and_ large walls of code. Prefer the pattern "byte sized description of one thing" -> "example code/picture/video contextualizing that one thing" -> repeat. Take readers on a journey step by simple step.
2. Don't use up reader's "mental bandwidth" without good reason. We can't afford page-long descriptions of minor bug fixes. If it isn't a "headliner change", keep the description short and sweet. If a change is self describing, let it do that (ex: We now support this new mesh shape primitive ... this is what it looks like). If it is a "headliner change", still try to keep it reasonable. We always have a lot to cover.
3. In slight competition with point (2), don't omit interesting technical information when it is truly fun and engaging. A good chunk of our users are highly technical and enjoy learning how the sausage is made. Try to strike a balance between "terse and simple" and "nerdy details".
4. When relevant, briefly describe the problem being solved first, then describe the solution we chose. This contextualizes the change and gives the feature value and purpose.
5. When possible, provide visuals. They create interest / keep people hooked / break up the monotony.
6. Record images and videos at the default bevy resolution (1280x720)
7. Provide an accurate listing of authors that meaningfully contributed to the feature. Try to sort in order of "contribution scale". This is hard to define, but try to be fair. When in doubt, ask other contributors, SMEs, and/or maintainers.
8. Provide numbers and graphs where possible.  If something is faster, use numbers to back it up. We don't (yet) have automated graph generation in blog post style, so send data / info to me (@cart) if you want a graph made.

## Headliners

[...]

PR #658, opened 2023-06-28
0.11 Section:  Schedule First APIs and Nested System Tuples

cart:
Part of #657

### DIFF
--- a/content/news/2023-07-07-bevy-0.11/index.md
+++ b/content/news/2023-07-07-bevy-0.11/index.md
@@ -17,11 +17,146 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 * **Feature**: description
 
-## Feature
+## Schedule-First ECS APIs
 
-<div class="release-feature-authors">authors: @todo</div>
+<div class="release-feature-authors">authors: @cart</div>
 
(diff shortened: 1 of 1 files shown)


hymm (approved):

content/news/2023-07-07-bevy-0.11/index.md
```diff
@@ -17,11 +17,146 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+```rust
+app.add_systems(Update,
+    (
+        (a, b, c).chain(),
+        (d, e),
+    ).chain()
+)
+```
+
+This will run a, then b, then c, then run d _and_ e in parallel.
```
NiklasEi, line 159: I don't see the use of this example. Isn't the ordering the same without the first `chain`?

cart: Hmmm yeah the intent was to illustrate the value of nested chaining but I guess this isn't the best illustrator of that.

cart: I've made the nested chaining example a bit more interesting / something that can't be trivially flattened out into a single chain.

PR #659, opened 2023-06-29
0.11 section  improve shader imports

robtfm:
i don't think the images add much except for breaking up the wall of text, happy to remove them.

also you may want to redact some of the "future possibilities" as there is some opinion in there 🙂 

Part of #657


robtfm:
also just to note - i would not object at all if this went below the more visually interesting items

content/news/2023-07-07-bevy-0.11/index.md
```diff
@@ -17,11 +17,39 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+The future possibilities are more exciting. Using naga IR opens the door to a bunch of nice features that we hope to bring in future releases:
+
+* Automatic bind slot allocation will let plugins extend the core view bindgroup, which means self-contained plugins for features like lighting and shadow methods, common material properties, etc become viable. This will allow us to modularise the core pipelines to make growing the codebase - while keeping support for multiple targets - more sustainable
+
+* "Virtual" shader functions will allow user modifications to core functions (like lighting), and potentially lead to a template-style material system, where users can provide "hooks" that will be called at the right point in the pipeline
+
+* Language interop: mix and match glsl and wgsl, so bevy's pbr pipeline features could be accessed from your glsl material shader, or utils written for glsl could be used in wgsl code. we're hopeful that this can extend to spirv (and rust-gpu) as well
```
JMS55, line 50:
```suggestion
* Language interop: mix and match glsl and wgsl, so bevy's pbr pipeline features could be accessed from your glsl material shader, or utils written for glsl could be used in wgsl code. We're hopeful that this can extend to spirv (and rust-gpu) as well
```

JMS55 (changes requested):

cart (approved):

PR #672, opened 2023-07-05
Add "Copy to clipboard" functionality to web-examples

ickk:
Users often go to the examples for quick-templates or just to play around with the code. Since we can't embed a bevy-playground in the browser yet, the next best thing is to make it easy for users to copy the examples to their local editor.

https://github.com/bevyengine/bevy-website/assets/17050131/e7c188aa-35f8-40f5-a8a1-693dccb4956c


ickk:
We could possibly add this same functionality to other places like the release post and migration guides where it would be useful.

PR #678, opened 2023-07-08
0.11 Section: SSAO

JMS55:
Huge thanks to @IceSentry for the screenshots as my PC is currently out of commission


cart:
Just noticed this is targeting main :)

PR #689, opened 2023-07-09
Add image-compare widget

ickk:
Adds a side-by-side image-compare widget for usage in news posts.

Usage in a document should look as follows
```html
<main>
  <div class="image-compare" style="aspect-ratio: 16 / 9" data-title-a="Apples" data-title-b="Oranges">
    <img class="image-a" alt="Apples" src="apples.png" />
    <img class="image-b" alt="Oranges" src="oranges.png" />
  </div>
</main>
<script type="module">
  import { enable_image_compare } from '/components.js';
  document.addEventListener("DOMContentLoaded", function() {
    enable_image_compare();
  });
</script>
```

The `aspect-ratio` should be set to match the aspect ratio of the images, but it will default to 16/9.

---

I tested this on Windows with Firefox, Chrome, & Edge. Additional testing for Safari & others would be appreciated.

### DIFF
--- a/sass/components/_image_compare.scss
+++ b/sass/components/_image_compare.scss
@@ -1,5 +1,6 @@
 div.image-compare {
-  display: block;
+  display: flex;
+  align-items: center;
   position: relative;
   border-radius: 10px;
   width: 100%;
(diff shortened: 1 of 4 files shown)


ickk:
It wasn't exactly clear in which file I should include the js.. so I made a new `components.js` file. I think it probably makes sense to rationalise our js utilities into only 1 or 2 different places.

cart (approved):

PR #690, opened 2023-07-09
Use image-compare widget

ickk:
Use the image-compare widget where it's obvious to do so.

This depends on #689 and cherry-picks #680

The Tonemapping section needs new images that are more suitable for this
~~and in general it would be good to standardise the placement of and size of titles in the images.~~ 
Thanks to @nicopap we can use markup to add titles to the images


nicopap:
Doesn't work on my end

nicopap:
Yeah. I cherry picked the commit, now it works. Sorry.

content/news/2023-07-09-bevy-0.11/index.md
```diff
@@ -284,7 +284,11 @@ Since normal maps do not "move" the shaded areas, merely shade them
 differently, we get those awkward specular highlights. With parallax mapping,
 they are gone.
 
-![A montage of the three preceding images, contrasting each effect](parallax_mapping_compare.jpg)
+**Parallax & Normal Mapping vs Normal Mapping Only**
+<div class="image-compare" style="aspect-ratio: 16 / 9">
+  <img class="image-b" alt="normal_mapping" src="parallax_mapping_normals.jpg"/>
+  <img class="image-a" alt="parallax_and_normal_mapping" src="parallax_mapping_depth.jpg"/>
+</div>
```
nicopap, line 291: It would be nice to have the two swapped, so that we have a before/after.

nicopap (commented):
Hmm. A shame we don't have the slider for RCAS off/on and msaa/taa. Those would also benefit from it.

content/news/2023-07-09-bevy-0.11/index.md
```diff
@@ -284,7 +284,11 @@ Since normal maps do not "move" the shaded areas, merely shade them
 differently, we get those awkward specular highlights. With parallax mapping,
 they are gone.
 
-![A montage of the three preceding images, contrasting each effect](parallax_mapping_compare.jpg)
+**Parallax & Normal Mapping vs Normal Mapping Only**
+<div class="image-compare" style="aspect-ratio: 16 / 9">
+  <img class="image-b" alt="normal_mapping" src="parallax_mapping_normals.jpg"/>
+  <img class="image-a" alt="parallax_and_normal_mapping" src="parallax_mapping_depth.jpg"/>
```
nicopap, line 290:
```suggestion
**normal mapping only** vs **parallax and normal mapping**
<div class="image-compare" style="aspect-ratio: 16 / 9">
  <img class="image-a" alt="normal_mapping" src="parallax_mapping_normals.jpg"/>
  <img class="image-b" alt="parallax_and_normal_mapping" src="parallax_mapping_depth.jpg"/>
```

ickk:
> Hmm. A shame we don't have the slider for RCAS off/on and msaa/taa. Those would also benefit from it.

Hopefully we do a pass on the images before release to standardize them (good to have text/titles in the images themselves I think), and that would be a good opportunity to add RCAS off/on (I need separate images that are full-width).

For msaa/taa I'm not sure whether it's appropriate to compare them. I think no-aa/msaa & no-aa/taa could make more sense?

nicopap:
The section titled "Temporal Anti-Aliasing" has an image that is a side-by-side comparison of TAA and MSAA. It would just be an improvement to replace that specific image with a slider comparison.

Whether it's a good comparison or not is orthogonal. But a good reason to compare MSAA and TAA would be a way to point out the various artifacts and improvements each algo have (typically MSAA is still bad with specular highlights, while TAA fixes them).

cart (approved):
I'll re-render some of the images and tweak the CSS a bit

Issue #613, opened 2023-03-16
Add a nice 404 template


rparrett:
Our current 404s look like this:
<img width="881" alt="image" src="https://user-images.githubusercontent.com/200550/225736095-717914a9-ca85-4a0f-9d5d-36279804799d.png">

This leaves users in an awkward spot if they catch a dead link over to us.

We should add a 404 template that includes our header/navigation, perhaps some sort of sad bird and apologetic message. We just need to add a `404.html` to the `templates` directory.

Here's an example:
<img width="881" alt="image" src="https://user-images.githubusercontent.com/200550/225735979-09f33d18-3206-4c79-a318-fd6698e259c5.png">

mockersf:
our 404 page should be a game built with Bevy!

... but it can definitely be just a static page that looks better for now

iprasandeep:
Hi! Could I work on this issue?

rparrett:
Please feel free to open a PR.

iprasandeep:
Hi!, @rparrett Thanks, Could I add the same page as the design given above or do I need to design my own 404 page?

rparrett:
The screenshot above is just an example.

Our 404 page should consist of our header and branding. But that basic layout would work.

The important part is that the header is included so a user has an easy path to navigate to bevy pages that do exist.

PR #707, opened 2023-07-11
Created a 404 not found page

ResponseTime:
Issue "Add a nice 404 template #613"
Created a 404 not found page and made it responsive as well.
This is how it looks
image


ResponseTime:
Can I fix it now. I have understood what to change.

rparrett:
With regard to the template error due to the missing `current_path` variable, it seems that this is expected behavior in Zola, so we need to fix it in our template.

One way might be to add something like
```
{% set current_path = current_path | default(value="/") %}
```
to our base template.

cart:
Something also seems off with Ferris' right eye. Has some transparency on the top part of the pupil and the bottom part is "cut off".

mockersf:
done in #737

PR #710, opened 2023-07-12
Add a shortcode for the image compare slider

nicopap:
This streamlines the process of creating image comparisons.

The downside is that macro invocations in zola do not admit trailing commas in argument position, so it is error prone.


templates/shortcodes/compare_slider.html
```diff
@@ -0,0 +1,10 @@
+<p class="image-compare-instruction">Drag this image to compare</p>
+<div
+    {% if start_slider_at %}style="--slider-value: {{start_slider_at}}"{% endif %}
```
nicopap, line 3: I removed  the `aspect-ratio: 16/9` since it looks like it is defined in the `_image_compare.scss` now.

ickk: `aspect-ratio` is still required in the general case. The CSS has 16/9 as a default value, but if the images have a different aspect ratio then this must be declared for the widget to behave correctly (I explicitly set the aspect-ratio even on 16/9 images so that future uses have an example that doesn't elide this detail).

Perhaps there's a way for zola to automatically divine the aspect ratio of the images?

nicopap: Fairly certain it is possible, let me add this

content/news/2023-07-09-bevy-0.11/index.md
```diff
@@ -290,11 +292,13 @@ Since normal maps do not "move" the shaded areas, merely shade them
...
-  <img class="image-a" alt="Normal Mapping Only" src="parallax_mapping_normals.jpg">
-  <img class="image-b" alt="Parallax & Normal Mapping" src="parallax_mapping_depth.jpg">
-</div>
+{{ compare_slider(
+    before_title="Normals Only",
+    before_image="parallax_mapping_normals.jpg",
+    after_title="Parallax & Normal Mapping",
+    after_image="parallax_mapping_depth.jpg",
+    start_slider_at="40%"
+) }}
```
ickk, line 301: These titles should match the `data-title-*` attributes rather than the alt-text; cart shortened those attributes so that the image titles would visually fit on small displays (phones, &c).

ickk:
A little bit of bikeshedding: I feel like maybe "left"/"right" might be better names than "before"/"after". For example, in the case of a comparison between msaa & taa "before"/"after" doesn't necessarily fit semantically.

templates/shortcodes/compare_slider.html
ickk: yeah I've thought about this problem a bit.. I think you could wrap the whole thing in one more div with a maximum height (maybe specified as an optional argument?). Without the short-code, adding yet-another-element to the markup is not great, but I think it's reasonable for the short-code to generate a container element for this purpose. 

I think a div with a max-width or max-height should more or less just be a drop in solution. I haven't tested it though

content/news/2023-07-09-bevy-0.11/index.md
ickk: oh maybe short-codes can't include optional arguments? that's a shame..

sass/components/_image_compare.scss
```diff
@@ -7,7 +7,6 @@ div.image-compare {
   --slider-min: 7%;
   --slider-max: 93%;
   --slider-value: 50%;
-  aspect-ratio: 16 / 9;
```
ickk, line 6: I think it's fine (preferred?) to leave this as a fallback in the css. It will just get overridden by the explicit styling in the html

content/news/2023-07-09-bevy-0.11/index.md
cart: short codes support optional arguments. the `rust_type` short code makes use of this regularly

nicopap: I'm worried about the complexity of the macro. It might intimidate potential contributors into glazing over the shortcode. But it's a relatively minor concern compared to more detailed alts

sass/components/_image_compare.scss
nicopap: I can see the benefit in keeping it. Nice suggestion.

nicopap:
Looks like building the site breaks when using the `page.colocated_path` parameter to evaluate the image's aspect ratio.

This seems to be an error with the version of Zola we use (0.16.1 if I'm not mistaken) there are mentions of fix for `colocated_path` in the Zola changelog of more recent versions: <https://github.com/getzola/zola/blob/master/CHANGELOG.md#0172-2023-03-19>

BD103:
Zola has finally been updated to 0.18 (per #1018, thanks Trial!). Does this PR need any other work to get merged?

nicopap:
I'll check it out in the following days and open if it's ready.

BD103 (changes requested):
I have a few requests, mainly to document this macro so others can use it. Could you also update to main, since #1026 was merged and may affect CI?

If you don't have the time, I can also adopt this PR. Your choice :)

ickk:
closing in favor of #1144

PR #703, opened 2023-07-10
Improve image-compare widget

ickk:
- add UI handle & cursor affordances to visually indicate interactivity to end-user,
- add padding fix to image-titles,
- add slight blur-shadow to image-titles to improve readability,
- allow widget-user to override slider configuration.

generally improves code quality

before & after: 
image

PR #717, opened 2023-07-15
fix range input min/max handling on vivaldi

ickk:
This addresses the issue raised in https://github.com/bevyengine/bevy-website/pull/703#pullrequestreview-1522440159 regarding the Vivaldi browser behaving incorrectly. 

The fix is simply to manually clamp the value, which is effectively a no-op on other (well-behaved) browsers


ickk:
honestly it's a bit surprising to me that this does work for vivaldi.. I would have expected it to introduce discontinuities

cart (approved):

PR #718, opened 2023-07-15
Improve note on const styles

SpecificProtagonist:
bevyengine/bevy#9095


Selene-Amanita:
There's already a mention of this in

```rust
### [Add CSS Grid support to `bevy_ui`](https://github.com/bevyengine/bevy/pull/8026)
[...]
- It is not possible to use the struct literal update syntax in const time with `Style` anymore, since one of its field implements `Drop`, doing so would raise a "the destructor for this type cannot be evaluated in constants" error. Implement all the fields or don't use `Style` in a `const` variable. See [this issue](https://github.com/bevyengine/bevy/issues/9095).
```

I think your code should be added there, also the migration guide usually links merged PRs for each section not issues I think?

cart (approved):

PR #720, opened 2023-07-16
Examples update

mockersf:
This PR:
- adds a new workflow to trigger manually that will build all examples in CI and upload them in cloudflare
- update the WebGL2 example pages to be like the WebGPU one

It needs a new secret for the cloudflare token. @cart I can either give you a token or show you how to setup the account. That secret is not needed to merge this PR, it will only be used when a new version of the examples is built

Building the examples is not done at each time the website is built anymore, but on demand. That should be triggered once for every new Bevy version.

Examples are built in a matrix job by batches of 50 examples. There are three batches done currently, so we're OK until we reach 150 examples, at which point a new batch will need to be added.

All the steps building are using the examples from `latest`, but the tooling from `main`. This is to be able to use the changes from https://github.com/bevyengine/bevy/pull/9168. CI won't work before it's merged.

This is what this workflow execution looks like:
Screenshot 2023-07-16 at 21 21 24

Why running on macos? Because GitHub runners on macos have more resources than on windows/linux, and it's significantly faster (around x2)


cart:
Just added the token under the CLOUDFLARE_PAGES variable!

mockersf:
the workflow has been merged in #721 and executed: https://github.com/bevyengine/bevy-website/actions/runs/5586192733

I updated this PR to use the new domains, should be good to go 👍

cart (approved):

Issue #623, opened 2023-04-18
Integration of new Bevy Book


Captain-Of-Coit:
After talking with @alice-i-cecile for a bit on Discord, I ended up writing a small document (first just for myself) to highlight the current progress on the new book, and how we could move forward with it. Sharing it here to get it signed off on, so I could proceed to actually implement it.

**End goal**
Bootstrap new Official Bevy Book and enable documentation contributions.

**Approach**
Setup a basic structure for each section of the new book and add versions so Bevy Book 0.11.0 corresponds to Bevy Engine 0.11.0 Merge book to main (but hidden from public website) that can be iterated on until it feels complete enough to publish to the public, hopefully together with 0.11.0.

**Motivation**
New book is a bit stuck in limbo. Better documentation is in dire need, but contributions to it are currently missing, and hard to organize. A small focus on integrating the new book into `main` asap will make it easier to organize contributions and get it into a initial version to be published.

**Related RFCs**

- RFC#23 - https://github.com/bevyengine/rfcs/blob/main/rfcs/23-quick_start_book.md
	- Information + Structure for Bevy Book and new Quickstart

**Branches**

- bevyengine/bevy-website#new-book - https://github.com/bevyengine/bevy-website/compare/main...new-book - initial changes by @alice-i-cecile for new structure
- akappel/bevy-website#new-book-redux -> new-book - https://github.com/bevyengine/bevy-website/compare/main...akappel:bevy-website:new-book-redux - some additions + split of quick-start and book by @akappel
    - PR here: https://github.com/bevyengine/bevy-website/pull/602

**Missing**

- [ ] Hide book from public for now, only accessible by URL
- [ ] Versioned by Bevy version
- [ ] Basic scaffolded structure
- [ ] Disallow indexing robots for new Bevy book for now (added after @alice-i-cecile comment below)
- [ ] Add Book contribution guidelines outlining workflow described in this issue's discussion (added after @cart comments below)
- [ ] Merge to `main`

**Implementation**
- Hide book from public:
	- Rename "The Bevy Book" link to "Bevy Quickstart"
	- Change URL (with redirect)
		- From: https://bevyengine.org/learn/book/introduction/
		- To:      https://bevyengine.org/learn/quick-start
- Versioned by Bevy version
	- Book should be accessible URL:
		- https://bevyengine.org/learn/book/v0.11.0/introduction/
	- Each Bevy version gets its own version of the book by copy-paste
[...]

PR #725, opened 2023-08-04
Add public draft feature

NiseVoid:
An implementation of the draft feature, as described in https://github.com/bevyengine/bevy-website/issues/623#issuecomment-1518887997

Since the built-in `draft` feature acts as if pages don't exist at all, this uses a `public_draft` field under `[extra]`


cart:
I think we'll want to sort out this bit from the comment before committing to this workflow:
image

Being able to discover + navigate to draft pages (especially during development) is important for dev workflows.

cart:
Doing it statically is fine (and makes this comparable to `draft` flags for dev workflows), but I think being able to do it dynamically both locally and for deployed sites would be best (ex: via url flags and/or a toggle-able cookie)

templates/docs-section.html
```diff
@@ -76,6 +88,10 @@ <h1>
                 <span class="docs-page-subtitle">({{section.extra.subtitle}})</span>
             {% endif %}
         </h1>
+        <a class="example__github" href="https://github.com/bevyengine/bevy-website/tree/main/content/{{ section.relative_path }}">
```
cart: I don't think we need this, given that we already have this at the bottom of each page (which I think is a better place for it). I think putting it here (especially unconditionally) is distracting and unnecessary.

templates/macros/public_draft.html
```diff
@@ -0,0 +1,16 @@
+{% macro warning(current=true, section) %}
+    {% if section and section.extra.public_draft %}
+        {% set public_draft = section.extra.public_draft %}
+        {% set issue = public_draft / 1 %}
+        {% if current %}
+            <h1 class="warning">Looks like this page is still under development</h1>
```
cart: I think the title here should be crisper. Something like "Draft Page". Too much noise for an `h1`. The description provides enough detail on its own.

cart (approved):
Looks good to me! Made a few tweaks / resolved my comments

Issue #450, opened 2022-08-21
Add filtering to asset page


TimJentzsch:
It would be nice to have some basic filters, like filtering by Bevy version or license.

For example, one might want to filter out outdated tutorials/plugins or require a compatible license to one's project.

BlackPhlox:
I think this is related to #232

d-bucur:
PR is merged and deployed

PR #734, opened 2023-08-22
Asset page filtering

d-bucur:
WIP for Issue #450 

I created a search bar that can filter cards in JS. The problem is that the cards are in a `grid` and when setting `display:none` it will leave around empty spaces. The only way I see to do it properly would be to change the display of the cards to `flex-row` instead of `grid` (which I think makes more sense for this layout anyway), but that requires quite a bit of work making sure everything is responsive and that there are no regressions.

image

Result when searching for "MIT"

image

Any thoughts on this?

### DIFF
--- a/sass/pages/_assets.scss
+++ b/sass/pages/_assets.scss
@@ -16,4 +16,19 @@
     .item-grid {
         margin-bottom: 52px;
     }
+
+    .assets-search {
+        margin-bottom: 20px;
+
+        &__input {
+            @include card;
+            font-size: 1.2rem;
+            padding: $asset-card-padding;
+            width: 100%;
+
+            &:focus-visible {
+                outline: none;
+            }
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


d-bucur:
Nevermind, setting display:none on grid elements does work properly, I was just doing it on the wrong element. So no need for big changes. Will update with images soon.

d-bucur:
And it's ready. Terms are separated by space and put in AND. Feedback welcome.

[Screencast from 22-08-2023 22:44:09.webm](https://github.com/bevyengine/bevy-website/assets/6576777/fc219a92-2714-47cb-b726-f132d44e33de)

cart (approved):
This is a solid win. Well done!

cart:
(I removed the submodule that you added)

d-bucur:
Thanks! Didn't notice that submodule slip by

PR #740, opened 2023-09-04
Align the type of the first System's Query to that of the first Component

skindstrom:
# What
Change the type of the query parameter from `Query<&Transform>` to `Query<&Position>`. 

# Why
The first Component that's introduced in the lines above is `Position`, not `Transform`. Seeing the difference in type got me a bit confused. Aligning the types in hopes that 

# Ref 
- Initially used `Position` in #2aa8193
- Changed to `Transform` in #73


alice-i-cecile (approved):

alice-i-cecile:
I think this is much clearer. Thanks for changing this for us :)

mockersf (approved):

PR #742, opened 2023-09-09
Add Search on Examples page

ameknite:
## Objective

Since #734 we have a search bar on the assets page, I think it will be good to add it to the example page as well.


IceSentry (approved):
I just tested it and it seems to work as expected.

cart (approved):
Amazingly straightforward code reuse!

PR #746, opened 2023-09-25
Add Troubleshooting: Unable to debug dynamically linked Bevy application in VSCode

alexmadeathing:
This change adds a troubleshooting section including a workaround for https://github.com/bevyengine/bevy/issues/2589

There may be a better solution to this issue and we might have to get in touch with the relevant VSCode extension maintainers but for now, this is a functional workaround.

It seems to be an issue with both cppvsdbg and codelldb.

### DIFF
--- a/content/learn/book/troubleshooting/_index.md
+++ b/content/learn/book/troubleshooting/_index.md
@@ -43,7 +43,7 @@ For `cppvsdbg`:
 Or for `codelldb`:
 ```json
 "env": {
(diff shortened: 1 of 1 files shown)


mockersf:
those instructions seems to be specific for Windows and should mention it

alexmadeathing:
I've pushed an update to mention windows. I do wonder if this is ever an issue on Linux. I'll set up a VM to test later.

cart (approved):

PR #751, opened 2023-10-05
Use `--depth 1` in getting started clone command

AgustinRamiroDiaz:
Cloning the full repo with all its history can take up a long time/space to get started.

By cloning just one commit we reduce the "time to wow"


alice-i-cecile (approved):
Great choice!

AgustinRamiroDiaz:
@alice-i-cecile thanks a lot for your review!

On a second thought, I'm not sure this is the best since in the next step we encourage users to `git checkout` other branches, and that would cause users to hit errors like [this one](https://stackoverflow.com/questions/23387057/git-checkout-new-remote-branch-when-cloning-with-depth-1-option)

I'll close the PR for now, feel free to copy/reopen by re-framing the next steps

PR #757, opened 2023-10-13
0.12: Add spatial audio API ergonomics section

rparrett:
@hymm @mockersf please let me know if you don't like what I've done with the authors block or have any other feedback.

I tried to keep this as short as possible. IMO, any other info folks need is in the migration guide. Maybe another sentence explicitly mentioning that `SpatialListener` exists would be nice?

### DIFF
--- a/content/news/2023-10-21-bevy-0.12/index.md
+++ b/content/news/2023-10-21-bevy-0.12/index.md
@@ -23,7 +23,7 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 <div class="release-feature-authors">authors: @rparrett, @hymm, @mockersf</div>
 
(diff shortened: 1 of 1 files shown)


hymm:
I probably don't deserve authorship here. I feel like I just had average-ish review comments.

Maybe add a sentence saying what spatial audio is. i.e. stereo, but not 3-D.

rparrett:
> I probably don't deserve authorship here.

I disagree -- the PR was in a very poor state when you arrived and your comments revealed some major issues that needed to be fixed! If you would rather not be mentioned though for any reason please let me know.

hymm:
I'm fine with being mentioned than. Thanks.

alice-i-cecile (approved):

cart (approved):

PR #762, opened 2023-10-15
Remove the bit about DefaultPlugins being equivalent to...

SIGSTACKFAULT:
S33n noticed in the Bevy Discord [here](https://discord.com/channels/691052431525675048/691052431974465548/1162928142403514368) that the plugins page page still references `CorePlugin`. 

Decided to remove the section entirely, instead of updating it because:

- DefaultPlugins contains so many things the information isn't really helpful
- so that we don't have to worry about it going outdated again.
- so noobs don't think they should inline it if they want to customize one of the default plugins (e.g., WindowPlugin)


rparrett:
It would probably be safe to write a shorter blurb that links off to https://docs.rs/bevy/latest/bevy/struct.DefaultPlugins.html but I don't personally disagree with your other points.

rparrett (approved):

alice-i-cecile (approved):

PR #763, opened 2023-10-16
0.12 - Limit Background FPS

maniwani:
Added section on the new throttling that applies when an app has no focused windows.


alice-i-cecile (changes requested):
Seems to be missing a link to the PR?

maniwani:
> Seems to be missing a link to the PR?

Where do I put it?

cart (approved):

PR #769, opened 2023-10-18
0.12: One-Shot Systems

Trashtalk217:
Writing is hard. Let me know if there's anything I should add/remove. I'm not quite sure how much space I'm allowed to hog.

Some things I left out:
- removing registered systems
- run_system_once_with

### DIFF
--- a/content/news/2023-10-21-bevy-0.12/index.md
+++ b/content/news/2023-10-21-bevy-0.12/index.md
@@ -45,7 +45,7 @@ fn foo(world: &mut World) {
 }
 ```
 
-There are three simple steps to using one-shot systems: register a system, store its `SystemId`, and then use either exclusive world access or commands to run the corresponding system. 
+There are three simple steps to using one-shot systems: register a system, store its `SystemId`, and then use either exclusive world access or commands to run the corresponding system.
 
 A lot becomes possible with just that, however `SystemId`s really start showing their power, when they're wrapped into components.
 
@@ -65,8 +65,8 @@ fn call_all(query: Query<&Callback>, mut commands: Commands) {
 
 One-shot systems can then be attached to UI elements, like buttons, actions in an RPG, or any other entity. You might even feel inspired to implement the bevy scheduling graph with one-shot systems and aery (let us know how that goes, by the way).
 
-One-shot systems are very flexible. 
-They can be nested, so you can call `run_system` from within a one-shot system. 
+One-shot systems are very flexible.
+They can be nested, so you can call `run_system` from within a one-shot system.
 It's possible to have multiple instances of one system registered at a time, each with their own `Local` variables and cached system state.
 It also plays nice with asset-driven workflows: recording a mapping from a string to an identifier in a serialized callback is much nicer than trying to do so with Rust functions!
 
@@ -96,7 +96,7 @@ world.run_system_once(increment); // prints 1
 world.run_system_once(increment); // prints 2
 ```
 
-This is great for unit testing systems and queries, and it's both lower overhead and simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments, change detection, or `EventReader`s. This state isn't saved between two `run_system_once` calls, creating odd behavior. The `Local`s reset every run, while change detection will *always* detect data as added/changed. Be careful and you'll be alright.
+This is great for unit testing systems and queries, and it's both lower overhead and simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments, change detection, or `EventReader`s. This state isn't saved between two `run_system_once` calls, creating odd behavior. The `Local`s reset every run, while change detection will _always_ detect data as added/changed. Be careful and you'll be alright.
 
 ## <a name="what-s-next"></a>What's Next?
 
--- a/content/news/2023-10-21-bevy-0.12/index.md
+++ b/content/news/2023-10-21-bevy-0.12/index.md
@@ -44,7 +44,7 @@ let _ = world.run_system(id); // prints 1
 let _ = world.run_system(id); // prints 2
 ```
 
-There is now a new way to run systems. First you need to register a system within a world, by calling `world.register_system(your_system)`. The returned `SystemId` can then be used to run that system, either within an exclusive system via `world.run_system(id)` or via a command `commands.run_system(id)`.
+There are three simple steps to using one-shot systems: register a system, store its `SystemId` somewhere, and then use either exclusive world access or commands to run the system corresponding to that id. 
 
 One-shot systems are very flexible. For example, you can have multiple instances of one system registered at the same time. One-shot systems can be nested (although not recursive), and you can also wrap `SystemId`s into components, making it possible to treat registered systems like entities.
 
--- a/content/news/2023-10-21-bevy-0.12/index.md
+++ b/content/news/2023-10-21-bevy-0.12/index.md
@@ -24,7 +24,7 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 <div class="release-feature-authors">authors: @alice-i-cecile @pascualex, @Trashtalk217, @Zeenobit</div>
 
 Ordinarily, systems run once per frame, as part of a schedule.
-But what if you don't want to poll for them regularly?
+But this isn't always the right fit.
 Maybe you're responding to a very rare event like in a complex turn-based game, or simply don't want to clutter your schedule with a new system for every single button.
 One-shot systems flip that logic on its head, and provide you the ability to run arbitrary logic on demand, using the powerful and familiar system syntax.
 
--- a/content/news/2023-10-21-bevy-0.12/index.md
+++ b/content/news/2023-10-21-bevy-0.12/index.md
@@ -44,7 +44,7 @@ let _ = world.run_system(id); // prints 1
 let _ = world.run_system(id); // prints 2
 ```
 
-There is now a new way to run systems. First you need to register a system within a world, by calling `world.register_system(your_system)`. The `SystemId` that is returned can then be used to run that system, either within an exclusive system via `world.run_system(id)` or via a command `commands.run_system(id)`.
+There is now a new way to run systems. First you need to register a system within a world, by calling `world.register_system(your_system)`. The returned `SystemId` can then be used to run that system, either within an exclusive system via `world.run_system(id)` or via a command `commands.run_system(id)`.
 
 One-shot systems are very flexible. For example, you can have multiple instances of one system registered at the same time. One-shot systems can be nested (although not recursive), and you can also wrap `SystemId`s into components, making it possible to treat registered systems like entities.
 
@@ -86,7 +86,7 @@ world.run_system_once(increment); // prints 1
 world.run_system_once(increment); // prints 2
 ```
 
-This is great for unit testing systems and queries, and it's both lower overhead and simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments or change detection, which will *always* detect data as added/changed. This state isn't saved between two `run_system_once` calls, this may create some unexpected behavior. Be careful and you should be fine.
+This is great for unit testing systems and queries, and it's both lower overhead and simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments or change detection. This state isn't saved between two `run_system_once` calls, creating odd behavior. The `Local`'s reset every run, while change detection will *always* detect data as added/changed. Be careful and you'll be alright.
 
 ## <a name="what-s-next"></a>What's Next?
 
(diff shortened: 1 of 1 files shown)


content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,69 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+    println!("{}", counter.0);
+}
+
+let mut world = World::new();
+world.init_resource::<Counter>();
+world.run_system_once(increment); // prints 1
+world.run_system_once(increment); // prints 2
+```
+
+This is great for unit testing systems and queries, and it's a lot simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments or change detection (`Changed`, `Added`). This state isn't saved between two `run_system_once` calls, this may create some unexpected behavior. Be careful and you should be fine.
```
SIGSTACKFAULT, line 83: are changes and additions *always* detected in a one-shot, or *never* detected?

Trashtalk217: never

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,69 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+fn call_all(query: Query<&Callback>, mut commands: Commands) {
+    for callback in query.iter() {
+        commands.run_system(callback.0);
+    }
+}
+```
+
+Currently, there are some drawbacks and limitations. Most importantly, one-shot systems are not ran in parallel, so running slow systems like this can become a bottleneck. Additionally, exclusive systems, systems with `In` arguments, and systems that return values, are not supported.
+
+Registering and running systems is useful and allows for a great degree of control. However, if you __just_ want to run a system - sans registration - there's `RunSystemOnce`, a new trait that does just that.
```
alice-i-cecile, line 64:
```suggestion
Registering and running systems is useful and allows for a great degree of control. However, if you _just_ want to run a system - sans registration - there's `RunSystemOnce`, a new trait that does just that.
```

content/news/2023-10-21-bevy-0.12/index.md
alice-i-cecile:
```suggestion
This is great for unit testing systems and queries, and it's both lower overhead and simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments or change detection, which will *always* detect data as added/changed. This state isn't saved between two `run_system_once` calls, this may create some unexpected behavior. Be careful and you should be fine.
```

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,69 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 <div class="release-feature-authors">authors: @author</div>
 
+## One-Shot Systems
+
+<div class="release-feature-authors">authors: @alice-i-cecile @pascualex, @Trashtalk217, @Zeenobit</div>
+
+Previously, in order to run a system you'd have to add it to a schedule. But while schedules are very useful, not every system needs that sort of overhead. Hence the introduction of one-shot systems.
```
alice-i-cecile, line 26:
```suggestion
Ordinarily, systems run once per frame, as part of a schedule.
But what if you don't want to poll for them regularly?
Maybe you're responding to a very rare event like in a complex turn-based game, or simply don't want to clutter your schedule with a new system for every single button.
One-shot systems flip that logic on its head, and provide you the ability to run arbitrary logic on demand, using the powerful and familiar system syntax.
```

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,69 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+// calling all callbacks!
+fn call_all(query: Query<&Callback>, mut commands: Commands) {
+    for callback in query.iter() {
+        commands.run_system(callback.0);
+    }
+}
+```
+
+Currently, there are some drawbacks and limitations. Most importantly, one-shot systems are not ran in parallel, so running slow systems like this can become a bottleneck. Additionally, exclusive systems, systems with `In` arguments, and systems that return values, are not supported.
```
alice-i-cecile, line 62:
```suggestion
One-shot systems are not without their limitations.
Currently, exclusive systems and systems designed for system piping (with either an `In` parameter or a return type) can't be used at all.
Additionally, one-shot systems are always evaluated sequentially, rather than in parallel.
While this reduces both complexity and overhead, this can be meaningfully slower than using a schedule with a parallel executor if you want to run a large number of heavy systems at once.
```

content/news/2023-10-21-bevy-0.12/index.md
Trashtalk217: Good to know that rhetorical questions are allowed.

content/news/2023-10-21-bevy-0.12/index.md
Trashtalk217: It seems I was being overly dense

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,75 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+    }
+}
+```
+
+One-shot systems are not without their limitations.
+Currently, exclusive systems and systems designed for system piping (with either an `In` parameter or a return type) can't be used at all.
+Additionally, one-shot systems are always evaluated sequentially, rather than in parallel.
+While this reduces both complexity and overhead, this can be meaningfully slower than using a schedule with a parallel executor if you want to run a large number of heavy systems at once.
+
+Registering and running systems is useful and allows for a great degree of control. However, if you _just_ want to run a system - sans registration - there's `RunSystemOnce`, a new trait that does just that.
```
alice-i-cecile, line 70:
```suggestion
Registering and running systems offers improved performance and correctness: you can have multiple copies of a single system, each with their own `Local` values and cached system state. 
It also plays nice with asset-driven workflows: recording a mapping from a string to an identifier in a serialized callback is much nicer than trying to do so with Rust functions!

However, when you're just prototyping or writing a unit test, it can be a real hassle: who wants all that boilerplate and indirection!
When those caveats don't bother you, just use the `World::run_system_once` method.
```

Trashtalk217: This does feel like you're stepping on my toes a bit, but then again, it's just better.

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,75 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+    println!("{}", counter.0);
+}
+
+let mut world = World::new();
+world.init_resource::<Counter>();
+world.run_system_once(increment); // prints 1
+world.run_system_once(increment); // prints 2
+```
+
+This is great for unit testing systems and queries, and it's both lower overhead and simpler to use. However, there is one caveat. Some systems have state, either in the form of `Local` arguments or change detection, which will *always* detect data as added/changed. This state isn't saved between two `run_system_once` calls, this may create some unexpected behavior. Be careful and you should be fine.
```
alice-i-cecile:
```suggestion
When you use this, be mindful that system state does not carry over from run-to-run! If you're using a system that either has a `Local` argument (like the one lurking behind `EventReader`) or uses change detection, unexpected behavior may occur.
`EventReaders` won't track event cursors correctly, and data will always be detected as added/changed.
```

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,79 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+}
+
+let mut world = World::new();
+world.init_resource::<Counter>();
+let id = world.register_system(increment);
+let _ = world.run_system(id); // prints 1
+let _ = world.run_system(id); // prints 2
+```
+
+There is now a new way to run systems. First you need to register a system within a world, by calling `world.register_system(your_system)`. The returned `SystemId` can then be used to run that system, either within an exclusive system via `world.run_system(id)` or via a command `commands.run_system(id)`.
```
alice-i-cecile, line 47:
```suggestion
There are three simple steps to using one-shot systems: register a system, store its `SystemId` somewhere, and then use either exclusive world access or commands to run the system corresponding to that id. 
```

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,79 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+let mut world = World::new();
+world.init_resource::<Counter>();
+let id = world.register_system(increment);
+let _ = world.run_system(id); // prints 1
+let _ = world.run_system(id); // prints 2
+```
+
+There are three simple steps to using one-shot systems: register a system, store its `SystemId` somewhere, and then use either exclusive world access or commands to run the system corresponding to that id. 
+
+One-shot systems are very flexible. For example, you can have multiple instances of one system registered at the same time. One-shot systems can be nested (although not recursive), and you can also wrap `SystemId`s into components, making it possible to treat registered systems like entities.
```
alice-i-cecile, line 49: I don't think that this is the right place to talk about one-shot system flexibility. That feels like it belongs below, with limitations.

alice-i-cecile, line 49: The callback pattern is really nice to demonstrate: I think that makes sense as the focus of this paragraph. We could also get into how it's a powerful tool for other complex abstractions if you want.

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,79 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+```rust
+#[derive(Resource, Default, Debug)]
+struct Counter(u8);
+
+fn increment(mut counter: ResMut<Counter>) {
+    counter.0 += 1;
+    println!("{}", counter.0);
+}
+
+let mut world = World::new();
```
rparrett, line 40: It's probably safe to say that most users will be interacting with this feature via commands or exclusive systems, right? 

I think maybe the front-and-center "non-unit-test-scenario" examples should probably not create a world manually.

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,79 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 <div class="release-feature-authors">authors: @author</div>
 
+## One-Shot Systems
+
+<div class="release-feature-authors">authors: @alice-i-cecile @pascualex, @Trashtalk217, @Zeenobit</div>
+
+Ordinarily, systems run once per frame, as part of a schedule.
+But what if you don't want to poll for them regularly?
```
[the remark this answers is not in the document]
alice-i-cecile: Hmm yeah, wording isn't clear. I'm trying to hint at there being two different ways to respond to events: push-based and pull-based.

In a push-based idiom, you read the event stream, and then dispatch work based on the events read. This is one-shot systems.

In a pull-based idiom, you have a bunch of listeners that are constantly checking (polling) for changes that they care about. This is ordinary scheduled systems.

Unfortunately that's a terrible explanation for this section of the post.

content/news/2023-10-21-bevy-0.12/index.md
alice-i-cecile: Agreed: I think an exclusive system is probably good.

content/news/2023-10-21-bevy-0.12/index.md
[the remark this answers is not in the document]
Trashtalk217: I was originally careful with wording it like 'being push-based' over 'being pull-based'. Since whenever you talk about: I want to run a system based on a rare event, the obvious thing to use is Events. And that would start a whole discussion on whether one-shot systems should be used rather then Events, and I didn't want to get into that.

alice-i-cecile (approved):

Trashtalk217:
Should I add a link to aery?

alice-i-cecile:
> Should I add a link to aery?

Nah, I don't think this is the spot for it.

rparrett:
It seems slightly odd to mention aery but not link it? Or maybe throw it in backticks so it doesn't look like a random out-of-place word? I don't think that readers will know what it is.

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -19,6 +19,85 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+// calling all callbacks!
+fn call_all(query: Query<&Callback>, mut commands: Commands) {
+    for callback in query.iter() {
+        commands.run_system(callback.0);
+    }
+}
+```
+
+One-shot systems can then be attached to UI elements, like buttons, actions in an RPG, or any other entity. You might even feel inspired to implement the bevy scheduling graph with one-shot systems and aery (let us know how that goes, by the way).
```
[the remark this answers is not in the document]
alice-i-cecile: I was totally wrong, upon rereading this should have a link.

cart (approved):

PR #771, opened 2023-10-20
Simplify generate_wasm_examples.sh

SIGSTACKFAULT:
the script was using an unnecessarily convoluted way to clone the bevy repo so i switched it to just `git clone --depth 1`. also added some comments.

(can squash 3aa2 and 3799 if wanted)

### DIFF
--- a/generate-wasm-examples/clone_bevy.sh
+++ b/generate-wasm-examples/clone_bevy.sh
@@ -1,6 +1,0 @@
-#!/bin/sh
-
-git init bevy
-cd bevy
-git remote add origin https://github.com/bevyengine/bevy
-git pull --depth=1 origin latest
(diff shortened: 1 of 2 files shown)


rparrett:
This isn't quite doing the same thing.

`clone_bevy.sh` is getting the `latest` branch. The next parts are retrieving `tools` from the `main` branch.

This is probably still necessary until `0.12` is released.

`clone_bevy.sh` also seems to be used in a few workflows.

SIGSTACKFAULT:
I see. this is my first time with github actions (fluent in gitlab actions) and couldn't tell if `act` was working because the upload jobs at the end always fail locally.

Turns out there's `git clone -b <branch>` which can be used instead of the previous complicated system.

should i start a new MR or just keep adding to this one and squash all my mistakes away?

rparrett:
Feel free to continue here.

SIGSTACKFAULT:
lost the adhd urge so i prolly won't finish this

PR #779, opened 2023-10-27
Check all PRs for migration guides

IceSentry:
Instead of looking for PRs with the `C-Breaking-Change` label, look for all PRs that have a Migration Guide section. Still check for PRs with the label to detect if there's any missing migration guide.

### DIFF
--- a/generate-release/src/migration_guide.rs
+++ b/generate-release/src/migration_guide.rs
@@ -37,13 +37,28 @@ As a result, the Minimum Supported Rust Version (MSRV) is "the latest stable rel
 
     let mut areas = BTreeMap::<String, Vec<(String, GithubIssuesResponse)>>::new();
 
-    let merged_breaking_prs = get_merged_prs(client, from, to, Some("C-Breaking-Change"))?;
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

generate-release/src/migration_guide.rs
```diff
@@ -85,10 +102,7 @@ As a result, the Minimum Supported Rust Version (MSRV) is "the latest stable rel
     }
     writeln!(&mut output, "</div>")?;
 
-    println!(
-        "\nFound {} breaking PRs merged by bors",
-        merged_breaking_prs.len()
-    );
+    println!("\nFound {} breaking PRs merged by bors", count);
```
alice-i-cecile, line 105:
```suggestion
    println!("\nFound {} breaking PRs merged", count);

```

IceSentry: Oh, nice catch, I didn't even notice, that's a flash from the past 😅

james7132 (approved):

PR #754, opened 2023-10-13
News: Release 0.12

cart:
## How this works

(This contains sections as of <https://github.com/bevyengine/bevy/commit/a042924a6ee520a0b0dc72bc287a6f5f5f305002> ... still plenty of open things to merge!)

For the Bevy 0.12 release blog post, other people (besides me) can write blog post sections. Specifically, if you worked on a feature in a substantial way and are interested in presenting it, you can now ask to claim a section by leaving a comment in this PR. If you claim a section, submit a pull request to the `release-0.12.0` branch in this repo. For the next week, we will be filling in sections (the release target is Friday October 20th or Saturday October 21st). Please don't claim a section if you don't plan on completing it within that timeline. Also don't claim a section if you weren't an active participant in the design and implementation of the change (unless you are a Maintainer or SME).

I will claim any unclaimed sections.

Try to match the style of previous release blog posts as much as possible.

1. Show, don't tell. Don't bombard people with information. Avoid large walls of text _and_ large walls of code. Prefer the pattern "byte sized description of one thing" -> "example code/picture/video contextualizing that one thing" -> repeat. Take readers on a journey step by simple step.
2. Don't use up reader's "mental bandwidth" without good reason. We can't afford page-long descriptions of minor bug fixes. If it isn't a "headliner change", keep the description short and sweet. If a change is self describing, let it do that (ex: We now support this new mesh shape primitive ... this is what it looks like). If it is a "headliner change", still try to keep it reasonable. We always have a lot to cover.
3. In slight competition with point (2), don't omit interesting technical information when it is truly fun and engaging. A good chunk of our users are highly technical and enjoy learning how the sausage is made. Try to strike a balance between "terse and simple" and "nerdy details".
4. When relevant, briefly describe the problem being solved first, then describe the solution we chose. This contextualizes the change and gives the feature value and purpose.
5. When possible, provide visuals. They create interest / keep people hooked / break up the monotony.
6. Record images and videos at the default bevy resolution (1280x720)
7. Provide an accurate listing of authors that meaningfully contributed to the feature. Try to sort in order of "contribution scale". This is hard to define, but try to be fair. When in doubt, ask other contributors, SMEs, and/or maintainers.
8. Provide numbers and graphs where possible.  If something is faster, use numbers to back it up. We don't (yet) have automated graph generation in blog post style, so send data / info to me (@cart) if you want a graph made.

Use [Bevy's Dev Docs](https://dev-docs.bevyengine.orgl) to create links to symbols in your code:

```md
[...]

PR #780, opened 2023-10-27
0.12 PCF Shadow Filtering

cart:
Part of #754


content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,45 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+![no pcf high resolution](no_pcf_high.png)
+
+Looking better! However this still isn't a perfect solution. Large shadowmaps aren't feasible on all hardware. And even if they are, you can still encounter this issue if you place an object in the wrong place, or point your light in the wrong direction. You can use Bevy's [Cascaded Shadow Maps](/news/bevy-0-10/#cascaded-shadow-maps) (which are enabled by default) to cover a larger area, with higher detail close to the camera and less detail farther away. However even under these conditions, you will still probably encounter these aliasing issues.
+
+**Bevy 0.12** introduces **PCF Shadow Filtering**, which is a popular technique that takes multiple samples from the shadow map and compares with an interpolated mesh surface depth projected into the frame of reference of the light. It then calculates the percentage of samples in the depth buffer that are closer to the light than the mesh surface. In short, this creates a "blur" effect that improves shadow quality, which is especially evident when a given shadow doesn't have enough "shadow map detail".
```
rparrett, line 34:
```suggestion
**Bevy 0.12** introduces **PCF Shadow Filtering**, which is a popular technique that takes multiple samples from the shadow map and compares with an interpolated mesh surface depth-projected into the frame of reference of the light. It then calculates the percentage of samples in the depth buffer that are closer to the light than the mesh surface. In short, this creates a "blur" effect that improves shadow quality, which is especially evident when a given shadow doesn't have enough "shadow map detail".
```

Is this how I'm supposed to read this series of words? :)

rparrett, line 34: Should we define this acronym here (or elsewhere)?

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,45 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+![no pcf](no_pcf.png)
+
+Those "jagged lines" in the shadow are the result of the shadow map being "too small" to accurately represent the shadow from this perspective. The shadow map above is stored in a 512x512 texture.
+
+One "solution" is to bump up the resolution. Here is what it looks like with a 4096x4096 shadow map.
+
+![no pcf high resolution](no_pcf_high.png)
+
+Looking better! However this still isn't a perfect solution. Large shadowmaps aren't feasible on all hardware. And even if they are, you can still encounter this issue if you place an object in the wrong place, or point your light in the wrong direction. You can use Bevy's [Cascaded Shadow Maps](/news/bevy-0-10/#cascaded-shadow-maps) (which are enabled by default) to cover a larger area, with higher detail close to the camera and less detail farther away. However even under these conditions, you will still probably encounter these aliasing issues.
```
JMS55, line 32: Large shadow maps take a lot of memory and more importantly rendering time. I think saying hardware capabilities is misleading, it's more about performance.

JMS55:
512x512 also looks a bit bad, even with PCF. Maybe you want to double it.

cart:
> 512x512 also looks a bit bad, even with PCF. Maybe you want to double it.

Hmmm yeah that was kind of the point. Show an "egregious" case. I was also worried that 1024 would benefit too much from "linear filtering" when sampling the scaled image in web browsers, making the effect less prominent.

Imo 1024 isn't _significantly_ better. And the difference is less stark (left is 512 right is 1024)
image

cart:
Resolved comments. I also explicitly called out that we don't use 512x512 shadowmaps by default / it was chosen to illustrate the point.

PR #781, opened 2023-10-29
update info on mold availability on Windows and MacOS.

miketwenty1:
mold is now available for MacOS as *sold*.
mold is planning to be released via *sold* as well for windows, but is not available yet, updated new github issue to track progress.

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -176,8 +176,8 @@ Bevy can be built just fine using default configuration on stable Rust. However 
   * **Ubuntu**: `sudo apt-get install mold`
   * **Fedora**: `sudo dnf install mold`
   * **Arch**: `sudo pacman -S mold`
-  * **Windows**: mold does not yet support Windows. [See this tracking issue](https://github.com/rui314/mold/issues/190) for more information.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

mockersf:
sold seems to not be updated anymore, and it seems it won't happen for windows, see comment https://github.com/rui314/mold/issues/1069#issuecomment-1653436823

miketwenty1:
Made another push @alice-i-cecile to pass linter check, and shortened text to be more concise.

miketwenty1:
> sold seems to not be updated anymore, and it seems it won't happen for windows, see comment [rui314/mold#1069 (comment)](https://github.com/rui314/mold/issues/1069#issuecomment-1653436823)

Ohh you're right looks like this isn't happening anymore. What you think? just update text to say "currently not planned for support"?

alice-i-cecile:
> What you think? just update text to say "currently not planned for support"?

Yeah, that sounds great.

alice-i-cecile (approved):

mockersf (approved):

mockersf:
unrelated to mold (or lld), but I think Apple shipped a new linker with latest Xcode version making those not very useful anymore. that should probably be (checked and) mentioned there too if you want a followup PR

PR #785, opened 2023-10-30
0.12 Deferred Lighting

cart:
Part of #754


content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,48 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+Bevy's renderer has historically been a "forward renderer". More specifically, it is a [Clustered Forward / Forward+](/news/bevy-0-7/#unlimited-point-lights) renderer, which means we break the view frustum up into clusters and assign lights to those clusters, allowing us to render many more lights than a traditional forward renderer.
+
+However, as Bevy has grown, it has slowly moved into "hybrid renderer" territory. In previous releases, we added a [Depth and Normal Prepass](/news/bevy-0-10/#depth-and-normal-prepass) to enable [TAA](/news/bevy-0-11/#temporal-anti-aliasing), [SSAO](/news/bevy-0-11/#screen-space-ambient-occlusion), and [Alpha Texture Shadow Maps](/news/bevy-0-10/#shadow-mapping-using-prepass-shaders). We also added a Motion Vector Prepass to enable TAA.
+
+In **Bevy 0.12** we added support for Deferred Rendering (building on the existing prepass work). Each material can choose whether it will go through the forward or deferred path, and this can be configured per-material-instance.
+
+For the PBR [`StandardMaterial`], the deferred prepass packs PBR information into the Gbuffer, which looks like this:
```
IceSentry, line 37: I'd add a note that this is completely optional and the `StandardMaterial` still uses forward by default right now.

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,48 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+<div class="release-feature-authors">authors: @DGriffin91</div>
+
+The two most popular "rendering styles" are:
+
+* **Forward Rendering**: do all material/lighting calculations in a single render pass
+  * **Pros**: Simpler to work with. Works on / performs better on more hardware. Supports MSAA. Handles transparency nicely.
+  * **Cons**: Lighting is more expensive / fewer lights supported in a scene, some rendering effects are impossible (or harder) without a prepass
+* **Deferred Rendering**: do one or more pre-passes that collect relevant information about a scene, then do material/lighting calculations in _screen space_ in a final pass after that.
+  * **Pros**: Enables some rendering effects that are not possible in forward rendering. This is especially important for GI techniques. Can support more lights in a scene
+  * **Cons**: More complicated to work with. Requires doing prepasses, which can be more expensive than an equivalent forward renderer in some situations (although the reverse can also be true), doesn't support MSAA, transparency is harder / less straightforward.
```
IceSentry, line 29: As far as I know transparency is generally implemented with a forward pass which is why it's not just forward vs deferred

DGriffin91: Agreed. Engines that primarily use deferred still use forward for alpha blend and effects.

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,48 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+
+<div class="release-feature-authors">authors: @DGriffin91</div>
+
+The two most popular "rendering styles" are:
+
+* **Forward Rendering**: do all material/lighting calculations in a single render pass
+  * **Pros**: Simpler to work with. Works on / performs better on more hardware. Supports MSAA. Handles transparency nicely.
+  * **Cons**: Lighting is more expensive / fewer lights supported in a scene, some rendering effects are impossible (or harder) without a prepass
+* **Deferred Rendering**: do one or more pre-passes that collect relevant information about a scene, then do material/lighting calculations in _screen space_ in a final pass after that.
+  * **Pros**: Enables some rendering effects that are not possible in forward rendering. This is especially important for GI techniques. Can support more lights in a scene
```
DGriffin91, line 28: Regarding the number of lights, this can be true, though we use the same clustered impl for deferred (and clustered deferred rendering is somewhat typical). I would expect that lots of lights are probably still faster with deferred since it's all just done in the one pass and there's no overdraw. I don't have an issue with what's said here, but I guess it's just more that the performance with lots of lights would be better rather than a limitation on how many can be used. (Sorry I'm mostly just thinking out loud here)

cart: > but I guess it's just more that the performance with lots of lights would be better rather than a limitation on how many can be used. (Sorry I'm mostly just thinking out loud here) 

Yeah I see these as two sides of the same coin. Ex: clustered doesn't inherently increase the number of lights _supported_ (provided you are using storage buffers, which is a separate concern), it just reduces the number of lights being processed per fragment, increasing the number of lights _you can reasonably use_. Same story for deferred lighting: it reduces the amount of work done by only lighting visible fragments, increasing the number of lights you can reasonably use.

I think "support more lights in a scene" is a reasonably terse way to express that to a layperson, but I'm open to alternatives!

DGriffin91: Makes sense to me, I like the current terseness  :)

content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,48 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 <!-- more -->
 
+## Deferred Rendering
+
+<div class="release-feature-authors">authors: @DGriffin91</div>
+
+The two most popular "rendering styles" are:
+
+* **Forward Rendering**: do all material/lighting calculations in a single render pass
```
JMS55, line 24: Forward rendering vs deferred:
* Forward can render transparencies, deferred cannot, also only forward can do MSAA (forward wins)
* Deferred saves on lighting costs - you only shade what's visible, where with forward you may be shading pixels that will be occluded later on (deferred wins)
* Deferred involves a lot of texture bandwidth and multiple render targets, which is prohibitive on mobile (forward wins)

In general I would say forward + small prepass (depth, +normals if using SSAO) is the way to go, but deferred opens up specific niches such as real-time GI (screenspace or otherwise) that needs a full G-buffer, and saves on costs when your material lighting shader is significantly expensive or would destroy register pressure in a forward shader.

"Number of lights" isn't that big a difference, you can do clustering in either one.

Visibility buffers are another possibility to further reduce overhead, as it _fully_ decouples "is this pixel visible" from _any_ sort of lighting, including material evaluation such as writing out a gbuffer pixel. Adding occlusion culling further cuts down on a lot of over-shading issues. I intend to experiment with this in my meshlet work for 0.13.

For now, forward is the better default, and users should experiment with deferred only for specific situations where it makes sense.

cart: > Forward can render transparencies, deferred cannot

I tried to avoid absolute statements here. I'm pretty sure order-independent transparency works for deferred?

content/news/2023-10-21-bevy-0.12/index.md
cart: I tried to avoid absolute statements here. I'm pretty sure order-independent transparency works for deferred?

content/news/2023-10-21-bevy-0.12/index.md
cart: Done!

cart:
Just addressed most comments

content/news/2023-10-21-bevy-0.12/index.md
JMS55: True, I don't know anything about OIT. Best to err on the side of caution.

cart:
Merging now in the interest of doing global changes / reordering in the 0.12 post. We've discussed adding more images for each gbuffer component, but we can do that in a followup.

PR #787, opened 2023-10-31
Foresight Update

aevyrie:
- Updates the logo and name.
- Moved logo to top of platinum list, as foresight's sponsor amount is double the next highest according to: https://bevyengine.org/sponsorship-pledge/


templates/index.html
```diff
@@ -275,11 +274,11 @@ <h3 class="sponsors__amount">$100 / month</h3>
...
-            <div>
-                Futurewei, Legion Labs, Striked, Metabuild 
-            </div>
+        <h2 class="past-sponsors-title">Past Sponsors</h2>
+        <div>
+            Futurewei, Legion Labs, Striked, Metabuild
         </div>
     </div>
 </div>
+</div>
```
rparrett, line 283: Indentation fix above is good, but I'm pretty sure this is an extra `</div>`.

rparrett:
Does the https://bevyengine.org/sponsorship-pledge/ page also need an edit? Foresight is mentioned there.

templates/index.html
aevyrie: oops, I think my auto-format extension got a bit excited. I'll take a closer look at the diff and try to minimize change.

aevyrie: Well this is spooky. When I compare my branch locally, this diff does not show up, only the foresight section is changed.

aevyrie: So, the code looks different in the deff viewer, but the indent is not highlighted... at all. That is concerning. Not sure what is going on there, but I removed the extra `div`.

aevyrie:
> Does the [sponsorship pledge](https://bevyengine.org/sponsorship-pledge/) page also need an edit? Foresight is mentioned there.

Ah, I missed some of the other mentions down the page. Thanks!

templates/index.html
rparrett: Also not sure what's going on here but in the current changeset the open/close tags seem balanced.

IceSentry (approved):

cart (approved):

PR #792, opened 2023-11-02
0.12: UI Outlines, Disjoint EntityMut, Unified configure_sets

cart:
Part of #754


content/news/2023-10-21-bevy-0.12/index.md
```diff
@@ -15,6 +15,80 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
...
+You can now mutably iterate all entities and access arbitrary components within them:
+
+```rust
+for mut entity in world.iter_entities_mut() {
+    let mut transform = entity.get_mut::<Transform>().unwrap();
+    transform.translation.x += 2.0;
+}
+```
+
+This required reducing the access scope of [`EntityMut`] to _only_ the entity it accesses (previously it had escape hatches that allowed direct [`World`] access).
```
alice-i-cecile, line 69: This would be a good spot to mention `EntityWorldMut` maybe.

PR #797, opened 2023-11-03
0.12: Cover Image

cart:
Part of #754 

Cover by @josfeenstra


josfeenstra:
I still feel so incredibly honored Cart, its crazy to see it sitting there on bevy's website, thank you!

In an ideal world, I think the .gif would indeed benefit from being a loop, and being even more 'chill'. 

However,. I can't really deliver major animation changes, as I've kinda lost control over the hacky branch i'm using. Several animation systems are 'fighting for control' over transforms, and the plugins are too split up to order them easily. 

Is there anything besides that I can do for you?

Also, I'm thinking the next version should have an actual bevy of birds fly by!

cart:
I think this is in a great spot. Thanks again!
Happy to swap it out if you come up with any changes, but I'm stoked on what we have now / no pressure.

PR #802, opened 2023-11-04
Remove black frame migration guide

IceSentry:


alice-i-cecile (approved):

alice-i-cecile:
Merging as trivial.

hymm:
Do users that use custom runners need updates because of this pr? I remember this needing a follow up pr for the run one runner.

alice-i-cecile:
Yeah, I think there's a good chance. I'd prefer to do that in a follow-up, but I'm not sure exactly what to put.

PR #803, opened 2023-11-04
Update book to Bevy 0.12

alice-i-cecile:
Just a quick version number bump.


cart (approved):

cart:
Skipping the queue in the interest of keeping the pipes open. Probably unnecessary, but better safe than sorry in this case.

Issue #825, opened 2023-11-08
0.12 Migration: mesh_position_world_to_clip was seemingly removed, but is not mentioned anywhere


rparrett:

IceSentry:
Right, it was part of the view transformations PR and is now `view_transformations::position_world_to_clip` but there wasn't a migration guide. I'll create one,

PR #826, opened 2023-11-08
Add view transformations migration guide

IceSentry:
Closes https://github.com/bevyengine/bevy-website/issues/825

### DIFF
--- a/content/learn/migration-guides/0.11-0.12/_index.md
+++ b/content/learn/migration-guides/0.11-0.12/_index.md
@@ -1531,3 +1531,13 @@ App::new()
 ```
 
 </div>
+
+### [View Transformations](https://github.com/bevyengine/bevy/pull/9726)
+
+`mesh_functions::mesh_position_world_to_clip` was moved and renamed to `view_transformations::position_world_to_clip`.
+
+You can import it like this:
+
+```rust
+#import bevy_pbr::view_transformations::position_world_to_clip;
+```
--- a/content/learn/migration-guides/0.11-0.12/_index.md
+++ b/content/learn/migration-guides/0.11-0.12/_index.md
@@ -1534,10 +1534,20 @@ App::new()
 
 ### [View Transformations](https://github.com/bevyengine/bevy/pull/9726)
 
-`mesh_functions::mesh_position_world_to_clip` was moved and renamed to `view_transformations::position_world_to_clip`.
-
-You can import it like this:
+`mesh_functions::mesh_position_world_to_clip` was moved and renamed to `view_transformations::position_world_to_clip`. It now also takes a `vec3` instead of a `vec4` so you will need to use `vec4.xyz` to get a `vec3`.
 
 ```rust
+// 0.11
+#import bevy_pbr::mesh_functions::mesh_position_world_to_clip
+fn mesh_position_local_to_clip(model: mat4x4<f32>, vertex_position: vec4<f32>) -> vec4<f32> {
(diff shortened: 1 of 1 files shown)


rparrett (approved):
Oh wow, was that the only breaking change there? Seems like it.

IceSentry:
Yeah, it was mostly just adding new functions that we had to manually write before. I missed the fact that it now uses vec3 instead of vec4 but I updated it now.

rparrett:
> Yeah, it was mostly just adding new functions that we had to manually write before. I missed the fact that it now uses vec3 instead of vec4 but I updated it now.

I actually just hit that right now while completing the migration I was working on, and ended up with the same thing. Cool.

content/learn/migration-guides/0.11-0.12/_index.md
```diff
@@ -1531,3 +1531,23 @@ App::new()
...
+// 0.11
+#import bevy_pbr::mesh_functions::mesh_position_world_to_clip
+fn mesh_position_local_to_clip(model: mat4x4<f32>, vertex_position: vec4<f32>) -> vec4<f32> {
+    let world_position = mesh_position_local_to_world(model, vertex_position);
+    return mesh_position_world_to_clip(world_position);
+}
+
+// 0.12
+#import bevy_pbr::view_transformations::position_world_to_clip;
+fn mesh_position_local_to_clip(model: mat4x4<f32>, vertex_position: vec4<f32>) -> vec4<f32> {
```
DGriffin91, line 1549: This is incorrect. There is no `mesh_position_local_to_clip` function in `bevy_pbr::view_transformations`. None of the functions in `bevy_pbr::view_transformations` take a mat as a parameter.

IceSentry: This is showing the migration of the `mesh_position_local_to_clip` function that uses `position_world_to_clip`. This isn't in `view_transformations` it was to show the import.

IceSentry: it was copied from `mesh_functions.wgsl`

rparrett: Yeah okay, checks out.

I wonder if the function signature is just a bit of noise we don't need here though.

DGriffin91: I see what you mean now. I only very briefly looked at this. The similarity in names between everything here may also make things confusing in the guide. idk though

TrialDragon: I agree that the function signature is confusing and noise.

rparrett: Maybe just show in diff format like this:
```diff
-#import bevy_pbr::mesh_functions::mesh_position_world_to_clip
+#import bevy_pbr::view_transformations::position_world_to_clip;
 fn mesh_position_local_to_clip(model: mat4x4<f32>, vertex_position: vec4<f32>) -> vec4<f32> {
     let world_position = mesh_position_local_to_world(model, vertex_position);
-    return mesh_position_world_to_clip(world_position);
+    return position_world_to_clip(world_position.xyz);
 }
 ```

alice-i-cecile (approved):
This is more useful than nothing. Merging now: feel free to clean up further in a follow-up.

alice-i-cecile:
Ugh, nasty merge conflicts... This needs to target the newly renamed file.

Issue #317, opened 2022-04-02
Multilingual: let's talking about translation strategies


ramon-bernardo:
[Zola](https://www.getzola.org/) now has basic support for [multiple languages](https://www.getzola.org/documentation/content/multilingual/), let's talking about translation strategies.

#### 
See [Zola example](https://github.com/getzola/zola/tree/master/test_site_i18n).

alice-i-cecile:
So, we've got a few areas we could target with translation:

1. Home page: reasonably high priority, this is the top of our funnel
2. Bevy Marketplace: helps the user experience, and eventually core to revenue. Descriptions for each plugin or asset would be hard though.
3. Book: very unstable still, but very important for learning.
4. Release notes: helpful to translate, but not useful until we have a sizable population of devs in that language who aren't comfortable with the English version 
5. Migration guides: as above, but perhaps less important, as these tend to be quite mechanical and technical
6. Assorted news posts: low priority, these are nice, but not critical content

cart:
As a note: before undertaking translation efforts we should sort out a way to track translation progress _and_ track "drift" when we make a change to the english "source". If we can't have a trustable view of the state of things, we risk having "wrong" documentation / website info in other languages, which is actively harmful. And considering that people will already have trouble finding non-english support (because english is where most of our "help" threads and github discussions happen), "wrong" docs are very dangerous / could waste a lot of time.

Additionally, given that Bevy docs / website content is still very "in flux", I think we should keep this hyper-scoped until things settle down (probably just translating the front page + general navigation).

PR #831, opened 2023-11-16
Add simplified chinese translation for home page

lomirus:
Added simplified chinese translation for home page. You can access it by http://127.0.0.1:1111/zh_cn/.

The links for images was changed to absolute paths from relative path in order to be able to be accessed in sub path like `/zh-cn/`.

This PR only adds the translation, but does not add the link for users to access this page (`/zh_cn/`) in website, because I think it should be another issue: the UI change (language switcher) for multi-language should be discussed further. 

#317 has discussed the translation strategies but has not provided a way to track translation progress yet, so maybe this PR can remain open until we sort out the way.

### DIFF
--- a/templates/index.zh_cn.html
+++ b/templates/index.zh_cn.html
@@ -211,62 +211,62 @@
             </div>
         </div> -->
         <div class="sponsors sponsors--platinum">
-            <h2 class="sponsors__title">铂金赞助商</h2>
-            <h3 class="sponsors__amount">$1,000 / 月</h3>
+            <h2 class="sponsors__title">Platinum Sponsors</h2>
+            <h3 class="sponsors__amount">$1,000 / month</h3>
             <div class="sponsors__content">
                 <a class="sponsors__link" href="https://www.foresightmining.com">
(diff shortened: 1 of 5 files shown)


templates/index.zh_cn.html
```diff
@@ -0,0 +1,275 @@
...
+            <h2 class="sponsors__title">黄金赞助商</h2>
+            <h3 class="sponsors__amount">$300 / 月</h3>
+            <div class="sponsors__content">
+                <a class="sponsors__link" href="https://playroids.com">
+                    <img class="sponsors__logo" src="/assets/roids_logo.png" alt="Roids 徽标" style="max-height: 130px" />
+                </a>
+            </div>
+        </div>
+        <div class="sponsors sponsors--silver">
+            <h2 class="sponsors__title">白银赞助商</h2>
```
cart, line 243: I'm not sure its viable for us to keep all translated sponsor lists in sync. It might be better to leave this out until we can automate it or something.

alice-i-cecile:
I don't think this is an initiative we can pursue right now <3 Closing.

PR #832, opened 2023-11-17
add version filter to assets

SET001:
This PR extends work that was done in [734](https://github.com/bevyengine/bevy-website/pull/734) and adds filter to select the latest supported version 

it does some version normalization:
- ignores any non digits
- pads version with 0

```
^0.9 -> 0.9.0
0.12 -> 0.12.0
0.12.* -> 0.12.0
~0.8 -> 0.8.0
>=0.9 -> 0.9.0
```

also, there are some cases of versions that I don't know how to treat: 'git', '*', 'main'. But all of them are shown when a filter is reset to `all`

P.S. seems like my IDE added some formatting to older code. Hope this will not be an issue )

asd1asdasdzsdasd

https://github.com/bevyengine/bevy-website/assets/30506/b9ab467d-35d1-44af-9cbc-264e39a8c3db

for some reason vide does not capture select box, but it is there. Believe me!

### DIFF
--- a/static/assets.js
+++ b/static/assets.js
@@ -1,18 +1,38 @@
+let filters_state = {
+    search_terms: [''],
+    version: '*'
+}
+
+const check_filters = (filters) => (node) => {
+    let results = filters.map(filter => filter(node));
+    return results.every(val => val)
+}
+
+const pass_filters = check_filters([
+    filter_search_terms,
+    filter_version
+]);
+
+function filter_assets() {
+    document.querySelectorAll('.asset-card').forEach(asset => {
+        asset.parentElement.style.display = pass_filters(asset) ? 'block' : 'none'
+    })
+}
+
+//  ------------    Search terms Filtering
+
 const searchElement = document.querySelector('#assets-search')
 
 searchElement.addEventListener("input", (_) => {
-    filterSearchTerms()
+    filters_state.search_terms = searchElement.value.toLowerCase().split(' ');
+    filter_assets()
     hideEmptySubSections()
     hideEmptySections()
 })
 
-function filterSearchTerms() {
-    const searchTerms = searchElement.value.toLowerCase().split(' ')
-    for (const asset of document.querySelectorAll('.asset-card')) {
-        const fullText = asset.text.toLowerCase()
-        const searchMatch = searchTerms.every((term) => fullText.includes(term))
-        asset.parentElement.style.display = searchMatch ? 'block' : 'none'
-    }
+function filter_search_terms(asset_node) {
+    const fullText = asset_node.text.toLowerCase()
+    return filters_state.search_terms.every((term) => fullText.includes(term))
 }
 
 function hideEmptySubSections() {
@@ -50,8 +70,21 @@ function sort_versions(a, b) {
 function normalize_version(raw_version) {
     let version = raw_version?.replace(/^[^\d]+/, '').replace(/[^\d]+$/, '');
     let normalized_version = version ? Array.from({ ...version.split('.'), length: 3 }, (v, i) => v ?? 0).join('.') : '*'
-
     return normalized_version;
+}
+
+function filter_version(asset_node) {
+    let name = asset_node.querySelector('.asset-card__title').innerHTML;
+
+    let tag = asset_node.querySelector('.asset-card__tags .asset-card__bevy-versions .asset-card__tag');
+    if (filters_state.version === 'all_versions') {
+        return true
+    }
+    else if (tag) {
+        let raw_item_value = tag.innerText;
+        let normalized_version = normalize_version(raw_item_value);
+        return [filters_state.version, ...version_always_show].includes(normalized_version);
+    } else return true
 }
 
 let versionsSelect = document.querySelector('#assets-filter');
@@ -78,20 +111,6 @@ if (versionsSelect) {
 document
     .querySelector('#assets-filter')
     .addEventListener("change", (item) => {
-        let selected_value = item.target.value;
-        for (const asset of document.querySelectorAll('.asset-card')) {
-            // let name = asset.querySelector('.asset-card__title').innerHTML;
-
-            let tag = asset.querySelector('.asset-card__tags .asset-card__bevy-versions .asset-card__tag');
-            if (selected_value === 'all_versions') {
-                asset.parentElement.style.display = 'block'
-            }
-            else if (tag) {
-                let raw_item_value = tag.innerText;
-                let normalized_version = normalize_version(raw_item_value);
-                // console.debug("<<<<", { name, normalized_version, selected_value }, [selected_value, ...version_always_show]);
-                const searchMatch = [selected_value, ...version_always_show].includes(normalized_version);
-                asset.parentElement.style.display = searchMatch ? 'block' : 'none'
-            }
-        }
+        filters_state.version = item.target.value;
+        filter_assets();
     })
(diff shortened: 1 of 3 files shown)


SET001:
> Does this default to "latest"?

it's default to `all`

SET001:
@alice-i-cecile I mean it shows all versions by default. Just like now by default. But additionally allows to filter by version

static/assets.js
```diff
@@ -36,4 +36,47 @@ function hideEmptySections() {
...
+//  ------------    Version Filtering
+function normalize_version(raw_version) {
+    let version = raw_version
+        .replace(/^[^\d]+/, '')
+        .replace(/[^\d]+$/, '');
+    return version ? Array.from({ ...version.split('.'), length: 3 }, (v, i) => v ?? 0).join('.') : null;
+}
+
+let versionsQuery = document.querySelectorAll('.asset-card .asset-card__tags .asset-card__bevy-versions .asset-card__tag');
+let versions = [...new Set([...versionsQuery]
```
cart, line 50: The versions list should be sorted / it should ensure 0.12 is sorted _above_ 0.2

static/assets.js
```diff
@@ -36,4 +36,47 @@ function hideEmptySections() {
...
+    .querySelector('#assets-filter')
+    .addEventListener("change", (item) => {
+
+        for (const asset of document.querySelectorAll('.asset-card')) {
+            let tag = asset.querySelector('.asset-card__tags .asset-card__bevy-versions .asset-card__tag');
+            if (item.target.value === 'all_versions') {
+                asset.parentElement.style.display = 'block'
+            }
+            else if (tag) {
+                const searchMatch = item.target.value === normalize_version(tag.innerText);
```
cart, line 76: When anything but `All` is selected, this currently filters out things like "Official Bevy Examples" and "Bevy Cheatbook", which leaves the impression that these are out of date / do not support the latest version. A lot of useful stuff is filtered out the second a filter is enabled. Imo the cost of filtering out useful things is _very_ high, so we should ensure things aren't falling through the cracks here. This is a problem we'll want to solve before shipping this feature.

static/assets.js
```diff
@@ -36,4 +36,47 @@ function hideEmptySections() {
...
+    let version = raw_version
+        .replace(/^[^\d]+/, '')
+        .replace(/[^\d]+$/, '');
+    return version ? Array.from({ ...version.split('.'), length: 3 }, (v, i) => v ?? 0).join('.') : null;
+}
+
+let versionsQuery = document.querySelectorAll('.asset-card .asset-card__tags .asset-card__bevy-versions .asset-card__tag');
+let versions = [...new Set([...versionsQuery]
+    .map(item => normalize_version(item.innerText))
+    .filter(i => i)
```
BlackPhlox, line 52: I would rather prefer if the list was sorted based on the version.
My js is a bit rusty, but this seems to work for me:
image

```suggestion
    .filter(i => i)
    .sort((a, b) => {
        let score = 0;
        let av = a.split('.');
        let bv = b.split('.');
        let ln = av.length < bv.length ? av.length : bv.length;
        for (let index = 0; index < ln-1; index++) {
            score += (parseInt(av[index]) - parseInt(bv[index])) * (ln - index);
        }
        return score;
    }).reverse()
```

BlackPhlox: It may need a comment or two 😅

cart (commented):
I like this idea! I just pushed a small change that adds styling to the dropdown consistent with the text input widget / bevy website generally. Preferable to using the default "os-provided" styling.

static/assets.js
BlackPhlox: I did the thing, however, a review might be needed

static/assets.js
SET001:
```
 Imo the cost of filtering out useful things is very high
```
Agreed! The problem here is how to treat versions like  'git', '*', 'main' and empty string.
I can make that `*` will go into all filters. I can replace `main` to `*` wich will do same. But what to do with `git` and empty string`?

static/assets.js
SET001: from what I see, it does not sort by minor version

BlackPhlox: It should be, though, it might not work if its "0.4" vs "0.4.1" , you can try to insert additional minor versions in normalize_version or the like

SET001:
maybe we can replace the minor version with `*` in select box (in normalize_version)? without it, we might end up with pretty long select box....

static/assets.js
BlackPhlox: There were some unhandled edge cases for patch, someone did it better:
https://medium.com/geekculture/sorting-an-array-of-semantic-versions-in-typescript-55d65d411df2
```suggestion
    .filter(i => i)
    .sort((a, b) => {
        // 1. Split the strings into their parts.
        let a1 = a.split('.');
        let b1 = b.split('.');
        // 2. Contingency in case there's a 4th or 5th version
        const len = Math.min(av.length, bv.length);
        // 3. Look through each version number and compare.
        for (let i = 0; i < len; i++) {
            const a2 = +a1[ i ] || 0;
            const b2 = +b1[ i ] || 0;
            if (a2 !== b2) {
                return a2 > b2 ? 1 : -1;        
            }
        }
        // 4. We hit this if the all checked versions so far are equal
        return b1.length - a1.length;
    }).reverse()
```

BlackPhlox: Heres with extra test versions: 
image

BlackPhlox:
> maybe we can replace pach version with `*` in select box (in normalize_version)? without it, we might end up with pretty long select box....

I think currently is a non-issue as it mostly depends on future bevy patches but is something we should keep in mind.

static/assets.js
SET001: optimised this a bit

static/assets.js
```diff
@@ -36,4 +36,52 @@ function hideEmptySections() {
...
+    .sort((a, b) => {
+        let a1 = a.split('.').map(i => i.padStart(3, '0')).join('');
+        let b1 = b.split('.').map(i => i.padStart(3, '0')).join('');;
```
BlackPhlox, line 55:
```suggestion
        let b1 = b.split('.').map(i => i.padStart(3, '0')).join('');
```

BlackPhlox (changes requested):

BlackPhlox (approved):

SET001:
still need some work to do to make this work with search bar

static/assets.js
alice-i-cecile: Empty string I think should map to *. Do the `git` fields link to a specific commit?

SET001: from what I see it is never more that just `git`
asdadcasda

SET001: Empty strings are mostly coming from `learning` category. I'm not sure where version should appear there but tutorials and videos are definitely might be outdated and would be good to have ability to filter them out

alice-i-cecile: Hmm. I think for this PR we should probably just mark them as *, and then fix that up later.

SET001: @cart `When anything but All is selected, this currently filters out things like "Official Bevy Examples"` The problem with `Official Bevy Examples` is that it is explicitly marked as 0.12.0 so it gets filtered out when other versions selected

SET001:
I think it is now finished and ready for review. Now it needs proper testing

DasLixou:
With Version normalizing, does that mean that a `>=0.9` asset doesn't show up under `0.12`?

SET001:
> With Version normalizing, does that mean that a `>=0.9` asset doesn't show up under `0.12`?

yes. It seems like correctly parsing and comparing all semver syntax would be too much of an effort here.

cart (approved):
I think this is in a good state now. Great work!

PR #835, opened 2023-12-04
added how to render a screen 'learn/contributing/plugins'

blikq:
Thought this would be a great addition to the rust bevy book.


content/learn/book/getting-started/plugins/_index.md
```diff
@@ -35,6 +35,37 @@ You should hopefully notice two things:
...
+The output of everything done to this point was only on console. But ofcourse bevy isn't restricted to that so, let's render a window. Change your importations to this:
+
+```rs
+use bevy::{prelude::*, window::*};
+```
+
+Then edit your `main` function with the following:
+
+```rs
+fn main() {
```
cart, line 48: I think this type of "window configuration" belongs in a dedicated window section. This brings in too many windowing details too early in the learning path. This page should just have the minimal details required to define a plugin.

cart: Probably best to hold off until we start working on the next book version. @alice-i-cecile any thoughts here?

alice-i-cecile:
Definitely a useful thing to teach about, agreed on "not quite the right learning path". Let's hold off: we can cover this in a section about configuring windows.

blikq:
Ok, noted

PR #842, opened 2024-01-03
Move parallelism note in ECS learn page

### LINKED ISSUE #638: Move note about parallelism above examples in The Book > Getting Started > ECS
Can the quick note, ""hello world!" might show up in a different order than it does above. This is because systems run in parallel by default whenever possible." be moved just above the output box at the bottom of The Book/Getting Started/ECS? Someone (and possibly more that did not publicly comment on it) did not see it until it was pointed out and it made the process of learning bevy unnecessarily confusing/frustrating.

It should be a 30 second change, I just don't have the setup to make a PR for this atm.

TrialDragon:
The purpose of this PR is to move a note about parallelism in the ECS learning page so that the reader is aware about this fact prior to the output in order to avoid confusion.

Closes #638


alice-i-cecile (approved):

james7132 (approved):

alice-i-cecile:
@TrialDragon can you resolve merge conflicts?

PR #845, opened 2024-01-04
Docs/mut query

TheBlckbird:
This adds an example for mutable queries to the ECS docs.  
I think this is an important topic that may be a bit difficult to find out in the beginning.

### DIFF
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -150,8 +150,8 @@ If we want to change the names of some people (perhaps they got married!), for e
 ```rs
 fn update_people(mut query: Query<&mut Name, With<Person>> {
     for mut name in &mut query {
-        if name.0 == "Elaine Proctor" {
-            name.0 = "Elaine Hume".to_string();
+        if name.0 == "Elaina Proctor" {
+            name.0 = "Elaina Hume".to_string();
             break; // We break out of the loop here, because we don’t need to change any other names
         }
     }
@@ -166,8 +166,7 @@ Don’t forget to add the system to the {{rust_type(type="struct", crate="bevy_a
 fn main() {
     App::new()
         .add_systems(Startup, add_people)
-        .add_systems(Update, hello_world)
-        .add_systems(Update, (update_people, greet_people).chain())
+        .add_systems(Update, (hello_world, (update_people, greet_people).chain()))
         .run();
 }
 ```
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -144,3 +144,32 @@ hello Zayna Nieves!
 Marvelous!
 
 **Quick Note**: "hello world!" might show up in a different order than it does above. This is because systems run in parallel by default whenever possible.
+
+If we want to change the names of some people because they get married, for example, we can do this using a mutable query:
+
+```rs
+fn update_people(mut query: Query<&mut Name, With<Person>> {
+    for mut name in &mut query {
+        if name.0 == "Elaine Proctor" {
+            name.0 = "Elaine Hume".to_string();
+            break; // We break out of the loop here, because we don’t need to change any other names
+        }
+    }
+}
+```
+
+The query parameter must be mutable and the components we want to change have to be a mutable reference for this to work.
+
+Don’t forget to add the system to the Update schedule:
+
+```rs
+fn main() {
+    App::new()
+        .add_systems(Startup, add_people)
+        .add_systems(Update, hello_world)
+        .add_systems(Update, (update_people, greet_people).chain())
+        .run();
+}
+```
+
+Note that we have used `chain()` on the two systems. This is because we want them two to run in exactly that order. If they weren’t, the name might change after we greet the people. But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs.
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -158,7 +158,7 @@ fn update_people(mut query: Query<&mut Name, With<Person>> {
 }
 ```
 
-The query parameter must be mutable and the components we want to change have to be a mutable reference for this to work.
+The `query` argument must be mutable and the components we want to change have to be a mutable reference (`&mut`) for this to work.
 
 Don’t forget to add the system to the {{rust_type(type="struct", crate="bevy_app", name="Update")}} schedule:
 
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -173,4 +173,4 @@ fn main() {
 
 Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
 If they weren’t, the name might change after we greet the people.
-But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. This way, bevy can run `hello_world` while the other systems are running.
+But we don’t add the `hello_world` system to the chain, because it doesn’t matter when it runs. This way, bevy can run `hello_world` in parallel while the other systems are running.
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -174,4 +174,4 @@ fn main() {
 
 Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
 If they weren’t, the name might change after we greet the people.
-But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. That way Bebmvy can run `hello_world` while the other systems are running.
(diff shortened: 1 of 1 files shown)


content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,32 @@ hello Zayna Nieves!
 Marvelous!
 
 **Quick Note**: "hello world!" might show up in a different order than it does above. This is because systems run in parallel by default whenever possible.
+
+If we want to change the names of some people because they get married, for example, we can do this using a mutable query:
```
alice-i-cecile, line 148:
```suggestion
If we want to change the names of some people (perhaps they got married!), for example, we can do this using a mutable query:
```

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,32 @@ hello Zayna Nieves!
...
+    for mut name in &mut query {
+        if name.0 == "Elaine Proctor" {
+            name.0 = "Elaine Hume".to_string();
+            break; // We break out of the loop here, because we don’t need to change any other names
+        }
+    }
+}
+```
+
+The query parameter must be mutable and the components we want to change have to be a mutable reference for this to work.
```
alice-i-cecile, line 161:
```suggestion
The `query` argument must be mutable and the components we want to change have to be a mutable reference (`&mut`) for this to work.
```

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,32 @@ hello Zayna Nieves!
...
+fn main() {
+    App::new()
+        .add_systems(Startup, add_people)
+        .add_systems(Update, hello_world)
+        .add_systems(Update, (update_people, greet_people).chain())
+        .run();
+}
+```
+
+Note that we have used `chain()` on the two systems. This is because we want them two to run in exactly that order. If they weren’t, the name might change after we greet the people. But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs.
```
alice-i-cecile, line 175:
```suggestion
Note that we have used `chain()` on the two systems. This is because we want them two to run in exactly that order: with `update_people` occurring before `greet_people`.
If they weren’t, the name might change after we greet the people.
But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs.
```

TheBlckbird:
Everything should resolved now

TheBlckbird:
One more question: AFAIK bevy runs systems in parallel if possible. Does this mean `hello_world` might run while the chained systems are running? Because if `hello_world` was a longer running process it would be more performant and I could add that to then end. I am just unsure and don't want to add wrong things.

alice-i-cecile:
Correct, `hello_world` might run during any of the other systems, since it doesn't require the same data :) Leaving it separate / unordered is fine though: there's no logical dependency between these systems.

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,34 @@ hello Zayna Nieves!
...
+fn main() {
+    App::new()
+        .add_systems(Startup, add_people)
+        .add_systems(Update, hello_world)
+        .add_systems(Update, (update_people, greet_people).chain())
+        .run();
+}
+```
+
+Note that we have used `chain()` on the two systems. This is because we want them two to run in exactly that order: with `update_people` occurring before `greet_people`.
```
alice-i-cecile, line 175:
```suggestion
Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
```

alice-i-cecile (approved):

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,34 @@ hello Zayna Nieves!
...
+The `query` argument must be mutable and the components we want to change have to be a mutable reference (`&mut`) for this to work.
+
+Don’t forget to add the system to the {{rust_type(type="struct", crate="bevy_app", name="Update")}} schedule:
+
+```rs
+fn main() {
+    App::new()
+        .add_systems(Startup, add_people)
+        .add_systems(Update, hello_world)
+        .add_systems(Update, (update_people, greet_people).chain())
```
nicopap, line 170: I think this might be a bit misleading. I would have written it this way
```rust
        .add_systems(
            Update,
            (hello_world, (update_people, greet_people).chain()),
        )
```

TheBlckbird:
No, I meant I could add that as an explanation

TheBlckbird:
Is this ok?

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,34 @@ hello Zayna Nieves!
...
+    for mut name in &mut query {
+        if name.0 == "Elaine Proctor" {
+            name.0 = "Elaine Hume".to_string();
+            break; // We break out of the loop here, because we don’t need to change any other names
+        }
+    }
+}
+```
+
+The `query` argument must be mutable and the components we want to change have to be a mutable reference (`&mut`) for this to work.
```
nicopap, line 161: I think this might be more readable. What do you think?
```suggestion
We need to make `query` mutable, and use a mutable reference (`&mut`) to the components we want to change.
```

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,34 @@ hello Zayna Nieves!
...
+        .add_systems(Startup, add_people)
+        .add_systems(Update, hello_world)
+        .add_systems(Update, (update_people, greet_people).chain())
+        .run();
+}
+```
+
+Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
+If they weren’t, the name might change after we greet the people.
+But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. That way Bebmvy can run `hello_world` while the other systems are running.
```
nicopap, line 177: this is technically wrong, as `hello_world` cannot run concurrently with `update_people`, as it accesses `Name` mutably.

nicopap, line 177:
```suggestion
But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. This way, bevy can run `hello_world` while the other systems are running.
```

nicopap (changes requested):
I think it's a good idea to add this section.

Just a typo on the last line that needs fixing. IMO

nicopap (approved):

TheBlckbird:
I think everything should be okay now.

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +144,33 @@ hello Zayna Nieves!
 Marvelous!
 
 **Quick Note**: "hello world!" might show up in a different order than it does above. This is because systems run in parallel by default whenever possible.
+
+If we want to change the names of some people (perhaps they got married!), for example, we can do this using a mutable query:
```
Vrixyz, line 148: Could be useful to add a title or [anchor/id](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_an_element_on_the_same_page) so readers can be referred to this section in particular ; but I wouldn't block on it.

Vrixyz (approved):
Small nitpick about being able to link to specific parts of the book https://github.com/bevyengine/bevy-website/pull/845/files#r1442684960 ; but might need more holistic approach to have consistent id throughout the page.

LGTM

content/learn/book/getting-started/ecs/_index.md
TheBlckbird: like this?

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -174,4 +174,4 @@ fn main() {
 
 Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
 If they weren’t, the name might change after we greet the people.
-But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. That way Bebmvy can run `hello_world` while the other systems are running.
+But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. This way, bevy can run `hello_world` while the other systems are running.
```
alice-i-cecile:
```suggestion
But we don’t add the `hello_world` system to this chain, because it doesn’t matter when it runs. This way, Bevy can run `hello_world` while the other systems are running.
```

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -144,3 +146,35 @@ hello Zayna Nieves!
...
+    App::new()
+        .add_systems(Startup, add_people)
+        .add_systems(Update, (hello_world, (update_people, greet_people).chain()))
+        .run();
+}
+```
+
+Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
+If they weren’t, the name might change after we greet the people.
+But we don’t add the `hello_world` system to the chain, because it doesn’t matter when it runs. This way, bevy can run `hello_world` in parallel while the other systems are running.
```
alice-i-cecile, line 180:
```suggestion
But we don’t add the `hello_world` system to the chain, because it doesn’t matter when it runs. This way, Bevy can run `hello_world` in parallel while the other systems are running.
```

PR #847, opened 2024-01-05
Change println! to info! in getting started tut

### LINKED ISSUE #108: Use info! instead of println! in tutorial
When debugging actual Bevy games, you almost always want to use `info!` instead of `println` in order to reduce blocking behavior.

This is not at all obvious to Rust beginners, and showing them how to use `println` misleads them into thinking it's a good pattern to replicate.

This was a simple find-and-replace change for me, and resulted in much faster looping.

TrialDragon:
This PR changes the use of `println!` to `info!` in the getting started tutorial to reflect to beginners the best practices when debugging actual Bevy games.

Closes #108

### DIFF
--- a/content/learn/book/getting-started/ecs/_index.md
+++ b/content/learn/book/getting-started/ecs/_index.md
@@ -29,7 +29,7 @@ Bevy ECS is Bevy's implementation of the ECS pattern. Unlike other Rust ECS impl
     ```rs
     fn print_position_system(query: Query<&Position>) {
         for position in &query {
-            println!("position: {} {}", position.x, position.y);
+            info!("position: {} {}", position.x, position.y);
         }
     }
     ```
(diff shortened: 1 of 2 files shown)


ItsDoot (approved):

content/learn/book/getting-started/ecs/_index.md
```diff
@@ -40,6 +40,8 @@ Bevy ECS is Bevy's implementation of the ECS pattern. Unlike other Rust ECS impl
     struct Entity(u64);
     ```
 
+**Note**: You may have noticed that in the above examples, and throughout this tutorial, we use `info!()` from Bevy rather than `println!()` to output text to the console. Why? `info!()` when used in a debug scenario reduces blocking behavior and runs faster compared to `println!()`.
```
alice-i-cecile, line 43: It would be nice to link to `info` here :)

alice-i-cecile (approved):

content/learn/book/getting-started/ecs/_index.md
TrialDragon: All references in that line, or just the first reference of each type?

alice-i-cecile: I think just the first is fine.

mockersf:
This logs won't work if one is following the tutorial, as the logger is not set before

TrialDragon:
Shoot, right. The LogPlugin isn't added until the Plugins chapter with the rest of the defaults, hmm.

Is it worth adding in the DefaultPlugins earlier, but have a note / footnote stating that it'll be covered or explained in a later section?

alice-i-cecile:
I think that this is mostly a distraction, so we should leave it out. It's not "best practice", but eh it'll do for beginners.

PR #859, opened 2024-01-10
Update Godot Rust link

conways-glider:
Update the Godot Rust link from gdnative to gdext, which is compatible with the latest Godot version. This also captures a change in ownership of both of these repos to the org https://github.com/godot-rust


alice-i-cecile (approved):

alice-i-cecile:
Thanks! Merging as trivial.

PR #860, opened 2024-01-12
Fix minor typo

davidepaci:
i understand if it gets denied because it's controversial

priority: EXTREME


ItsDoot (approved):

mockersf:
Could you change the PR title so that it shows correctly in the fit log?

PR #862, opened 2024-01-13
Flatten book layout

### LINKED ISSUE #160: Flatten book layout structure
Rather than using `page-title/_index.md`, it would be very convenient to just be able to use `page-title.md`.

Unfortunately, I couldn't figure out how to get Zola to do this.

TrialDragon:
Makes the book use pages where before sections with just an _index.md file were for developer usability sake. Also does a minor change to the migration guides urls because apparently dots can't be in page urls turning it from `0.4-0.5` to `0-4-0-5`. Now they're like `0-4-to-0-5` for readability sake.

Closes #160


TrialDragon:
Still need to fix next / previous page navigation.

TrialDragon:
I'd like to know opinions on the whole migration guide's url thing. Should those just be reverted to sections to avoid that issue at the cost of losing minor usability benefits moving to pages bring, or is it worth it to just keep them as pages and deal with the url change? (I already tried change the slug and such; pages unlike sections just can't have dots in the page url apparently.)

TrialDragon:
(This comment is more of a note to myself to look into this) This PR may require changing the migration guide stub generator since it may be generating the files as sections still (presuming migration guide stuff isn't switched back to sections.)

alice-i-cecile:
> I'd like to know opinions on the whole migration guide's url thing. Should those just be reverted to sections to avoid that issue at the cost of losing minor usability benefits moving to pages bring, or is it worth it to just keep them as pages and deal with the url change? (I already tried change the slug and such; pages unlike sections just can't have dots in the page url apparently.)

I'd like to keep this consistent across the project and swap to pages. We should set up redirects though.

TrialDragon:
> I'd like to keep this consistent across the project and swap to pages. We should set up redirects though.

By redirects you mean setup dummy section pages that go to regular pages since the sections can have dots in url?

TrialDragon:
Ope, ignore the previous comment; aliases just work for redirects apparently? How did I not notice this when I was looking for solutions before... I'll commit that ASAP and get the conflicts solved.

alice-i-cecile (approved):

content/learn/book/contributing/_index.md
```diff
@@ -3,10 +3,11 @@ title = "Contributing"
 weight = 4
 sort_by = "weight"
 template = "docs-section.html"
-page_template = "docs-section.html"
+page_template = "docs-page.html"
 insert_anchor_links = "right"
 [extra]
 subtitle = "join the bevy"
+weight = 5
```
cart, line 10: Why is duplicating the "weight" field necessary? Remembering to add it twice is extra work that needs to be done, and it creates potential for the weight to be out of sync.
This one is out of sync with the "actual weight". Was that a mistake or intentional?

content/learn/book/next-steps.md
```diff
@@ -2,9 +2,9 @@
 title = "Next Steps"
 weight = 3
 sort_by = "weight"
-template = "docs-section.html"
-page_template = "docs-section.html"
 insert_anchor_links = "right"
+[extra]
+weight = 4
```
cart, line 7: This is out of sync with the "actual weight" (see my other comment about this).

cart (commented):
Love the flattened structure. Just one thought / concern.

content/learn/book/contributing/_index.md
TrialDragon: In the docs menu macro we can't access the actual weight thru the zola pages / sections objects. It isn't a field. But I needed to resort them by weight due to the fact there isn't a mixed pages and sections array and I had to make one myself.

The out of sync wasn't intended.

Tbh, I'm 99% sure we can just replace zola's weight with the custom weights and be fine. I just left them in on the off chance I missed one esoteric thing that does need weights and the code I added doesn't replace in effective functionality and as such breaks things unexpectedly.

pablo-lua (commented):
Looks really interesting changes in general, really flattened the book.

content/learn/book/contributing/_index.md
TrialDragon: Okay, have removed the duplication.

alice-i-cecile:
@BD103, could you take a look at this PR? I'd like to get this merged in before we start working on the book content further.

BD103:
> @BD103, could you take a look at this PR? I'd like to get this merged in before we start working on the book content further.

Yeah, I'll give it my best shot! :D

content/learn/advanced-examples/sudoku/_index.md
```diff
@@ -1,5 +1,8 @@
 +++
 title = "Sudoku"
-weight = 2
 template = "docs-section.html"
+page_template = "docs-page.html"
```
BD103, line 4: I may be wrong, but isn't `page_template` inherited? Since it is set in `advanced-examples/_index.md`, does it also need to be set here?

BD103 (commented):
From a very brief first pass, there don't appear to be any visible changes from browsing the built website. (Which is good, since messing up URLs would be bad.)

My only question is about setting the template for every single page / section, when the docs mentions that it might be inheritable? Could this be reduced so that there is less copy-and-pasting when creating new pages?

content/learn/advanced-examples/sudoku/_index.md
TrialDragon: Honestly not sure. I'll do some tests later to check. Hmm.

content/learn/quick-start/_index.md
BD103, line 1: Does this need an `extra.weight` property, even though it is empty?

content/learn/quick-start/next-steps.md
```diff
@@ -1,11 +1,9 @@
 +++
 title = "Next Steps"
-weight = 12
-sort_by = "weight"
-template = "docs-section.html"
-page_template = "docs-section.html"
 insert_anchor_links = "right"
 aliases = ["learn/book/next-steps"]
+[extra]
+weight = 3
```
BD103, line 6: This weight was changed from 12 to 3. Was that purposeful?

BD103 (approved):
I think this looks good! (Though there were a _lot_ of changed files, so I may have missed something.) I noticed a few things and wanted to make sure they were intentional, so check out my comments.

content/learn/book/plugin-development/_index.md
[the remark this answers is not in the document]
TrialDragon: Yes. That's an issue from the new book merge. That file was never supposed to be left, and I noticed it during this PR. I can, and probably should, spin this out into its own issue and PR tho.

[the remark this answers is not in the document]
BD103: Ok. If the file was never meant to be in the book in the first place, then you should probably make a separate PR deleting it. At the very least, that PR will likely get merged faster than this one.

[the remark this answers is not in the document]
TrialDragon: Did my best to restore it; will open issue and PR to remove it.

content/learn/quick-start/next-steps.md
TrialDragon: Yes and no. It was originally 4, then made 3 on main during some re-ordering with plugin guidelines, the fact that it is 12 on main means the original new book to main merge fucked up some weights and wasn't caught. It's still supposed to be 3.

Hmm. Either time to open up an issue and fix that on main, or just wait until this is merged to avoid having to solve _another_ merge conflict.

content/learn/advanced-examples/sudoku/_index.md
TrialDragon: Ope, yep. I'll remove all the unnecessary `page_template`'s across the content then.

content/learn/quick-start/_index.md
TrialDragon: I don't think so. The way stuff works it gets all of `learn/{current_section`'s subsections those are the ones with extra.weight called on them. If we ever, for whatever reason, have `quick-start` as a child of a `docs-section` (for example converting the learn page to one) then yes, it would need a weight, but that can be handled then since the site will refuse to compile at that point until solved.

BD103: Ok, just making sure!

content/learn/quick-start/next-steps.md
BD103: I don't think this is a blocking issue, so it can probably be merged as-is.

alice-i-cecile:
Thanks a ton! This is much nicer to work with :)

Issue #870, opened 2024-01-18
Integrate new Bevy book into main


TrialDragon:
Current plan for integrating new Bevy book into bevy-website/main. Moved from #623 per https://github.com/bevyengine/bevy-website/issues/623#issuecomment-1518887997

- [x] Synchronize `new-book` with `main` to make these changes easier
- [x] #874
- [x] #875
- [x] #876
- [x] #877
- [x] #885 
- [x] #878
- [ ] Merge into main and any final merge conflict resolution not covered in prior PRs

PR #871, opened 2024-01-18
Sync new book to main

TrialDragon:
Synchronizes new book to main and fixes issues that came up as the result of resolving merge conflicts. 

Changes usage of rust type shortcode to reference links since that was removed from new books.

Change book_section to docs_section because book_section was changed to docs_section in main a while ago.

Create a quick-start section. This one is more prep work for #870 

Is apart of #870 in order to make further changes easier; especially when maintainers merge new book into main.


alice-i-cecile (approved):
Commit-by-commit view looks great. Thank you!

alice-i-cecile:
Merging unilaterally: this is really valuable, janitorial and will get a final review or 7 before finally being shown to end users.

PR #873, opened 2024-01-19
Discourage beginners from using nightly Rust

alice-i-cecile:
# Objective

The advice in the section on optimizing compile times frequently leads beginners down the wrong path.

If you are using Rust for the first time (like about half of our new users), you *should not* follow this advice.

Furthermore, the actual gains from `nightly` alone are marginal, and cause new users to stumble when nightly is broken with Bevy (like right now).

# Solution

1. Remove the advice to use nightly completely.
2. Modify the language to make it even clearer that this is both optional and advanced.

Long-term, I'd like to move this section far away from the start of the book, but that's a more involved change that can wait.

### DIFF
--- a/content/learn/book/getting-started/setup/_index.md
+++ b/content/learn/book/getting-started/setup/_index.md
@@ -143,7 +143,7 @@ You might think to simply develop in release mode instead, but we recommend agai
 
 ### Enable Fast Compiles (Optional)
 
-Bevy can be built just fine using default configuration on stable Rust. However for maximally fast iterative compiles, we recommend the following configuration:
+Bevy can be built just fine using Rust's default configuration. However, if you know your way around a compiler and find that Bevy's iterative compilation times are still too slow for you, try fine-tuning the following options:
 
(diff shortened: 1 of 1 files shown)


rodolphito (approved):

rparrett (approved):
Big thumbs-up from me all around.

content/learn/book/getting-started/setup/_index.md
```diff
@@ -189,20 +189,9 @@ Bevy can be built just fine using default configuration on stable Rust. However
...
-    [toolchain]
-    channel = "nightly"
-    ```
-
-    For more information, see [The rustup book: Overrides](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file).
-
 * **Generic Sharing**: Allows crates to share monomorphized generic code instead of duplicating it. In some cases this allows us to "precompile" generic code so it doesn't affect iterative compiles. This is only available on nightly Rust.
 
-To enable fast compiles, install the nightly rust compiler and LLD. Then copy the contents of [this file](https://github.com/bevyengine/bevy/blob/main/.cargo/config_fast_builds) to `YOUR_WORKSPACE/.cargo/config.toml`. For the project in this guide, that would be `my_bevy_game/.cargo/config.toml`.
+To enable generic sharing, first install the nightly rust compiler and LLD. Then copy the contents of [this file](https://github.com/bevyengine/bevy/blob/main/.cargo/config_fast_builds) to `YOUR_WORKSPACE/.cargo/config.toml`. For the project in this guide, that would be `my_bevy_game/.cargo/config.toml`.
```
IceSentry, line 194: Technically the linked file does more than just setup generic sharing

alice-i-cecile: I think that this is clearer in the context of the section.

PR #883, opened 2024-01-21
Document Rename

janhohenheim:
I spent a few minutes looking for why my `HandleUntyped`s weren't working anymore and want to save the next person some trouble :)


alice-i-cecile (approved):
Thanks!

janhohenheim:
Added `AssetId` -> `UntypedAssetId` as well, since it is not documented either and related to my first change.

PR #884, opened 2024-01-21
Update untyped handle usage

janhohenheim:
I just commited a change that calls the new behavior a simple rename. I have however made a mistake and the involved types need to be used differently. I've documented this now :)

### DIFF
--- a/content/learn/migration-guides/0.11-0.12/_index.md
+++ b/content/learn/migration-guides/0.11-0.12/_index.md
@@ -192,7 +192,7 @@ Also, as you may have noticed, the set of events has changed. The most important
 
 #### `UntypedHandle`
 
-`HandleUntyped` and `HandleId` have been renamed to `UntypedHandle` and `UntypedAssetId`. 
+`HandleUntyped` and `HandleId` have been renamed to `UntypedHandle` and `UntypedAssetId`.
 The new way to construct an untyped handle looks like this:
 
 ```rust
@@ -200,7 +200,7 @@ The new way to construct an untyped handle looks like this:
 const MESH_HANDLE: HandleUntyped =
     HandleUntyped::weak_from_u64(Mesh::TYPE_UUID, 0x1f40128bac02a9b);
 // 0.12
-const MESH_HANDLE: UntypedHandle = 
+const MESH_HANDLE: UntypedHandle =
     UntypedHandle::Weak(UntypedAssetId::Uuid { type_id: TypeId::of::<Mesh>(), uuid: Uuid::from_u128(0x1f40128bac02a9b) });
 ```
 
--- a/content/learn/migration-guides/0.11-0.12/_index.md
+++ b/content/learn/migration-guides/0.11-0.12/_index.md
@@ -192,7 +192,16 @@ Also, as you may have noticed, the set of events has changed. The most important
 
 #### `UntypedHandle`
 
-`HandleUntyped` has been renamed to `UntypedHandle`.
+Replace `UntypedHandle` usage with the corresponding `Handle<T>`:
+```rust
+// 0.11
+const MESH_HANDLE: UntypedHandle =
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

thepackett (changes requested):
I'm not very experienced with the bevy asset system prior to Assets V2, but I'm not entirely sure this is correct. Even looking through the asset system prior to 0.12, `Handle<A>` still existed back then, and the [bevy unofficial cheat book](https://bevy-cheatbook.github.io/assets/assetserver.html) even shows using `Handle<A>` as far back as 0.9.

From my knowledge of the asset system in 0.12, `Handle<A>` and `UntypedHandle` both have their separate uses, and neither replace the other. The typed handle `Handle<A>` provides a nicer user facing api, but can be cumbersome when dealing with assets of unknown or heterogeneous types. This is when `UntypedHandle` is preferable (and is likely the reason why `UntypedHandle` is so common inside the `AssetServer`'s internal implementation).

Similarly to the above, `AssetId<T>` is to `UntypedAssetId` what `Handle<A>` is to `UntypedHandle`; they are interchangeable depending on whether a typed or untyped api is preferred.


Looking at the code change from 0.11 to 0.12, it seems like what really changed was:

- `HandleUntyped` was renamed to `UntypedHandle`

- The internal structure of `UntypedHandle` was changed from a struct with private fields to a public enum, and so there was no longer a need for a constructor function and so it was removed. It can still be used more or less the same way as before though, since every member is public. I.e. the code below works (as long as you use the crate level `#![feature(const_type_id)]` attribute, but that's a std thing for `TypeId`s in const).

```rust 
const MESH_HANDLE: UntypedHandle = 
UntypedHandle::Weak(UntypedAssetId::Uuid { type_id: TypeId::of::<Mesh>(), uuid: Uuid::from_u128(0x1f40128bac02a9b) });
```

For this PR, my suggestion would be to just document the new way of constructing an untyped handle since that is a significant change that ought to be noted.

It would be valuable to add additional documentation somewhere about why the typed and untyped apis both exist for those learning the asset system, but I don't think that the 0.11 -> 0.12 migration guide is the right place for that sort of general knowledge.

janhohenheim:
@thepackett thanks for correcting me! Let me fix that real quick.

janhohenheim:
@thepackett is it correct like this?

content/learn/migration-guides/0.11-0.12/_index.md
```diff
@@ -192,7 +192,17 @@ Also, as you may have noticed, the set of events has changed. The most important
 
 #### `UntypedHandle`
 
-`HandleUntyped` has been renamed to `UntypedHandle`.
+`HandleUntyped` and `HandleId` have been renamed to `UntypedHandle` and `UntypedAssetId`.
```
thepackett, line 195: One last nitpick:

I would put "`HandleUntyped` renamed to `UntypedHandle`" on its own line, and then on the next line put:
"`HandleId` has been replaced with `UntypedAssetId` and its typed equivalent `AssetId<T>`."

The justification for saying "replaced" is that it was a bit more than just a rename since type information is now stored in the id (whereas before it was stored in the handle).

thepackett (changes requested):

content/learn/migration-guides/0.11-0.12/_index.md
janhohenheim: Good point. Sec.

janhohenheim: There ya go :)

thepackett (approved):
Alright, everything looks good to me! 👍

janhohenheim:
Thanks for your help @thepackett  :)

alice-i-cecile:
Thank you @thepackett for the detailed review and guidance <3

PR #886, opened 2024-01-22
Implement todo shortcode

### LINKED ISSUE #876: Define, and implement, todo / work-in-progress section (headers) styling and wording for `new-book`
Apart of #870 

A way to style work in progress sections, or headers, in the Bevy book needs to be made. Implementation wise the easiest initial solution would be a short-code made in the same way that the `incorrect_code_block` short-code is done.

As for what it should look and read like? No discussion has been made yet. My initial concept of how it could look is a yellow background with darker shade / contrasting rounded border, all de-saturated, with the actual header/section being a higher contrasting orange or red, perhaps prefixed with `TODO:` or `WIP:` so that there is ample warning both visual and textual that this is not ready for the public's eyes yet. 

Another idea is just take a page directly out of the public drafting's style and only have the headers be yellow, with maybe the prefix in addition.

Refine and discuss the style as you please tho; those were just initial thoughts for a base.

TrialDragon:
Adds a todo shortcode to act as a list of sections that need to be done in a file.

Closes #876


templates/shortcodes/todo.html
```diff
@@ -0,0 +1,3 @@
+<div class="todo">
+	<h2 class="todo-header">TODO</h2>{{ body | markdown | safe }}
```
rparrett, line 2:
```suggestion
    <h2 class="todo-header">TODO</h2>
    {{ body | markdown | safe }}
```
Spaces and newline for content

sass/components/_todo.scss
```diff
@@ -0,0 +1,13 @@
+.todo {
+  background-color: hsla(50, 50, 50, 0.3);
+  border: double hsla(50, 75, 50, 0.5) 6px;
+  border-radius: 10px;
+  padding: 4px;
+
+  h2.todo-header {
```
rparrett, line 7: The `todo-header` class seems unnecessary.

sass/components/_todo.scss
```diff
@@ -0,0 +1,13 @@
+.todo {
+  background-color: hsla(50, 50, 50, 0.3);
+  border: double hsla(50, 75, 50, 0.5) 6px;
```
rparrett, line 3: Just commentary, I don't want to block on a subjective matter in a non-user-facing context.

I personally find this double border a bit funky. I think it would be nice to have a consistent "callout" style for these sorts of things with styling similar to what was done for the "bad code blocks."

A few css classes that we could use to produce these sorts of blocks would be handy.
image

rparrett (changes requested):

sass/components/_todo.scss
TrialDragon: I had to make it more specific to override unwanted huge margin styling from else where.

rparrett: I see, okay.

rparrett (approved):
Looks good to me, other than the new markdown lint failure from the ios-android swaparoo merge.

Took a quick browse through all the pages with TODOs and didn't see anything crazy.

PR #887, opened 2024-01-22
Switch ios and android pages right way around

TrialDragon:
Apparently these got put in wrong pages when book structure was made. This trivial PR fixes that.


mockersf:
why did you reverse the weight between Android and web?

mockersf:
it's OK, I was just surprised to see the change without it being mentioned 👍

alice-i-cecile (approved):

PR #895, opened 2024-01-30
0.13: Add Dynamic Queries section

nicopap:
For #891 announcement post.

I also added a subsection on relations, since it's a primary motivator for dynamic queries.

But, it feels like it's just too much for this section. Maybe push it to the "what's next?" section… Could use it as a synthesis for all the changes that brings us closer to relations.

@james-j-obrien, @jakobhellermann, @suficio, or anyone. Please tell if anything is wrong or misleading.

I think I got it alright: context + motivation in two sentences, despite the subject being complex and not immediately useful to end users. Then spin it in a way that is interesting (I think) thanks to the amazing projects people are already building on top of dynamic queries, since **those projects** are useful for end users.

PR #891, opened 2024-01-24
0.13 announcement post

alice-i-cecile:
Since @cart is drowning in admin paperwork for the legal org already (thanks!), I'm going to lead the effort here this cycle.

Like in #754, I'll stub out the major sections here, and then folks can tackle and refine them in their own PRs to this branch :)

TODOs will be tracked in the post, with the handy `TODO` marker.

PR #896, opened 2024-01-30
Fix build of the news section

nicopap:
When writing my section of the 0.13 news post, I found that zola refused to generate the news section. After digging, it seems that this change fixes it.

I notice build fails on #895  and #891, it might fix it.


mockersf (changes requested):
this could work, but the issue is that the document initiated for the 0.13 release doesn't have a summary because unlike the other news, it doesn't have a first few paragraphs delimited with `<!-- more -->`

PR #899, opened 2024-01-31
0.13: Add Extensionless Asset Support Section

bushrat011899:
For #891 announcement post. I tried to split the section into a couple of smaller chunks to help with reading. This might be a bit too lengthy, so please feel free to make suggestions!

### DIFF
--- a/content/news/2024-02-03-bevy-0.13/index.md
+++ b/content/news/2024-02-03-bevy-0.13/index.md
@@ -13,6 +13,7 @@ image_subtitle_link = "TODO"
 Thanks to **TODO** contributors, **TODO** pull requests, community reviewers, and our [**generous sponsors**](/community/donate), we're happy to announce the **Bevy 0.13** release on [crates.io](https://crates.io/crates/bevy)!
 
 For those who don't know, Bevy is a refreshingly simple data-driven game engine built in Rust. You can check out our [Quick Start Guide](/learn/book/getting-started/) to try it today. It's free and open source forever! You can grab the full [source code](https://github.com/bevyengine/bevy) on GitHub. Check out [Bevy Assets](https://bevyengine.org/assets) for a collection of community-developed plugins, games, and learning resources.
+And to see what the engine has to offer hands-on, check out the entries in the [latest Bevy Jam](https://itch.io/jam/bevy-jam-4/entries), including the winner [That's a lot of beeeeees](https://andrewb330.itch.io/thats-a-lot-of-beeeeees)
 
 To update an existing Bevy App or Plugin to **Bevy 0.13**, check out our [0.12 to 0.13 Migration Guide](/learn/migration-guides/0.12-0.13/).
 
--- a/content/news/2024-02-03-bevy-0.13/index.md
+++ b/content/news/2024-02-03-bevy-0.13/index.md
@@ -164,7 +164,7 @@ In Bevy 0.12 and earlier, all assets needed a file extension which uniquely mapp
 let settings = asset_server.load("data/settings.json");
 ```
 
-Peeking behind the curtain, this is achieved by passing the **path** to the `AssetServer`, which selects an `AssetLoader` based on the file extension. This is possible because every `AssetLoader` is required to declare up-front what file types it supports.
+Peeking behind the curtain, this is achieved by passing the **path** to the [`AssetServer`], which selects an [`AssetLoader`] based on the file extension. This is possible because every [`AssetLoader`] is required to declare up-front what file types it supports.
 
 ```rust
 impl AssetLoader for JsonAssetLoader {
@@ -216,7 +216,7 @@ The workaround for this is to create a compound file extension for each unique a
 
 ### Relaxing constraints
 
-But we can do better! In Bevy 0.13, the asset type can now used to infer the `AssetLoader`.
+But we can do better! In Bevy 0.13, the asset type can now used to infer the [`AssetLoader`].
 
 ```rust
 // Uses AudioSettingsAssetLoader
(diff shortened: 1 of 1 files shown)


content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -154,9 +154,157 @@ TODO.
...
+let bang_again = asset_server.load::<AudioSource>("sound/bang.ogg");
+```
+
+Note that the above example uses turbofish syntax for clarity. In practice, it's not required, since the type of asset loaded can usually be inferred by surrounding context at the call site.
+
+```rust
+fn setup(mut effects: ResMut<SoundEffects>, asset_server: Res<AssetServer>) {
+    effects.bang = asset_server.load("sound/bang.ogg");
+    effects.bang_blob = asset_server.load("sound/bang.ogg");
+}
```
tbillington, line 302:
```suggestion
#[derive(Resource)]
struct SoundEffects {
    bang: AudioSource,
    bang_blog: Blob,
}

fn setup(mut effects: ResMut<SoundEffects>, asset_server: Res<AssetServer>) {
    effects.bang = asset_server.load("sound/bang.ogg");
    effects.bang_blob = asset_server.load("sound/bang.ogg");
}
```

Would suggest the definition of `SoundEffect` be included in the snippet. If you're going to the effort of making this specific example it should be super obvious that these are different types, not just different fields/variable names :)

bushrat011899: Good call!

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -154,9 +154,170 @@ TODO.
 
 ## Extensionless asset support
 
-<div class="release-feature-authors">authors: @TODO</div>
+<div class="release-feature-authors">authors: @bushrat011899</div>
 
-TODO.
+### Some background
```
hymm, line 159: I think this background section should be cut. And we just need a sentence that says that before the AssetLoader to use was inferred from the file extension.

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -154,9 +154,170 @@ TODO.
...
+
+The workaround for this is to create a compound file extension for each unique asset type, such as `.audio.json` and `.graphics.json`.
+
+### Relaxing constraints
+
+But we can do better! In Bevy 0.13, the asset type can now used to infer the [`AssetLoader`].
+
+```rust
+// Uses AudioSettingsAssetLoader
+let audio_settings = asset_server.load("data/audio.json");
```
hymm, line 223: This example feels a little weird since there isn't code showing where the AudioSettingsAssetLoader type is inferred from. Maybe have a resource they're inserted into that has the type?

hymm (commented):
How does this feature interact with meta files? Like before you could override the default loader by specifying in the meta file which loader to use.

TrialDragon (commented):
I'm having trouble understanding a bit in this section. How is the asset type inferred? Like, I can't make logical sense of where or even how this is possible. How it can differentiate two different files as different types logically.

content/news/2024-02-03-bevy-0.13/index.md
bushrat011899: Perfectly reasonable, I decided to err on the side of verbosity since it's easier to cut in editing. Happy to remove that section as described.

content/news/2024-02-03-bevy-0.13/index.md
bushrat011899: Agreed, I should've been more explicit here

bushrat011899:
> I'm having trouble understanding a bit in this section. How is the asset type inferred? Like, I can't make logical sense of where or even how this is possible. How it can differentiate files as different types logically, especially when they're using the same underlying data structure like JSon or Toml.

This is a good point where I should clarify _how_ inference works. I'm going to be removing the background section and I'll expand on the full pipeline. The short version of the explanation is inference is purely a code-side choice, and has nothing to do with the name or contents of the asset itself. Instead, the type of the asset to be loaded (the `A` in `AssetServer::load<A>`) is used to select the `AssetLoader`. This makes the assumption that a Rust type representing the asset to be loaded is unique, which was unfortunately not a good assumption that I'm working on resolving.

Essentially, instead of throwing an error (oh no, the asset loaded doesn't match the type), we can do the inverse (this asset _must_ be this type, pick a loader to match)

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -154,9 +154,130 @@ TODO.
 
 ## Extensionless asset support
 
-<div class="release-feature-authors">authors: @TODO</div>
+<div class="release-feature-authors">authors: @bushrat011899</div>
 
-TODO.
+In prior versions of Bevy, the default way to choose an [`AssetLoader`] for a particular asset was entirely based around file extensions. The recent addition of [`meta` files] allowed for specifying more granular loading behavior, but file extensions were still required. In Bevy 0.13, the asset type can now used to infer the [`AssetLoader`].
```
TrialDragon, line 159:
```suggestion
In prior versions of Bevy, the default way to choose an [`AssetLoader`] for a particular asset was entirely based around file extensions. The recent addition of [`meta` files] allowed for specifying more granular loading behavior, but file extensions were still required. In Bevy 0.13, the asset type can now be used to infer the [`AssetLoader`].
```

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -154,9 +154,130 @@ TODO.
...
+```
+
+When loading an asset, the loader is chosen by checking (in order):
+
+1. The asset `meta` file
+2. The type of `Handle<A>` to return
+3. The file extension
+
+```rust
+// This will be inferred from context to be a Gltf asset, ignoring the file extension
```
TrialDragon, line 209: Proper capitalization of glTF.
```suggestion
// This will be inferred from context to be a glTF asset, ignoring the file extension
```

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -154,9 +154,130 @@ TODO.
...
+let gltf_handle = asset_server.load("models/cube/cube.gltf");
+
+// This still relies on file extension due to the label
```
TrialDragon, line 212: I presume the label mentioned is the `#Mesh0/Primitive0` ? Am unsure.

TrialDragon (commented):
Mostly minor grammar fixes and a question, otherwise is easier to comprehend compared to before and LGTM!

content/news/2024-02-03-bevy-0.13/index.md
bushrat011899: Correct, I've updated this example to add a comment highlighting it as the cause of extension-based resolution.

TrialDragon (approved):

alice-i-cecile (approved):

PR #901, opened 2024-02-02
0.13: Write 'Type-safe labels for `RenderGraph`' section

DasLixou:
For #891 announcement post.

### DIFF
--- a/content/news/2024-02-03-bevy-0.13/index.md
+++ b/content/news/2024-02-03-bevy-0.13/index.md
@@ -189,7 +189,7 @@ When you need dynamic labels for render nodes, those can still be achieved via e
 pub struct MyDynamicLabel(&'static str);
 ```
 
-The nice thing is, that this don't has to be a string, it can be any time, also numbers etc.
+The nice thing is, that this don't has to be a string, it can be any type, also numbers etc.
 
 ## Camera-driven UI
 
--- a/content/news/2024-02-03-bevy-0.13/index.md
+++ b/content/news/2024-02-03-bevy-0.13/index.md
@@ -182,6 +182,13 @@ render_app
     );
 ```
 
+When you need dynamic labels for render nodes, those can still be achieved via e.g. tuple structs:
+```rust
+#[derive(Debug, Hash, PartialEq, Eq, Clone, RenderLabel)]
+pub struct MyDynamicLabel(&'static str);
+```
+The nice thing is, that this don't has to be a string, it can be any time, also numbers etc.
+
 ## Camera-driven UI
 
 <div class="release-feature-authors">authors: @TODO</div>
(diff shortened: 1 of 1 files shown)


TrialDragon (commented):
Needs some minor grammar and readability changes, but looks good to me.

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -126,9 +126,70 @@ TODO.
 
 ## Type-safe labels for the `RenderGraph`
 
-<div class="release-feature-authors">authors: @TODO</div>
-
-TODO.
+<div class="release-feature-authors">authors: @DasLixou</div>
+
+Bevy uses Rust's type system at many points like for `SystemSets` which gives it a modern feel.
```
alice-i-cecile, line 131:
```suggestion
Bevy uses Rust's type system extensively when defining labels, letting developers lean on tooling to catch typos and ease refactors.
```

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -126,9 +126,70 @@ TODO.
...
+But this didn't apply to Bevy's render graph. In the render graph, hard-coded—and potentially overlapping—strings were used to define nodes and sub-graphs.
+
+```rust
+// Before 0.13
+impl MyRenderNode {
+    pub const NAME: &'static str = "my_render_node"
+}
+```
+
+We are now introducing a more robust way to name rendernodes and -graphs with the help of Rust's type system and Bevy's label system.
```
alice-i-cecile, line 141:
```suggestion
In Bevy 0.13, we're using a more robust way to name render nodes and render graphs with the help of Rust's type system and Bevy's label system.
```

content/news/2024-02-03-bevy-0.13/index.md
```diff
@@ -126,9 +126,70 @@ TODO.
...
+```
+
+When you need dynamic labels for render nodes, those can still be achieved via e.g. tuple structs:
+
+```rust
+#[derive(Debug, Hash, PartialEq, Eq, Clone, RenderLabel)]
+pub struct MyDynamicLabel(&'static str);
+```
+
+The nice thing is, that this don't has to be a string, it can be any type, also numbers etc.
```
alice-i-cecile, line 192:
```suggestion
This is particularly nice because we don't have to store strings in here: we can use integers, custom enums or any other hashable type.
```

DasLixou:
Sorry, totally forgot about this. Also weird, I commited a change because I wrote `time` instead of `type` but GitHub seems to just be like "meh, don't care"

alice-i-cecile (approved):
Awesome, thank you!

PR #904, opened 2024-02-03
Improve table styling

Jondolf:
Currently, tables on news articles would look like this:

Old tables

This doesn't look great and discourages people from using tables in posts.

This PR adds styling to make tables look like this:

New tables


alice-i-cecile (approved):

sass/pages/_news.scss
```diff
@@ -23,6 +23,25 @@
         border-radius: $border-radius;
         max-width: 100%;
     }
+
+    table {
+        border-spacing: 0;
+        border: 2px solid $subtitle-color;
```
Jondolf, line 29: Using `$subtitle-color` is a bit strange here, but there isn't an existing color variable that would work

alice-i-cecile:
Non-blocking: could we apply this to the site as a whole, rather than just the News section? I'd like to be able to use tables in the book as well, and a consistent style would be better.

Jondolf:
Yep, I can make a `sass/elements/table.scss`. Not sure if there's some other better place

sass/site.scss
```diff
@@ -54,4 +55,4 @@
 @import "pages/features";
 @import "pages/content";
 @import "pages/migration_guide";
-@import "pages/404";
+@import "pages/404";
```
alice-i-cecile, line 58:
```suggestion
@import "pages/404";

```

Jondolf: My SCSS formatter doesn't like this for some reason, I need to configure that

TrialDragon:
Stupid question, but shouldn't this be targeting the 0.13-blog branch or whichever 0.13 branch it is based on? Maybe I'm missing something, but if this merged into main as it is, won't it also merge the WIP article in a public manner to the site?

Jondolf:
This doesn't have the article. It was just an example to visualize what it looks like

TrialDragon:
O, I see now. I got confused by the two different branches named quite similarly on your fork, one based on 0.13 stuff, another on main.

TrialDragon (approved):

PR #914, opened 2024-02-05
Fix a link error

cuppar:
Fix a link error


alice-i-cecile (approved):

alice-i-cecile:
Merging as trivial, thanks! We changed that recently and forgot to fix the link.

PR #915, opened 2024-02-05
Fix lines not being hidden in indented code blocks

rparrett:
# Objective

From Discord:

> doup — Today at 2:33 AM
> there is a CLI tool in the repo that adds the required zola annotations to the code blocks, go to write-rustdoc-hide-lines folder and run ./write_rustdoc_hide_lines.sh… it doesn't support indented code blocks (e.g. ones inside ul/ol lists)

I don't believe there are actually any code blocks like this currently in the repo. I added a test though, and tried it out on one of the book pages.

## Solution

Use a regex that accounts for whitespace rather than `starts_with` throughout the code.

## Discussion

Rust tests pass, but it would be nice to ensure this is producing identical output.

### DIFF
--- a/write-rustdoc-hide-lines/Cargo.lock
+++ b/write-rustdoc-hide-lines/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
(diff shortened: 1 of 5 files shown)


alice-i-cecile (approved):

rparrett:
Oops, this seems to be nuking attributes in code blocks.

rparrett:
Had a few more minutes due to car trouble. Added annotations to the main / indented tests cases and checked for identical output vs main.

doup (changes requested):
I found a small issue, should be easy to fix.

alice-i-cecile (approved):

doup (approved):

PR #852, opened 2024-01-07
Add bevy dev docs link to learn page

TrialDragon:
Adds a link to the bevy developer docs so that they're discoverable for anyone unaware.

Closes #775

Issue #910, opened 2024-02-04
Remove old book contributing section


TrialDragon:
All relevant contributing details for this repo should be moved to `CONTRIBUTING.md` and from there if need be reference main repo `CONTRIBUTING.md` for things such as code contributions in order to avoid maintaining duplicate resources.

PR #919, opened 2024-02-05
Link to dev-docs on contributing page

### LINKED ISSUE #775: Link to dev docs (for git main version of bevy)
The API docs hosted at https://dev-docs.bevyengine.org are quite undiscoverable. I only know about them from discord discussions? I am at a new computer and tried to find the link, because I don't have it in my browser history on this machine, and I could not find it anywhere. Google only links to docs.rs when searching for bevy dev docs, the website doesn't offer any way to reach it, I had a hard time finding it via discord, ...

Please consider adding a link to it somewhere on the website (probably on the "Learn") page.

BD103:
Closes #775.

This adds links on the contributing pages of the website, both for code and for docs. This is similar to #852, but I believe my approach avoids the confusion @mockersf warns about since it is targeted towards programmers developing the engine.

_I don't want to imply that #852 is bad, just that my approach is a compromise and easier to merge._

### DIFF
--- a/content/learn/quick-start/contributing/code/_index.md
+++ b/content/learn/quick-start/contributing/code/_index.md
@@ -16,6 +16,7 @@ Would you like to contribute code to Bevy?  Here's how!
 
 1. Fork the [`bevyengine/bevy` repository on GitHub][bevy], you'll need to create a GitHub account if you don't have one already.*
 2. Make your changes in a local clone of your fork
+   - You can view the unstable developer documentation [here](https://dev-docs.bevyengine.org), generated from the main branch on Github.
 3. For a higher chance of CI passing the first time, consider locally running `cargo run -p ci`. You can run the commands manually:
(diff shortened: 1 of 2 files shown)


TrialDragon:
Main concern is that since we're planning on removing the contributing section per #910 this would probably be removed within a week or two from now. Also since the `CONTRIBUTING.md` here is intended to only have info relevant to the site and book in order to minimize resource duplication this wouldn't fit properly (not sure there is a good reason to list the dev docs in a documentation or site related area since any links we need will usually be using latest stable, or earlier versions, via `docs.rs`). The intended place for this then would be the main `bevy` repo's `CONTRIBUTING.md`—since we direct code-related contributions to it in the site's `CONTRIBUTING.md—, but that does already have this link in it's docs section.

BD103:
That's a fair point, especially if the contributing section is going to be removed. I'll close this then, if no one else has any objections.

PR #921, opened 2024-02-05
Use Environmental Files instead of `set-output` command

### LINKED ISSUE #920: Migrate CI from `set-output` to Environmental Files
The [`generate-assets`](https://github.com/bevyengine/bevy-website/blob/22c34147ca28c6037f98ca2424ff978942ac8955/.github/workflows/ci.yml#L12) jobs in CI raises a warning because it uses the `set-output` command, while it should be migrated to Environmental Files. `set-output` is used [here](https://github.com/bevyengine/bevy-website/blob/22c34147ca28c6037f98ca2424ff978942ac8955/.github/workflows/ci.yml#L28-L30). Here is a blog post on [migrating to Environmental Files](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and the docs on [workflow commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files).

image

BD103:
Closes #920. This should fix the final warning emitted by the CI. :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Get cache key
         id: cache-key
-        run: echo "::set-output name=key::$(date +'%Y-%m-%d')"
+        run: echo "key=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
 
       - name: Get crates.io datadump from cache
         uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


BD103:
Just forced pushed because I didn't realize `deploy.yml` also uses `set-output`. It should be all good now :)

alice-i-cecile (approved):

PR #922, opened 2024-02-06
Generate Zola Rust code block hidden lines annotations in the book

TrialDragon:
Generates the hidden lines Zola annotation in the book and quick-start and adds a bit to `CONTRIBUTING.md` so that contributors know that this is a tool in their belt to be used.

Fixes #911

PR #923, opened 2024-02-06
Improve write-rustdoc-hide-lines

### LINKED ISSUE #924: Ensure that code block lines have been hidden in CI
These are an important QC check for content. See [this advice](https://github.com/bevyengine/bevy-website/pull/922/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R176) from #921 for bread crumbs on how to run this step.

Running this operation should result in no change in our files. A similar check exists in the main Bevy repo for validating generated examples.

BD103:
Closes #924.

There are a few things I would like to accomplish in this PR:

- [x] Allow passing multiple folder paths to `write-rustdoc-hide-lines`
  - This prevents PRs like #922 from having to write out `cargo run ...` multiple times (96759b7)
- [x] Allow running `write_rustdoc_hide_lines.sh` from outside its current directory.
- [x] Allow running `write-rustdoc-hide-lines` in check mode, so it can be used in CI to ensure all Markdown files are up to date before merging.


rparrett:
This workflow from the bevy repo is a way of getting the `check` command for "free": https://github.com/bevyengine/bevy/blob/main/.github/workflows/ci.yml#L285 by checking out the repo, running the tool, and checking `git diff`.

BD103:
I believe this PR is feature-complete. The diff for `main.rs` and `formatter.rs` is pretty messy, since I heavily modified / redid a lot of things. You can now either format or check multiple folder paths. I changed the program to check the entire website instead of just the book, but that caused some weird invisible changes to the migration guides and various other files.

I improved the terminal output, but I eventually want to add optional color output. That will likely be a separate PR, though, since this one is large enough.

Please reach out if you need any explanations for why I did certain things! You can also check out the [commits tab](https://github.com/bevyengine/bevy-website/pull/923/commits), I usually add longer descriptions and reasonings for individual changes there.

BD103:
In order to make reviewing this change easier, I reverted d28809e because of line-ending inconsistencies. #939 should be merged before this. Until then, the CI will fail.

TrialDragon (changes requested):
Looked over all the files and ran it on local. It all seems to work good, even when dealing with page files rather than sections which was my main concern if that broke or had issues.

(Maybe these changes I've suggested are too much for this PR and maybe better on their own, hmm.)
One maybe change I suggest is updating the `write-rustdoc-hide-lines/README.md` for the tool to reflect the facts:
1. You can `check` the files via `cargo run`
2. That the `.sh` script can not be passed directories or files as command args (in case people think you're supposed to use it for that rather than `cargo run`).
3. That, at this time, the `.sh` script runs across all `.md` files in the `content` directory, not just the `content/learn/book` directory like it did before.
4. That `cargo run` is available for formatting as well (or just period; I guess this is sort of the same as point 1 but rephrased) if you need more specific targeting of directories and that there is a difference between `cargo run` and the `.sh` script.
5. Also mention that it goes for the lines with a `#` and a space , not just `#`.
6. Mention it also works on `rs` code blocks, not just `rust`.

And that's all I can think of at this moment.

BD103:
I added your suggestions to `README.md` in aa12eeb6371653b874dc7fd49393c94ed18aad21. (Good idea!) Please tell me if you want me to rephrase something. :)

alice-i-cecile (approved):

BD103:
Any idea how to silence [this lint](https://github.com/bevyengine/bevy-website/actions/runs/7919568079/job/21620641402?pr=923#step:4:104)? It doesn't like `# ` because there is a space.

rparrett:
That seems like an okay lint to me overall. I would disable it for that line with
`<!-- markdownlint-disable-next-line MD038 -->`.

TrialDragon (approved):

PR #927, opened 2024-02-06
Stop public draft pages from being indexed by search engines, hopefully

### LINKED ISSUE #926: Keep public draft pages out of search engine indices
We tried to do this with `robots.txt` and disallowing pages, however it has come to attention that this was not the proper way and didn't have the intended effect. Instead each public draft page should have metatags to make them not indexed. `<meta name="robots" content="noindex">` for stopping plain indexing, and `<meta name="robots" content="noarchive">` for caching a link to a search result. 

Both Google and Bing seem to support this, unknown if Safari does. Should stop most indexing of these pages.

TrialDragon:
Removes the apparently useless robots.txt and adds in meta tags that should, hopefully, block indexing by the major search engines like Google, Bing, DuckDuckGo, or Safari.

Note: The hopefully in the title is because I have no means of testing this prior to deployment.

Fixes #926


NthTensor (commented):
looks reasonable. may as well leave the sitemap, idk.

TrialDragon:
> looks reasonable. may as well leave the sitemap, idk.

Sitemap is iffy. Like looking into it further it doesn't change the actual indexing occurred, just discoverability by crawlers by a small margin on occasion. Any of the specific useful stuff that it does do is already handled by Zola, like keeping error links, redirects, and duplicates out.

NthTensor (approved):

nicopap:
I would recommend keeping the `robots.txt`.

RomainMazB:
Yeah, I agree with @nicopap .

Even if it's not the #1 robots.txt we have seen, it's not accurate to delete it to de-index the pages.

For the other stuff, LGTM and as you said @TrialDragon : we only can validate the success of this modification when some of the indexed paged will be removed from the search engine.

We should keep track of some of them but it can be long. Weeks or months, it depends of the crawler frequency.

TrialDragon:
> Even if it's not the #1 robots.txt we have seen, it's not accurate to delete it to de-index the pages.

The main reason I was going to do so was because Zola's default robots.txt it adds in when there is no custom robots.txt would work just fine, allowing all crawlers which should allow the search engines to see not to index the public draft pages via the metatags. However I've gotten disagreement over the decision overall to remove the custom robots.txt even tho I personally feel it may hinder removing the public draft pages from index since it would prevent the search engine crawlers from checking out the pages and removing them from the index per the noindex tag and overall use of robots.txt for prevention of indexing is considered a footgun / unintended and faulty way to solve that issue by the major search engines with the noindex tag, and official tools provided by each search engine to remove pages, being the official solution(s).

So I'll revert that part and I guess open another issue regarding robots.txt and the fixing of it since it is broken at the moment or overall deciding what to do with it so that it doesn't block this PR.

BD103:
[robots.txt](https://github.com/bevyengine/bevy-website/files/14202781/robots.txt)

If you look at the generated `robots.txt` file, there appears to be a lot of weird whitespace and indentation. Perhaps its not working properly because the formatting is wrong.

```
User-agent: *
Sitemap: https://bevyengine.org/sitemap.xml

     # 4 spaces
         # 8 spaces
             # 12 spaces
     # 4 spaces
         # 8 spaces
     # 4 spaces
        Disallow: /learn/advanced-examples/
```

NiseVoid (approved):
Code seems fine, the meta tag looks correct, I tested it and the tags get served correctly ... I don't think we actually need the `noarchive` but it doesn't do any harm either

PR #931, opened 2024-02-08
Fix broken fast build file link

Henriquelay:
Sorry, I didn't fully read CONTRIBUTING.md)but this is a very trivial link path change. Just spotted it as I was going through get-started.


rparrett (approved):

BD103 (approved):
Thanks!

PR #932, opened 2024-02-08
Added a comment for linux users trying the examples

### LINKED ISSUE #894: Following the quick-start guide led me to a cargo compile error
## How can Bevy's documentation be improved?
Please have the "quick-start" mention everything I need to get started, including a link to any dependencies. Otherwise, it's a slow start, and the main Setup page would have been a quicker start.

For context, I'm brand new to Rust, and decided to try porting a medium-complexity project into Bevy for my first project&nbsp;&mdash; as one does. Also, I'm using an ancient version of Ubuntu (20.04 Focal), so that's a potential factor. However, I'm confident that even the latest Ubuntu won't install all the dependencies for me.

Below is the issue I had. I figured out how to solve it, but it shattered my expectations for a quick start.

This [Quick Start](https://bevyengine.org/learn/book/getting-started/#quick-start) didn't work for me. I got the following error when running `git checkout v0.12.1 && cargo run --example breakout`:

```
  pkg-config exited with status code 1
  > PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags alsa

  The system library `alsa` required by crate `alsa-sys` was not found.
  The file `alsa.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
  The PKG_CONFIG_PATH environment variable is not set.
```
Here's some logging context: 
[build-crash.txt](https://github.com/bevyengine/bevy/files/14076006/build-crash.txt)

Clearly, I need to install the dependencies. I believe a link to [Installing Linux Dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) is warranted here, to avoid new users giving up too soon.

~~Also, the Linux dependencies documentation mentions alsa-lib for Fedora, but not Ubuntu. I'm sure it's an oversight, but I'm not sure which versions of Ubuntu you want to support. For my older system (Ubuntu 20.04), I solved this issue by installing `libalsaplayer-dev`, which [exists in _some_ more recent Ubuntu releases, including Jammy, the latest LTS][some releases] but [not in the latest release, Ubuntu 22.10 Kinetic][latest-ubuntu]. I'm sure the name has simply changed.~~

Actually, it looks like the `libasound2-dev` dependency, which is mentioned in the docs, does include the right files. My "Ubuntu" instincts failed me! :-) I'm glad I found the Linux Dependencies documentation.

I'm going to check for other hidden dependencies, and update if I find any not listed on the dependencies page.

[some releases]: https://packages.ubuntu.com/search?keywords=libalsaplayer-dev&searchon=names&suite=all&section=all
[latest-ubuntu]: https://packages.ubuntu.com/search?keywords=libalsaplayer-dev&searchon=names&suite=kinetic&section=all

BrunaoW:
Fixes #894 

When trying Bevy for the first time through the [Quick Start - Getting started](https://bevyengine.org/learn/quick-start/getting-started/) guide in a linux distro, you will get some errors when trying to `cargo run` the examples. That's because the dependencies listed [here](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) are needed for Bevy, and maybe not be installed in the user's system, which was in my case.

### DIFF
--- a/content/learn/quick-start/getting-started/_index.md
+++ b/content/learn/quick-start/getting-started/_index.md
@@ -45,6 +45,8 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
     cargo run --example breakout
     ```
 
+    If you are using a Linux distro, [install the dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) needed for bevy before trying the examples
+
 ### Add Bevy as a Dependency
 
 Bevy is [available as a library on crates.io](https://crates.io/crates/bevy).
--- a/content/learn/quick-start/getting-started/_index.md
+++ b/content/learn/quick-start/getting-started/_index.md
@@ -45,7 +45,7 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
     cargo run --example breakout
     ```
 
-    If you are using a Linux distro, [install the dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) needed for bevy before trying the examples
+    If you are using Linux, be sure to [install the dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) needed for Bevy before trying the examples.
 
 ### Add Bevy as a Dependency
 


content/learn/quick-start/getting-started/_index.md
```diff
@@ -45,6 +45,8 @@ Note: the "fast compiles" setup is on the next page, so you might want to read t
     cargo run --example breakout
     ```
 
+    If you are using a Linux distro, [install the dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) needed for bevy before trying the examples
```
alice-i-cecile, line 48:
```suggestion
    If you are using Linux, be sure to [install the dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) needed for Bevy before trying the examples.
```

alice-i-cecile (approved):

rparrett:
> ... before trying the examples

This phrasing feels slightly odd since this is **after** we instructed the reader to run the examples.

IMO, we should adjust the wording under the `## Quick Start` header instead. It mentions a "a working Rust setup" but that isn't enough on Linux.

alice-i-cecile:
I went with the approach in #1014. Thank you!

PR #936, opened 2024-02-11
Fix content in blog

DasLixou:
Im not sure if I put the ui text to the right title, pls check before merge


alice-i-cecile (approved):
Looks correct, thanks!

PR #912, opened 2024-02-05
Remove bors.toml

BD103:
I believe Bors is no longer being used for the website (and [all of Bevy](https://bevyengine.org/news/bevy-0-11/#no-more-bors)). This PR deletes a leftover `bors.toml` file.

Additionally, there appear to be a few other references to Bors in the codebase. Do you want me to remove them too?

- [`github_client::get_merged_prs`](https://github.com/bevyengine/bevy-website/blob/0a3c92f6163797661d28a442b7200d473e7db73d/generate-release/src/github_client.rs#L165-L168)
- [`github_client::get_merged_prs_by_page`](https://github.com/bevyengine/bevy-website/blob/0a3c92f6163797661d28a442b7200d473e7db73d/generate-release/src/github_client.rs#L227-L233)
- [`helpers::get_pr_title_from_commit`](https://github.com/bevyengine/bevy-website/blob/0a3c92f6163797661d28a442b7200d473e7db73d/generate-release/src/helpers.rs#L58-L64)
- [`main::Args`](https://github.com/bevyengine/bevy-website/blob/0a3c92f6163797661d28a442b7200d473e7db73d/generate-release/src/main.rs#L18-L28)
- [`release_notes_website::generate_release_notes_website`](https://github.com/bevyengine/bevy-website/blob/0a3c92f6163797661d28a442b7200d473e7db73d/generate-release/src/release_notes_website.rs#L39-L44)
- [`release_notes::generate_release_note`](https://github.com/bevyengine/bevy-website/blob/0a3c92f6163797661d28a442b7200d473e7db73d/generate-release/src/release_notes.rs#L39-L44)

PR #939, opened 2024-02-12
Normalize line endings

BD103:
MacOS / Linux represents newlines with the `\n` LF character, while Windows represents them with the `\r\n` CRLF characters. This repository uses a mixture of both, which led to formatting issues in #912 (See https://github.com/bevyengine/bevy-website/pull/912#pullrequestreview-1876172575, 7800a5e9716512df29afa2133ca44d06bd9b535c).

This PR creates a `.gitattributes` file which enforces a consistent line ending, no matter the OS. The file is copied directly from the main [Bevy repository](https://github.com/bevyengine/bevy/blob/2d90b2093adae679377c3273ff799dc9f3979853/.gitattributes), inspired greatly by [a guide in Github's docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings).

Additionally this fixes the existing line endings in 87c3193509c69c1257c49b7f3db5b305d1e3a17f using:

```shell
$ git rm -rf --cached .
$ git reset --hard HEAD
```

I do not know whether this will cause merge conflicts or issues during the transition period. Does anyone else have experience with this? Additionally, should Markdown files be explicitly set to `LF` or `CRLF`, or is `auto` fine?


alice-i-cecile:
Let's do LF for `md` files, just arbitrarily.

BD103:
> Let's do LF for `md` files, just arbitrarily. I think it's nice to standardize this, and I really don't have strong feelings.

Done. The diff (835cc3b7ea0aef4ca05b7b795c22e3ab255360a0) surprisingly does not affect many Markdown files, I thought more than 4 would be affected. :v

alice-i-cecile (approved):

.gitattributes
```diff
@@ -0,0 +1,20 @@
...
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.rs text eol=lf
+*.toml text eol=lf
+*.frag text eol=lf
+*.vert text eol=lf
+*.wgsl text eol=lf
```
JMS55, line 11: Probably want scss too

.gitattributes
```diff
@@ -0,0 +1,20 @@
...
+*.md text eof=lf
+
+# Declare files that will always have CRLF line endings on checkout.
+*.sln text eol=crlf
```
IceSentry, line 15: I don't think we need this. I doubt we'll ever have a .sln in this repo

.gitattributes
IceSentry: should probably add html and js too

.gitattributes
BD103: I agree, fixed in e32cf6a14d21b404ff09af0a405af4c986d86795.

BD103:
Here's a list of files currently used in the repository:

```
{
    "png",
    "xml",
    "rs",
    "sh",
    "yml",
    "js",
    "jpg",
    "html",
    "mjs",
    "svg",
    "pack",
    "eot",
    "idx",
    "lock",
    "toml",
    "md",
    "webp",
    "woff2",
    "ttf",
    "gif",
    "woff",
    "rev",
    "txt",
    "sample",
    "scss",
    "mp4",
}
```

<details>
  <summary>Generated with this program</summary>

```rust
use std::{fs, path::Path, collections::HashSet};

fn main() {
    let mut file_types = HashSet::new();

    visit_dir(Path::new("."), &mut |path| {
        match path.extension() {
            Some(ext) => {
                file_types.insert(ext.to_os_string());
            },
            None => {
                // eprintln!("No extension: {:?}", path);
            },
        }
    });

    println!("{:#?}", file_types);
}

fn visit_dir(dir: &Path, cb: &mut dyn FnMut(&Path)) {
    if !dir.is_dir() {
        return;
    }

    for entry in fs::read_dir(dir).unwrap() {
        let entry = entry.unwrap();
        let path = entry.path();

        if path.is_dir() {
            visit_dir(&path, cb);
        } else {
            cb(&path);
        }
    }
}
```

</details>

The code may have caught some generated files, so we probably don't need to set the line endings for every file type.

BD103:
I added every tracked file type to `.gitattributes`, though I don't know if it is good practice to list them all.

If this PR is successful, then this could also be applied to the main Bevy repository. (Based on a message in [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1206723467823550484) from someone whose name I cannot pronounce.)

kristoff3r (approved):

PR #893, opened 2024-01-25
Merge `new-book` into `main`

TrialDragon:
Merges the `new-book` branch into `main` resolving conflicts along the way. 

Closes #870 
Closes #602

PR #951, opened 2024-02-14
Fix extra whitespace in example code block

rparrett:
I think this indentation was added mistakenly in #893. It's already correct in `example-webgpu.html`.

### Before / After
image


alice-i-cecile (approved):

alice-i-cecile:
Merging as trivial, thanks.

PR #956, opened 2024-02-15
Fix CRLF line endings for markdown lint files

BD103:
These files were not caught by #939, surprisingly enough.


BD103:
@alice-i-cecile, would you mind reviewing?

alice-i-cecile (approved):
Merging as trivial, thanks

PR #962, opened 2024-02-15
Fix migration guide page navigation ordering consistency

### LINKED ISSUE #863: Migration guide page navigation and docs menu ordering inconsistency
While working on #862 I have noticed that in the migration guide the page navigation is going in the normal order in spite of the docs page menu being reversed. This causes a weird inconsistency where the next page is marked as the previous.

The page navigation will need a few if blocks to reverse the order of subsections or pages if it is a migration guide page methinks. Should be a relatively simple fix.

TrialDragon:
Changes the ordering of the navigation guide pages to be reversed since the weights are reversed for contributor usability.

Fixes #863


alice-i-cecile (approved):

BD103 (approved):
Looks good, though the description of this PR is a bit confusing.

For future reviewers, this fixes the "Previous" and "Next" buttons at the bottom of the migration guide so it matches with the navigation list. :)

Before:

image

After:

image

PR #966, opened 2024-02-16
Let shell scripts be called from any directory

BD103:
This calls:

```shell
$ cd $(dirname $0)
```

at the beginning of all shell scripts in this repository. This guarantees the assumption that a script will be operating in the same directory it lives in. Namely, the following commands will do the same thing:

```shell
(bevy-website) $ ./generate-assets/generate_assets.sh
(bevy-website/generate-assets) $ ./generate_assets.sh
```

Technically a portion of this was implemented for specifically `write-rustdoc-hide-lines` in #923, but that is not yet merged and this widens the change to all scripts. There will be merge conflicts, but I'll fix them. :)

<details>
  <summary>But what is the command actually doing?</summary>

```shell
$ cd $(dirname $0)
```

Let's break it down.

- `cd` stands for "change directory." It changes the current directory to whatever path is given to it.
- `$(...)` is an evaluation block. It will execute any command within it, then return the output as a string.
- `dirname` accepts a file path and returns the full path to the directory of that file. For instance, `dirname path/to/foo.txt` would return `path/to`.
- `$0` is a special variable that represents the path to the current script being executed. If `./bin/dostuff.sh` was executed, `$0` would be `./bin/dostuff.sh`.

So in summary, the command is finding the folder of the script being run and `cd`-ing into it. Tada! :D

</details>

### DIFF
--- a/generate-community/generate_community.sh
+++ b/generate-community/generate_community.sh
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+# Switch to script's directory, letting it be called from any folder.
+cd $(dirname $0)
+
 git init bevy-community
 cd bevy-community
 git remote add origin https://github.com/bevyengine/bevy-community
--- a/generate-community/generate_community.sh
+++ b/generate-community/generate_community.sh
@@ -3,6 +3,8 @@
 # Switch to script's directory, letting it be called from any folder.
 cd $(dirname $0)
 
+# Download a copy of the Bevy community repository.
+# FIXME: Can this be shortened to git clone --depth=1?
 git init bevy-community
 cd bevy-community
 git remote add origin https://github.com/bevyengine/bevy-community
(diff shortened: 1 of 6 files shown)


alice-i-cecile (approved):
Nice explanation! I really love seeing that sort of stuff in PR descriptions: it really helps the team learn.

generate-community/generate_community.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Switch to script's directory, letting it be called from any folder.
+cd $(dirname $0)
+
+# Download a copy of the Bevy community repository.
+# FIXME: Can this be shortened to git clone --depth=1?
```
BD103, line 7: In my quick testing, using the current method only downloads 0.02 KB less of data. I have no idea what that data is, though. Are the small performance benefits worth a more complicated script?

alice-i-cecile: Probably not currently no.

TrialDragon (approved):
LGTM

PR #968, opened 2024-02-16
Add public draft support to news section

TrialDragon:
Adds public draft support to the news section so that we can have the release notes as public draft and be incrementally worked on during a release cycle rather than leave it as a huge chore for the end.

Relevant discord discourse (and it's surrounding messages): https://discord.com/channels/691052431525675048/692572690833473578/1208084275069722664

Related to #969

### DIFF
--- a/templates/news.html
+++ b/templates/news.html
@@ -4,6 +4,7 @@
 {% block content %}
 <div class="padded-content item-grid">
   {% for page in section.pages %}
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

cart (approved):
I'm very down to experiment with this process!

Issue #396, opened 2022-07-18
Add accessibility lints to CI


TimJentzsch:
I'm not sure if there is an easy way to add this to Zola projects, but it would be nice to have basic accessibility checks in CI.
Stuff like "every `img` element has an `alt` attribute" should be enough for the start.

If Zola has anything built-in we could use that, otherwise maybe it's possible to do the linting on the generated HTML?

TrialDragon:
Did a little bit of research on this. [`pa11y`](https://github.com/pa11y/pa11y) seems to be a good candidate and has a version explicitly for CI: [`pa11y-ci`](https://github.com/pa11y/pa11y-ci).

I've done a local test with the tool and once installed it does work to catch accessibility issues on the site; technically more than even the Firefox accessibility tools showed me. Showed missing image alt text and told me the recommended color to use for each contrast issue.

PR #970, opened 2024-02-16
Make the site higher contrast for accessibility

### LINKED ISSUE #3: Improve accessibility by using higher contrast on the website.

TrialDragon:
Makes the site higher contrast to address accessibility issues with the site.

Going thru this I think #396 is a good idea to catch any accessibility issues since at the moment I'm just checking the Firefox accessibility tools manually.

Fixes #3


TrialDragon:
A side note for myself; should probably do a cleanup of the scss at some point. Remove duplicates, centralize more universal variables in `_vars.scss` (mainly the blue color variable, makes no good sense for it to be limited to the buttons if it's meant to be a general blue color, especially when the pink color for donate already is in `_vars.scss`), et cetera.

alice-i-cecile:
@TimJentzsch if you have time and interest I'd love your opinion on this: you did a great job helping me out with website a11y in the past.

cart:
Personally not a huge fan of the new button palette ("donate" and "get started"). Feels a bit jarring and out of place with the other colors on the site. "Balanced saturation/lightness" does to a degree imply "less accessible", but I suspect we can find something that fits a bit better.

Old:
image
New:
image

cart:
(I do support this general idea though)

TrialDragon:
Hmm; I modified it again to make the borders stand out less.
Before: 
Screenshot from 2024-02-16 21-24-43
After: 
Screenshot from 2024-02-16 21-25-56

I'll see if I can make it a bit less saturated so it doesn't pop out in a jarring way like it does now.

TrialDragon:
Okay, here it is a bit more desaturated, but not too much it starts veering into hard to read again.
Screenshot from 2024-02-16 21-35-29

TrialDragon:
Okay, I think I've got all the contrast issues based on manual checking. Should note that until #973 is merged the anchor links will still show as having issues.

TimJentzsch:
> A side note for myself; should probably do a cleanup of the scss at some point. Remove duplication, centralize more universal variables in `_vars.scss` (mainly the blue color variable, makes no good sense for it to be limited to the buttons if it's meant to be a general blue color, especially when the pink color for donate already is in `_vars.scss`), et cetera.

Agreed. I think it would also be better to use semantic variable names like primary and secondary instead of stuff like blue and pink, that would also make it easier to introduce themes (e.g. a light theme).

TimJentzsch:
> Personally not a huge fan of the new button palette ("donate" and "get started"). Feels a bit jarring and out of place with the other colors on the site. "Balanced saturation/lightness" does to a degree imply "less accessible", but I suspect we can find something that fits a bit better.
> 
> Old: image New: image

I made a website for exactly this issue: Finding a higher contrast color while staying as close as possible to the original: <https://color-id.timjen.net/?color=%23bb799c&secondary=%23ececec#accessibility>

Looks like it suggests `#9f517a` if we want to keep the `#ececec` text color in this case, for AA contrast.

TrialDragon:
Okay, here is how it looks after I used color-id.timjen.net to get the colors for buttons.
image

TrialDragon:
Also reverted the white color change since I remembered off-white is preferable.

TimJentzsch (approved):
Personally, I like the changes.
Accessibility by default is always good to pursuit and this would also make it better e.g. when viewing the page on mobile in bright sunlight.

cart (approved):
Alrighty I think this is in a reasonable spot. I opted to lighten the border instead of darken it because I think it makes the buttons pop / feel less drab. Thanks for putting this together!

cart:
Actually one sec lets not merge yet. I've just noticed the forestgreen image-comparison color :laughing:

TrialDragon:
> Actually one sec lets not merge yet. I've just noticed the forestgreen image-comparison color 😆

Feel free to change it to whichever color—that is generically contrasting enough—that suits your fancy. That was just the one I found that seemed to work; not attached to it.

cart:
I strongly suspect that it is somehow confused due to how we are rendering that text. The text is white with a black outline, which should make it pop in pretty much every scenario. The "best case" scenario (black background, white text) fails the firefox test contrast test, despite being significantly _more_ contrast-ey than the rest of the site:

image

I'm going to revert this change for now in the interest of moving forward quickly. If you want to make your case for changing this one, lets do it in a separate PR.

TrialDragon:
Alright. I suspect it will be forced to be dealt with once accessibility linting is added; in that case CI will complain about it until a solution is found.

PR #978, opened 2024-02-17
Add tools to Cargo workspace

BD103:
This adds all of the tools written in Rust to a [Cargo workspace], which centralizes `Cargo.lock`, `target/`, lints, and other configuration.

[Cargo workspace]: https://doc.rust-lang.org/cargo/reference/workspaces.html

Originally discussed [on Discord](https://discord.com/channels/691052431525675048/695741366520512563/1208202278046015508).

PR #979, opened 2024-02-17
Remove deprecated `chrono` usage

### LINKED ISSUE #977: `generate-release` uses deprecated `chrono` functions
```shell
$ cargo check
warning: use of deprecated method `chrono::NaiveDate::and_hms`: use `and_hms_opt()` instead
   --> generate-release/src/github_client.rs:172:76
    |
172 |         let naive_datetime = NaiveDate::parse_from_str(since, "%Y-%m-%d")?.and_hms(0, 0, 0);
    |                                                                            ^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `chrono::DateTime::<Tz>::from_utc`: Use TimeZone::from_utc_datetime() or DateTime::from_naive_utc_and_offset instead
   --> generate-release/src/github_client.rs:173:45
    |
173 |         let datetime_utc = DateTime::<Utc>::from_utc(naive_datetime, Utc);
    |                                             ^^^^^^^^
```

BD103:
Fixes #977 and unblocks #978.

- `NaiveDate::and_hms` is deprecated in favor of `NaiveDate::and_hms_opt`, which returns an `Option` instead of panicking.
- `DateTime::from_utc` is deprecated in favor of `TimeZone::from_utc_datetime`, which I believe does the same thing with a few extra checks.

The deprecated warnings are not shown in CI because `generate-release`'s `Cargo.lock` file is out of date. The reason the error showed up in #977 is because is regenerated the lockfile, causing dependencies to be updated.

### DIFF
--- a/generate-release/src/github_client.rs
+++ b/generate-release/src/github_client.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 
 use anyhow::bail;
-use chrono::{DateTime, NaiveDate, Utc};
+use chrono::{DateTime, NaiveDate, TimeZone, Utc};
 use serde::Deserialize;
 
 #[derive(Deserialize, Clone, Debug)]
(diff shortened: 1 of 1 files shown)


TrialDragon (approved):
LGTM; it uses functions that exists in the current version, so nothing breaks.

alice-i-cecile (approved):

james7132 (approved):

PR #986, opened 2024-02-17
0.13: Fix author attribution in Entity Opt/Query for_each/TableRow sections

Bluefinger:
Fix for #891 

Does what it says on the tin.


alice-i-cecile (approved):
Thank you! I must have messed this up when refactoring. Merging as trivial.

PR #991, opened 2024-02-17
Update "Unload render assets from RAM"

afonsolage:
Closes #984.

Sending PR to correct branch, since #987 was on outdated branch.

PR #992, opened 2024-02-17
Removed outdated statements about asset_usage

brianreavis:
Incorporates #991 and includes additional updates to address outdated information.

### DIFF
--- a/content/learn/migration-guides/0.12-to-0.13.md
+++ b/content/learn/migration-guides/0.12-to-0.13.md
@@ -1222,7 +1222,8 @@ The increased `Exposure::ev100` means that all existing 3D lighting will need to
     <div class="migration-guide-area-tag">Assets</div>
 </div>
 
-- Asset loaders (GLTF, etc) now load meshes and textures without `asset_usage`. These assets will be removed from `Assets<Mesh>` and `Assets<Image>` once `RenderAssets<Mesh>` and `RenderAssets<Image>` contain the GPU versions of these assets, in order to reduce memory usage. If you require access to the asset data from the CPU in future frames after the GLTF asset has been loaded, modify all dependent `Mesh` and `Image` assets and set `asset_usage` to `RenderAssetUsages::default()`, which implies `RenderAssetUsage::MAIN_WORLD | RenderAssetUsage::RENDER_WORLD`.
+Render assets can now be automatically unloaded from the main world after being prepared for rendering in the render world. This is controlled using `RenderAssetUsages` bitflags. To mimic old behavior and keep assets around in the main world, use `RenderAssetUsages::default()`.
+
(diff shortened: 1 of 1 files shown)


afonsolage:
The main branch already has this modifications, I added this on #991. Or did I miss something?

brianreavis:
> - Asset loaders (GLTF, etc) now load meshes and textures without `asset_usage`. These assets will be removed from `Assets<Mesh>` and `Assets<Image>` once `RenderAssets<Mesh>` and `RenderAssets<Image>` contain the GPU versions of these assets, in order to reduce memory usage. If you require access to the asset data from the CPU in future frames after the GLTF asset has been loaded, modify all dependent `Mesh` and `Image` assets and set `asset_usage` to `RenderAssetUsages::default()`, which implies `RenderAssetUsage::MAIN_WORLD | RenderAssetUsage::RENDER_WORLD`.

This statement is inaccurate for two reasons:

1. Assets aren't loaded by default anymore (https://github.com/bevyengine/bevy/pull/11212)
2. "Asset loaders (GLTF, etc) now load meshes and textures without `asset_usage`" doesn't really make sense.

@alice-i-cecile Can you reopen this?

alice-i-cecile:
Ah, got it. Can you resolve merge conflicts so we can see the true diff?

brianreavis:
All set

alice-i-cecile (approved):
Ah much clearer now. Thanks!

content/learn/migration-guides/0.12-to-0.13.md
```diff
@@ -1231,7 +1232,7 @@ The increased `Exposure::ev100` means that all existing 3D lighting will need to
   - The trait now requires `Clone`.
   - The `ExtractedAsset` associated type has been removed (the type itself is now extracted).
   - The signature of `prepare_asset()` is slightly different
-  - A new `asset_usage()` method is now required (return `RenderAssetUsage::RENDER_WORLD` to match the previous behavior).
+  - A new `asset_usage()` method is now required (return `RenderAssetUsages::default()` to match the previous behavior).
```
afonsolage, line 1235: The original migration guide stated the `asset_usage` should return `RenderAssetPersistencePolicy::Unload` to match the previous `AssetRender` behavior, which maps to `RenderAssetUsage::RENDER_WORLD` and then unloading the asset from CPU.

By changing it to `RenderAssetUssages::default()` the asset will be kept both in CPU and GPU. This was a mistake on original PR or something has changed in another PR?

afonsolage (changes requested):

content/learn/migration-guides/0.12-to-0.13.md
rparrett: Seems like a mistake in the original guide to me.

afonsolage (approved):

PR #993, opened 2024-02-17
Explicit add deprecated shapes on migration guide

### LINKED ISSUE #989: Migration Guide 0.13 - `Capsule::depth` -> `Capsule3d::half_length`
- bevyengine/bevy/pull/10466 Added new shapes and included new `Capsule`
- bevyengine/bevy/pull/11585 Renamed `Capsule` to `Capsule3d`
- bevyengine/bevy/pull/11773 Deprecated all shapes in `bevy_render::mesh::shape`

So at the end, the old `Capsule` from `bevy_render::mesh::shape` is totally different from `Capsule3d`. I don't think the deprecation warning is enough. Maybe should we explicit add something like bellow?
- `bevy_render::mesh::Capsule` is deprecated use `bevy_math::primitives::dim3::Capsule3d` instead;
- `bevy_render::mesh::Cylinder` is deprecated use `bevy_math::primitives::dim3::Cylinder` instead;
- `bevy_render::mesh::Icosphere` is deprecated use `bevy_math::primitives::dim3::Sphere` instead;
..etc

afonsolage:
Closes #989 

The idea is to help users when they are just searching for their deprecated primitives, which will help to find the PR that introduced the change.

### DIFF
--- a/content/learn/migration-guides/0.12-to-0.13.md
+++ b/content/learn/migration-guides/0.12-to-0.13.md
@@ -1153,6 +1153,18 @@ If you were using the pipelined rendering channels, `MainToRenderAppSender` and 
 
 Bevy has previously used rendering-specific types like `UVSphere` and `Quad` for primitive mesh shapes. These have now been deprecated to use the geometric primitives newly introduced in version 0.13.
 
+- `bevy_render::mesh::Capsule` is deprecated use `bevy_math::primitives::dim3::Capsule3d` instead;
+- `bevy_render::mesh::Cylinder` is deprecated use `bevy_math::primitives::dim3::Cylinder` instead;
+- `bevy_render::mesh::Icosphere` is deprecated use `bevy_math::primitives::dim3::Sphere` instead;
+- `bevy_render::mesh::Cube` is deprecated use `bevy_math::primitives::dim3::Cuboid` instead;
+- `bevy_render::mesh::Box` is deprecated use `bevy_math::primitives::dim3::Cuboid` instead;
+- `bevy_render::mesh::Quad` is deprecated use `bevy_math::primitives::dim2::Rectangle` instead;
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

content/learn/migration-guides/0.12-to-0.13.md
```diff
@@ -1153,6 +1153,18 @@ If you were using the pipelined rendering channels, `MainToRenderAppSender` and
 
 Bevy has previously used rendering-specific types like `UVSphere` and `Quad` for primitive mesh shapes. These have now been deprecated to use the geometric primitives newly introduced in version 0.13.
 
+- `bevy_render::mesh::Capsule` is deprecated use `bevy_math::primitives::dim3::Capsule3d` instead;
```
TrialDragon, line 1156: Shouldn't this also mention `bevy_math::primitives::dim3:Capsule2d` since from my understanding `bevy_render::mesh::Capsule` was dimension-agnostic and could work for both 2d and 3d—even if primarily intended for 3d—, and was probably used by someone for 2d? Or is this unwanted since there is already a minor section mentioning that Capsule is now Capsule3d and any 2d uses should use Capsule 2d?

afonsolage: The deprecated [Capsule](https://github.com/bevyengine/bevy/blob/release-0.12.1/crates/bevy_render/src/mesh/shape/capsule.rs) was intended for 3d only, was we can see on [`Into<Mesh>` impl](https://github.com/bevyengine/bevy/blob/22e39c4abf6e2fdf99ba0820b3c35db73be71347/crates/bevy_render/src/mesh/shape/capsule.rs#L47)

TrialDragon (approved):

PR #1017, opened 2024-02-18
Update screenshot job

mockersf:
This is needed to update the example screenshots on the website:
- add a new page as there are more examples
- remove temp workaround not needed anymore
- fix dependencies


TrialDragon (approved):

BD103 (approved):
Looks good! What was the reason for the workaround in the first place?

mockersf:
> Looks good! What was the reason for the workaround in the first place?

the first time this was used, there were issues with the version of the tool matching the Bevy released version, I don't remember which but I would say something with paths, and it was not possible to fix in the released version, so it was using the one from main

PR #719, opened 2023-07-16
Update zola

nicopap:
The 0.16.1 version of zola has a bug when using colocated_path, this prevents usage of image resolution in #710

Solution: upgrade the zola action we use.

The diff of the github action is available at:

https://github.com/shalzz/zola-deploy-action/compare/v0.16.1-1...v0.17.2

Unblocks #710

PR #1018, opened 2024-02-18
Update Zola to 0.18.0

TrialDragon:
Updates Zola to 0.18.0.

Adopted #719 
Unblocks #710 , probably


TrialDragon:
Okay, cool. So the issue we were having with updating Zola was https://github.com/shalzz/zola-deploy-action/issues/71 and https://github.com/shalzz/zola-deploy-action/issues/78 : an issue with zola-deploy-action where the proper solution prior to `v0.18.0` would've been to use the hotfix version `v0.17.2-1` per https://github.com/shalzz/zola-deploy-action/issues/81#issuecomment-1837382452 .

TrialDragon:
Thanks @BD103 for mentioning the slash as division deprecation / removal in Sass.

TrialDragon:
To address other updating concerns:
* The new `authors` content variable seems to be just for feeds and isn't accessible in templates; so we can't use this first class variable in place of our current `extra.author` variable.
* Zola's Built-in Shortcodes removal; we luckily don't use them.
* `index.md` and `_index.md` can't be in same folder; luckily we don't have that.
* Files that get process—primarily images—now incorporate the stem; we do use resize_image which this affects, but this shouldn't break anything as far as I'm aware.

TrialDragon:
Manually ran and checked: Examples, Assets, Errors, and Community (people). They render correctly and are not broken.
Per https://discord.com/channels/691052431525675048/695741366520512563/1208861386097172500 addressing concerns mentioned on Discord.

doup (changes requested):
I'm pretty sure using SASS `math.div` inside CSS `calc()` without SASS interpolation is a bug. Actually, it's unnecessary, as the browser supports division. In those cases you can just leave it as it was.

sass/components/_asset-card.scss
```diff
@@ -14,7 +15,7 @@ $asset-card-padding: 0.4rem;
     grid-area: banner;
     display: block;
     position: relative;
-    aspect-ratio: 16/9;
+    aspect-ratio: math.div(16, 9);
```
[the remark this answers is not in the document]
TrialDragon: O, I didn't know that. Will revert the aspect ratio changes then.

sass/components/_icon.scss
```diff
@@ -1,36 +1,38 @@
...
-        }
-    }
+  &--#{nth($icon, 1)} {
+    height: var(--icon-height);
+    width: calc(var(--icon-height) * #{$ratio});
+    -webkit-mask-image: url('/assets/icon-#{nth($icon, 1)}.svg');
+    mask-image: url('/assets/icon-#{nth($icon, 1)}.svg');
+  }
 }
+}
```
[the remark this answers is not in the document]
TrialDragon: There is an open PR for that #1013

sass/components/_image_compare.scss
```diff
@@ -17,8 +19,8 @@ div.image-compare {
   outline-offset: -1px;
 
   // calculated
-  --clip-L-geo: inset(0 calc(100% - var(--slider-value) + var(--gap) / 2) 0 0);
-  --clip-R-geo: inset(0 0 0 calc(var(--slider-value) + var(--gap) / 2));
+  --clip-L-geo: inset(0 calc(100% - var(--slider-value) + math.div(var(--gap), 2)) 0 0);
+  --clip-R-geo: inset(0 0 0 calc(var(--slider-value) + math.div(var(--gap), 2)));
```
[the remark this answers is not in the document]
TrialDragon: Ope, yep. Image compare looks broken. Will see if removing that fixes it.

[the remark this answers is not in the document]
doup: OK, I can confirm this breaks the image comparison thingy:

image

sass/components/_image_compare.scss
```diff
@@ -1,3 +1,5 @@
+@use "sass:math";
```
doup, line 1: Not needed anymore.

doup (approved):

sass/components/_asset-card.scss
```diff
@@ -14,7 +15,7 @@ $asset-card-padding: 0.4rem;
     grid-area: banner;
     display: block;
     position: relative;
-    aspect-ratio: 16/9;
+    aspect-ratio: 16 / 9;
```
BD103, line 18: Is this meant to be a list or division?

TrialDragon: > **Nit**: `math.div` is not needed in this case, CSS accepts ratios in `x / y` form for `aspect-ratio`: https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio
> 
> I mean, you could use `math.div` as the property also accepts decimal numbers… but (for me at least) it just adds noise.

What Doup said above; so more like a list than division I think.

BD103: Whoops, didn't see that above. Makes sense, though.

doup: If I understand correctly, is similar to saying `16:9`, but [CSS choose](https://drafts.csswg.org/css-values-4/#ratios) `/` to separate width/height (?).

You can also write it as `1.7777` which CSS would interpret as `1.7777 / 1` (?). The only thing… that's not as useful/familiar as saying `16:9`.

alice-i-cecile:
I'm going to wait on cleanup, then do a final test and merge :)

BD103:
Could you update the README to mention that we're building with Zola 0.18?

https://github.com/bevyengine/bevy-website/blob/bbc525d0deb1dc2f99dc7fbba207f681ccb79b33/README.md?plain=1#L13

BD103 (approved):
I tested this locally and it looked good! Approved :)

RomainMazB (approved):
LGTM too :)

PR #1021, opened 2024-02-18
Add back in necessary prelude boilerplate to getting started

TrialDragon:
Makes a couple prelude boilerplate statements visible again and notifies readers that further use of it will be hidden. Caught after a user experienced issues with the tutorial on Discord: https://discord.com/channels/691052431525675048/691052431974465548/1208911025223761920

PR #1022, opened 2024-02-19
Follow up #1021

TrialDragon:
A change to satisfy a review was made right after #1021 was added to merge queue and as such was lost.


rparrett:
Just to bring the full context into this PR, My review can be considered resolved if the `ommitted` typo gets fixed, but I think something like this wording might be slightly more clear / straight-to-the-point about why the line might be omitted / who is doing the omitting.

rparrett (approved):

alice-i-cecile (approved):

PR #1026, opened 2024-02-19
Make HTML formatting consistent

TrialDragon:
A part of #975 and a follow-up PR to #1013 

The aim of this PR is to make the formatting of indentation and such consistent across all HTML files. This is done via the editorconfig checker tool included in Super Linter. 

This PR makes the presumption that a 2-space wide indentation is preferred similar to #1013. 

~~One unfortunate tradeoff of changing the code to follow this is that when formatting the formatter doesn't consider indentation of Tera DSL code as proper indentation—only HTML element indentation. This means the Tera code in a lot of macros get flattened and become hard to read. Not sure if this will be mandated by CI or not tho.~~


IceSentry:
I really don't agree with unformatting all those templates. It makes it much harder to read. Unless we have tools that can handle this I'm not sure what's the best course of action here.

TrialDragon:
> I really don't agree with unformatting all those templates. It makes it much harder to read. Unless we have tools that can handle this I'm not sure what's the best course of action here.

I don't like it either. I'm going to see if CI will actually complain if we manually format those files with desired indentation

TrialDragon:
Okay, so CI won't complain; code editors will just fight to format this still.

BD103:
It may be worth looking into [djlint](https://www.djlint.com/) or another linter that supports templates. Super Linter doesn't seem to support what we're looking for.

TrialDragon:
> It may be worth looking into [djlint](https://www.djlint.com/) or another linter that supports templates. Super Linter doesn't seem to support what we're looking for.

Nice; I couldn't find a tool like that earlier. Looks like it supports a few of the template languages Tera is based on, so it could work for this.

doup:
I think you're mixing stuff. Linting != Formatting.

I suggested to use `super-lint` given that adding a formatter would probably introduce tooling complexity (up to you if you want research this). What `.editorconfig` _super-linter_ will check is that your code is indented by a multipe of X spaces, that's it. So, it's just a little safeguard to keep indentation consistent. But each contributor still needs to format manually (or with the help of their IDE if it support jinja-like template… and doesn't add extraneous changes).

It's nice if you find a better linter that forces consistent indentation inside template blocks (if/for/block…) in the templates… but I think that's an extra (although, it could also be tiring if it's too picky and the only way of checking code validity is in a pipeline).

The real problem is what you mention [in this comment](https://github.com/bevyengine/bevy-website/pull/1026#issuecomment-1953123121)… you're fighting your editor and forcing your editor's formatting taste into the website project. If your editor doesn't know how to handle Zola templates, I suggest you disable auto-formating.

Sorry, I don't mean to be rude (you already apologized because some 2 space indented code slipped in the book PR 🙏), but I think that this PR (as https://github.com/bevyengine/bevy-website/pull/1013) in the end comes from the fact that your editor wasn't aligned with the bevy-website style. The project already had (somewhat) a coding style.

**Anyway, back to something more practical**, this should be enough to have some consistency:
- Fix flattened templates so we have proper indentation inside `if/for/block`s and such blocks
- Change indentation to 2 where needed
- Just add the `.editorconfig` linter for HTML files
- Disable your editor autoformatting for this project :P

doup (changes requested):
Hi! I still see cases where the blocks are not indented, for example:
- https://github.com/bevyengine/bevy-website/blob/a146baf053f249e588055fd250077e730a0fde42/templates/community.html#L6-L7 (and also the `if`s in the same file)
- https://github.com/bevyengine/bevy-website/blob/a146baf053f249e588055fd250077e730a0fde42/templates/community.html#L6-L7

Then there is also the question of what to do with root level `block/macro` tags? They define blocks, so the content should be indented? That's what _djlint_ thinks.

Maybe we should just run `djlint` on the codebase, so we "reset" the formatting. And then just use the `editorconfig` lint (or add `djlint` in the pipeline). The only thing, I don't like is that it removes blank spaces before/after `for/if` blocks… but that's better than going file by file searching for un-indented code…

BD103:
One thing I do want to point out is that [`disallow.html`](https://github.com/bevyengine/bevy-website/blob/d65f219147b56ba8503a5f4be6ec1423c821fc70/templates/macros/disallow.html) is specifically formatted to make `robots.txt` render correctly. It should be skipped by `djlint`, or made to work with both.

TrialDragon:
I'll add djlint via mega linter (hard fork of super linter with more functionality) and see how it goes.

TrialDragon:
Huh, either we need to configure `djlint` or start fixing files; it has found _plenty_ of issues in the HTML.

TrialDragon:
Went with nunjucks since it was the most comprehensive without a bunch of templating language unique complaints.

TrialDragon:
Okay, so djlint can reset our indentation, but it doesn't use pure 2 space indentation. It has situations like this:
```
<element attr=""
         attrtwo=""
         attrthree=""
```
which conflicts with editor configs strict multiple of 2 only allowed.

BD103:
Tip for reviewers: enable "Hide whitespace" to only see actual changes. :)

templates/macros/disallow.html
```diff
@@ -1,12 +1,12 @@
...
-{{ self::generate_disallows(disallow=true, section=get_section(path=s)) }}
-{%- endfor -%}
-{%- else -%}
-{%- for s in section.subsections -%}
-{{ self::generate_disallows(disallow=false, section=get_section(path=s)) }}
-{%- endfor -%}
-{%- endif -%}
+  {%- if disallow or section and section.extra and section.extra.public_draft -%}
+    Disallow: {{ section.path }}
+    {% for s in section.subsections -%}
```
BD103, line 4: This should not be formatted, it breaks `robots.txt` generation. A `robots.txt` file should not have any indentation. (I added `{%-` and `-%}` to everything else, so those are fine.)

TrialDragon: Just removed all indentation for that file; it's a special case that hopefully shouldn't be touched often and I want to be sure this PR doesn't break it.

templates/macros/disallow.html
```diff
@@ -1,12 +1,13 @@
+{# This file is supposed to not be indented; robots.txt shouldn't be indented and this is used in one #}
```
BD103, line 1:
```suggestion
{# This macro generates robots.txt, which should not be indented #}
```

BD103 (approved):
Looks good! Just one adjustment to the comment in `disallow.html`. This should hopefully help with consistency in the future.

templates/docs-page.html
```diff
@@ -1,28 +1,18 @@
...
-{% block docs_page_content %}
-  {{ page.content | safe }}
-{% endblock %}
-
-{% block docs_page_relative_path %}
-{{page.relative_path}}
-{% endblock %}
+  {% if page.extra.subtitle %}<span class="docs-page-subtitle">{{ page.extra.subtitle }}</span>{% endif %}
+{% endblock page_title %}
+{% block docs_page_content %}{{ page.content | safe }}{% endblock docs_page_content %}
```
IceSentry, line 17: Would it be possible for the formatter to keep this multi line?

IceSentry (approved):
I didn't see anything obviously wrong, and I like having a formatter in CI but I only skimmed the PR

templates/docs-page.html
TrialDragon: The formatter was just a once run thing to fix the indentation after stuff had messed it up; only a linter (which doesn't actually care about indentation) will be run in CI. So we'll have to manually keep the formatting of the indentation ourselves.  I'll manually fix this.

PR #1037, opened 2024-02-20
0.13 migration: Remove name serialization section

### LINKED ISSUE #1030: 0.13 migration: `Name` serialization section is incorrect
https://bevyengine.org/learn/migration-guides/0-12-to-0-13/#fix-reflected-serialization-deserialization-on-name-component

Following the current advice would produce a broken scene.

I'm not totally sure what's going on here. This needs more investigation. It might be a bug in Bevy.

What the guide tells you to do:
```
"bevy_core::name::Name": "My Entity Name",
```

What actually works (same as in 0.12, as far as my testing shows, and as reported on [discord](https://discord.com/channels/691052431525675048/1019697973933899910/threads/1209244908372627507)):
```
"bevy_core::name::Name": (
  hash: 15171155951088048606,
  name: "My Entity Name",
),
```
Or this also works fine.
```
"bevy_core::name::Name": (
  name: "My Entity Name",
),
```

rparrett:
This migration isn't required, and following it would break your scenes.

Fixes #1030

### DIFF
--- a/content/learn/migration-guides/0.12-to-0.13.md
+++ b/content/learn/migration-guides/0.12-to-0.13.md
@@ -746,14 +746,6 @@ If you were making use `collide_aabb::Collision`, see the new `collide_with_side
 - Users of `#[reflect(Component)]` and `#[reflect(Bundle)]` may now want to also add `FromWorld` to the list of reflected traits in case their `FromReflect` implementation may fail.
 - Users of `ReflectComponent` will now need to pass a `&TypeRegistry` to its `insert`, `apply_or_insert` and `copy` methods.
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

rparrett:
If https://github.com/bevyengine/bevy/pull/12024 is merged, this should be reverted.

But with some added nuance like

If you are getting `Could not parse RON: Expected string` errors and utilize Bevy's `serialize` feature (not enabled by default), update your scenes like so:

```diff
- "bevy_core::name::Name": (
-   hash: 13195797433709767748,
-   name: "joe",
- ),
+ "bevy_core::name::Name": "joe",
```

PR #1045, opened 2024-02-22
0.13 migration: Use clippy-approved ReceivedCharacter advice

rparrett:
And other improvements. I went in there to fix the clippy thing and things sort of snowballed.

- Moved the text to a subsection of the PR that caused the breakage
- Added some context about why it's not a `char` and what else might be in there.
- Added a sentence about a new alternative to `ReceivedCharacter` that folks may want to use instead.
- Modified the previous advice so that it is Clippy-approved.

  The previous advice would result in:
  
  ```
  warning: called `.nth(0)` on a `std::iter::Iterator`, when `.next()` is equivalent
     --> src/typing.rs:126:30
      |
  126 |             let Some(char) = event.char.chars().nth(0) else {
      |                              ^^^^^^^^^^^^^^^^^^^^^^^^^ help: try calling `.next()` instead of `.nth(0)`: `event.char.chars().next()`
      |
      = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#iter_nth_zero
      = note: `#[warn(clippy::iter_nth_zero)]` on by default
  ```

  Clippy suggests `.next()`, but based on the [winit docs](https://docs.rs/winit/0.29.10/winit/event/struct.KeyEvent.html#structfield.text), it seems like `.last()` might be the way to go to me. In the cases where there's more than one char, folks probably want the last one.

Some of this stuff should probably make its way into Bevy's docs or a bevy issue.

### DIFF
--- a/content/learn/migration-guides/0.12-to-0.13.md
+++ b/content/learn/migration-guides/0.12-to-0.13.md
@@ -1450,6 +1450,19 @@ Common examples of the updated variants are as follows:
 
 See the relevant [documentation](https://docs.rs/bevy/0.13.0/bevy/input/keyboard/enum.KeyCode.html) for more information.
 
+#### ReceivedCharacter changes
+
+The `char` field of [`ReceivedCharacter`] is now a [`SmolStr`], and _could_ contain multiple characters. See these [winit docs] for details.
+
+A simple workaround if you need a `char` is to call `.chars().last()`.
+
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

rparrett:
Opened docs issue here: https://github.com/bevyengine/bevy/issues/12040

PR #1050, opened 2024-02-23
Improve `generate_wasm_examples.sh`

BD103:
There are several small things that can be done to improve `generate_wasm_examples.sh`, the main entrypoint for generating the WebGL2 and WebGPU examples pages.

### Summary

- Adds more logging about what the script is doing.
  - There are now a few `echo` commands scattered throughout the script.
- Adds comments and update existing ones.
- Deletes `clone_bevy.sh`, merging its contents into the main script.
  - `clone_bevy.sh` was likely created so that multiple scripts could all use it for shared functionality. It was only used in CI, so I merged the two scripts into one and updated Github Actions.
- Fetches latest commit if Bevy was already cloned.
  - This is primarily for local copies. If you have the `bevy` folder cloned already, it will try to fetch the latest commits. Since it targets the `latest` branch, this will only update if a new Bevy release is published.
- Deletes built example folders if they already exist.
  - `rm -rf ../../content/examples`. Without this being called, files could be leftover from old builds.
- Removes temporary main branch tools patch.
  - As commented by @mockersf, this is [no longer necessary](https://github.com/bevyengine/bevy-website/pull/1017#issuecomment-1951464244).
- ~~Removes unecessary `mv content` call because `cargo run` supports `--content-folder`.~~
  - ~~There's no point in build the examples to a specific folder if you're going to immediately `mv` it somewhere else. This instead directly builds it to the destination folder.~~
  - This is needed due to a quirk in `showcase-examples`. See d1043083b546d638764f66184542636721a624a1.
- Removes caching of `generate-wasm-examples` in CI.
  - `generate-wasm-examples` only takes 30 seconds, as compared to `generate-assets` that takes 7 minutes. Caching isn't saving time because other jobs take longer.
  - A cache miss adds about 10 seconds to the runtime, and it happens [relatively frequently](https://github.com/bevyengine/bevy-website/actions/caches?query=key%3ALinux-generate-wasm+).
  - A cache hit has similar performance to generating the examples list from scratch, especially when the cache is outdated.
  - Caching also makes it harder to diagnose CI issues, especially when code that generates the cached content gets changed. I don't think it's worth keeping in this specific circumstance.

You may find [this website](https://devhints.io/bash) useful for Bash scripting, especially the odd `if` / `else` syntax. :)

### DIFF
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -12,7 +12,6 @@ if [[ -d bevy ]]; then
     cd bevy
 
     # Attempts to fetch the latest commits, which should only happen every Bevy release.
-    git switch latest
     git pull --depth=1
 else
     echo Bevy folder does not exist, cloning repository.
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 echo Generating WASM example list...
 
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -12,6 +12,7 @@ if [[ -d bevy ]]; then
     cd bevy
 
     # Attempts to fetch the latest commits, which should only happen every Bevy release.
+    git switch latest
     git pull --depth=1
 else
     echo Bevy folder does not exist, cloning repository.
--- a/generate-wasm-examples/generate_wasm_examples.sh
+++ b/generate-wasm-examples/generate_wasm_examples.sh
@@ -1,26 +1,39 @@
 #!/bin/sh
 
+echo Generating WASM example list...
+
 # Switch to script's directory, letting it be called from any folder.
 cd $(dirname $0)
 
-./clone_bevy.sh
+# If Bevy folder already exists, pull the latest changes.
+if [[ -d bevy ]]; then
+    echo Bevy folder already exists, attempting to fetch latest changes.
 
-# temporary: fetch tools from main branch
-git clone --depth=1 https://github.com/bevyengine/bevy bevy-tools
+    cd bevy
 
-rm -rf bevy/tools
-cp -r bevy-tools/tools bevy
-rm -rf bevy-tools
-cd bevy
+    # Attempts to fetch the latest commits, which should only happen every Bevy release.
+    git pull --depth=1
+else
+    echo Bevy folder does not exist, cloning repository.
 
-cargo run -p example-showcase -- build-website-list --content-folder content --api webgl2
-mv content ../../content/examples
+    # Clone Bevy's latest branch from scratch, only downloading the latest commit.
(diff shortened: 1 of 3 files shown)


BD103:
CI seems to be failing due to an outdated cache. @alice-i-cecile, would you mind deleting `Linux-generate-wasm-examples-` from [here](https://github.com/bevyengine/bevy-website/actions/caches)? I'm hoping the issue is an invalid cache, but I'll check my code again.

TrialDragon:
I think it's more than that.  You're trying to pull when not on a branch with `git pull --depth 1`.  In the clones you're using branch latest.
> You are not currently on a branch.
> Please specify which branch you want to merge with.
> See git-pull(1) for details.
>
>   git pull \<remote\> \<branch\>

BD103:
> I think it's more than that. You're trying to pull when not on a branch with `git pull --depth 1`. In the clones you're using branch latest. You should probably run a `git switch latest` in the line before.

I tried that in 54d4371b3ff9ad757a29ea71664138cf7a92b66e, but it didn't work. When running the script fresh it should be checking out the `latest` branch, so I don't understand why it's not working on CI. It works on my machine, the cache is the only difference that I'm aware of.

I'm going to try temporarily disabling the cache in 3e82320212399fc1bfdf9aeeba9cd1a4407bd4c6 and see if it makes a difference.

generate-wasm-examples/generate_wasm_examples.sh
```diff
@@ -1,26 +1,39 @@
...
 
-cargo run -p example-showcase -- build-website-list --content-folder content --api webgl2
-mv content ../../content/examples
+    cd bevy
+fi
 
-rm -rf content
+# Build WebGL2 example list.
+rm -rf ../../content/examples
+cargo run -p example-showcase -- build-website-list --content-folder ../../content/examples --api webgl2
```
BD103, line 27: This line seems to be raising the error, specifically the `--content-path` argument. Going to do more testing to figure out what is going wrong...

BD103: d1043083b546d638764f66184542636721a624a1 should fix this...

alice-i-cecile:
> @alice-i-cecile, would you mind deleting Linux-generate-wasm-examples- from [here](https://github.com/bevyengine/bevy-website/actions/caches)?

Going to wait on @mockersf for this: I'm not familiar with this area of the codebase.

mockersf:
@BD103 cache cleaned!

mockersf:
> Adds more logging about what the script is doing.
> * There are now a few `echo` commands scattered throughout the script.

you can use `set -x` at the top of the script, it will automatically echo all commands executed https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

BD103:
> you can use `set -x` at the top of the script, it will automatically echo all commands executed https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

I tried it out, but I feel that `set -x` clutters up stdout with confusing noise, especially with the amount of file handling calls. I prefer using explicit `echo`s instead.

TrialDragon (approved):
LGTM, and it runs per the CI checks on this PR.

mockersf (approved):

PR #1057, opened 2024-02-24
Add typos - Source code spell checker

ameknite:
Same as https://github.com/bevyengine/bevy/pull/12036

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,22 @@ jobs:
       - name: Check clippy
         run: cargo clippy --workspace -- -Dwarnings
 
+  typos:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


ameknite:
I decided to ignore the news and migration-guides from 0.1 to 0.13 to avoid changing the name of old pull requests and keep the Changelog.

This will allow to use typos for the next bevy versions, but the downside is that we will need to fix pull requests names in the next releases.

Or the alternative is just ignore for completely all the release news and migration guides.

alice-i-cecile (approved):

alice-i-cecile:
I'm merging the sibling PR over on bevyengine/bevy. I'd like to get this merged here too, and enable it in branch protection. Typos here are much more severe, and this will serve as a nice lower stakes testing grounds.

BD103 (approved):
Looks good!

One thing I want to point out is how the typos check is not required by any of the other jobs. This is a good thing, since it can be run in parallel and does not block anything else.

PR #1061, opened 2024-02-27
Increase CI parallelism

BD103:
This change removes the requirement for `super-linter`, `test-code-examples`, `lint-tools`, and `check-hide-lines` to be run before any of the `generate-*` jobs. While those linters are important, they should not prevent the generation examples from starting.

The only job requirement is `build-website`, while requires `generate-*` to ensure the assets, community, and other pages are built correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
       - 'trying'
 
 jobs:
+  # Lints Markdown and SCSS files
   super-linter:
     runs-on: ubuntu-latest
     steps:
@@ -24,6 +25,7 @@ jobs:
           VALIDATE_EDITORCONFIG: true
           DEFAULT_BRANCH: main
 
+  # Runs the code-validation tool
   test-code-examples:
     runs-on: ubuntu-latest
     steps:
@@ -38,6 +40,7 @@ jobs:
       - name: Build & run doc tests
         run: cd code-validation && cargo test
 
+  # Runs the write-rustdoc-hide-lines tool
   check-hide-lines:
     runs-on: ubuntu-latest
     steps:
@@ -49,6 +52,7 @@ jobs:
       - name: Check hide-lines
         run: cd write-rustdoc-hide-lines && cargo run --release -- check ../content
 
+  # Runs Clippy and rustfmt
   lint-tools:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


BD103:
After initial testing, this seems to decrease CI from [12 minutes](https://github.com/bevyengine/bevy-website/actions/runs/8058411503) to [9 minutes](https://github.com/bevyengine/bevy-website/actions/runs/8058573242?pr=1061) on the dot! These are good improvements!

Now the only bottleneck is `generate-assets` with its 8 minute runtime 😅

BD103:
Added some small comments on the roles of the different jobs in CI. This is ready for review :)

alice-i-cecile (approved):

mockersf:
increasing parallelism also means reducing the number of PRs across the Bevy org that can run in parallel, as the quota on executor is on the org

BD103:
> increasing parallelism also means reducing the number of PRs across the Bevy org that can run in parallel, as the quota on executor is on the org

I see. A free-tier organization can have [20 concurrent jobs](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits). The main repo's CI has 15 different jobs, of which only 12 will be running at a time. The website's CI has 10 different jobs, where before this only 5 would run at a time. After this PR, 9 would run at a time. This does not include multiple PRs running checks at the same time, where it would begin to really add up.

This now makes sense why the lints are run first. The only real solution is to make each job as fast as possible, so it doesn't hold up the concurrency limit for too long. Perhaps this could be helped by cancel-in-progress, like in https://github.com/bevyengine/bevy/pull/11130.

I'm convinced after looking at the problem closer, but it's frustrating to let a 3-minute time save just sit there. Perhaps it could be argued that the website CI isn't run enough to be a problem, but we'd need actual data on that. Darn.

mockersf:
I would be more interested in reducing the duration of each job. here, the longest one is `generate-assets`, because it's querying for fresh data every time even though it probably doesn't make sense to do it on a code change in the website.

Instead of querying crates.io db / GitHub each time, we could handle a cache in the repo, and only query the external sources when the cache is not present. then have a daily job to refresh the cache

BD103:
That's a good idea. I created a new issue for it, so either I or someone else can work on it.

I'm going to close this PR, since it would hurt the main Bevy repo's CI.

PR #1068, opened 2024-03-01
Fix examples

### LINKED ISSUE #1067: Shader examples will not load. Unexpected token '}'
When looking at the shader examples the web console gives the following error and the example fails to load: 
Uncaught SyntaxError: Unexpected token '}' (at animate-shader/:304:3)

Here is the page that I was looking at specifically: 
https://bevyengine.org/examples/Shaders/animate-shader/

rparrett:
Fixes #1067.

By removing some stray `}` chars.

And fixes the first-line indentation that showed up again in the code blocks on the example pages.

image

### DIFF
--- a/templates/example-webgpu.html
+++ b/templates/example-webgpu.html
@@ -8,7 +8,7 @@
       <h2 class="example__title">{{ category.title }} / {{ page.title }}</h2>
       <a class="example__back" href="/examples-webgpu"><i class="icon icon--chevron-left"></i> Back to examples</a>
       <a class="example__github"
-         href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
+          href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
         <i class="icon icon--github"></i> View in GitHub
       </a>
     </div>
--- a/templates/example-webgpu.html
+++ b/templates/example-webgpu.html
@@ -8,7 +8,7 @@
       <h2 class="example__title">{{ category.title }} / {{ page.title }}</h2>
       <a class="example__back" href="/examples-webgpu"><i class="icon icon--chevron-left"></i> Back to examples</a>
       <a class="example__github"
-          href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
+         href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
         <i class="icon icon--github"></i> View in GitHub
       </a>
     </div>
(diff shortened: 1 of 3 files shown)


rparrett:
I'm not sure if I'm running djlint properly, so I am curious to see if CI is okay with this.

templates/example-webgpu.html
```diff
@@ -8,7 +8,7 @@
       <h2 class="example__title">{{ category.title }} / {{ page.title }}</h2>
       <a class="example__back" href="/examples-webgpu"><i class="icon icon--chevron-left"></i> Back to examples</a>
       <a class="example__github"
-          href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
```
TrialDragon, line 11: This was intentional. Editor config requires the tab amount to be a multiple of 2 to keep tab consistency.

templates/example.html
```diff
@@ -8,7 +8,7 @@
       <h2 class="example__title">{{ category.title }} / {{ page.title }}</h2>
       <a class="example__back" href="/examples"><i class="icon icon--chevron-left"></i> Back to examples</a>
       <a class="example__github"
-          href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
```
TrialDragon, line 11: What I said above.

alice-i-cecile:
Looks like the linter is complaining: perhaps we should just ignore this file?

templates/example-webgpu.html
rparrett: My local copy of djlint removed the space. Will add it back in manually.

TrialDragon (approved):

templates/layouts/base.html
```diff
@@ -41,6 +41,8 @@
     {% set show_nav_toggle = true %}
   {% endif %}
 {% endif %}
+{% set newline = "
```
BD103, line 44: What is the point of this line? (Non-blocking)

BD103 (approved):
Looks goods! I just have one quick question, but I tested it locally and it works. :)

templates/layouts/base.html
rparrett: This is part of the fix for the weird indentation in the code blocks.

As far as I can tell, a literal newline is the only way to get a newline into a string in Zola's templates. Previously, we put that literal newline in the code block generation code. But because the new CI is forcing us to indent that code, we're getting some spaces along with out newline.

So I set up this `newline` variable so I could get the newline in with string concatenation.

rparrett: See also: https://github.com/bevyengine/bevy-website/pull/951

PR #1070, opened 2024-03-01
Group `typos` with linters in CI

BD103:
In CI, all of the linters are grouped together and run before any of the `generate-*` jobs, except for typos.

typos is by itself here

Typos was added in #1057. I [originally praised it](https://github.com/bevyengine/bevy-website/pull/1057#pullrequestreview-1901511010) for being separate from the rest of the linters, to increase parallelism. Unfortunately, increasing parallelism is not a viable solution ([as pointed out](https://github.com/bevyengine/bevy-website/pull/1061#issuecomment-1969742951) by mockersf in #1061) because only a limited amount of job runners are allowed per organization. Increasing parallelism for the website would only steal runners from the main Bevy repo, which arguably needs them more.

Since lints should all be run before the longer `generate-*` jobs, this PR adds `typos` to their `needs` attributes. Realistically this change is only visual, because `typos` is the fastest lint we have (most recently finishing in only 10 seconds).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,7 @@ jobs:
           echo 'You can find the extension here: https://marketplace.visualstudio.com/items?itemName=tekumara.typos-vscode'
 
   generate-assets:
-    needs: [mega-linter, test-code-examples, lint-tools, check-hide-lines]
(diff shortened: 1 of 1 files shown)


BD103:
This is how the job graph looks after this PR, for a comparison:

image

TrialDragon (approved):

alice-i-cecile (approved):

PR #1071, opened 2024-03-01
Update local CI running instructions for contributors

TrialDragon:
Before it only had markdownlint, but we've since added djlint and typos, so this needs to be updated.


TrialDragon:
I can't find an easy way to manually run `editorconfigchecker` outside of CI. Changed the contributing file accordingly.

BD103 (approved):
This is definitely needed, with all the recent changes.

Consider showing the lint commands as code blocks, so `&&` is not necessary?

```shell
$ markdownlint -f -c --idk-any-options-lol
$ djlint .
$ typos
```

TrialDragon:
The main issue with that is throwing in multi-line command blocks in markdowns numbered lists angers the linters from past experience; AFAIK they can't handle multi-line lists. I will still try tho.

TrialDragon:
Mm, passes CI, but that majorly broke visually, hmm.

BD103:
Looks good! @alice-i-cecile would you mind reviewing if you get the chance?

alice-i-cecile (approved):

PR #1073, opened 2024-03-03
Improve Bevy news feed

### LINKED ISSUE #1072: Add an RSS feed!
## What problem does this solve or what need does it fill?

I sadly don't have much time to use Bevy but I do like to live vicariously through the Bevy news. I check manually multiple times a week to see if there is any new news. I'd love a way to get notified if there is anything new!

## What solution would you like?

I think adding an RSS feed file to the news would be a pretty good, easy solution, vs some kind of email subscription list or something.

## What alternative(s) have you considered?

N/A

## Additional context

Nope!

TrialDragon:
Adds a standard feed link to the footer of the page and adds the authors variable to the news pages so that the news feed will have attribution to the authors of articles. It is currently using the rss icon provided by feather icons: <https://github.com/feathericons/feather/blob/main/icons/rss.svg>

image
image

Closes #1072

A note: we can't use the `authors` variable in place of our current `extra.author` variable because it isn't available to templates. It's only for feeds as far as I know.


mockersf:
I don't think RSS is used enough to get a place in the header

BD103:
I'm not a fan of the bright orange RSS logo; it clashes a lot with the current website design. What if we used a black + white variant that could use Bevy's theme?

> I don't think RSS is used enough to get a place in the header

What if the RSS logo was only shown on the [News](https://bevyengine.org/news/) page and at the bottom of posts? That way it is out-of-the-way, but still discoverable. Additionally a website footer could be made, which could include the RSS link.

TrialDragon:
> I don't think RSS is used enough to get a place in the header
> 
> From [#1072 (comment)](https://github.com/bevyengine/bevy-website/issues/1072#issuecomment-1975910422), this shouldn't be needed. We could add a title in the html header link.

We already have the atom link in the HTML head and that feature, AFAIK, isn't supported by default in any of the modern browsers (that article you mentioned in the issue is from almost 20 years ago and contains outdated info regarding browser support) and requires external extensions and tooling. I can try to find some place to put it at <https://bevyengine.org/news/>, but that'll take a bit more effort to make it look good. 



> I'm not a fan of the bright orange RSS logo; it clashes a lot with the current website design. What if we used a black + white variant that could use Bevy's theme?

That can work, yeah. There should be some freely available feed icon for use in that color scheme, or we can just make one ourselves if need be.

mockersf:
> What if the RSS logo was only shown on the [News](https://bevyengine.org/news/) page and at the bottom of posts? That way it is out-of-the-way, but still discoverable. Additionally a website footer could be made, which could include the RSS link.

Sounds good 👍

TrialDragon:
Updated it to use a neutral colored svg and reside in a website footer.

sass/components/_footer.scss
```diff
@@ -0,0 +1,20 @@
...
+    flex-direction: row;
+    justify-content: center;
+    gap: 10px;
+    padding: 20px;
+  }
+
+  &__social {
+    img {
+      height: 32px;
+      width: auto;
```
BD103, line 13: Is the width and height necessary here if it is required for all the `` tags?

BD103 (approved):
I'm not sure if I like the footer in its current form, but I'm not really sure how to improve it.

I'm going to mark this as approved, but in the future I think it would be good to revisit this and see if we can rethink what goes in to the footer. (Maybe important links / contact info?)

sass/components/_footer.scss
TrialDragon: Mm, I guess not since I could just mark all of them as 32 x 32 in the tag rather than their actual sizes. I will fix this in a bit.

TrialDragon:
> I'm not sure if I like the footer in its current form, but I'm not really sure how to improve it.
> 
> I'm going to mark this as approved, but in the future I think it would be good to revisit this and see if we can rethink what goes in to the footer. (Maybe important links / contact info?)

Agreed. I considered adding more, but that would require a lot of design and was out of scope for this specific PR (tbh, switching to the footer was already toeing the line of the scope imo even tho I did go with it). I'll open up an issue (or maybe a discussion?) if this merged so that the footer can have design decisions made and then be improved.

cart (approved):
I've resisted the footer for awhile but I do think it is probably inevitable as we expand the site (and run out of space in the headerbar).

I did change the icons to grayscale as I don't think the colors were a good fit.

I removed the "author" and "authors" redundancy by embracing "authors". I also made it default to Bevy Contributors (and removed the explicit definitions in those cases), although now that I'm thinking about it, that would probably break the RSS attribution in those cases. I'll add those lines back.

TrialDragon:
> I removed the "author" and "authors" redundancy by embracing "authors". I also made it default to Bevy Contributors (and removed the explicit definitions in those cases), although now that I'm thinking about it, that would probably break the RSS attribution in those cases. I'll add those lines back.

I did not realize that would work, otherwise I would've done that back when we updated to Zola 0.18.0. `authors` was documented only as something that can be set in front matter and not in templates (I guess I have something to look into and contribute to the Zola documentation now); I thought it was just like the `weight` variable

PR #1078, opened 2024-03-06
Improve LLD fast compile instructions

### LINKED ISSUE #727: MacOS "fast compile" instructions could be improved
The problem is that the instructions for the `config.toml` are not included in the section for setting up `lld` on MacOS, yet it's required to get the alternative linker working.

See https://github.com/bevyengine/bevy/discussions/9397

TrialDragon:
Adds relevant config file info for using LLD linker in Rust and Bevy to avoid confusion.

Closes #727

### DIFF
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -25,7 +25,7 @@ Once this is done, you should have the ```rustc``` compiler and the ```cargo``` 
 
 #### Linux
 
-Follow the instructions at [Linux Dependencies](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md)
+Follow the instructions at [Linux Dependencies](https://github.com/bevyengine/bevy/blob/latest/docs/linux_dependencies.md)
 
 #### Windows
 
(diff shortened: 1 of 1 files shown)


mockersf:
all platforms need to look at this file, not just macos.

TrialDragon:
I know that, it's why I'm not moving the entire thing up under MacOS; I'm simply adding a reference to point towards the appropriate smaller part of that file that is relevant to MacOS usage of LLD so as to avoid the confusion the original issue / discussion had where users may not realize there is more needed to get LLD working for MacOS.

BD103 (approved):
Looks good! Do you think it's also worth linking to [`config_fast_builds.toml`](https://github.com/bevyengine/bevy/blob/f67b17d80d410045f0fb0af668521e2f91569159/.cargo/config_fast_builds.toml) from the main repository?

mockersf:
I think copying the file content in the markdown will be an extra place that's not kept in sync, I would prefer to have that in only one place. No strong opinion though if it should be in the Bevy repo or in the website

content/learn/quick-start/getting-started/setup.md
```diff
@@ -202,7 +223,7 @@ Bevy can be built just fine using default configuration on stable Rust. However
 
 * **Generic Sharing**: Allows crates to share monomorphized generic code instead of duplicating it. In some cases this allows us to "precompile" generic code so it doesn't affect iterative compiles. This is only available on nightly Rust.
 
-To enable fast compiles, install the nightly rust compiler and LLD. Then copy the contents of [this file](https://github.com/bevyengine/bevy/blob/main/.cargo/config_fast_builds.toml) to `YOUR_WORKSPACE/.cargo/config.toml`. For the project in this guide, that would be `my_bevy_game/.cargo/config.toml`.
+To enable fast compiles, install the nightly rust compiler and LLD. Then copy the contents of [this file](https://github.com/bevyengine/bevy/blob/main/.cargo/config_fast_builds.toml) to `<YOUR_WORKSPACE>/.cargo/config.toml`. For the project in this guide, that would be `my_bevy_game/.cargo/config.toml`.
```
ickk, line 226: Should we point at the `latest` ref, maybe? unlikely to matter in practice

ickk (approved):

content/learn/quick-start/getting-started/setup.md
BD103: Agreed, the website should probably prefer using `latest` unless mentioning a specific commit or unstable feature.

BD103:
It may be worth taking a look at https://github.com/bevyengine/bevy/pull/13136, where I updated `config_fast_builds.toml` with far more documentation. If there's anything there that can be reused, feel free :)

BD103:
@ickk I re-added `S-Ready-For-Final-Review` because Trial fixed the change requested. If you have another request, feel free to remove it again. Overall I'm happy with this PR, though. ✅

alice-i-cecile:
This should live here, but we should definitely work towards deduplicating this.

msvbg:
I see this was merged already, but this advice for macOS is outdated. The new default linker on macOS is faster than lld, which is also why the macOS version of mold (called sold) is no longer a paid product. See aevyrie's recent stats: https://discord.com/channels/691052431525675048/691052431974465548/1214452387071397958.

BD103:
> I see this was merged already, but this advice for macOS is outdated. The new default linker on macOS is faster than lld, which is also why the macOS version of mold (called sold) is no longer a paid product.

I created #1160. If you would like to open a pull request, feel free!

msvbg:
PR: https://github.com/bevyengine/bevy-website/pull/1161

PR #1081, opened 2024-03-08
Cache `generate-assets` output in CI

### LINKED ISSUE #1066: Reduce `generate-assets` duration
> I would be more interested in reducing the duration of each job. here, the longest one is `generate-assets`, because it's querying for fresh data every time even though it probably doesn't make sense to do it on a code change in the website.
>
> Instead of querying crates.io db / GitHub each time, we could handle a cache in the repo, and only query the external sources when the cache is not present. then have a daily job to refresh the cache

_Originally posted by @mockersf in https://github.com/bevyengine/bevy-website/issues/1061#issuecomment-1970050641_

BD103:
Closes #1066.

This adds a cache for `generate-assets`, which is the CI job currently takes the longest at ~7 minutes, and reduces its runtime to 11 seconds. The cache is updated ever time the website is deployed using `deploy.yml`. The cache is then used in `ci.yml` when no run in the merge queue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,27 +86,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/cache@v4
+      - name: Restore cached assets
+        id: restore-cached-assets
+        uses: actions/cache/restore@v4
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            generate-assets/target/
-          key: ${{ runner.os }}-generate-assets-${{ hashFiles('generate-assets/Cargo.toml') }}
-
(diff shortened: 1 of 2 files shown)


TrialDragon (approved):
It looks good from the workflow. I'm still a bit hesitant since we can't test saving the cache until deployment.

mockersf (approved):

BD103:
One thing I do need to check is whether `actions/cache/save` overwrites the previous cache of the same key. If it doesn't, I'll need to manually delete the old cache before creating a new one. (And that could lead to data races with multiple runners, if not handled properly.)

BD103:
We did some testing (thanks @TrialDragon!). Saving and using the cache worked well, but we were unable to test using it with merge queues because it is a feature restricted to Github orgs and private repos.

Another note: using the cache reduces `generate-assets` times to 11 seconds!

image

TrialDragon (approved):

PR #1083, opened 2024-03-08
Update bevy version on `learn` chapters

EmiOnGit:
I've noticed that bevy is still not updated to `0.13` in the learning sections. A comment to check for new versions is always included but I still think, that this section should use the newest major bevy release


BD103 (approved):
Awesome, good catch!

PR #1084, opened 2024-03-09
Cancel in progress CI jobs on PRs

TrialDragon:
At the moment if you commit multiple times to a PR while CI is still going it takes up the amount of jobs that can be run at once. Instead we should cancel in progress jobs on new pushes to a PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,10 @@ on:
       - 'staging'
       - 'trying'
 
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: ${{github.event_name == 'pull_request'}}
+
 jobs:
   mega-linter:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,10 @@ on:
       - 'trying'
 
 concurrency:
-  group: ${{github.workflow}}-${{github.ref}}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,6 +8,10 @@ on:
       - 'staging'
       - 'trying'
 
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: ${{github.event_name == 'pull_request'}}
```
BD103, line 13:
```suggestion
concurrency:
  # Only the same workflows on the same branch can cancel each other.
  group: ${{ github.workflow }}-${{ github.ref }}
  # Only cancel runs in a PR; we don't want to cancel commit checks on main.
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

BD103 (changes requested):
Similar PR in the main Bevy repository: https://github.com/bevyengine/bevy/pull/11130

I'm personally for this change, since the less time an action runner spends doing a specific job means the more jobs in total it can do. I have a few requested changes, but they're non-blocking since they only add comments and formatting.

BD103 (approved):

BD103:
@mockersf would you mind reviewing this?

ickk (approved):
Looks like a straight forward implementation of github actions [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency). 

Might help keep things moving in CI at times when lots is happening, such as release-prep

Issue #1059, opened 2024-02-26
Add `README.md` to all tools


BD103:
There are multiple tools in this repository used for generating the site. Unfortunately, not all of them are well documented. A good step in this direction is to add a `README.md` file to each directory, documenting what the tool does and how to use it.

- [ ] `code-validation` (#1052)
- [ ] `generate-assets` (#1052)
- [ ] `generate-community`
- [ ] `generate-errors`
- [ ] `generate-release`
- [ ] `generate-wasm-examples`
- [x] `write-rustdoc-hide-lines`

Issue #789, opened 2023-11-01
Generated site contains a lot of invalid pages


Kanabenki:
While testing some accessibility linting tools for #396, I noticed that `sitemap.xml` contains lots of url pointing to the default zola html for non-rendered section/page (see [here](https://bevyengine.org/assets/assets/ai/big-brain/) for example).

It seems this is due to how the `generate_*` tooling generates the content, as we end up with lots of markdown files that actually do not correspond to a section/page in `content` (they are used to build lists [like here](https://bevyengine.org/assets/) instead.

A possible solution that I tested would be to add `render = false` to the generated sections front matter, and manually exclude the generated pages from the sitemap. There is no equivalent to  `render = false` for pages, so this need to be done with an extra attribute and a custom sitemap template, but this does not completely solves the problem as the invalid html is still built. The simplest way to fix that would probably to have the `render` attribute be implemented by Zola for pages, otherwise we would need to change how list pages are generated to avoid going through some intermediate markdown.

rparrett:
A while back the author of Zola suggested not using "pages" at all for this sort of content in #601.

Kanabenki:
Thanks for pointing that out, indeed it seems that would properly solve the problem.

PR #1089, opened 2024-03-11
Rewrite `generate-errors` and improve `learn/errors`

TrialDragon:
Part of #789 and touches #1059

This removes the use of invalid pages the `learn/errors` section had and stops using a section as a page which is bad practice.

Now `learn/errors` uses the `docs-*` templates and is more like the book or migration guides or quick start guide. This feels better imo since you don't have to scroll one long page to find your specific error; you now have specific pages for errors.

I ended up rewriting the whole `generate-errors` tool because the old one was poorly documented, seemingly esoteric / maybe over-engineered in it's ways of doing things making it hard to work on, and just  a bit hard to retrofit to work as a `docs-*` based page rather than the bad practice section as a page.

Tl:dr: Was easier to rewrite than fix the old.

At least it has documentation now. (Speaking of which _please_ mention anywhere the documentation may be lacking; I don't want this to be hard to work with in the future potentially).


generate-errors/README.md
```diff
@@ -0,0 +1,17 @@
+# Generate Errors
+
+This website utility tool takes the documentation of Bevy-unique error codes from the Bevy engine repo and makes them valid Zola pages.
+
+For most uses run the bash script from any directory:
+`./generate-assets/generate-errors.sh`
```
BD103, line 6:
```suggestion
`./generate-errors/generate-errors.sh`
```

I'll do a full review soon :)

BD103 (changes requested):
Good work, this is definitely needed!

This seems to break syntax highlighting in the generated pages. I believe it's due to `should_panic` and the order of the code block attributes.

Generated with your branch, broken:

    ```rust,should_panic

Generated with main, works:

    ```rust

You can also fix it by making `rust` last:

    ```should_panic,rust

TrialDragon:
Ah, I see. The old solution just thru away the info, it just needs to be reordered. I'll fix that later.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,174 @@
...
+    let mut results_map: HashMap<String, String> = HashMap::new();
 
-    Ok(())
+    for path in error_page_paths {
+        let file_name: String = path
+            .file_name()
+            .ok_or(anyhow!("An error page path has an invalid file stem"))?
+            .to_string_lossy()
+            .into_owned();
+        let content = String::from_utf8(fs::read(path)?)?;
```
TrialDragon, line 54: Looking with fresh eyes, i probably should've just used `fs::read_from_string`, will fix later.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,174 @@
...
+            key.strip_suffix(".md").unwrap_or(key),
+            // Since the introduction page takes the
+            // zeroth position we need to treat the keys
+            // like they're one indexed to not have
+            // conflicting weights which have
+            // undefined behavior.
+            index + 1,
+            pages
+                .get(*key)
+                .ok_or(anyhow!("The page content for {key} isn't valid!"))?
```
TrialDragon, line 132: This error should probably say doesn't exist for clarity.

TrialDragon (commented):
Personal notes questioning yesterday's me's decisions.

alice-i-cecile (approved):

generate-errors/Cargo.toml
BD103, line 1: You can probably change the version to 0.2.0, since this is a major rewrite. (Or set it to 0.0.0, if we don't want to keep track of versions.)

generate-errors/Cargo.toml
```diff
@@ -8,10 +8,10 @@ authors = [
 license = "MIT"
 edition = "2018"
```
BD103, line 9:
```suggestion
edition = "2021"
```

This doesn't change anything in the code, but it is probably good to have. ([Rust 2021 migration guide](https://doc.rust-lang.org/edition-guide/rust-2021/index.html))

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
+/// supplied by user via
+/// a path to the directory containing
+/// the original Bevy error files.
+pub fn get_error_pages(bevy_errors_path: &Path) -> anyhow::Result<HashMap<String, String>> {
+    // Guard clause that determines
+    // if a path exists and is valid
+    // and not a broken symbolic link.
+    // Otherwise it errors and returns.
+    if !bevy_errors_path.try_exists()? {
+        return Err(anyhow!("The path ({bevy_errors_path:?}) is invalid"));
```
BD103, line 19:
```suggestion
        bail!("The path ({bevy_errors_path:?}) is invalid");
```

`bail` does the same thing, and it is even provided by `anyhow`!

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
 
-#[derive(Serialize)]
-pub struct FrontMatterErrorCode {
-    pub title: String,
-    pub weight: usize,
-}
+    let mut error_page_paths: Vec<PathBuf> = vec![];
+
+    let entries = bevy_errors_path.read_dir()?;
+    let regex = Regex::new(r"B[0-9]{4}")?;
```
BD103, line 25:
```suggestion
    // Matches Bevy error codes, such as B0001, B0002, etc.
    let regex = Regex::new(r"B[0-9]{4}")?;
```

Regex can be confusing, especially to those who have not worked with it before. A small comment can help with that.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
 
-impl From<&ErrorCode> for FrontMatterErrorCode {
-    fn from(asset: &ErrorCode) -> Self {
-        FrontMatterErrorCode {
-            title: asset.code.clone(),
-            weight: 0,
+    for entry in entries {
+        // You can't propagate a value multiple times
+        // due to values moving, so it has to be put
+        // in a variable to be used multiple times.
```
BD103, line 30: I find these comments are confusing. The move semantics are a symptom of not being able to `.unwrap()` a `Result` more than once (or technically propagate using `?`).

I don't think the comment here is necessary, but if anything it should just say how it propagates any errors reading the directory.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
+        let entry = entry?;
+
+        if entry.metadata()?.is_dir() {
+            continue;
+        }
+
+        if entry
+            .file_name()
+            .to_str()
+            .is_some_and(|value| regex.is_match(value))
```
BD103, line 40:
```suggestion
            // Only add Bevy error codes
            .is_some_and(|value| regex.is_match(value))
```

The wording on that comment could be improved, but I think it is important to show that it is filtering out unwanted files.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
+        // The error pages already have a header built-in
+        // but Zola provides its own title header
+        // so we need to remove this for proper formatting
+        let regex = Regex::new(r"# B[0-9]{4}")?;
+        let regex_content = regex.replace(content.as_str(), "");
+
+        // Code blocks will be invalid unless
+        // the annotations are before the language
+        // like `should_panic,rust` or `no_run,rust`.
+        let mut content: String = String::new();
```
BD103, line 64:
```suggestion
        let mut content = String::new();
```

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
+    let introduction_content = r#"+++
+title = "Introduction"
+[extra]
+weight = 0
++++
+
+These pages document Bevy's error codes for the _current release_.
+
+In case you are looking for the latest error codes from Bevy's main branch, you can find them in the [Bevy engine repository](<https://github.com/bevyengine/bevy/tree/main/errors>). 
+"#;
```
BD103, line 132: These can probably be converted into `const`s, but I won't block on it. This function only gets called once and Rust likely optimizes it anyways.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
+    for (index, key) in keys.iter().enumerate() {
+        let page_content = format!(
+            r#"+++
+title = "{}"
+[extra]
+weight = {}
++++
+
+{}"#,
+            key.strip_suffix(".md").unwrap_or(key),
```
BD103, line 162: This could use a comment explaining how it is extracting the error code from the filename, and just using the filename for the title if stripping fails.

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
 #[cfg(test)]
-mod test_regex {
-    use regex::Regex;
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_error_page_content() {
+        let error_content: anyhow::Result<HashMap<String, String>> =
+            get_error_pages(Path::new("./bevy/errors"));
```
BD103, line 187:
```suggestion
        let error_content = get_error_pages(Path::new("./bevy/errors"));
```

generate-errors/src/lib.rs
```diff
@@ -1,114 +1,213 @@
...
-        assert!(!error_code_pattern.is_match("B0000.txt"));
-        assert!(!error_code_pattern.is_match("Cargo.toml"));
-        assert!(!error_code_pattern.is_match("README.md"));
+    fn test_write_section() {
+        // Fake content folder
+        let output_path = Path::new("content/learn");
+
+        let result: anyhow::Result<()> = write_section(output_path);
+        assert!(result.is_ok());
+    }
```
BD103, line 198: Should this delete the directory afterwards?

Additionally, this and `test_write_pages` might conflict if run in parallel. I don't know if `cargo-test` runs tests in parallel by default, but it might be a possibility in the future. Instead, could each of these tests create their own custom folder so they don't interfere with one another?

There are also some crates for doing filesystem tests, which may be a more maintainable solution.

generate-errors/src/lib.rs
TrialDragon: Switching to const breaks stuff because we use non-constant values and error handling in the formatting. Not worth it imo.

generate-errors/src/lib.rs
TrialDragon: Mm, I'll just create separate folders for either of the tests for now. Adding in crates for filesystem tests sounds like a bit much for the current problem, now if these tests relied on results of previous ones or other complex things, then I'd go for filesystem testing crates.

generate-errors/src/lib.rs
TrialDragon: Mm, I mainly had that comment there to explain to future me or maintainers why my original solution didn't work and to avoid minimizing lines of code fruitlessly. A clarification of intent if you will.

I'll add a dual comment that first explains what it does, then adds proper clarification of the why regarding unwrapping.

generate-errors/.gitignore
```diff
@@ -1,2 +1,4 @@
 # Bevy repository
 bevy/
+# Fake content folder for tests
+content/
```
BD103, line 4:
```suggestion
# Fake content folder for tests
section_content/
pages_content/
```

BD103 (approved):
Please update the `.gitignore` for the separate content folders, but beyond that it looks good!

PR #1090, opened 2024-03-11
Use Javascript `mega-linter` flavor

### LINKED ISSUE #1087: Improve speed of `mega-linter`
`mega-linter` is a tool that we use to check all HTML, Markdown, and Sass files. It currently takes around 3 minutes to run, which is the longest of the 5 initial lint jobs. Since it is run first, it blocks the `generate-*` and `build-website` jobs from starting. Increasing the speed of `mega-linter` would increase the speed of the entire CI.

The slowest part of using `mega-linter` is not running it, but instead downloading it. It requires downloading a 3.5 GB docker image, which contains a whole host of unused linters.

`mega-linter` allows using custom [flavors](https://megalinter.io/latest/flavors/), which are slim versions of the default image. If the CI could be configured to use one of those flavors instead, the CI times would be decreased. We currently only use 3 linters, so only they are required:

https://github.com/bevyengine/bevy-website/blob/cbd158c930dcb524b1992a34682deec19bcc1c63/.mega-linter.yml#L1-L4

I additionally [asked in Github Discussions](https://github.com/oxsecurity/megalinter/discussions/3412) how to create a custom flavor of `mega-linter` that just includes the linters we need. If that's not possible and a provided flavor is not found, it may be worth just installing each linter manually.

BD103:
Closes #1087.

This configures the `mega-linter` CI check to use the [Javascript](https://megalinter.io/latest/flavors/javascript/) flavor, as suggested by Trial in https://github.com/bevyengine/bevy-website/issues/1087#issuecomment-1987113660. Flavors are customized version of the default image are smaller and more specialized.

~~The goal of this PR is to decrease the `mega-linter` runtime from 3 minutes to at least 2 minutes, so it will no longer be the slowest job.~~ This appears to decrease the runtime for the `mega-linter` job from 3 minutes to <1 minute!

### DIFF
--- a/.mega-linter.yml
+++ b/.mega-linter.yml
@@ -2,3 +2,6 @@ ENABLE_LINTERS:
   - HTML_DJLINT
   - MARKDOWN_MARKDOWNLINT
   - EDITORCONFIG_EDITORCONFIG_CHECKER
+
+# We use the Javascript flavor, which does not include all of the linters.
+FAIL_IF_MISSING_LINTER_IN_FLAVOR: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           # Full git history is needed to get a proper list of changed files within `mega-linter`
           fetch-depth: 0
       - name: Run Mega Linter
-        uses: oxsecurity/megalinter@v7
+        uses: oxsecurity/megalinter/flavors/javascript@v7
         env:
           VALIDATE_ALL_CODEBASE: false
           DEFAULT_BRANCH: main


mnmaita:
You beat me to it 😆 let's see how much this improves the CI times! Good callout btw.

mnmaita:
Should we add the `FAIL_IF_MISSING_LINTER_IN_FLAVOR` variable in the YAML config? It says:

```
If set to true, MegaLinter fails if a linter is missing in the selected flavor
```

which could be handy if new linters are added later down the road (although that might be unlikely).

BD103:
This appears to decrease the runtime to <1 minute, which is a definite win!

> Should we add the `FAIL_IF_MISSING_LINTER_IN_FLAVOR` variable in the YAML config? It says:
> 
> ```
> If set to true, MegaLinter fails if a linter is missing in the selected flavor
> ```
> 
> which could be handy if new linters are added later down the road (although that might be unlikely).

That sounds like something we should definitely enable, good idea!

mnmaita (approved):
I don't have any objections to this change but let's see if someone else has any additional input.

.mega-linter.yml
```diff
@@ -2,3 +2,6 @@ ENABLE_LINTERS:
   - HTML_DJLINT
   - MARKDOWN_MARKDOWNLINT
   - EDITORCONFIG_EDITORCONFIG_CHECKER
+
+# We use the Javascript flavor, which does not include all of the linters.
+FAIL_IF_MISSING_LINTER_IN_FLAVOR: true
```
mockersf, line 7:
```suggestion
```
It's better to have it fail if we use a linter that's not in this flavour

TrialDragon: I'm confused. Isn't the setting you're removing explicitly that? Otherwise, it _won't fail_ when we add linters that aren't present, which is the default. This makes sure we get errors and know if we're adding a linter not a part of the flavor.

mockersf: :facepalm: I read a negation somewhere in that, never mind

PR #1095, opened 2024-03-11
Fix cursor for corporate tier button

### LINKED ISSUE #1094: On the donate page, the cursor is incorrect when hovering over the "corporate tiers/individual tiers" button
Basically what it says on the tin. When hovering the button, the cursor shows as either normal or text, instead of a link cursor as one would expect.

pkupper:
Fixes #1094


NiklasEi:
Couldn't https://github.com/bevyengine/bevy-website/blob/bfa144bc70c01bd45f9d6fd331cca1ce386187d3/templates/donate.html#L118 be a "real" button element?

TrialDragon:
Agreed. Using a div for this feels wrong, especially for accessibility given none of the relevant aria elements are implemented (I wouldn't be surprised if the accessibility linter is going to throw a fit when I merge main into branch next). 

This should be a regular `<input>` styled imo.

pkupper:
Agreed.
When using <button>, some CSS properties have to be set to inherit to maintain the styling. Also, the cursor apparently still needs to be set explicitly. Should these perhaps be set globally for all buttons (e.g. in elements/_button.scss)?

TrialDragon:
I think it's a safe thing to set the default for buttons to use the pointer cursor. The `.button-square.scss` already does so (tho I don't know where this component is even being used).

sass/elements/_button.scss
```diff
@@ -0,0 +1,6 @@
+button {
+    cursor: pointer;
+    color: inherit;
+    font-size: inherit;
+    font-family: inherit;
+}
```
TrialDragon, line 6:
```suggestion
}

```
Should be a final newline.

TrialDragon (changes requested):

TrialDragon (approved):

BD103 (approved):
Thank you!

PR #1106, opened 2024-03-12
Make the Query struct link work in getting started ECS page

TrialDragon:
The reference style markdown links break without new line it seems.


BD103 (approved):
Marking as trivial, looks good!

james7132 (approved):

PR #1111, opened 2024-03-13
typo: fix "Add Bevy as a dependency" section

xyzst:
Changed version of bevy in sample cargo.toml file to include patch number (0.13 -> 0.13.0)


james7132:
This is already correct in that it will pick up the latest semver-compatible version. The patch version here would only limits it to 0.13.0 and newer, which would match every possible 0.13.x version.

BD103:
I agree with James here. While I appreciate you taking the time to do this, keeping the version as 0.13 means that we don't have to change it every time a patch version is released.

PR #1116, opened 2024-03-16
Sync Donors from bevy-donors repo

cart:
[bevy-donors](https://github.com/bevyengine/bevy-donors) is now the source of truth for Bevy Donor information. This grabs the latest bevy-donors release, containing the latest metrics.toml, donors.toml, and donor logos, and adds it to the relevant section of the website.


alice-i-cecile (approved):
I've taken a look over this and https://github.com/bevyengine/bevy-donors/blob/main/src/main.rs, and it all seems reasonable. Strongly in favor of automating this.

A bit more robustness in the linked code would be good, but it's overall clear and maintainable. I'm content to ship this in its current form and see what needs love.

As you said, we should make the `bevy-donors` release action trigger a `bevy-website` rebuild soon after this to finalize the last major step in the automation chain.

PR #1121, opened 2024-03-18
Improve logo size configurability and fix mobile scale

cart:
1. Moves base logo sizing config to `tiers.toml`
2. Uses css `calc()` per-logo to calculate the max-height
3. Adds `square_logo` to sponsor config to give square logos a fair area bump relative to wider aspect ratios. This is nice because it removes the need for direct style overrides and makes the rules consistent across sponsors.
4. Adds `logo_scale` which will further scale the final result. This can be configured per-sponsor
5. Fixes mobile clipping introduced in #1120 

In general this allows us to move away from direct style overrides (although they are still supported)

### DIFF
--- a/templates/individual-sponsors.html
+++ b/templates/individual-sponsors.html
@@ -45,14 +45,20 @@
     <div class="sponsors__content">
         {% for sponsor in sponsors_in_tier %}
         {% if tier.reward_logo %}
+            {% set_global logo_height = tier.logo_height %}
+            {% if sponsor.square_logo %}
+                {% set_global logo_height = logo_height * 1.666 %}
+            {% endif %}
+            {% if sponsor.logo_scale %}
+                {% set_global logo_height = logo_height * sponsor.logo_scale %}
+            {% endif %}
         <a class="sponsors__link" href="{{ sponsor.link }}">
             <img
                 class="sponsors__logo"
-                style="height:
-                    calc({{ tier.logo_height }}{% if sponsor.square_logo %}* 1.666{% endif %}{% if sponsor.logo_scale %} * {{ sponsor.logo_scale }}{% endif %});
-                    {% if sponsor.style %}
-                        {{ sponsor.style }}
-                    {% endif %}"
+                height="{{ logo_height }}"
+                {% if sponsor.style %}
+                style="{{ sponsor.style }}"
+                {% endif %}
                 src="/assets/sponsors/{{ sponsor.image }}"
                 alt="{{ sponsor.name }} logo"
             />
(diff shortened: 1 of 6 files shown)


templates/individual-sponsors.html
```diff
@@ -46,8 +46,16 @@ <h3 class="sponsors__amount">${{ tier.amount }} / month</h3>
         {% for sponsor in sponsors_in_tier %}
         {% if tier.reward_logo %}
         <a class="sponsors__link" href="{{ sponsor.link }}">
-            <img class="sponsors__logo" {% if sponsor.style %}style="{{ sponsor.style }}"{% endif %} src="/assets/sponsors/{{ sponsor.image }}"
-                alt="{{ sponsor.name }} logo" />
+            <img
+                class="sponsors__logo"
+                style="height:
```
TrialDragon, line 51: I disagree with using inline styling for setting the height for an ``. This should probably instead use the template language to calculate a value for the height attribute in a way like:

NOTE: I'm not sure the exact syntax that Tera uses for variable calculations, so this may be slightly off.
```
<img src="..." height="
  {%- if sponsor.square_logo and sponsor.logo_scale -%}
    {{ tier.logo_height * 1.666 * sponsor.logo_scale }}
  {%- elif sponsor.square_logo and not sponsor.logo_scale -%}
    {{ tier.logo_height * 1.666 }}
  {%- elif sponsor.logo_scale and not sponsor.square_logo -%}
    {{ tier.logo_height * tier.logo_scale }}
  {%- endif -%} />
```

TrialDragon (changes requested):

templates/individual-sponsors.html
cart: I'm assuming you mean that you disagree with using css `calc()`, as your suggestion also uses "inline styling". Assuming that, I agree that pre-calculating makes sense.

TrialDragon: My solution isn't using "inline styling" for inline styling is, for all intents and purposes, non-semantic CSS styling in the `style` attribute. This is instead using the semantic height attribute of the image tag that has more meaning to the HTML and rendering.

cart: Ah yeah I missed the part of your suggestion where you set the `height` attribute directly on the `img`. Just pushed similar changes.

I rephrased your suggested logic to be less combinatorically nasty when it comes to boolean logic.

I also kept the optional style override as a break-glass solution if we need to tweak a donor logo in a non-standard way.

cart:
I'm going to merge now as the mobile fix is time-sensitive.

PR #1124, opened 2024-03-20
wasm examples: update the GitHub actions

mockersf:
update the script to not use the tools from main. the tools from latest should work, and main has patches that no longer work on latest

compute the number of pages to run the build in a required job

### DIFF
--- a/.github/workflows/build-wasm-examples.yml
+++ b/.github/workflows/build-wasm-examples.yml
@@ -7,13 +7,31 @@ env:
   PER_PAGE: 50
 
 jobs:
+  prepare-pages:
+    name: Get Pages
+    runs-on: ubuntu-latest
+    outputs:
+      pages: ${{ steps.pages.outputs.pages }}
+    steps:
+      - name: Checkout Bevy main branch
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-wasm-examples.yml
```diff
@@ -7,13 +7,31 @@ env:
...
+        uses: actions/checkout@v4
+        with:
+          repository: 'bevyengine/bevy'
+          ref: 'latest'
+      - name: Get Pages
+        id: pages
+        run: |
+          example_count=`cat Cargo.toml | grep '\[\[example\]\]' | wc -l`
+          page_count=$((example_count / ${{ env.PER_PAGE }} + 1))
+          echo "pages=`python -c \"import json; print(json.dumps([i for i in range($page_count)]))\"`" >> $GITHUB_OUTPUT
```
BD103, line 26: I'm not sure I like using `python` here. Could it perhaps use `jq`'s [`range`](https://jqlang.github.io/jq/manual/#range) function instead?

BD103 (approved):

rparrett (approved):

rparrett:
Tools on `latest` seem to work, and tested the pages stuff in isolation. Looks good to me.

PR #1130, opened 2024-04-06
Use nouns in singular form to match the corresponding data types

lyrakisk:


BD103:
I disagree with this change. `Components` and `Resources` are internal data structures used by `World` and should not be mentioned in the getting started section, even if the links go to the user-facing traits. The only reason why `Entities` is like this is because `Entity`s is not proper English, though I honestly prefer that. I appreciate you going out of your way to contribute, but I think there is likely a better solution that will lead to less confusion.

lyrakisk:
Hello, 

Thank you for the feedback, I wasn't aware of these data structures. What if we update the text like this:

> The [Entity](https://docs.rs/bevy/latest/bevy/ecs/entity/struct.Entity.html) and [Component](https://docs.rs/bevy/latest/bevy/ecs/component/trait.Component.html) data types are great for representing complex, query-able groups of data. But most Apps will also require "globally unique" data of some kind. In Bevy ECS, we represent globally unique data using the [Resource](https://docs.rs/bevy/latest/bevy/ecs/system/trait.Resource.html) trait.

> Here are some examples of data that could be encoded as a [Resource](https://docs.rs/bevy/latest/bevy/ecs/system/trait.Resource.html):

BD103:
I like what you are suggesting, I'd much prefer switching it to ``[`Entity`]s`` and leaving it at that. @TrialDragon, do you have any thoughts on this?

lyrakisk:
`[Component]s` is what caused me confusion in the first place, making me think it was a typo. I think that switching it to `[Entity]s` will cause some head scratches to newcomers that don't have enough context and most probably they will also treat this as a typo.

TrialDragon:
Hmm, there is merit to both. 

Given the struct types and such are proper nouns, I believe the grammar for pluralizing is a suffix of `s`, or `'s` if the proper noun already ends with s.

Yet this doesn't feel obvious to everyone, so appending data type, or struct, or trait, and pluralizing that honestly may be clearer for more people.

The big issue is that if we go with the former then the links will look weird since the pluralization suffix would be outside of it normally. So in that case the latter is more readable.

So off my little tangent, I'd probably say lyrakisk's latter solution of avoiding directly pluralizing the proper nouns of data types for clarity.

lyrakisk:
I updated the page, do you think it's good now? 

image

BD103:
Yes, thank you!

TrialDragon (commented):
I like the changes, but there are also similar issues in the `apps.md` and `plugins.md` sections of getting started that I feel should be fixed by this PR for consistency sake. Both are `App`s pluralization issues early on in both pages.

lyrakisk:
I updated those as well

image  

image

BD103 (approved):

TrialDragon (approved):

PR #1137, opened 2024-04-19
Use documentation MegaLinter flavor

BD103:
We currently use the [Javascript](https://megalinter.io/latest/flavors/javascript/) MegaLinter flavor, which is 857 MB in size and takes 58s ± 2s in CI.

This PR switches it for the [Documentation](https://megalinter.io/latest/flavors/documentation/) flavor, which not only is more fitting for the project but also is only 844 MB. (13 less than Javascript!) When used in CI, it improves speed by about 1 second (lol).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           # Full git history is needed to get a proper list of changed files within `mega-linter`
           fetch-depth: 0
       - name: Run Mega Linter
-        uses: oxsecurity/megalinter/flavors/javascript@v7
+        uses: oxsecurity/megalinter/flavors/documentation@v7
         env:
           VALIDATE_ALL_CODEBASE: false
           DEFAULT_BRANCH: main


mockersf:
don't we want this to also lint the javascript in this repo?

BD103:
> don't we want this to also lint the javascript in this repo?

We don't currently have the Javascript linter enabled. It is certainly an option, but right now it's unused.

TrialDragon:
Mm, we do use javascript in this repo, but not the linter currently, tho I've tossed the idea around of enabling such once or twice in my head before. I'd say a change of subtracting 13 MB of size and 1 second from CI isn't worth losing the future proofing of having the option to enable javascript linting at some point in the future.

But at that point, it's not a hassle to go back to a javascript megalinter flavor in order to enable it in the future. So it's a minute enough change I'm not entirely sure it ultimately matters either way in the end; bike-shedding really.

BD103:
Ok, I'm going to close this. After further thought, I agree that a 1 second change isn't worth sacrificing the future-proofing. (Especially since MegaLinter is not the bottleneck in our CI.)

PR #1139, opened 2024-04-21
Update ecs.md

mrexcessive:
The main() definition after introduction of the greet_people Update function was incorrect, still calling hello_world, but greet_people would be a less confusing thing to refer to at this point in tutorial.

### DIFF
--- a/content/learn/quick-start/getting-started/ecs.md
+++ b/content/learn/quick-start/getting-started/ecs.md
@@ -116,7 +116,7 @@ Now register the startup system like this:
 fn main() {
     App::new()
         .add_systems(Startup, add_people)
-        .add_systems(Update, hello_world)
+        .add_systems(Update, greet_people)
         .run();
 }
 ```


mockersf (approved):

content/learn/quick-start/getting-started/ecs.md
```diff
@@ -116,7 +116,7 @@ Now register the startup system like this:
 fn main() {
     App::new()
         .add_systems(Startup, add_people)
-        .add_systems(Update, hello_world)
+        .add_systems(Update, greet_people)
```
TrialDragon, line 119: At this point in the code, `greet_people` hasn't been introduced, and `hello_world` is removed explicitly in the resources chapter of quick start. I'm not sure this change is warranted per se.

SIGSTACKFAULT: Yeah i think we should leave it

PR #1140, opened 2024-04-22
Remove stray word from bevy tutorial

chrisguidry:


mockersf (approved):

content/learn/quick-start/getting-started/ecs.md
```diff
@@ -198,7 +198,7 @@ fn main() {
 }
 ```
 
-Note that we have used `.chain()` on the two systems. This is because we want them two to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
+Note that we have used `.chain()` on the two systems. This is because we want them to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
```
TrialDragon, line 201:
```suggestion
Note that we have used `.chain()` on the two systems. This is because we want both of them to run in exactly the order they're listed in the code: with `update_people` occurring before `greet_people`.
```

TrialDragon (changes requested):
Good catch, this was originally fixed months ago (Janurary I think) but the whole new book merge overwrote it and wasn't caught. It'd probably be best to just go back to the original fixed wording of `both of them`.

Reference to original fix: #864

TrialDragon (approved):

PR #1141, opened 2024-04-26
add linux dependency instructions and some placeholder headers

SIGSTACKFAULT:
Decided to put the linux dependencies on the installation page but no reason they couldn't be on their own page if y'all want.

Mostly copied from [bevyengine/docs/linux_dependencies.md](https://github.com/bevyengine/bevy/blob/main/docs/linux_dependencies.md) but I added a Nix Flake.

### DIFF
--- a/content/learn/book/getting-started/installation.md
+++ b/content/learn/book/getting-started/installation.md
@@ -217,7 +217,8 @@ Try this `flake.nix`:
     );
 }
 ```
-Activate with `nix develop`, or consider using `direnv`. You may need to re-launch your IDE from inside the flake's environment.
+Activate with `nix develop`, or consider using [`direnv`](https://direnv.net/).
+You may need to re-launch your IDE from inside the flake's environment.
 </details>
 
 #### [OpenSUSE](https://www.opensuse.org/)
--- a/content/learn/book/getting-started/installation.md
+++ b/content/learn/book/getting-started/installation.md
@@ -148,6 +148,9 @@ And enter it by just running `nix-shell`.
 You should be able compile Bevy programs using `cargo run` within this nix-shell.
 You can do this in one line with `nix-shell --run "cargo run"`.
 
+[Here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/games/jumpy/default.nix)
+is an example of packaging a Bevy program in nix.
+
 #### Nix Flake
 
 Try this `flake.nix`:
@@ -199,9 +202,6 @@ Try this `flake.nix`:
 }
 ```
 Activate with `nix develop`, or consider using `direnv`.
-
-[Here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/games/jumpy/default.nix)
-is an example of packaging a Bevy program in nix.
 
 ### [OpenSUSE](https://www.opensuse.org/)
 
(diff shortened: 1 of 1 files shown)


TrialDragon (commented):
It'd probably make more sense to have a level 2 heading of `OS Dependencies` with a level 3 heading for `Windows`, `MacOS`, and `Linux`. Other than that, this is a fine change.

(Tho, it makes me think we should maybe look into having the option of collapsible headings / sections for usability or navigation in the future, since the Linux dependencies are so long and not everyone will be using them. Or maybe just having them as a separate page like they are now, but not sure where would even be a good place to categorize that page or put it).

mockersf:
this used to be in the book, but it was removed to keep only the file in the git repo because they weren't updated in sync.

why do you want to duplicate that information?

TrialDragon:
From my understanding, the intent is to move it from git to the book, migrating applicable documents from `bevy/docs` to the book.

SIGSTACKFAULT:
Maybe we should delete it from bevy/docs to keep the info in one place.

SIGSTACKFAULT:
> Collapsible headings

html [`<details>`](https://www.w3schools.com/TAGS/tag_details.asp) seems to work fine.

NthTensor (approved):

ickk:
> this used to be in the book, but it was removed to keep only the file in the git repo because they weren't updated in sync.

Maybe we could consider creating a mechanism to #include files from the latest bevy release? This has come up a few times in the form of the `config.toml`, now dependencies, and I'm sure there are other fitting places; i.e. in-repo documentation such as `bevy/docs/cargo_features.md`

SIGSTACKFAULT:
The plan is to move all of bevy/docs to bevy-website/learn/book. bevy/docs/cargo_features is next on my hit list.

as soon as this merges i'll make a mr replacing bevy/docs/linux_dependencies with a link to this (to avoid information duplication).

ickk:
Some of these documents really require being kept in sync with `main` (cargo-features for instance), and the website just needs a checkpoint of what's on `latest`, so I'm not sure if moving them to the website repo is ideal

NthTensor:
> why do you want to duplicate that information?

Posting docs to the website will boost visibility and make it easier to on-board new users. It also just straight-up looks way more professional. I'm not suggesting we remove this from the repo, it's important we keep up-to-date install instructions for main. But what's the problem with duplication? Can't we just make sure this gets updated to be current every release (just like all the other docs on the website will to be updated)?

TrialDragon (approved):
This LGTM, but I will note that yeah some of the `bevy/docs` files, like `cargo-features` as the primary example, should probably stay in `bevy/docs` since it is tracking main and tightly coupled with the current state of the Cargo.toml and such.

If we truly _do_ want to, in some way, duplicate those certain files that must stay in `bevy/docs` so that they may be more visible in the official book, then I suggest we make a `generate-bevy-docs` (with a better name) tool similar to how we currently fetch from the community repo in `generate-community` or from `bevy/errors` in `generate-errors`. That way it will automatically stay in sync with the latest Bevy version.

SIGSTACKFAULT:
Yeah cargo features should probably be strongly coupled. But I don't think *this* needs to be.

PR #1145, opened 2024-05-02
correct name of the dynamic_linking feature

SIGSTACKFAULT:
for some reason book/development-practices/fast-compiles.md refers to the "dynamic" feature instead of the actual name "dynamic_linking". so i fixed it.

### DIFF
--- a/content/learn/book/development-practices/fast-compiles.md
+++ b/content/learn/book/development-practices/fast-compiles.md
@@ -11,20 +11,20 @@ status = 'hidden'
 * Explain why you might want faster compiles
 {% end %}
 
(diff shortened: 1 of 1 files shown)


IceSentry:
A bit of history, the reason is that at some point I confused the dynamic plugin and dynamic_linking so we renamed the feature to be more specific to what it is. Apparently we never renamed it in the book.

IceSentry (approved):

ickk (approved):

cart (approved):

SIGSTACKFAULT:
<3

PR #1147, opened 2024-05-02
Add "On This Page" navigation

SIGSTACKFAULT:
(replaced original mr comment with actual up-to-date viz of what it looks like)

https://github.com/bevyengine/bevy-website/assets/14184826/74397650-dc7d-46c2-b35d-4425656dbd31


SIGSTACKFAULT:
https://github.com/bevyengine/bevy-website/assets/14184826/29dd7eaf-eb00-4957-9d5f-2132d894f887

SIGSTACKFAULT:
I think we should either increase `$max-width` to 1400px, or move On This Page outside the container.

doup (commented):
Sorry I think I didn't express myself clearly on Discord. Unless you think there is a case for supporting this new menu in more pages I think these changes should go in `docs-base.html`; which is solely used in the docs. `page-with-menu` is used everywhere there is a sidebar. 

The TOC should probably go here in `docs-base.html`:
image

You will need to change `.docs-page` from flex to grid, and define `grid-template-areas` for the doc content, doc title and TOC. But apart from that I hope it's not much big of an issue.

You won't have access to `$sidebar-width` in that context, but I think it's OK, I would give it a little bit less space than to the sidebar. You still can shave some space from the `ul` margin/padding, reduce `li` indentation, and maybe reduce font size.

<details>
<summary>Also, maybe this is due for some extra refactor…</summary>

Please, you can ignore me… 😆

I would refactor also `docs-page`, `docs-content` and some other classes. Maybe just have a `doc` BEM class, so I would add a `doc__title`, `doc__content`, `doc__toc`, `doc__footer`. With their own `grid-area`.

Probably `.docs-content` can be removed, so it's one less level of nesting.

I would leave `media-content` and `docs-footer` classes, I would just compose with them (similar to what you have done with `page-nav`). So it would look like:

```
<div class="doc__content media-content">
...
<div class="doc__footer docs-footer">
```
</details>

doup:
I forgot the most important thing! It's looking great! :-)

doup:
- regarding `max-width`, we could increase it and check how it looks in other pages. The only one that "worries" me are the blog posts, as it will increase the line length which tends to worsen text legibility.
- after seeing the last video I think I like the TOC sidebar width as it is… if you want to share the width between the two CSS files you can move `$sidebar-width` to `_vars.scss`
- one thing I noticed in the first video, right before changing to the mobile layout the content column looks too narrow, maybe you could use a custom breakpoint that changes the TOC layout earlier?

SIGSTACKFAULT:
Away from my PC for a few days but if anyone wants to work this feel free to add this branch to your personal form and I'll cherry-pick them when I get home

SIGSTACKFAULT:
(rebase)

SIGSTACKFAULT:
tried putting it above or below the menu but wasn't a fan. at the top, causes the menu to jump around as the "On This Page" length changes. at the bottom, you've gotta *scroll* and many won't find it

SIGSTACKFAULT:
i didn't like this idea before but it's starting to grow on me
image

doup:
I had also my initial doubts but after seeing it live I think it works great for the pages.

Although, **it might be confusing for the section pages** (the ones with the chevron like: [ECS intro](https://bevyengine.org/learn/book/ecs/)). There might be some confusion between pages and TOC. And then, there is the toggling of the section, the TOC should be visible even when the section is collapsed. I think we won't know until we see it. 🤔 

Btw, maybe something like this would make it clearer that this is the TOC of the page?

image

SIGSTACKFAULT:
> something like this would make it clearer that this is the TOC of the page?

Ooh, I like that

SIGSTACKFAULT:
coded it like this by accident and decided i liked it even better :)

image
image

sass/components/_on-this-page.scss
```diff
@@ -0,0 +1,36 @@
+/// internal page navigation
+/// i.e., the "On This Page" section
+
+
+.tree-menu .on-this-page {
+    grid-row: 2;
```
doup, line 6: I would create a `tree-menu__toc` (in `_tree-menu.scss`) and move this `grid-row` rule there. I think it composes better that way; that way `.on-this-page` could be used in another context.

So the idea is to have two classes:

```
<nav aria-label="On This Page" class="on-this-page tree-menu__toc">
```

- `on-this-page` will control the internal styling
- `tree-menu__toc` will control the layout in relation to `tree-menu`

sass/components/_on-this-page.scss
```diff
@@ -0,0 +1,36 @@
...
+    // border: 3px solid rgba($color-white, 0.1);
+    border: 3px solid transparent;
+    border-width-top: 0;
+    box-sizing: border-box;
+    background-color: $color-grey-800;
+}
+
+.page-with-menu .tree-menu .on-this-page {
+    // desktop menu uses grey-800 as the bg
+    background-color: $color-grey-900;
```
doup, line 17: I would add this as a "modifier", like `.on-this-page--mobile` and apply that class to the mobile menu. `docs_menu_page_row` macro can be refactored to accept an extra parameter `is_mobile: bool` and based on that you could apply `.on-this-page--mobile`.

If we add `is_mobile` to `docs_menu_page_row` then `prefix` parameter could be removed and computed internally based on `is_mobile`. So we would have a couple of calls like:

```
{{ docs_macros::docs_menu(is_mobile=true, root=get_section(path=root_section_path) ) }}
{{ docs_macros::docs_menu(is_mobile=false, root=get_section(path=root_section_path) ) }}
```

templates/macros/docs.html
```diff
@@ -100,3 +103,25 @@
     {% endif %}
   </ul>
 {% endmacro %}
+{% macro on_this_page(toc) %}
+<nav aria-label="On This Page" class="on-this-page media-content">
```
doup, line 107: I think `on-this-page` should define it's own styling for `<a>` elements and not rely on `media-content` for this.

templates/macros/docs.html
```diff
@@ -100,3 +103,25 @@
...
+  {# screen readers will read this as 'Navigation: On This Page' #}
+  <!-- <b class="page-nav__on-this-page">On This Page</b> -->
```
doup, line 109: Remove?
```suggestion
```

templates/macros/docs.html
```diff
@@ -100,3 +103,25 @@
...
+          {% if h1.children %}
+              <ul>
+                  {% for h2 in h1.children %}
+                      <li data-fragment="{{ h2.id }}">
+                          <a href="{{ h2.permalink | safe }}">{{ h2.title }}</a>
+                      </li>
+                  {% endfor %}
+              </ul>
+          {% endif %}
+      </li>
```
doup, line 123: Change indentation to 2 spaces.

templates/macros/docs.html
```diff
@@ -69,6 +69,9 @@
   <li class="{{ class }}">
     <div class="{{ label_class }}">
       <a class="tree-menu__link" href="{{ menu_page.path }}">{{ menu_page.title }}</a>
+      {% if is_active and menu_page.toc %}
+      {{ self::on_this_page(toc=menu_page.toc) }}
+      {% endif %}
```
doup, line 74: Nit:
```suggestion
      {% if is_active and menu_page.toc %}
        {{ self::on_this_page(toc=menu_page.toc) }}
      {% endif %}
```

doup (changes requested):
Hi! I've done a review… but I think it got a little bit out hand. Let me know if it's too much and **ignore as you wish**. I can decrease the code review dial for the next time. 😅

General:
- Can you sync with `main`? There are relevant [changes in the `tree-menu`](https://github.com/bevyengine/bevy-website/pull/1153).
- The TOC doesn't show in section pages. Try adding some dummy headers in `/content/learn/book/ecs/_index.md` then go to `/learn/book/ecs/`. We should consider how the TOC will behave when the `tree-menu` collapses.

UX:
- Change `<a>` to `display: block` so the whole line is clickable (not only over the text).
- Maybe increase `<a>` line-height (or add tiny bit of vertical padding) so it has a bigger hit area in mobile?

Styling suggestions:
- Change horizontal padding and subheading indentation to 12px to be aligned with `.tree-menu`
- `overflow: hidden` on `.tree-menu__label` so the it's border-radius is also applied to `.on-this-page`

And some more in the in-lined comments.

SIGSTACKFAULT:
(rebase)

SIGSTACKFAULT:
- [x] rebase
- [ ] make toc show in section pages
- [ ] change <a> to display: block
- [ ] consider <a> hit area
- [ ] change horizontal padding to 12px
- [ ] `overflow: hidden` on .tree-menu__label

sass/components/_on-this-page.scss
```diff
@@ -0,0 +1,36 @@
+/// internal page navigation
+/// i.e., the "On This Page" section
+
+
+.tree-menu .on-this-page {
+    grid-row: 2;
+    padding-inline: 0.5em;
+    // border: 3px solid rgba($color-white, 0.1);
+    border: 3px solid transparent;
+    border-width-top: 0;
```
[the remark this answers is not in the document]
SIGSTACKFAULT: yep just leftover when i was trying to have an outline but i was trying to avoid having to actually put the colour in

sass/components/_on-this-page.scss
```diff
@@ -0,0 +1,36 @@
+/// internal page navigation
+/// i.e., the "On This Page" section
+
+
+.tree-menu .on-this-page {
```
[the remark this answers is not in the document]
SIGSTACKFAULT: nav.on-this-page was originally to gain more specificity because .media-content was winning but now that i've removed .media-content it's not a problem.

sass/components/_on-this-page.scss
SIGSTACKFAULT: done :)

templates/macros/docs.html
SIGSTACKFAULT: done

sass/components/_on-this-page.scss
SIGSTACKFAULT: changed it so that it just uses a translucent background so there's no distinction between the mobile and desktop version

SIGSTACKFAULT:
https://github.com/bevyengine/bevy-website/assets/14184826/74397650-dc7d-46c2-b35d-4425656dbd31

sass/components/_on-this-page.scss
```diff
@@ -0,0 +1,36 @@
...
+    }
+    display: block;
+    // copied from .media-content
+    text-decoration: none;
+    color: $link-color;
+    &:hover {
+      text-shadow: 0 0 0.9px $link-hover-shadow-color, 0 0 0.9px $link-hover-shadow-color;
+    }
+  }
+}
```
doup, line 36: https://github.com/bevyengine/bevy-website/assets/188612/563020aa-03b9-4632-ac0e-f72eb203c07a

Changed:
- `.on-this-page { font-size }`
- `a { padding, text-wrap }`

```suggestion
.on-this-page {
  $h-padding: 12px;
  font-size: 0.95em;
  padding-block: calc($h-padding/2);
  ul {
    list-style-position: outside;
    list-style-type: none;
    padding-inline-start: $h-padding;
  }
  > ul {
    padding-inline: $h-padding;
  }
  li[data-active=true] {
    font-weight: bolder;
    li {
        font-weight: normal;
    }
  }
  a {
    @media ($bp-tablet-portrait-down) {
        // increase hit area for touchscreens
        padding-block: 3px;
    }
    display: block;
    // copied from .media-content
    text-decoration: none;
    color: $link-color;
    padding: 2px 0;
    text-wrap: balance;
    &:hover {
      text-shadow: 0 0 0.9px $link-hover-shadow-color, 0 0 0.9px $link-hover-shadow-color;
    }
  }
}
```

doup (approved):
I really like how this shaped out. The section TOC works great also.

I've left a final suggestion that –for me– improves legibility and helps with tapping on mobile (since `<a>` will have more area). In any case, I'm approving the PR as it's already in a good shape.

I'm not super fan of the blue links, but it's OK. I prefer to someone else to pitch in and give their opinion.

Great work! And thanks for the patience. 😅

sass/components/_on-this-page.scss
SIGSTACKFAULT: I deleted the font-size rule and added text-wrap. I already added extra padding for mobile on line 26

doup: I was suggesting it also for desktop, for me it's too dense otherwise. But it's OK.

SIGSTACKFAULT: I see. I don't think destkop needs it.

SIGSTACKFAULT: gave it 1px

SIGSTACKFAULT:
If there's no other objections, I think this can be tagged ready-for-final

TrialDragon (commented):
On the core of things, this is good. Regarding colors of the sidebar navigation, personally I prefer a more amber / yellow / orange in the same vein as the `#e4c151` we use for code block links. Or just a white or grey for an uncontroversial choice. But the color in the end is inconsequential since the current one has contrast and such.

I should note that this PR has merge conflicts (probably with the whole merging `docs-page`, `docs-section`, and `docs-base` into just `docs`) that need to be resolved before this can be merged.

SIGSTACKFAULT:
(rebase)

alice-i-cecile (approved):
Honestly, I wasn't convinced initially. Seemed somewhat busy. But comparing the experience both before and after I quite like it!

It helps get an overview at a glance of what's covered on each page, and helps the reader stay oriented on where they are.

PR #1149, opened 2024-05-03
Merge `docs-section`/`docs-page` templates into `docs`

doup:
Merge `docs-section`/`docs-page` templates into `docs` leveraging [`default` filter](https://keats.github.io/tera/docs/#default) (allows removal of repetitive `if section … elseif page …` code).


TrialDragon (approved):
LGTM.

I wonder if we could use this same feature to merge `docs-page` and `docs-section` into `docs-base` and just rename it to `docs.html`. I want to say the primary reason we even have a separate section and page templates for the docs / book portion of the site is because it was easier and more readable than writing even more if else conditions for page and section variables, but this solves that issue.

doup:
Hehe, that's actually something I wanted to ask in Discord as it seems that those are just repeated. I was going to open a PR removing those templates and just leaving a `docs.html` (former `docs-base`).

doup:
I wanted to try an extra simplification by moving the `section_or_page` (previously `current`) variable declaration to the root layout (`base.html`). And after seeing it worked fine I took the extra step of also removing the `docs-section`/`docs-page` templates.

This increases the size of the PR, but it's quite easy to follow as it's mostly repetitive changes. Let me know if you prefer this in two separate PRs, I can revert the changes.

TrialDragon:
The scope is fine imo.

TrialDragon (approved):
LGTM. I almost thought you also had to change a template variable in `generate-release`, but I guess we currently don't include that in it's front matter and add it when we move it to be an actual page.

doup:
I'm tempted to slip this changes in `layout/base.html`, opinion? 😇
https://github.com/doup/bevy-website/pull/3/files

TrialDragon:
Mm, tentatively I'll say that perhaps the `layout/base.html` changes can be it's own PR for sake of keeping the PR from slowly growing too big, but I approve of them either way.

doup:
I'll leave this PR as it is and I'll open the `base.html` afterwards.
Thanks for the feedback!

SIGSTACKFAULT (approved):

TrialDragon:
This PR needs to be updated to account for the `contributing` section added in 11f5816 since it is still using `docs-section.html`.

PR #1150, opened 2024-05-03
Fix docs footer navigation

doup:
There are some cases where the footer links don't show up, or show the incorrect link, for example:

- Previous Page link not showing: https://bevyengine.org/learn/book/getting-started/why-bevy/
- Next page link incorrect: https://bevyengine.org/learn/book/getting-started/bevy-community/

This PR fixes these issues. I've double checked in all the other pages and it seems to work correctly (e.g. `Breakout` draft page doesn't show up).

### Demo

On this branch.

First case (previous page correct):

image

Second case (next page correct):

image

### DIFF
--- a/templates/docs-base.html
+++ b/templates/docs-base.html
@@ -97,8 +97,6 @@
         {% if found_current %}
           {% if p.extra and p.extra.public_draft %}
             {% continue %}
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

BD103 (approved):
Looks good!

PR #1155, opened 2024-05-06
add more styles to <details>

SIGSTACKFAULT:
(sneak peek of the page on custom queries i'm writing)

image

image

makes it more obvious what's part of the `<details>` and what's not


NthTensor (approved):
Looks great!

IceSentry (approved):

doup (approved):

cart:
Love this! Some thoughts:

1. This now creates a "perceived indent" for expandable widgets, making them feel unaligned with sibling content when they aren't expanded.
2. The "white" expanded color feels slightly too white / too loud. I think we should use `$subtitle-color` like we do for things like table widgets.

A solution to (1) is to always draw a full outline (with rounded corners matching the border radius used elsewhere). This also serves to make the content look more "clickable". But it also makes it a bit "louder". I'm not fully sold. Just throwing ideas out.

image

image

```css
details {
    padding: 8px;
    border: 2px solid;
    border-color: $subtitle-color;
    border-radius: $border-radius;
    & > summary {
        // this should be the default
        cursor: pointer;
    }
}
```

SSebigo:
> Love this! Some thoughts:

We should check how it plays out when there are many of these in a page, feels like it would make the page bloated.

doup:
Bloating can be reduced by using a darker outline that contrasts less. In any case, right now `<details>` is only used twice in two posts… so I wouldn't worry for now. 😅

NthTensor:
Could we try a 10% lighter/darker background color of an outline, or perhaps with a feint 15% lighter/darker outline.

cart:
Works for me!

SIGSTACKFAULT:
image

NthTensor:
I really like how this looks.

sass/elements/_details.scss
```diff
@@ -1,4 +1,64 @@
...
+  // border: $border;
+  border-radius: $border-radius;
+  padding: 0 $h-padding;
+  margin-block: 1em;
+  background-color: rgba($color-white, 0.1);
+  overflow: hidden; // prevent clobbering corner radii
+
+  >summary {
+    cursor: pointer; // this should be the default
+    font-weight: bold;
```
cart, line 13: Given that now has a visually eye-catching background, I think this might be too "loud" as a default. Can we remove this in favor of the default boldness?

cart (commented):
Love the new style. Very clean!

sass/elements/_details.scss
SIGSTACKFAULT: sure; i waffled about it a few times myself.

PR #1157, opened 2024-05-09
create logging page

SIGSTACKFAULT:
I figured that there should be a page on logging. so i added one under Development Practices.

image

it's just a placeholder; i just wanted to get it in the book so it's not forgotten.


alice-i-cecile (approved):
No objections to adding this stub. It should definitely exist.

SIGSTACKFAULT:
I might also write the page but that can be seperate

TrialDragon (approved):

PR #1162, opened 2024-05-14
Add custom `atom.xml` template to hide public draft news articles

### LINKED ISSUE #1103: Hide public draft news pages from atom news feed
The default `atom.xml` news feed template still adds the public draft news articles and publishes them. We should probably make our own custom template mostly the same as the default, but with the exclusion of public draft articles being published.

Relevant page of the Zola docs: https://www.getzola.org/documentation/templates/feeds/

TrialDragon:
Fixes #1103 

Adds an `atom.xml` template file that hides pages which are public draft from the news feeds, primary line of code relevant being `Line 23`.  Note, this doesn't modify the last updated date, so that is still tracking the latest file regardless of public draft, but that is a minor issue that I think is irrelevant to our situation so far.

### DIFF
--- a/templates/atom.xml
+++ b/templates/atom.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="{{ lang }}">
+    <title>{{ config.title }}
+    {%- if term %} - {{ term.name }}
+    {%- elif section.title %} - {{ section.title }}
+    {%- endif -%}
+    </title>
+    {%- if config.description %}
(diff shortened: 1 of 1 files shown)


doup (approved):
OK, I've poked around with Zola trying to find a workaround[0] but it looks like until [this issue in Zola](https://github.com/getzola/zola/issues/2391) this is the best option.

**Approved 👍**.

0: [Zola docs](https://www.getzola.org/documentation/templates/feeds/) state `Only pages with a date will be available.`… but I couldn't make posts disappear while keeping the public draft preview. There is a `zola build --drafts` flag, but it comes with it's own issues.

IceSentry (approved):

PR #1164, opened 2024-05-14
Minor On This Page Script Changes

SIGSTACKFAULT:
minor followup to #1147 

I realized that the `on-this-page.js` still referred to it as "page content navigation" or "page_nav" or etc, from before i decided that the one true name:tm: of the feature was "on this page". then i got a bit carried away with jsdoc, and improved the initialization logic.


alice-i-cecile:
@doup could I get your review here too? :)

TrialDragon (approved):

alice-i-cecile (approved):

Issue #1163, opened 2024-05-14
Parallelizable release prep


alice-i-cecile:
# Parallelizable release prep

## Current release strategy

We need to prepare several parallel-but-distinct documents for each release:

1. The release notes
2. The migration guide
3. The complete change log
4. The list of contributors

The release notes historically have been created ad-hoc, by looking through the list of PRs merged, and adding a stub section. This has been improved already for the 0.14 cycle, through the addition of the `C-Needs-Release-Note` label. This will give us a good initial list, but it will need to be checked for omissions, unwarranted inclusions and the ability to consolidate related work.

First, the migration guides are scraped from Github, looking at the `C-Breaking-Change` label. These are then compiled into a single list using the `generate-release` tool. Next, the individual migration guide entries need to be manually checked for errors, reversions, and general quality. Unlike the release notes, these may have last-minute additions in order to address critical bugs or problems that required breaking changes.

The change log is a simple list of all PRs, sorted by area, that were merged between the last Bevy release and the current one. This includes those shipped in patch releases as well! This step is largely mechanical, although it has historically been organized by area.

Finally, the list of contributors is a simple scraped, randomized list of the complete set of PR authors.

## Motivation

Preparing a new Bevy release is a *ton* of work, and often ends up bottlenecked on a single central PR with hundreds of comments that's prepared at the last minute.

This responsibility usually lands on a single person, who authors the PR and coordinates the release.

Keeping track of all of the work is painful, but the single most serious problem is how this interacts with parsed / generated migration guide data. Currently, `generate-release` simply scrapes the relevant section from each PR, compiles it into a list, and then the person running the tool submits a PR with the migration guide in a single file.

If we need to regenerate that list (due to changes), any manual cleanup is lost (unless an inordinate amount of care is taken).

The consequences of this are:

1. We can't start generating and editing the migration guide until the very end of the release cycle, once we have a feature freeze.
2. Extending this to the release notes is fundamentally infeasible, as it *all* needs to be carefully written and revised. Individual PRs simply don't have the context required to frame a feature, and the audience is different.

## Core Idea

The core problem here is that both Git and any tooling we might build really struggles to work with highly structured text documents in a single file.

Instead, we should have one *folder* per document we're producing (release notes / migration guide), and then store each section in individual files. Then, in the central file, import the various component files during the generation of the website according to human-friendly annotations.

This allows us to isolate changes from each other correctly and generate / regenerate sections of the migration guides and release notes independently.

[...]

PR #1165, opened 2024-05-15
Generate multiple files for migration guide

IceSentry:
The `generate-release --migration-guide` tool now generates a file for each PR with a migration guide.

I also added a new `--combine-migration-guides` tool to combine all those files into a single large file. The intent is for that file to be generated in CI and be used directly as a zola page.


alice-i-cecile:
Part of #1163.

stan4cb (approved):
looks good

generate-release/src/main.rs
```diff
@@ -15,33 +17,26 @@ mod release_notes_website;
...
-/// * Generate the list of merge PR
-///
 /// Requires a valid `GITHUB_TOKEN` environment variable, you can use a .env file or use your preferred method of passing env arguments.
 ///
-/// Example used to generate for 0.9:
-/// cargo run -- migration-guide --from v0.9.0 --to main --title "0.9 to 0.10" --weight 6
-/// cargo run -- release-note --from v0.9.0 --to main
+/// Example used to generate the 0.14 release:
+/// cargo run -- migration-guide --from v0.13.0 --to main --release-version 0.14
+/// cargo run -- combine-migration-guides --title "0.13 to 0.14" --weight 9 --release-version 0.14
```
alice-i-cecile, line 24: Are we able to automate generation of the "weight" parameter?

generate-release/src/main.rs
```diff
@@ -59,10 +70,20 @@ enum Commands {
...
+        release_version: String,
+
         /// Path used to output the generated file. Defaults to ./migration-guide.md
         #[arg(short, long)]
-        path: Option<std::path::PathBuf>,
+        output_path: Option<std::path::PathBuf>,
     },
+    /// Release notes:
+    /// * Gets all merged PRs
+    /// * Collect each author of closed PRs (Should this just list all contributors?)
```
alice-i-cecile, line 83:
```suggestion
    /// * Collect each author of merged PRs
```
We want to be accurate for "contributors for this release"

alice-i-cecile (approved):
On board with this as a next step :)

generate-release/src/main.rs
IceSentry: Not easily, that would require parsing all previous guides to figure out the latest and also updating introduction.md

It's technically possible, but way out of scope

generate-release/src/main.rs
IceSentry: I think I added that comment because this doesn't account for reviewers and I really want reviewers to be mentioned in things like that, but yeah, right now it uses the merged PRs only.

generate-release/src/main.rs
alice-i-cecile: Yep, definitely for follow-up!

PR #1170, opened 2024-05-16
Fix On This Page algorithm

SIGSTACKFAULT:
scope-creep edition!

- sort by the actual Y position. and not the top border width.
- sort in the correct direction
- handle the case where there are no `<h2>`s (which is the case on migration guides)
- compare `<a href>` which need to be there anyway instead of `<li data-fragment>`

### DIFF
--- a/sass/components/_on-this-page.scss
+++ b/sass/components/_on-this-page.scss
@@ -15,25 +15,27 @@
     padding-inline: $h-padding;
   }
 
-  li[data-active="true"] > a {
-    color: #ececec;
-  }
-
   a {
+    $color: #868686;
     display: block;
     text-wrap: balance;
     padding-block: 4px;
-    color: #868686;
+    color: $color;
     text-decoration: none;
     word-break: break-word;
-
-    &:hover {
-      color: lighten(#868686, 20);
-    }
 
     @media ($bp-tablet-portrait-down) {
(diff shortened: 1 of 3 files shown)


TrialDragon (approved):
LGTM. Tho why do we use `<h3>` instead of `<h2>` as the next step down from a `<h1>` in the migration guides I wonder. Semantically weird, and style-wise a `<h2>` can be styled to look correctly.

IceSentry:
>  Tho why do we use <h3> instead of <h2>

The code from that is pretty old, so my guess is just that cart liked the look of a h3 better and nobody questioned it since, but yeah, we should just use h2 and update the styling

SIGSTACKFAULT:
scope creep >.<

TrialDragon (approved):

SIGSTACKFAULT:
Will rebase when I get home tomorrow

SIGSTACKFAULT:
Also request from @Jondolf because after rebasing i had to mess with your styles

alice-i-cecile (approved):

alice-i-cecile:
@SIGSTACKFAULT once merge conflicts are resolved I'll merge this in for you.

SIGSTACKFAULT:
@alice-i-cecile resolved!

PR #1171, opened 2024-05-16
Fix TOC overflow in Migration Guides

doup:
Right now TOC is not adding line breaks in the Migrations pages so the text is overflowing:

image

---

This is after the fix:

image

### DIFF
--- a/sass/components/_on-this-page.scss
+++ b/sass/components/_on-this-page.scss
@@ -31,6 +31,7 @@
     // copied from .media-content
     text-decoration: none;
     color: $link-color;
+    word-break: break-word;
 
     &:hover {
       text-shadow: 0 0 0.9px $link-hover-shadow-color, 0 0 0.9px $link-hover-shadow-color;
--- a/sass/components/_tree-menu.scss
+++ b/sass/components/_tree-menu.scss
@@ -23,7 +23,7 @@
 
     &__label {
         display: grid;
-        grid-template-columns: auto min-content;
+        grid-template-columns: 1fr min-content;
         border-radius: $border-radius;
         overflow: hidden; // prevent the radii from getting messed up
         margin-bottom: 2px;


TrialDragon (approved):
LGTM. Tho, looking at the screenshots it's hard to tell where one entry ends and another begins; perhaps the line spacing between entries should be increased. (Tho, that is probably suited for its own PR).

SIGSTACKFAULT:
Spent the last hour fiddling with overflow settings trying to fix this XD

SIGSTACKFAULT (approved):

doup:
> LGTM. Tho, looking at the screenshots it's hard to tell where one entry ends and another begins; perhaps the line spacing between entries should be increased. (Tho, that is potentially suited for its own PR).

I agree, for me also it's too crowded. I suggested to increase the padding, but we ended up with 2px separation between elements (`.on-this-page a { padding-block: 1px; `).

These are with 2px (current), 4px (`padding-block: 2px;`) & 6px (`padding-block: 3px;`) separation:

image

image

image

---

These are with 2px (current), 4px & 6px separation:

image

image

image

@SIGSTACKFAULT  maybe we can reconsider this? 😇 

> Spent the last hour fiddling with overflow settings trying to fix this XD. I forgot that frs are fractions of the remaining space while writing this apparently.

Haha, I was just lucky then, it's the first one I tried as I usually work with just `fr` and `px` on Grids.

TrialDragon:
Mm, looking at that I, at least, prefer a `6px` separation; it feels obvious as to where entries begin and end compared to the other two.

doup:
@TrialDragon @SIGSTACKFAULT I noticed that one the sections of the migration guide was still overflowing because it had a couple of "words" bigger than the TOC width so I had to slap a `word-break: break-word;` there. This is the ofender:

image

---

Regarding the spacing, I'll open a PR with 6px. Let's discuss there (in case @SIGSTACKFAULT disagrees).

PR #1175, opened 2024-05-17
Increase TOC spacing between entries

doup:
TOC is quite dense and sometimes is difficult to identify the different entries. This PR increases the space between the TOC entries so it's more legible.

See discussion and screenshots in https://github.com/bevyengine/bevy-website/pull/1171#issuecomment-2116859193.

### DIFF
--- a/sass/components/_on-this-page.scss
+++ b/sass/components/_on-this-page.scss
@@ -27,7 +27,7 @@
   a {
     display: block;
     text-wrap: balance;
-    padding-block: 1px;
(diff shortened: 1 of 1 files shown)


IceSentry:
Could be done in another PR, but are there any plans to remove the link styling? It kinda makes sense for the headers since they are actual links, but for the TOC it's mostly just ugly I think.

IceSentry (approved):

TrialDragon (approved):

PR #1176, opened 2024-05-17
Change text colors for "On This Page" section

Jondolf:
The new "On This Page" section for the sidebar has blue links, which feel out of place and look a bit distracting.

This PR changes the text colors to gray for non-active items and a near-white for active items, also removing the on-hover text shadow in favor of simply lightening the text color.

Before:

https://github.com/bevyengine/bevy-website/assets/57632562/4a3948a2-543b-48f8-9e3f-4d7526d3b119

After:

https://github.com/bevyengine/bevy-website/assets/57632562/d7bdc83f-eb5d-4df7-98a1-df7ced88dcc6

Note that the whole website appears lighter in the above videos for some color encoding reasons, and don't perfectly reflect what it actually looks like.

Here are close-up screenshots with more accurate colors:

kuva
kuva

(ignore that black border on the right and bottom, not sure what that's about)

### DIFF
--- a/sass/components/_on-this-page.scss
+++ b/sass/components/_on-this-page.scss
@@ -1,7 +1,6 @@
 /// internal page navigation
 /// i.e., the "On This Page" section
 
-
 .on-this-page {
   $h-padding: 12px;
   padding-block: calc($h-padding/2);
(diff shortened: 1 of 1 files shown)


sass/components/_on-this-page.scss
```diff
@@ -12,29 +11,24 @@
...
 
-  li[data-active=true] {
-    font-weight: bolder;
-
-    li {
-      font-weight: normal;
-    }
+  li[data-active="true"] > a {
+    color: #ececec;
   }
```
Jondolf, line 20: I tried this with bold too, but with white it looked too distracting in my opinion. I can change it back to bold or some in-between weight (if the font supports it) in case others prefer it though

sass/components/_on-this-page.scss
```diff
@@ -12,29 +11,24 @@
...
     padding-block: 3px;
-    // copied from .media-content
+    color: #868686;
     text-decoration: none;
-    color: $link-color;
     word-break: break-word;
 
     &:hover {
-      text-shadow: 0 0 0.9px $link-hover-shadow-color, 0 0 0.9px $link-hover-shadow-color;
+      color: lighten(#868686, 20);
```
Jondolf, line 31: I'm not entirely sure if there's an existing color that would work for this (a color between the gray and white), so I just used `lighten`. Let me know if there's a better fit

sass/components/_on-this-page.scss
doup: I wouldn't use bold as in some rare cases the text might jump. Just using color should be enough.

sass/components/_on-this-page.scss
doup: I've tried it locally and for me it looks perfect, subtle yet visible. 👍

doup (approved):

TrialDragon (approved):

PR #1178, opened 2024-05-17
Contributor's Guide: Create "Using Github Issues"

BD103:
Part of #1146. This covers both a guide on when to create issues, as well as a more general-purpose triage reference.

PR #1179, opened 2024-05-20
Hard-code possible docs template root sections

TrialDragon:
Hopefully should unblock #1178 

The previous, more procedural, way of determining docs root sections did not take into consideration possible docs sections outside of `/learn/*`. I would've done a better procedural way, but the lack of while loops in Tera threw a wrench in that plan, so for now this is the best I can do without further blocking progress on the contributing guide.

### DIFF
--- a/templates/layouts/base.html
+++ b/templates/layouts/base.html
@@ -7,6 +7,18 @@
   {% set_global section_or_page = section | default(value=page) %}
 {% endif %}
 
+{% if section_or_page %}
+  {# Possible docs root section paths. A temporary workdown. TODO: find a better solution #}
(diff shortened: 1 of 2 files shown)


templates/layouts/base.html
```diff
@@ -7,6 +7,18 @@
   {% set_global section_or_page = section | default(value=page) %}
 {% endif %}
 
+{% if section_or_page %}
+  {# Possible docs root section paths. A temporary workaround. TODO: find a better solution #}
```
alice-i-cecile, line 11: Make an issue for "find a better solution" and link it here please :)

BD103 (approved):
Thank you for working on this!

PR #1180, opened 2024-05-20
Fix wrong docs section name showing up in the mobile page menu

TrialDragon:
Before in the mobile page menu the page name / docs section name for the menu was showing as Migration Guides for all but the Book since the page name hadn't been updated as new docs sections has been added.

### DIFF
--- a/templates/docs.html
+++ b/templates/docs.html
@@ -1,5 +1,6 @@
 {% extends "layouts/page-with-menu.html" %}
 {% import "macros/docs.html" as docs_macros %}
+{% import "macros/base.html" as base_macros %}
 
 {% block head_extensions %}
   {% set_global ancestor_is_public_draft = false %}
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

BD103 (approved):
Looks good!

SIGSTACKFAULT (approved):

mockersf:
This seems a partial copy of https://github.com/bevyengine/bevy-website/blob/4147812fa2d8e69f7b9da58cb659a1d7f65f7cae/templates/layouts/base.html#L101-L125

Should all the cases be copied here? Instead of it being duplicated, could it be done only once and reused?

TrialDragon:
> This seems a partial copy of
> 
> https://github.com/bevyengine/bevy-website/blob/4147812fa2d8e69f7b9da58cb659a1d7f65f7cae/templates/layouts/base.html#L101-L125
> 
> Should all the cases be copied here? Instead of it being duplicated, could it be done only once and reused?

Hmm, I guess we could make a general use macro that outputs the same text and call it to avoid duplicating this info and keep it one place. Tho obviously there would be unreachable cases in the `Features` and `News` cases in regards to the docs template.

I will update this PR to do this in a bit, until then removing `Ready-for-final-review` and going into draft.

BD103 (approved):
Looks good! (Sorry for letting this sit for so long 😅)

PR #1182, opened 2024-05-21
add print styling

SIGSTACKFAULT:
- Hides everything but the content
- asks the browser nicely to not put page breaks after headers, inside headers, inside paragraphs, or inside code
- hides anchor links (`#`es beside headers)
- Shows link URLs
- fully outlines callouts

image

example of a whole printed page: [Setup.pdf](https://github.com/bevyengine/bevy-website/files/15383780/Setup.pdf)

### DIFF
--- a/sass/_print.scss
+++ b/sass/_print.scss
@@ -1,0 +1,24 @@
+@media print {
+  .print\:hidden {
+    // stolen from tailwind in case we ever switch
+    display: none !important;
+  }
+
+  .callout {
+    border: 4px solid var(--callout-accent-color);
+  }
+
+  .anchor-link {
+    display: none;
+  }
+
+  h1, h2, h3 {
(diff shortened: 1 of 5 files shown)


TrialDragon (commented):
This may be my ignorance speaking, but _why_ is this useful and _what_ does it do on a bigger scale? My best guess is it's for printing out the site in some way, but I fail to see how a hidden styling would improve that.

SIGSTACKFAULT:
It makes it look nicer if for some reason you decide to print out one of the pages. That's all.

TrialDragon (approved):

alice-i-cecile (approved):
Good: I think this is useful enough to exist and the end result is solid.

I can definitely see users deciding that reading the book (or even release notes) in print form was easier for them.

PR #1194, opened 2024-05-23
Generate authors list for release notes

### LINKED ISSUE #1192: `release-notes` CLI option should try its best to list authors
Scraping the authors of the commits would go a long way to automating this process. See https://github.com/bevyengine/bevy-website/pull/1191 to see the current results.

Part of #1163.

IceSentry:
Closes #1192

### DIFF
--- a/generate-release/src/migration_guides.rs
+++ b/generate-release/src/migration_guides.rs
@@ -114,7 +114,7 @@ fn get_prs_by_areas(
 
 /// Generates the metadata markdown section for the given values
 fn generate_metadata_block(
-    title: &String,
+    title: &str,
     file_name: &String,
     areas: &[String],
     pr_number: i32,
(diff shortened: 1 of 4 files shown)


doup (approved):
I was going to mention that maybe this could go in the TOML file… but then so should the title, and blah, blah, blah; looks like a path of pain. **This is good.** 😅

doup:
Although, we already do it for the Migration Guides. Any reason why we shouldn't do the same here? So the release notes MD files would only contain the "body" text. 🤔

I was discarding it in the previous comment because I thought that then the TOC wouldn't work, and so on. I think I'll give it a try…

doup:
I couldn't resist: https://github.com/IceSentry/bevy-website/pull/14 🙈

alice-i-cecile (approved):

Issue #1195, opened 2024-05-23
Generated migration guide is inconsistent


alice-i-cecile:
> Bit of a nit, but I'm not a fan of all the white space inconsistency in the migration guide. How easy would it be to call `str::trim` on the file contents, then write an extra `\n`?

_Originally posted by @BD103 in https://github.com/bevyengine/bevy-website/issues/1191#issuecomment-2126861114_
            
> Btw, the first thing I did pretty much every release was capitalize every title correctly and standardize all the before/after code blocks to use // bevy 0.13 / // bevy 0.14 people use a bunch of different ways to do that but it's pretty easy to just do a search and replace. Although now you'll need to do a search replace in the folder I guess

- @IceSentry

alice-i-cecile:
We should be able to do a basic formatting pass on the contents of the migration guide before generating them :)

Release notes don't need this step: the starting content needs to be completely rewritten anyways.

PR #1191, opened 2024-05-22
Add initial migration guides and release notes for 0.14

alice-i-cecile:
As part of #1188, we need to generate the initial release notes and migration guide for the release.

These files will be manually edited and revised, as described in #1163, before finally being published.

# Note to reviewers

These are simple stubs: they should be merged without review of the content.

Instead, this is largely a test of the new process outlined in #1163.

PR #1196, opened 2024-05-26
Normalize migration guide whitespace

alice-i-cecile:
> Bit of a nit, but I'm not a fan of all the white space inconsistency in the migration guide. How easy would it be to call str::trim on the file contents, then write an extra \n?

- @BD103 in https://github.com/bevyengine/bevy-website/pull/1191#issuecomment-2126861114

Progress towards #1195.


BD103 (approved):
Thank you! Looks good :)

jgayfer (approved):

PR #1202, opened 2024-05-26
Release notes for bevy_color

alice-i-cecile:

PR #1203, opened 2024-05-26
Cut some unhelpful release notes

alice-i-cecile:
Some of the PRs with the `C-Need-Release-Notes` label are not very relevant to Bevy's end users, and don't deserve a section in the release notes. Once this PR is merged, I'll go back to the original PR and remove the label so they don't get re-added next time we run the tool.

My criteria for inclusion in the release notes are:

1. Relevant to the average Bevy user.
2. Technically impressive.
3. Visually impressive.

I want to ensure that we can write something interesting for every section we have listed.

Suggestions on more to cut, or disagreement on any PRs that I have cut is welcome!

Note that there are quite a few related PRs that should be grouped still (hi gizmos and primitives). These will be handled in a later pass: we're only looking for work that we don't need to call out in the grouped section.

P.S. Yes I'm aware I added the label to most of these PRs myself! Hindsight is 20-20 :p


jgayfer (approved):

release-content/0.14/release-notes/12331_We_must_have_googly_eyes_new_Game_example.md
BD103, line 1: I think this should be kept. While it is important to look professional, I think including this will showcase how fun and inviting the Bevy ecosystem is. Yes, we are competing to make a great game engine that AAA studios can use, but we also have fun while doing it. This message is very important, because it encourages outsiders to join in. All are welcomed here, and each new person benefits the engine as a whole.

BD103 (changes requested):

release-content/0.14/release-notes/12331_We_must_have_googly_eyes_new_Game_example.md
IceSentry: I don't think it's a matter of being fun or not, I think it just doesn't add much. The release is already going to be massive so that feels like bloat to me. I don't really mind either way tbh.

alice-i-cecile: Yeah, that was my rationale. There's *so* much to cover, and I want to be respectful of our readers' time.

BD103: Ok, I will leave it up to you to make the decision. I'm giving this my approval.

BD103 (approved):

ChristopherBiscardi (approved):
makes sense to me. There's nothing being cut here that I would feel the need to call out separately when talking about the 0.14 release. color is getting its own in #1202 and meshing isn't necessary if the primitives are already mentioned imo

alice-i-cecile:
I've unlabelled the PRs removed here :)

PR #1205, opened 2024-05-27
Initial release notes for computed and substates

alice-i-cecile:
This is an unusually complex feature: please complain if you can't follow the example or don't understand the point of the feature!

### DIFF
--- a/release-content/0.14/release-notes/11426_Computed_State__Sub_States.md
+++ b/release-content/0.14/release-notes/11426_Computed_State__Sub_States.md
@@ -1,1 +1,107 @@
-TODO
+Bevy's `States` are a simple but powerful abstraction for managing the control flow of your app.
+
+```rust
+use bevy_state::prelude::*;
+use bevy_ecs::prelude::IntoSystemConfigs;
+use bevy_ecs::system::ResMut;
+
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, Default, States)]
+enum GameState {
+ #[default]
+  Menu,
+  InGame,
+}
+
+fn handle_escape_pressed(mut next_state: ResMut<NextState<GameState>>) {
+    if escape_pressed {
+        next_state.set(GameState::Menu);
+    }
+}
+
+fn open_menu() {
+    // Show the settings menu...
+}
+
+App::new()
+    .init_state::<GameState>()
+    .add_systems(Update, handle_escape_pressed.run_if(in_state(GameState::Menu)));
+    .add_systems(OnEnter(GameState::Menu), open_settings_menu);
+```
+
+But as users' games (and non-game applications!) grew in complexity, their limitations became more apparent.
+What happens if we want to capture the notion of "in a menu", but then have different states corresponding to which submenu should be open?
+What if we want to ask questions like "is the game paused", but that question only makes sense while we're within a game?
+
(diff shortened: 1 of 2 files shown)


lee-orr:
This is an awesome write-up! My only question is whether we should mention the two potentially breaking changes:
- if you are using `default-features = false`, you need to manually enable the feature / if using `bevy_ecs` directly you need to import `bevy_states`
- if you had multiple places where you triggered `apply_state_transition`, you should now instead run the `StateTransition` schedule

release-content/0.14/release-notes/11426_Computed_State__Sub_States.md
```diff
@@ -1 +1,107 @@
...
+App::new()
+   .init_state::<GameState>()
+   .add_computed_state::<InGame>()
+   .add_substate::<GamePhase>()
+```
+
+Just like any other state, computed states and substates work with all of the tools you're used to:
+the `State` and `NextState` resources, `OnEnter`, `OnExit` and `OnTransition` schedules and the `in_state` run condition.
+
+The only exception is that, for correctness reasons, computed states *cannot* be mutated manually.
```
MiniaczQ, line 101:
```suggestion
The only exception is that, for correctness, computed states *cannot* be mutated through `NextState`.
```
Bit more direct?

release-content/0.14/release-notes/11426_Computed_State__Sub_States.md
```diff
@@ -1 +1,107 @@
...
+Instead, they are strictly derived from their input states, and added, removed and updated automatically during state transitions based on the provided `compute` method.
+
+All of Bevy's state tools are now found in a dedicated `bevy_states` crate, which can be controlled via a feature flag.
```
MiniaczQ, line 104: Note about a breaking change here?

MiniaczQ (approved):
Looks correct and inviting

alice-i-cecile:
I think the breaking change is best relegated to the migration guide: using both pure bevy_ecs and still using states is a minority of a minority situation.

release-content/0.14/release-notes/11426_Computed_State__Sub_States.md
```diff
@@ -1 +1,107 @@
...
+    }
+}
+
+fn open_menu() {
+    // Show the settings menu...
+}
+
+App::new()
+    .init_state::<GameState>()
+    .add_systems(Update, handle_escape_pressed.run_if(in_state(GameState::Menu)));
```
theotherphil, line 28: Should this say `in_state(GameState::InGame)` or do I just not understand Bevy states?

alice-i-cecile: Yes, it should.

release-content/0.14/release-notes/11426_Computed_State__Sub_States.md
```diff
@@ -1 +1,107 @@
...
+```rust
+/// Let's begin with our simple state above
+#[derive(States, Clone, PartialEq, Eq, Hash, Debug, Default)]
+enum GameState {
+    #[default]
+    Menu,
+    // The addition of this field means that simply checking for `GameState::InGame` doesn't work:
+    // the values are different based on its value and we may want to distinguish between game systems that run
+    // when the game is paused or not!
+    InGame {
```
B-Reif, line 49: I think it's a little confusing to have this enum case share a name with our derived state example. Maybe we could rename this to `Running` or something?

B-Reif (commented):
Looks good, a couple things I don't understand.

PR #1206, opened 2024-05-27
Remove bevy_dev_tools from release notes

alice-i-cecile:
This was missed in #1203.

Unfortunately this crate is currently just a stub: there's nothing particularly interesting or relevant to tell users about this cycle.


BD103 (approved):
I agree, though we need to make sure to include it in next release's notes.

pablo-lua (approved):
as bevy_dev_tools is not yet completed, agree here that its better to cut it off

matiqo15 (commented):
I also don't think it is a good idea to make a whole section for it, but we are mentioning the ui debug overlay that lives in this crate - maybe we could add a short paragraph in that section?

PR #1207, opened 2024-05-27
Write up dev tools abstraction future work

alice-i-cecile:
Follow-up to #1206.


pablo-lua (approved):
Good and linking to the RFC so we can get more reviews, all good!

rewin123:
I think the purpose of bevy_dev_tool is not quite clear from the text 

I've rewritten to shift the focus a bit. What do you think of the option below? 

"We want to create a place for tools and overlays that speed up game development such as performance tracking, fly camera, ci testing, fast object spawning. We have already created a crate bevy_dev_tool for these and are working on creating a "dev tool abstraction"(link). An abstraction that will allow a unified way to enable/disable, customise and group them in toolboxes to create something like Quake console or VSCode Command Palette with customizable dev tools set! "

alice-i-cecile:
Yeah, I think that yours is a bit clearer :) I'll swap to something like that.

rewin123 (approved):

PR #1208, opened 2024-05-27
Remove taffy upgrade from release notes

alice-i-cecile:
@nicoburns, I'd appreciate your review here for correctness and context :)

### DIFF
--- a/release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
+++ b/release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
@@ -1,5 +1,0 @@
-[`taffy`](https://crates.io/crates/taffy), the UI layout crate that we rely on, has had a major update.
-In addition to various bug fixes, this brings two new pieces of functionality to users of `bevy_ui`:
-
-1. `Display::Block`: use the [CSS block layout model](https://www.w3schools.com/css/css_display_visibility.asp) to layout UI nodes. Elements will start on a new line, and take up the whole width. Great for in-line images!
-2. `Overflow::Hidden`: if elements overflow their container, they will be completely hidden, rather than shown or clipped.
--- a/release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
+++ b/release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
@@ -1,5 +1,5 @@
 [`taffy`](https://crates.io/crates/taffy), the UI layout crate that we rely on, has had a major update.
 In addition to various bug fixes, this brings two new pieces of functionality to users of `bevy_ui`:
 
-1. `Display::Block`: use the [CSS block layout model] to layout UI nodes. Elements will start on a new line, and take up the whole width. Great for in-line images!
+1. `Display::Block`: use the [CSS block layout model](https://www.w3schools.com/css/css_display_visibility.asp) to layout UI nodes. Elements will start on a new line, and take up the whole width. Great for in-line images!
 2. `Overflow::Hidden`: if elements overflow their container, they will be completely hidden, rather than shown or clipped.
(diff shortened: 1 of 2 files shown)


release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
```diff
@@ -1 +1,5 @@
-TODO
+[`taffy`](https://crates.io/crates/taffy), the UI layout crate that we rely on, has had a major update.
+In addition to various bug fixes, this brings two new pieces of functionality to users of `bevy_ui`:
+
+1. `Display::Block`: use the [CSS block layout model] to layout UI nodes. Elements will start on a new line, and take up the whole width. Great for in-line images!
```
Themayu, line 4: Are you missing a link here?

Themayu: I-

This is specific to line 4. _Really_ don't know why it included lines 1 through 3.

alice-i-cecile: If you select multiple lines on the left you can leave comments on the whole block :) Easy to do by accident if you're not aware of the feature.

release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
```diff
@@ -1 +1,5 @@
-TODO
+[`taffy`](https://crates.io/crates/taffy), the UI layout crate that we rely on, has had a major update.
+In addition to various bug fixes, this brings two new pieces of functionality to users of `bevy_ui`:
+
+1. `Display::Block`: use the [CSS block layout model](https://www.w3schools.com/css/css_display_visibility.asp) to layout UI nodes. Elements will start on a new line, and take up the whole width. Great for in-line images!
```
nicoburns, line 4: This is not great for inline images! We need "inline" layout for that. This just does vertical block layout. IMO the primary use case would be matching the layout of an existing UI ported from web. Otherwise you're better off using Flexbox.

nicoburns: For inline we'll need something like https://github.com/linebender/parley/pull/67 (or similar support in cosmic text)

release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
```diff
@@ -1 +1,5 @@
...
+2. `Overflow::Hidden`: if elements overflow their container, they will be completely hidden, rather than shown or clipped.
```
nicoburns, line 5: There isn't really any difference between "hidden" and "clipped" in this sentence. In Taffy difference between `overflow: hidden` and `overflow: clip` is primarily that `overflow: hidden` turns off the automatic minimum size of a flexbox or grid child. Again, the main benefit here is that `overflow: hidden` will match `overflow: hidden` on the web (`overflow: clip` is arguably often what you want but isn't widely used on the web because it's much newer than "hidden").

release-content/0.14/release-notes/10690_Upgrade_to_Taffy_04.md
```diff
@@ -1 +1,5 @@
-TODO
+[`taffy`](https://crates.io/crates/taffy), the UI layout crate that we rely on, has had a major update.
```
nicoburns, line 1: Doesn't affect content in this PR (except the filename), but it's worth noting that the upgrade is now to 0.5 rather than 0.4

alice-i-cecile:
> Yeah, I would agree. Changes in 0.4 were mostly under the hood (improvements to the low-level api traits). And 0.5 was just bugfix release that required a breaking change.

Removing entirely, per discussion with @nicoburns [on Discord](https://discord.com/channels/691052431525675048/743663673393938453/1246885765963583559)

PR #1209, opened 2024-05-27
Add initial release notes for commands + exclusive world access

alice-i-cecile:
I've agglomerated the changes from https://github.com/bevyengine/bevy/pull/11823 and added one of the convenient side effects of https://github.com/bevyengine/bevy/pull/10756 to round it out.

These felt related to me, and the new `World::commands` API is something I've wanted for years :)


james-j-obrien:
Looks good to me!

(hopefully one day with command batching using the command queue will become the more performant solution)

PR #1210, opened 2024-05-27
Write up deprecate dynamic plugins

alice-i-cecile:


BD103 (approved):
Great working, thanks! 🚀 (Also I'm assuming we're still operating on the one-approval system for now, before we can do a final pass.)

PR #1217, opened 2024-05-28
`Node2d` migration guide

BD103:
Original pull request: https://github.com/bevyengine/bevy/pull/12038

The migration guide originally mentioned the struct `bevy_mikktspace::generated::SGroup`, but that actually is not public API so I removed it.


BD103:
I'm not sure how to mute the Typos check for that specific line, since it is intentional misspelled. Any ideas?

alice-i-cecile (changes requested):
Add the file itself to the `typos.toml` ignore list :)

PR #1222, opened 2024-05-29
Automatically generate an issue for each `C-Needs-Release-Note` PR merged to Bevy

### LINKED ISSUE #1220: Automatically generate issues for each uncompleted release note
Two of the largest remaining challenges involved in #1163 are:

1. Prompting feature authors to write release notes for the corresponding sections.
2. Keeping track of what release note sections still need to be written, especially as more sections are completed.

[Previously](https://github.com/bevyengine/bevy-website/pull/891#issuecomment-1914655314), this was done by manual suggestions and tracking. This is a) a lot of work to generate and b) generates an endless stream of notifications for everyone involved.

There are two proposals for how we could handle and automate this.

# Proposal 1: big tracking issue

As part of a tracking issue for the release (e.g. #1188), we could generate a checkbox list like before, suggesting that all of the authors listed on the PR handle the process.

## Advantages

- no new issues generated.

## Disadvantages

- everyone gets pinged for every update
- the thread gets very long as people chime in to claim issues
- only 50 people can be pinged in a single comment
- only maintainers and the comment author can update the status of each ticket
- discoverability is poor
- doesn't work terribly well when done on a continual basis: we need to make the tracking issue for the next release immediately and constantly edit it?

# Proposal 2: a million issues

Instead, for each release note entry generated (aka each merged PR with a `C-Needs-Release-Note` entry, a bot makes an issue with a  "Write release notes for PR TITLE" with a link to the PR and pings the authors. This issue is added to the milestone.

## Advantages

- discussions are separated, and thus much less noisy
- much easier to discover work that needs to be done
- PRs can say "Closes #X" or the triage team can update this and it will work great without maintainer intervention
- we can add advice on how to write release notes to every issue
- issues can be generated dynamically as PRs are merged throughout the cycle

## Disadvantages

- makes a large number of issues! This will be noisy for 0.14 due to the last-minute nature of the work
- need to figure out more complex bot workflows to write this tool

alice-i-cecile:
Fixes #1220.

As discussed there, this PR will cause running the `generate-release` tool to automatically generate an issue for each release note section, pinging the authors to help write it.

While this will be noisy initially, this should be very helpful in coordinating the release process, and ease the notification burden on individual contributors by providing separate places where they can discuss each release note.

## Sample output

See https://github.com/alice-i-cecile/bevy-website/issues/10

## To do

- [x] make sure not to open duplicate issues
- [x] get the full list of issues, not just one
- [x] fix JSON conversion error
- [x] validate that duplication checking works as intended
- [x] validate that issues can be created successfully

## Future work

- look up the correct milestone for the release, and add the generated issue to it
- add a daily job that runs this workflow, opening PRs to update the release notes if needed


BD103:
Managed to debug while this was erroring: It was due to some issues having `"closed_at": null`. We only expected a `DateTime`, not an `Option<DateTime>`. I fixed it with this patch, though I'm not sure if the `.unwrap()` logic I used was correct:

```patch
diff --git a/generate-release/src/github_client.rs b/generate-release/src/github_client.rs
index 3ad666c..c898803 100644
--- a/generate-release/src/github_client.rs
+++ b/generate-release/src/github_client.rs
@@ -96,7 +96,7 @@ pub struct GithubIssuesResponse {
     pub body: Option<String>,
     pub labels: Vec<GithubLabel>,
     pub user: GithubUser,
-    pub closed_at: DateTime<Utc>,
+    pub closed_at: Option<DateTime<Utc>>,
     pub pull_request: Option<GithubIssuesResponsePullRequest>,
 }
 
@@ -204,11 +204,16 @@ impl GithubClient {
             page += 1;
             if let Some(pr) = prs.last() {
                 if let Some(datetime_utc) = datetime_utc {
-                    if pr.closed_at < datetime_utc {
-                        println!(
-                            "\x1b[93mSkipping PR closed before the target datetime {}\x1b[0m",
-                            pr.closed_at
-                        );
+                    if let Some(closed_at) = pr.closed_at {
+                        if closed_at < datetime_utc {
+                            println!(
+                                "\x1b[93mSkipping PR closed before the target datetime {}\x1b[0m",
+                                closed_at
+                            );
+                            break;
+                        }
+                    } else {
+                        println!("Skipping open PR.");
                         break;
                     }
                 }
@@ -222,7 +227,7 @@ impl GithubClient {
                 datetime_utc
             );
 
-            prs.retain(|pr| pr.closed_at < datetime_utc);
+            prs.retain(|pr| pr.closed_at.is_some_and(|closed_at| closed_at < datetime_utc));
         }
 
         Ok(prs)

```

generate-release/src/release_notes.rs
```diff
@@ -107,3 +147,68 @@ file_name = "{file_name}.md"
...
+
+    let authors = authors
+        .iter()
+        .map(|author| format!("{}", author))
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    let issue_body = format!(
+        "{pr_url} needs release notes for the upcoming Bevy release!
+        Please reply below if you'd like to volunteer do so, whether or not you're the author of the PR.
```
BD103, line 182:
```suggestion
        Please reply below if you'd like to volunteer do so, even if you're the author of the PR.
```

This version is a little less wordy, while still clear that anyone can contribute.

generate-release/src/release_notes.rs
```diff
@@ -107,3 +147,68 @@ file_name = "{file_name}.md"
...
+        
+        1. Clearly motivate the change.
+        2. Be written in a way that is understandable by the average Bevy user: some programming background and a general understanding of games.
+        3. Show off the coolest features of the PR. Screenshots are awesome, but elegant APIs are also welcome!
+        4. If this was a perf-centric PR, quantify the performance improvements. Graphs and statistics work well for this.
+
+        We can help you revise the release notes: a rough draft alone is incredibly useful :)
+        Your expertise is invaluable for contextualizing the changes; we'll work with you to bring the technical writing up to par.
+
+        To submit your release notes, modify `{file_path}.md` and submit a PR.
```
BD103, line 194:
```suggestion
        To submit your release notes, modify `{file_path}` and submit a PR.
```

`.md` is already in this filename, so it gets duplicated with this!

generate-release/src/release_notes.rs
```diff
@@ -107,3 +147,68 @@ file_name = "{file_name}.md"
...
+    }
+
+    let pr_url = format!("https://github.com/bevyengine/bevy/pull/{pr_number}",);
+    let file_path = file_path.to_string_lossy();
+
+    let authors = authors
+        .iter()
+        .map(|author| format!("{}", author))
+        .collect::<Vec<_>>()
+        .join(", ");
```
BD103, line 178:
```suggestion
    let authors = authors.join(", ");
```

Clippy doesn't like this line, so here's a simplified version.

alice-i-cecile:
Issues can't be created correctly currently: we need to double check the API docs and make sure we're hitting the right end point.

release-content/0.14/release-notes/_release-notes.toml
```diff
@@ -1,355 +0,0 @@
-[[release_notes]]
```
alice-i-cecile, line 1: This should not have been changed.

generate-release/src/release_notes.rs
```diff
@@ -107,3 +147,64 @@ file_name = "{file_name}.md"
...
+        3. Show off the coolest features of the PR. Screenshots are awesome, but elegant APIs are also welcome!
+        4. If this was a perf-centric PR, quantify the performance improvements. Graphs and statistics work well for this.
+
+        We can help you revise the release notes: a rough draft alone is incredibly useful :)
+        Your expertise is invaluable for contextualizing the changes; we'll work with you to bring the technical writing up to par.
+
+        To submit your release notes, modify `{file_path}` and submit a PR.
+        In that PR, please mention this issue with the `Fixes #{pr_number}` keyphrase so it gets closed automatically.
+
+        Pinging: {authors}
```
MrGVSV, line 193: Nit: we should probably indicate why we're pinging these users. Maybe something like:
```suggestion
        Original Authors: {authors}
```

MrGVSV, line 193: I wonder if we should give some indication here that the PR authors are welcome to write the release notes themselves, act as a consultant, or do nothing if they don't want to participate in this step for whatever reason.

I feel like it's sorta implied already, but should we make it more explicit?

jgayfer (approved):
I haven't tested this, but went through it all and looks sane to me.

generate-release/src/main.rs
```diff
@@ -93,12 +96,16 @@ fn main() -> anyhow::Result<()> {
             &client,
             overwrite_existing,
         )?,
-        Commands::ReleaseNotes { overwrite_existing } => generate_release_notes(
+        Commands::ReleaseNotes {
```
BD103, line 99: (Nit) Could you add docs here describing that this command creates a bunch of issues? (Also recommend that users add `--dry-run` when testing, so we don't get spammed.)

BD103 (approved):
Looks good! (Would it be possible to test creating only a single issue, to ensure the output looks good in Github? If not then that's fine.)

alice-i-cecile:
https://github.com/bevyengine/bevy-website/issues/1225 :fire:

alice-i-cecile:
Duplication checking isn't working properly: see #1267 and #1247.

generate-release/src/release_notes.rs
```diff
@@ -107,3 +135,72 @@ file_name = "{file_name}.md"
...
+    // The weird indentation is intentional.
+    // Otherwise, the text is not formatted correctly on GitHub as the tabs are copied into the issue body.
+    let issue_body = format!(
+        "{pr_url} needs release notes for the upcoming Bevy release!
+
+Original authors: {authors}
+
+Please reply below if you'd like to write these notes. 
+While the author(s) of the PR often have the context, knowledge and motivation to draft the release notes for their feature,
+anyone can contribute release notes!
```
MrGVSV, line 173: I'm pretty sure GitHub-flavored Markdown doesn't ignore newlines, so this is actually spread across two separate lines.
```suggestion
While the author(s) of the PR often have the context, knowledge and motivation to draft the release notes for their feature, anyone can contribute release notes!
```

MrGVSV: Yeah it does separate the lines: 
Screenshot of the text being rendered across two lines instead of one

generate-release/src/github_client.rs
```diff
@@ -305,8 +336,84 @@ query {{
...
+            .post(&format!(
+                "https://api.github.com/repos/bevyengine/{repo}/issues"
+            ))
+            .set("Authorization", &format!("Bearer {}", self.token))
+            .set("Accept", "application/vnd.github+json")
+            .set("X-GitHub-Api-Version", "2022-11-28")
+            .send_json(ureq::json!({
+                "title": issue_title,
+                "body": issue_body,
+                // TODO: add the milestone.
```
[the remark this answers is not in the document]
alice-i-cecile:
```suggestion
                // TODO: add the milestone. Tracked in https://github.com/bevyengine/bevy-website/issues/1269
```

generate-release/src/release_notes.rs
```diff
@@ -107,3 +135,72 @@ file_name = "{file_name}.md"
...
+) {
+    let pr_number = pr.number;
+    let issue_title = format!("Write release notes for PR #{pr_number}: {title}");
+
+    // Check if this issue already exists
+    // If it does, we don't want to spam the repo with duplicate issues
+    if issue_titles.contains(&issue_title) {
+        println!("Issue already exists for PR #{}: {}", pr_number, title);
+        return;
+    }
```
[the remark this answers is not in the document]
alice-i-cecile: I'm not concerned enough about that happening a) within the cycle and b) after the PR is already merged to try and do a more complex deduplication strategy. At least in this PR :)

We could parse it to grab the number I suppose?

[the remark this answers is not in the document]
MrGVSV: Yeah definitely not a blocker for this PR but possibly something to consider for future work

MrGVSV (approved):

mockersf (changes requested):
we can't automatically mention people, that is a very bad idea and most likely against GitHub terms of services
automated mentions without having explicit consent if spam

generate-release/src/github_client.rs
```diff
@@ -305,8 +336,83 @@ query {{
...
+
+/// The status of an issue or PR on Github.
+#[derive(Debug, Clone, Copy)]
+#[allow(dead_code)]
+pub enum IssueState {
+    Open,
+    Closed,
+    Merged,
+    All,
+}
```
BD103, line 402:
```suggestion
#[derive(Debug, Clone, Copy)]
pub enum IssueState {
    #[allow(dead_code)]
    Open,
    #[allow(dead_code)]
    Closed,
    Merged,
    All,
}
```

This makes it still lint a warning if `IssueState` is never used at all, but ignore any warnings about `Open` and `Closed` specifically.

generate-release/src/github_client.rs
```diff
@@ -305,8 +336,83 @@ query {{
...
+
+impl IssueState {
+    /// The string representation of the issue state.
+    ///
+    /// This is requested by the Github API,
+    /// as documented [here](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues).
+    fn as_str(&self) -> &str {
```
BD103, line 409:
```suggestion
    fn as_str(&self) -> &'static str {
```

It may be worth marking this explicitly `'static` so consumers know that the results are constant. On the other hand, it doesn't matter much because the lifetime is elided. Your choice 🤷‍♂️

generate-release/src/main.rs
```diff
@@ -55,10 +55,18 @@ enum Commands {
         #[arg(short, long)]
         overwrite_existing: bool,
     },
+    /// Generates release notes for all PRs merged with the `C-Needs-Release-Note` label.
+    ///
+    /// This will also open an issue for each PR that doesn't have a release note.
+    /// While duplicate issues will not be opened, be sure to use the `dry-run` flag
```
BD103, line 61:
```suggestion
    /// While duplicate issues will not be opened, be sure to use the `--dry-run` flag
```

Adding the double-dash makes it more clear that this flag is supposed to be used in the CLI, not somewhere else. This is a stylistic choice, though, and I won't block on it. :)

generate-release/src/release_notes.rs
```diff
@@ -1,17 +1,24 @@
...
 
 use crate::{
-    github_client::GithubClient,
+    github_client::{GithubClient, GithubIssuesResponse, IssueState},
     helpers::{get_contributors, get_merged_prs},
 };
-use std::{collections::HashSet, io::Write as IoWrite, path::PathBuf};
+use std::{
+    collections::HashSet,
+    io::Write as IoWrite,
```
BD103, line 9:
```suggestion
    io::Write as _,
```

We never mention `IoWrite` explicitly, so we don't need to name it in the imports.

BD103 (approved):
Looks good! I have a few nits, but they're inconsequential stylistic opinions that can be skipped :)

alice-i-cecile:
I'm going to swap to leaving a comment in the merged PRs.

generate-release/src/release_notes.rs
alice-i-cecile:
```suggestion
    io::Write,
```

generate-release/src/release_notes.rs
```diff
@@ -107,3 +127,81 @@ file_name = "{file_name}.md"
...
+            .unwrap();
+        println!("Opened issue for PR #{}: {}", pr_number, title);
+        // Pause between opening issues to avoid getting rate-limited.
+        // See https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#pause-between-mutative-requests
+        std::thread::sleep(std::time::Duration::from_secs(2));
+
+        // Leave a comment on the PR linking to the new issue
+        let issue_url = response.html_url;
+
+        let comment = format!("Thanks for making and reviewing this PR! This work is relatively important and needs release notes! Head over to {issue_url} to help out.",);
```
MrGVSV, line 202: Nit: This wording is a bit strange to me since it almost sounds like its intended target is someone who both made and reviewed the PR (that might just be my take though lol). I also feel like the comment might come across as a demand rather than a polite request/suggestion.

Maybe we could do something like this?
```suggestion
        let comment = format!("Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to {issue_url} if you'd like to help out.",);
```

generate-release/src/release_notes.rs
```diff
@@ -107,3 +127,81 @@ file_name = "{file_name}.md"
...
+To submit your release notes, modify `{file_path}` and submit a PR.
+In that PR, please mention this issue with the `Fixes #ISSUE_NUMBER` keyphrase so it gets closed automatically.");
+
+    let labels = vec!["A-Release-Notes", "C-Content", "S-Ready-For-Implementation"];
+
+    if dry_run {
+        println!("Would open issue on GitHub:");
+        println!("Title: {}", issue_title);
+        println!("Body: {}", issue_body);
+        println!("Labels: {:?}", labels);
```
MrGVSV, line 188: Nit: Could we include the comment output here as well?

alice-i-cecile: Not really unfortunately. The critical information (which issue to link to) isn't available until we actually create the issue and get the response, which we can then parse for the link.

MrGVSV: Makes sense! It can always be added in a future PR (i.e. printing out the comment with a dummy link or something).

MrGVSV (approved):

generate-release/src/github_client.rs
```diff
@@ -194,52 +212,74 @@ impl GithubClient {
...
+                    .is_some_and(|closed_at| closed_at >= datetime_utc)
+            });
+        }
+
+        Ok(prs)
     }
 
-    // Returns all PRs from the main branch that are merged.
-    pub fn get_merged_prs_by_page(
+    // Request issues and PRs by the page returned by the Github API
```
kristoff3r, line 245:
```suggestion
    /// Request issues and PRs by the page returned by the Github API
```

kristoff3r (approved):

generate-release/src/main.rs
```diff
@@ -55,10 +55,18 @@ enum Commands {
...
+    /// This will also open an issue for each PR that doesn't have a release note.
+    /// While duplicate issues will not be opened, be sure to use the `--dry-run` flag
+    /// if you want to test the release notes generation without spamming the repo.
     ReleaseNotes {
         /// Use this if you want to overwrite existing files
         #[arg(short, long)]
         overwrite_existing: bool,
+        /// If this is set, no issues will be opened.
+        #[arg(short, long)]
+        dry_run: bool,
```
djeedai, line 69: nit: this will still overwrite any previously generated release notes, _e.g._ if you run it twice in a row. So it's not really a dry run (which normally means "no mutation").

generate-release/src/release_notes.rs
```diff
@@ -107,3 +127,81 @@ file_name = "{file_name}.md"
...
+        // See https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#pause-between-mutative-requests
+        std::thread::sleep(std::time::Duration::from_secs(2));
+
+        // Leave a comment on the PR linking to the new issue
+        let issue_url = response.html_url;
+
+        let comment = format!("Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to {issue_url} if you'd like to help out.",);
+
+        // Unwrap to warn the user if the comment fails
+        client.leave_comment("bevy", pr_number, &comment).unwrap();
```
djeedai, line 205: Can we avoid the `unwrap()` and instead log an error to whoever is running that command? Management scripts crashing is a plague of software maintenance 😅

djeedai (approved):
Only minor comments, looks good to me

generate-release/src/github_client.rs
```diff
@@ -162,30 +172,38 @@ impl GithubClient {
...
-    pub fn get_merged_prs(
+    /// Gets a filtered list of issues and PRs from `bevyengine/{repo}`.
+    ///
+    /// If `since` is provided, it should be a date in the YYYY-MM-DD format.
+    pub fn get_issues_and_prs(
         &self,
-        since: &str,
+        repo: &str,
+        state: IssueState,
+        since: Option<&str>,
```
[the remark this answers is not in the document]
alice-i-cecile: Agreed, but I don't want to touch this further in this PR.

generate-release/src/github_client.rs
```diff
@@ -305,8 +345,118 @@ query {{
...
+    Merged,
+    All,
+}
+
+impl IssueState {
+    /// The string representation of the issue state.
+    ///
+    /// This is requested by the Github API,
+    /// as documented [here](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues).
+    fn as_str(&self) -> &'static str {
```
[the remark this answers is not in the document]
alice-i-cecile: Agreed!

generate-release/src/release_notes.rs
```diff
@@ -107,3 +127,81 @@ file_name = "{file_name}.md"
...
+
+    if dry_run {
+        println!("Would open issue on GitHub:");
+        println!("Title: {}", issue_title);
+        println!("Body: {}", issue_body);
+        println!("Labels: {:?}", labels);
+    } else {
+        // Open an issue on the `bevy-website` repo
+        let response = client
+            .open_issue("bevy-website", &issue_title, &issue_body, labels)
```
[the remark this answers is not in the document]
alice-i-cecile: Agreed.

generate-release/src/release_notes.rs
alice-i-cecile: Can do!

generate-release/src/github_client.rs
```diff
@@ -114,11 +124,11 @@ impl GithubClient {
         Self { agent, token, repo }
     }
 
-    fn get(&self, path: &str) -> ureq::Request {
+    /// Submits a request to `bevyengine/{repo}`
```
BD103, line 127:
```suggestion
    /// Submits a GET request to `bevyengine/{repo}`
```

There are multiple HTTP requests types, such as GET, POST, DELETE, etc. This makes it clear which method is used.

generate-release/src/main.rs
BD103: `--no-issues` is another option, though I don't have a strong preference.

BD103 (approved):
Awesome! I have a few small nits, but overall 👍.

alice-i-cecile:
@djeedai I've done one better than constants: I've swapped to an `BevyRepo` enum and changed the signatures to make it way harder to accidentally mess up.

generate-release/src/main.rs
```diff
@@ -55,10 +56,18 @@ enum Commands {
         #[arg(short, long)]
         overwrite_existing: bool,
     },
+    /// Generates release notes for all PRs merged with the `C-Needs-Release-Note` label.
+    ///
+    /// This will also open an issue for each PR that doesn't have a release note.
+    /// While duplicate issues will not be opened, be sure to use the `--dry-run` flag
```
BD103, line 62:
```suggestion
    /// While duplicate issues will not be opened, be sure to use the `--local` flag
```

djeedai:
> @djeedai I've done one better than constants: I've swapped to an `BevyRepo` enum and changed the signatures to make it way harder to accidentally mess up.

C++ got me again! This was obviously the better solution! Thanks

djeedai (approved):

mockersf (approved):

PR #1224, opened 2024-05-30
More migration guides! (Yippee)

BD103:
Here's another round of migration guides! I'm going to get a few more done before marking it as ready-for-review.

## Open questions

- Shortcodes don't seem to work in migration guides. I tried using `{% callout() %}`, but it just rendered as plain text. This should probably be turned into an issue, though I don't know how easily it would be solved.
- I don't know much about Bevy's animation, and thus cannot accurately provide code samples for `11707_Rework_animation_to_be_done_in_two_phases.md`. Would someone with more experience be able to help me here? (cc @pcwalton, if you have the time 😄)


BD103:
Requesting a review from @NthTensor so they can look over the color section I wrote. :)

alice-i-cecile (approved):

PR #1276, opened 2024-06-02
Use pngcrush on all images

msvbg:
I noticed some images were quite big in here. This PR reduces the repo size by about ~156 MB and will probably have a positive impact on bandwidth usage. It's a big PR in terms of files changed, but all I did was to run:

```
find . -type f -iname "*.png" -exec pngcrush -ow {} \;
```

`pngcrush` optimizes the compression settings for png files to minimize file size. It is lossless, so all images should look the same.


alice-i-cecile:
Would it be possible to automate this somehow? This seems like a generally useful step we should apply by default.

alice-i-cecile (approved):

msvbg:
Yeah, it might be possible to set this up as a cron with https://github.com/calibreapp/image-actions or similar.

mockersf:
it should be done before merging png files to the repo, or it will increase the repo size a lot.

I think we were under the impression that Zola automatically compress images when moving them to the built website but I can't find that in the docs

BD103 (approved):
Looks good, though a few images seem to have increased in size. Do you know why that is?

msvbg:
I think I'll be closing this for now. It'd probably be better to do something more holistic/future-proof here.

BD103:
> > Do you know why that is?
> 
> Not sure, I suppose they were already optimized better without pngcrush.

I don't know if that's true. A significant number of images decreased in file size, I was just curious more than anything.

> I think I'll be closing this for now. It'd probably be better to do something more future-proof here.

I appreciate your concern for this, but I still think this should be merged. If anything is going to be implemented, it will be a request to optimize images _before_ they even get merged into the website. I find it unlikely that existing images will ever be optimized if not done manually.

If you don't have the time to follow through this PR then that's fine, but I still feel this is a valuable improvement. Latency is very important on the web, and decreasing image size is a great way to improve it.

PR #1334, opened 2024-06-03
Add gizmos line joints release notes

lynn-lumen:
Add release notes for 12252_Gizmo_line_joints


mockersf:
this and #1336 are making opposite affirmations. gizmos are either lines or line strips, but they're not both

alice-i-cecile:
This should get rolled into #1336.

PR #1336, opened 2024-06-03
Add gizmos line styles release notes

lynn-lumen:
Adds release notes for 12394_Gizmo_line_styles


release-content/0.14/release-notes/12394_Gizmo_line_styles.md
```diff
@@ -1 +1,54 @@
...
+fn draw_gizmos(mut gizmos: Gizmos) {
+    gizmos.line_2d(Vec2::ZERO, Vec2::splat(-80.), RED);
+}
+
+fn setup(mut commands: Commands, mut config_store: ResMut<GizmoConfigStore>) {
+    commands.spawn(Camera2dBundle::default());
+
+	// Get the config for you gizmo config group
+    let (config, _) = config_store.config_mut::<DefaultGizmoConfigGroup>();
+	// Set the line joints for this config group
```
alice-i-cecile, line 36: Comment doesn't match.

alice-i-cecile (changes requested):
We should tackle both the line styles and the joint types in the same section here.

lynn-lumen:
> We should tackle both the line styles and the joint types in the same section here.

Should I leave them in seperate files or can I merge the contents to something more coherent?
If so, are there rules for what that file should be named?

alice-i-cecile:
I've removed the tag from https://github.com/bevyengine/bevy/pull/12252 accordingly :)

alice-i-cecile (approved):
I really like the unification. Thanks!

alice-i-cecile:
@lynn-lumen you need to update the `_release_notes.toml` to remove the offending section as well.

alice-i-cecile (approved):
I'm giving in: I'll let the format-on-save win this round.

lynn-lumen:
Ooh, I didn't notice all the new whitespace :/ I'm sorry for that 😓

alice-i-cecile:
Ha no, I've been fighting this for a week. I think it's better this way, and since there's very little else in flight this won't cause merge conflicts.

PR #1340, opened 2024-06-04
0.14 migration: Delete some guides for non-breaking PRs

rparrett:
Did a pass just looking for guides that were generated for non-breaking PRs.

The only one of these that seems questionable to me is https://github.com/bevyengine/bevy/pull/13211. The PR itself didn't bump the "effective MSRV" of anything, and we have a big MSRV warning at the top of the guide. So I think it can/should be deleted.


IceSentry (approved):

BD103 (approved):
Looks good, and I agree with removing https://github.com/bevyengine/bevy/pull/13211.

rparrett:
Noticed/fixed just one more while making another pass for a different PR.

PR #1218, opened 2024-05-28
fixedbitset 0.5 migration guide

BD103:
Original PR: https://github.com/bevyengine/bevy/pull/12512

PR #1366, opened 2024-06-05
Update migration guide

BD103:
I re-ran the `generate-release` tool to create new stubs for the migration guide. Furthermore, I:

- Reverted the filenames of migration guides that I edited to their original form. (From #1217, #1218)
- Prevented deleted migration guides from #1340 being re-added.


BD103:
In order to get typos to pass, I had to exclude `_guides.toml` from checking. This means that all titles for migration guides will no longer be automatically checked. Is this ok?

alice-i-cecile:
Yeah that's fine IMO.

alice-i-cecile (approved):

PR #1378, opened 2024-06-08
Mention StatesPlugin in the migration docs

cBournhonesque:
I was blocked for a while when migrating to 0.14 because I didn't realize that on top of adding the `bevy_state` feature you need to add the `StatesPlugin`

I'm actually a bit confused as to why adding the `StatesPlugin` was needed:
- https://github.com/bevyengine/bevy/blob/main/crates/bevy_state/src/app.rs#L193 it doesn't seem to do much apart from inserting an empty `StateTransition`  schedule
- but then inserting a new state (https://github.com/cBournhonesque/bevy/blob/a4f324fdb490d75d808fd8cb54690a7b35af2f9d/crates/bevy_state/src/state/transitions.rs#L148-L148) doesn't add the sets inside the `StateTransition` schedule

Is it correct that we need the  `StatesPlugin` for states to work? It seemed to me that states did not work if I was just using the MinimalPlugins

### DIFF
--- a/release-content/0.14/migration-guides/13216_Separate_state_crate.md
+++ b/release-content/0.14/migration-guides/13216_Separate_state_crate.md
@@ -4,4 +4,4 @@ Projects that use `bevy_ecs` directly and use states will need to add the `bevy_
 
 Projects that use `bevy_app` directly and use states will need to add the `bevy_state` feature.
 
-To enable state functionality to your app, you will need to add the `StatesPlugin`.
+If you do not use `DefaultPlugins`, you will need to add the `StatesPlugin` manually to your app.
--- a/release-content/0.14/migration-guides/13216_Separate_state_crate.md
+++ b/release-content/0.14/migration-guides/13216_Separate_state_crate.md
@@ -3,3 +3,5 @@ States were moved to a separate crate which is gated behind the `bevy_state` fea
 Projects that use `bevy_ecs` directly and use states will need to add the `bevy_state` crate as a dependency.
 
 Projects that use `bevy_app` directly and use states will need to add the `bevy_state` feature.
+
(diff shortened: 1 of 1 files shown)


BD103:
`StatesPlugin` [appears to be part of](https://github.com/bevyengine/bevy/blob/3bfc42766621d12749c09eca3bfd9911dc0dc3f5/crates/bevy_internal/src/default_plugins.rs#L141-L144) `DefaultPlugins`, does your app not use that?

cBournhonesque:
> `StatesPlugin` [appears to be part of](https://github.com/bevyengine/bevy/blob/3bfc42766621d12749c09eca3bfd9911dc0dc3f5/crates/bevy_internal/src/default_plugins.rs#L141-L144) `DefaultPlugins`, does your app not use that?

Most of my examples/apps use MinimalPlugins

release-content/0.14/migration-guides/13216_Separate_state_crate.md
```diff
@@ -3,3 +3,5 @@ States were moved to a separate crate which is gated behind the `bevy_state` fea
 Projects that use `bevy_ecs` directly and use states will need to add the `bevy_state` crate as a dependency.
 
 Projects that use `bevy_app` directly and use states will need to add the `bevy_state` feature.
+
+To enable state functionality to your app, you will need to add the `StatesPlugin`.
```
BD103, line 7:
```suggestion
If you do not use `DefaultPlugins`, you will need to add the `StatesPlugin` manually to your app.
```

BD103 (approved):
Looks good! I'll likely do a second pass on the migration guide, merging all of the `bevy_state` related notes together, but for now this is good.

PR #1391, opened 2024-06-12
Migration guides (I need more creative text to add to these titles)

BD103:
These are more migration guides, but there are also 3 commits of note.

- 9d78d973e88e507ba42645c1cf85124e5c6b23f9 and e55c1a32966346c31050cc995d550b9a113bbd79 remove the migration guides for all new color spaces.
- f1de2331b35452107a2c6560797c65d8399857a9 is a hack that uses `rust` syntax highlighting for `ron` data, even though its technically incorrect. I chose prettiness over correctness in this scenario, though left a comment.


rparrett (approved):
Looks good to me.

PR #1392, opened 2024-06-12
0.14 migration: Slightly more help for users who had their colors changed

rparrett:
These are simple changes to migrate, but we are asking users to do a lot of research and could be helping them out a bit more.

I think these three colors were the extent of the damage for "color values getting changed," though maybe we could also throw "colors that simply had their names changed" into the table as well.

Do we have a handy list of those? I thought they were listed in some iteration of the migration guide previously.

For reference:

https://github.com/bevyengine/bevy/pull/12451
https://github.com/bevyengine/bevy/pull/12333
https://github.com/bevyengine/bevy/pull/12328


alice-i-cecile (approved):

BD103 (approved):
Like this change, thank you! I have one nit about capitalizing CSS, but the rest is good!

PR #1400, opened 2024-06-13
Mention RustRover, an IDE for Rust, in Getting Started

bravit:
Hi! I suggest mentioning RustRover as a dedicated IDE for Rust in the Getting Started docs.

### DIFF
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -44,7 +44,7 @@ Install the Xcode command line tools with `xcode-select --install` or the [Xcode
 
 ### Code Editor / IDE
 
-You can use any code editor you want, but we highly recommend one that has a [rust-analyzer](https://github.com/rust-lang/rust-analyzer) plugin. It's still in development, but it already provides top-tier autocomplete and code intelligence. [Visual Studio Code](https://code.visualstudio.com/) has an officially supported [rust-analyzer extension](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer). Alternatively, you can use [RustRover](https://www.jetbrains.com/rust/), a JetBrains IDE for Rust.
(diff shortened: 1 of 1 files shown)


MrGVSV (approved):

BD103 (changes requested):
Could you please mention how RustRover is only free under a non-commercial license? I'm not against it being included, but you would not be able to use it if you want to sell your game.

mockersf:
The free licence is only for non-commercial use, so you're not eligible if you intend to sell a game, or receive sponsorship.
Commercial is defined by:
> As per our [license agreement](https://www.jetbrains.com/legal/docs/toolbox/license_non-commercial/), any activity from which you, as a product user, obtain regular direct or indirect income would be classified as commercial usage.

And it's free as in "mandatory telemetry"
> It’s also important to note that if you’re using a non-commercial license, you cannot opt out of the collection of anonymous usage statistics.

bravit:
@BD103 I've mentioned that it is free for non-commercial use. Would it make sense to provide more licensing details on this page?

BD103 (approved):
I think I'm good with this, thanks!

SIGSTACKFAULT (approved):

alice-i-cecile:
We shouldn't do this. I have a few arguments for why:

1. RustRover's licensing terms aren't appropriate for our users, either for Bevy engine developers or commercial teams. Only hobbyists can safely use it free, and only if they're okay with telemetry.
2. For beginners who actually need advice on IDE selection, listing multiple options is likely to leave them overwhelmed and confused, rather than presenting a list of options that they can choose between.
3. There's no clear place to draw the line. Why RustRover and VsCode, not VsCodium, Zed, Helix, Sublime Text, Emacs and Vim?

We know that VSCode works well, it's suitable for all of our users, and my dominant concern here is a single, clear path for the new users who need this most.

PR #1401, opened 2024-06-13
Fix top of table borders

### LINKED ISSUE #1271: Top corners of tables in blog and migration guide are clipped
I noticed this while adding a table to a migration guide here: https://github.com/bevyengine/bevy-website/pull/1268

image

But this can also be seen in this section of the Bevy 0.13 blog post:

https://bevyengine.org/news/bevy-0-13/#primitive-shapes

(testing on chrome / macos)

rparrett:
## Objective

Fixes #1271

## Solution

Also apply border radius to first/last `th` elements.

## Testing

Tested in chrome and firefox on macos.

### Before

Screenshot 2024-06-13 at 11 52 10 AM

### After

Screenshot 2024-06-13 at 11 51 57 AM

### DIFF
--- a/sass/elements/_table.scss
+++ b/sass/elements/_table.scss
@@ -16,3 +16,11 @@ tr td {
 tr td:not(:first-child) {
   border-left: 2px solid $subtitle-color;
 }
+
+table tr:first-child th:first-child {
+  border-radius: $border-radius;
+}
+
+table tr:first-child th:last-child {
+  border-radius: $border-radius;
+}
--- a/sass/elements/_table.scss
+++ b/sass/elements/_table.scss
@@ -17,10 +17,15 @@ tr td:not(:first-child) {
   border-left: 2px solid $subtitle-color;
 }
 
-table tr:first-child th:first-child {
-  border-radius: $border-radius;
-}
-
-table tr:first-child th:last-child {
-  border-radius: $border-radius;
+table tr:first-child {
+  border-top-left-radius: $border-radius;
+  border-top-right-radius: $border-radius;
+  
+  th:first-child {
+    border-top-left-radius: $border-radius;
+  }
+  
+  th:last-child {
+    border-top-right-radius: $border-radius;
+  }
 }


alice-i-cecile (approved):

sass/elements/_table.scss
```diff
@@ -16,3 +16,11 @@ tr td {
...
   border-left: 2px solid $subtitle-color;
 }
+
+table tr:first-child th:first-child {
+  border-radius: $border-radius;
+}
+
+table tr:first-child th:last-child {
+  border-radius: $border-radius;
+}
```
Jondolf, line 26: Not sure if we want to make all of the corners rounded, just the top left and top right should be enough. Not that it matters that much since it doesn't really make a visual difference with the current styling.
```suggestion
table tr:first-child {
  border-top-left-radius: $border-radius;
  border-top-right-radius: $border-radius;
  
  th:first-child {
    border-top-left-radius: $border-radius;
  }
  
  th:last-child {
    border-top-right-radius: $border-radius;
  }
}
```
Alternatively, just setting `overflow: hidden` for the whole `table` fixes the issue, no extra rounding needed. Of course it would also mean that if we did have actual content overflowing for some reason, it'd get clipped (for better or for worse)

Jondolf (commented):
Ah, this seems somewhat browser-specific. I use Firefox so I didn't notice this initially.

Firefox (before this PR):



Chrome (before this PR):



Good to fix this though!

rparrett:
> I use Firefox so I didn't notice this initially.

Ah, I only tested the changed version in firefox.

sass/elements/_table.scss
rparrett: Makes sense / looks good on my end.

Jondolf (approved):

BD103:
Would you mind updating this PR from main so that CI passes? (Once it does, it should auto-merge.)

alice-i-cecile:
@BD103 FYI maintainers get a nice "update branch" button that can be used to do that without author intervention :)

PR #1404, opened 2024-06-14
Bump `code-validation` to Bevy 0.14.0-rc.2

BD103:
We're in the release stage! 🥳 Because of this, we should begin testing all website code on the latest release candidate. I additionally updated the comment to take into account the release candidate process.

When a new release candidate (or stable 0.14!) is published, this should be updated again.

I rebased this on #1403, so consider this blocked until that is merged. If you're reviewing before then, just take a look at b581c219e2938028aa9b0a66cbf7205d1811e5c4.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.23"
+version = "0.2.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
Can you add a reminder to do this to https://github.com/bevyengine/bevy/blob/main/docs/release_checklist.md as well?

BD103:
> Can you add a reminder to do this to https://github.com/bevyengine/bevy/blob/main/docs/release_checklist.md as well?

Yup! https://github.com/bevyengine/bevy/pull/13849

alice-i-cecile (approved):
Merging as trivial

PR #1406, opened 2024-06-14
Set `publish = false` for all tools

BD103:
All of the Cargo projects in this repository are internal and should never be published on <https://crates.io>. This PR sets the [`publish`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) field to false to prevent `cargo publish` from accidentally succeeding.

Furthermore, this also removes the `version` field. These tools do not need a version field because nothing outside of the repository should be depending on them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1508,7 +1508,7 @@ checksum = "98cc8fbded0c607b7ba9dd60cd98df59af97e84d24e49c8557331cfc26d301ce"
 
 [[package]]
 name = "code-validation"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "bevy",
 ]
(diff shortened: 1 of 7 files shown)


alice-i-cecile (approved):
TIL you can remove the version field completely.

BD103:
> TIL you can remove the version field completely.

Yup!

> To prevent a package from being published to a registry (like crates.io) by mistake, for instance to keep a package private in a company, you can omit the [version](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field) field. If you’d like to be more explicit, you can disable publishing:
>
> ```toml
> [package]
> # ...
> publish = false
> ```

Technically I could just remove the `version` field, but I think being explicit here is better.

mockersf:
well there's no reason to release anything from the website repo... 🤷

SIGSTACKFAULT (approved):

PR #1408, opened 2024-06-16
Auto exposure release notes

Kurble:
I've written some release notes for auto exposure

### DIFF
--- a/release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
+++ b/release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
@@ -1,4 +1,7 @@
-<!-- Implement Auto Exposure plugin -->
-<!-- https://github.com/bevyengine/bevy/pull/12792 -->
+Since Bevy 0.13, you can configure the the EV100 of a camera, which allows you to adjust the exposure of the camera in a physically based way. This also allows you to dynamically change the exposure values for various effects. However, this is a manual process and requires you to adjust the exposure values yourself.
 
-<!-- TODO -->
+Bevy 0.14 introduces a new `AutoExposurePlugin` that automatically adjusts the exposure of your scene based on the brightness of the scene. This can be useful when you want to create the feeling of a very high dynamic range, since your eyes also adjust to large changes in brightness. Note that this is not a replacement for hand-tuning the exposure values, rather an additional tool that you can use to create dramatic effects when brightness changes rapidly.
+
+[insert image here]
+
+Auto exposure is implemented by making a histogram of the scene's brightness in a post processing step. The exposure is then adjusted based on the average of the histogram. A histogram is used rather than a simple average so that additional filtering can be applied, such as ignoring very bright or very dark pixels, or giving more weight to certain parts of the image. Read the docs of the `AutoExposureSettings` component if you want to know more!
--- a/release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
+++ b/release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
@@ -4,10 +4,10 @@ Since **Bevy 0.13**, you can configure the the EV100 of a camera, which allows y
 
 <video controls><source src="auto_exposure.mp4" type="video/mp4"/></video>
 
-Bevy's Auto Exposure is implemented by making a **histogram** of the scene's brightness in a **post processing** step. The exposure is then adjusted based on the average of the histogram. Because the histogram is calculated using a **compute shader**, Auto Exposure is **not available on WebGL**. It's also not enabled by default, so you need to add the [`AutoExposurePlugin`](https://docs.rs/bevy/0.14/bevy/core_pipeline/auto_exposure/struct.AutoExposurePlugin.html) to your app.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
Writing looks great! @mockersf or @rparrett may have advice on video size / format though.

Definitely think we need a video here though: the whole feature is about being dynamic.

release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
```diff
@@ -1,4 +1,13 @@
-<!-- Implement Auto Exposure plugin -->
-<!-- https://github.com/bevyengine/bevy/pull/12792 -->
+Since **Bevy 0.13**, you can configure the the EV100 of a camera, which allows you to adjust the exposure of the camera in a physically based way. This also allows you to dynamically change the exposure values for various effects. However, this is a manual process and requires you to adjust the exposure values yourself.
 
-<!-- TODO -->
+**Bevy 0.14** introduces **Auto Exposure**, which automatically adjusts the exposure of your scene based on the brightness of the scene. This can be useful when you want to create the feeling of a very high dynamic range, since your eyes also adjust to large changes in brightness. Note that this is not a replacement for hand-tuning the exposure values, rather an additional tool that you can use to create dramatic effects when brightness changes rapidly.
+
+<video controls><source src="auto_exposure.mp4" type="video/mp4"/></video>
```
SIGSTACKFAULT, line 5: add a link to the auto exposure example (which i assume this is a video of) https://github.com/bevyengine/bevy/blob/53910e07aefc40a9f9ca089f643d731b748c699f/examples/3d/auto_exposure.rs

SIGSTACKFAULT (changes requested):

release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
```diff
@@ -1,4 +1,13 @@
-<!-- Implement Auto Exposure plugin -->
-<!-- https://github.com/bevyengine/bevy/pull/12792 -->
+Since **Bevy 0.13**, you can configure the the EV100 of a camera, which allows you to adjust the exposure of the camera in a physically based way. This also allows you to dynamically change the exposure values for various effects. However, this is a manual process and requires you to adjust the exposure values yourself.
 
-<!-- TODO -->
+**Bevy 0.14** introduces **Auto Exposure**, which automatically adjusts the exposure of your scene based on the brightness of the scene. This can be useful when you want to create the feeling of a very high dynamic range, since your eyes also adjust to large changes in brightness. Note that this is not a replacement for hand-tuning the exposure values, rather an additional tool that you can use to create dramatic effects when brightness changes rapidly.
```
JMS55, line 3:
```suggestion
**Bevy 0.14** introduces **Auto Exposure**, which automatically adjusts the exposure of your camera based on the brightness of the scene. This can be useful when you want to create the feeling of a very high dynamic range, since your eyes also adjust to large changes in brightness. Note that this is not a replacement for hand-tuning the exposure values, rather an additional tool that you can use to create dramatic effects when brightness changes rapidly.
```

release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
```diff
@@ -1,4 +1,13 @@
...
+
+<video controls><source src="auto_exposure.mp4" type="video/mp4"/></video>
+
+Bevy's Auto Exposure is implemented by making a **histogram** of the scene's brightness in a post processing step. The exposure is then adjusted based on the average of the histogram. Because the histogram is calculated using a compute shader, Auto Exposure is **not available on WebGL**. It's also not enabled by default, so you need to add the [`AutoExposurePlugin`](https://docs.rs/bevy/0.14/bevy/core_pipeline/auto_exposure/struct.AutoExposurePlugin.html) to your app.
```
JMS55, line 7: As an aside, I'm curious why it's a non-default plugin. Usually, we tend to make all plugins part of PbrPlugin and enabled by default, and simply early-exit if there's no corresponding camera setting. The only exception is stuff like TAA which disables MSAA when you add the plugin.

Kurble: That's a really good question, I did this because it wouldn't work on the webgl backend, but it could of course be turned off by feature gates instead. Would it be too late to fix this for 0.14?

alice-i-cecile: Nope, submit a PR to change this.

JMS55: Doesn't need feature gating either. Just note in the docs that it's not WebGL2 compatible, and it's up to users to not use it on those platforms. Feature gating makes it harder to find, and there's already a ton of features.

Kurble: Alright, I don't have a lot of time at the moment, I will hopefully do this around the weekend

Kurble: Resolved when this PR gets merged: https://github.com/bevyengine/bevy/pull/13963

release-content/0.14/release-notes/12792_Implement_Auto_Exposure_plugin.md
Kurble: Added the link!

alice-i-cecile (approved):

alice-i-cecile:
If we get a video for this, so be it. But that can be done if a follow-up issue.

PR #1419, opened 2024-06-18
Fixed FAQ to mention auto credits updates

### LINKED ISSUE #1410: FAQ tells that the sponsors list is updated manually, but it isn't true
From the FAQ:

> ### I sponsored but the metrics / names / logos on the donate page haven't updated!
>
> This _isn't automated_ (yet!). For now ... bear with us. We try to update these as quickly as possible.

At the same time `bevy-donors` is scheduled using a `cron` job, and the site is deployed every hour with this data fetched.

YohDeadfall:
Fixes #1410.


alice-i-cecile (approved):

alice-i-cecile:
Thanks!

PR #1421, opened 2024-06-18
Added style for kbd element

### LINKED ISSUE #1390: Add CSS styling `<kbd>` tag
[`<kbd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) is the keyboard element in HTML, used to represent keys such as <kbd>Ctrl</kbd> and <kbd>Shift</kbd>. Right now there is no custom styling for it in our SCSS, so it just looks like monospace text.

image

In the above image, <kbd>Alt+F4</kbd> and <kbd>Command+Q</kbd> use the `<kbd>` element, but have no special styling. I think it would be good if, at the very least, `<kbd>` elements were treated as inline `code blocks`. They can receive custom styling above that, but this seems like the easiest route. :)

YohDeadfall:
Fixes #1390.

Preview:

image

### DIFF
--- a/sass/elements/_kbd.scss
+++ b/sass/elements/_kbd.scss
@@ -1,9 +1,6 @@
 kbd {
   font-size: 0.8rem;
-  padding-left: 0.1rem;
-  padding-right: 0.1rem;
-  padding-top: 0.1rem;
-  padding-bottom: 0.1rem;
+  padding: 0.1rem;
   line-height: 1;
   border-color: grey;
   border-radius: 0.3rem;
--- a/sass/elements/_kbd.scss
+++ b/sass/elements/_kbd.scss
@@ -1,0 +1,12 @@
+kbd {
+  font-size: 0.8rem;
+  padding-left: 0.1rem;
+  padding-right: 0.1rem;
+  padding-top: 0.1rem;
+  padding-bottom: 0.1rem;
+  line-height: 1;
+  border-color: grey;
+  border-radius: 0.3rem;
+  border-style: solid;
+  border-width: 1px;
+}
--- a/sass/site.scss
+++ b/sass/site.scss
@@ -17,6 +17,7 @@
 @import "elements/table";
 @import "elements/button";
 @import "elements/details";
+@import "elements/kbd";
 
 // CSS utility classes
 @import "utils";


sass/elements/_kbd.scss
```diff
@@ -0,0 +1,12 @@
+kbd {
+  font-size: 0.8rem;
+  padding-left: 0.1rem;
+  padding-right: 0.1rem;
+  padding-top: 0.1rem;
+  padding-bottom: 0.1rem;
+  line-height: 1;
+  border-color: grey;
+  border-radius: 0.3rem;
```
SIGSTACKFAULT, line 9: should use `$border-radius` (defined in _vars.scss)

SIGSTACKFAULT (approved):

sass/elements/_kbd.scss
YohDeadfall: Should it be done for migration tags too then?

SIGSTACKFAULT: actually on second thought that's way too much radius. maybe `calc($border-radius / 2)`

SIGSTACKFAULT: > Should it be done for migration tags too then?

oh, probably. that can be a separate MR.

sass/elements/_kbd.scss
```diff
@@ -0,0 +1,12 @@
+kbd {
+  font-size: 0.8rem;
+  padding-left: 0.1rem;
+  padding-right: 0.1rem;
+  padding-top: 0.1rem;
+  padding-bottom: 0.1rem;
```
BD103, line 6:
```suggestion
  padding: 0.1rem;
```

I believe these can be merged.

sass/elements/_kbd.scss
```diff
@@ -0,0 +1,12 @@
...
+  line-height: 1;
+  border-color: grey;
+  border-radius: 0.3rem;
+  border-style: solid;
+  border-width: 1px;
```
MrGVSV, line 11: Bikeshed: I kind of like how GitHub styles their `kbd` elements (seen in the issue). It makes it look kinda like a key or button. Would adding a larger bottom width look good?

SIGSTACKFAULT: <kbd>test</kbd> it actually uses `box-shadow: inset 0 -1px 0 var(--borderColor-neutral-muted, var(--color-neutral-muted));`

sass/elements/_kbd.scss
YohDeadfall: I checked all of that and it seems that `$border-radius` is used only for cards and menus, but not for anything else. In the tree it's even overridden to be `4px`. Therefore, I'm not sure that it should be calculated.

A possible solution for that issue is making additional variables to be used instead of primary and which might use `$border-radius`, but might not.

sass/elements/_kbd.scss
YohDeadfall: So, what's about it? As for me, the flat style looks more clear. Plus I don't remember any shadow on the site, even assets used there, I mean the rendered one, are flat too.

BD103 (approved):

PR #1436, opened 2024-06-19
Render layers

torsteingrindvik:
Fixes https://github.com/bevyengine/bevy-website/issues/1316

Hard to get creative on this one. I opted to show a snippet for context, then explain what's new.

Let me know if a note about the possible slight perf regression should be mentioned

### DIFF
--- a/release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
+++ b/release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
@@ -1,12 +1,11 @@
 <!-- #12502 Remove limit on RenderLayers. -->
 <!-- https://github.com/bevyengine/bevy/pull/13317 -->
 
-In Bevy you are able to assign entities spawned to different layers.
+Render layers are used to quickly toggle the visibility of sets of objects, and control which objects can be seen by which cameras.
+This can be useful for things like debug views, gear preview screens, toggleable diegetic UI and so on.
 
-You can make 3D models, meshes, lights, etc. all have their own if you wish.
 
-You can tell cameras which layers to render, and only objects which are part of those layers get rendered to that camera's target.
-
-There is no longer any practical limit to how many layers you can define!
-Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available. We've made sure to keep the common case fast, but now use a growable mask that will allocate space for additional layers as needed. Remember, there's still a cost to check visibility per layer, but this allows for more dynamic uses where layers can be created on demand without worrying about going over a limit.
+Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available.
+Now, There is no longer any practical limit to how many layers you can define, which is particularly helpful for creative coding applications like [nannou](https://nannou.cc/)! 
+We've made sure to keep the common case fast, but now use a growable mask that will allocate space for additional layers as needed. Remember, there's still a cost to check visibility per layer, but this allows for more dynamic uses where layers can be created on demand without worrying about going over a limit.
 
--- a/release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
+++ b/release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
@@ -23,7 +23,7 @@ commands.spawn((
 But this is old news, you've already been able to do this! What's new?
 
 There is no longer any practical limit to how many layers you can define!
-Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available.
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
@tychedelia, pinging you for review here :)

release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
```diff
@@ -1,4 +1,31 @@
...
+commands.spawn((
+    Camera3dBundle::default(),
+    RenderLayers::default().with(DEBUG_EXTRAS),
+));
+```
+
+But this is old news, you've already been able to do this! What's new?
+
+There is no longer any practical limit to how many layers you can define!
+Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available.
```
tychedelia, line 26:
```suggestion
Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available. We've made sure to keep the common case fast, but now use a growable mask that will allocate space for additional layers as needed. Remember, there's still a cost to check visibility per layer, but this allows for more dynamic uses where layers can be created on demand without worrying about going over a limit.
```

tychedelia (commented):
LGTM! As per my suggestion, I think if we wan to include an example (which tbh may not be necessary), we should show something like a mock `RenderLayerManager` resource that hands out new layers dynamically in a system, etc. This is more compelling than just adding a layer 100 (which you probably shouldn't do anyway, since it will be inefficient.

release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
```diff
@@ -1,4 +1,31 @@
...
+
+You can make 3D models, meshes, lights, etc. all have their own if you wish.
+
+Then you can tell cameras which layers to render, and only objects which are part of those layers get rendered to that camera's target.
+
+```rust
+// This camera shows everything by default- membership in layer 0 is implicit
+commands.spawn(Camera3dBundle::default());
+
+const DEBUG_EXTRAS: usize = 100;
```
tychedelia, line 14: This probably doesn't *actually* matter in practice, but layer 100 will require 2 bitmasks, which maybe isn't great to model for users.

torsteingrindvik: I'm going to update this PR later today when I get home- since for most users the change is simply _use as many render layers as you want_ I'm thinking of just removing code examples from this PR and just having a few sentences (including what you suggested).

Perhaps `RenderLayerManager` etc. could grow into a proper example in Bevy itself in the future, afaik we have no dedicated example for render layers.

tychedelia: Great! Yeah, I think a. example isn't really necessary here. :)

tychedelia (approved):

release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
```diff
@@ -1,4 +1,12 @@
 <!-- #12502 Remove limit on RenderLayers. -->
 <!-- https://github.com/bevyengine/bevy/pull/13317 -->
 
-<!-- TODO -->
+In Bevy you are able to assign entities spawned to different layers.
+
+You can make 3D models, meshes, lights, etc. all have their own if you wish.
+
+You can tell cameras which layers to render, and only objects which are part of those layers get rendered to that camera's target.
```
alice-i-cecile, line 8:
```suggestion
Render layers are used to quickly toggle the visibility of sets of objects, and control which objects can be seen by which cameras.
This can be useful for things like debug views, gear preview screens, toggleable diegetic UI and so on.
```

release-content/0.14/release-notes/13317_12502_Remove_limit_on_RenderLayers.md
```diff
@@ -1,4 +1,12 @@
...
+
+There is no longer any practical limit to how many layers you can define!
+Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available. We've made sure to keep the common case fast, but now use a growable mask that will allocate space for additional layers as needed. Remember, there's still a cost to check visibility per layer, but this allows for more dynamic uses where layers can be created on demand without worrying about going over a limit.
```
alice-i-cecile, line 11:
```suggestion

Before Bevy 0.14 the membership was defined by a bitmask which had limited slots available.
Now, There is no longer any practical limit to how many layers you can define, which is particularly helpful for creative coding applications like [nannou](https://nannou.cc/)! 
We've made sure to keep the common case fast, but now use a growable mask that will allocate space for additional layers as needed. Remember, there's still a cost to check visibility per layer, but this allows for more dynamic uses where layers can be created on demand without worrying about going over a limit.
```

alice-i-cecile (changes requested):
Some improvements to flow for you, with a bit more context on "what are render layers again" :)

alice-i-cecile (approved):

PR #1443, opened 2024-06-20
0.14 migration: Fix an error in colors table and add more colors

rparrett:
I included a mistake in this table when I added it -- `GREEN`'s new equivalent is `LIME`, not `LIME_GREEN`. `LIME_GREEN` isn't even a thing anymore, it's `LIMEGREEN`. 😅

Also alphabetized the list and added a few more examples. This may still not be totally comprehensive, but I did a quick side-by-side scan of the color constants before/after.


rparrett:
CI error unrelated

alice-i-cecile (approved):

PR #1411, opened 2024-06-16
Add 3D rendering release notes

alice-i-cecile:
This PR aims to tackle all of the assorted outstanding 3D rendering patch notes all at once. These are almost entirely features by @pcwalton, but there's one from @aevyrie as well :)

Like always, I'm still relatively novice at rendering, so corrections are extremely welcome.

## To Do

I've opened this PR very early in order to more easily track work that needs to be done.

- [x] grab PR descriptions for the required PRs to work off of
- [x] finish drafting the initial notes
- [x] edit titles of each section
- [x] add images
- [x] make sure that every post links to an example if possible
- [x] self-editing pass

## Linked PRs

- Fixes #1298
- Fixes #1305
- Fixes #1314
- Fixes #1318
- Fixes #1346
- Fixes #1360
- Fixes #1361
- Fixes #1362

PR #1448, opened 2024-06-21
Fix compare slider

### LINKED ISSUE #1424: Tera Templates don't seem to work in new release notes
image

_Originally posted by @BD103 in https://github.com/bevyengine/bevy-website/pull/1411#discussion_r1644646131_
        
 This is a pretty serious problem: it needs to have a resolution in some form before we can ship the notes.

kristoff3r:
Fixes #1424

This is based on #1411, but I set that as base for some reason, so it will need to be cherry-picked or rebased


content/news/draft-bevy-0.14/index.md
```diff
@@ -22,7 +22,7 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 <!-- more -->
 
-{{ release_notes(version="0.14") }}
+{{ release_notes(version="0.14", post_path="news/draft-bevy-0.14/") }}
```
doup, line 25: Second though, I think this can be removed:

```suggestion
{{ release_notes(version="0.14") }}
```

And in `release_notes.md` use `page.colocated_path`:

```
replace(from='POST_PATH', to=page.colocated_path)
```

doup (commented):
Idea, can you try just doing `replace(from='POST_PATH', to=page.colocated_path)` instead of passing `post_path` to `release_notes()`? I think that should work 🤞 

Also, the PR target should be #1411?

kristoff3r:
> Idea, can you try just doing `replace(from='POST_PATH', to=page.colocated_path)` instead of passing `post_path` to `release_notes()`? I think that should work 🤞
> 
> Also, the PR target should be #1411?

It only allowed me to target `master` and `gh-pages`. Maybe it's a permission problem?

kristoff3r:
Ah, it's probably because the branch is on alice's fork, so I'd have to make this PR there instead. Maybe we should just merge it first and then I can rebase?

doup (approved):

alice-i-cecile (approved):
Oh lovely, that was way less tricky than I was worried about. @cart see https://github.com/bevyengine/bevy-website/pull/1448/commits/6b6635addeff006d815be219391b957b229791c9

I'm going to hold off merging this PR until my PR is merged, which is waiting on new images, which @IceSentry has offered to generate for us this weekend :)

cart:
Ahhh so shortcodes do still work and this was an issue with the format / evaluation of them.

kristoff3r:
This is rebased now, I tested it locally and the sliders work. I also grepped for `compare_slider` in the `release-content` directory and made sure all of them use the path variable.

alice-i-cecile (approved):

Issue #1200, opened 2024-05-26
_release_notes.toml formatting does not have standard formatting


alice-i-cecile:
This creates nasty diffs during format-on-save.

![image](https://github.com/bevyengine/bevy-website/assets/3579909/17521d14-6824-4904-9a9a-ccee53f0913e)

Missing spaces.

![image](https://github.com/bevyengine/bevy-website/assets/3579909/cc6450bb-400a-4924-ab5d-3c40b0b928ed)

Line breaks.

IceSentry:
I think the easiest thing to do would be to make sure to save the file with a formatter before opening the first PR. We can add the spaces for the list, but the newlines on longer list I don't really know what the heuristic would be for that.

IceSentry:
oh, but that file is regenerated every time, right....

rparrett:
Does the `toml` crate's pretty serialization match the expected format? 

I'll try to remember to look into that at some point.

PR #1452, opened 2024-06-21
Add spaces to match formatting by taplo in guide and notes generators

rparrett:
This will make generated toml by these tools more closely match the formatting that is produced by `taplo` (as recommended in `CONTRIBUTING.md`). This should result in less of a mess when someone's editor auto-formats them.

There are some places where this still won't match because there are many entries in the array and they won't fit on one line. We should probably be using the `toml` crate to generate this instead of building it manually.

```diff
[[guides]]
title = "Decouple `BackgroundColor` from `UiImage`"
url = "https://github.com/bevyengine/bevy/pull/11165"
- areas = ["Rendering","UI"]
+ areas = ["Rendering", "UI"]
file_name = "11165_Decouple_BackgroundColor_from_UiImage.md"
```

```diff
[[release_notes]]
title = "Upstreaming bevy_color."
- authors = ["@viridia","@mockersf"]
+ authors = ["@viridia", "@mockersf"]
url = "https://github.com/bevyengine/bevy/pull/12013"
file_name = "12013_Upstreaming_bevy_color.md"
```

### DIFF
--- a/generate-release/src/migration_guides.rs
+++ b/generate-release/src/migration_guides.rs
@@ -130,7 +130,7 @@ file_name = "{file_name}.md"
             .iter()
             .map(|area| format!("\"{area}\""))
             .collect::<Vec<_>>()
-            .join(","),
+            .join(", "),
         title = title.trim().replace('"', "\\\"")
     )
 }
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
Progress towards #1200.

alice-i-cecile (approved):

PR #1453, opened 2024-06-22
0.14: Improve titles and begin section reordering

cart:
For my first pass over the post, I've:
1. Renamed many of the titles to be clearer and / or more concise. The general idea is to reframe changes as Concept Names instead of sentences. Note that these use "Concept Name" style instead of "Concept name", both to match previous blog posts and add emphasis. It would be worthwhile to update our guidance to encourage people to use this titling convention.
2. Started reordering based on "interest level" (this is just a first pass / not the final ordering). This is an art, not a science, but the general rule of thumb is to put big new features, highly visual, and/or high cool factor features higher. Ordering by area is a secondary concern. Worth doing when possible, but it should not override "interest level".


rparrett (approved):

jgayfer (approved):
Makes sense.

I like the new titles. More punchy this way.

BD103 (approved):
Looks good! :)

release-content/0.14/release-notes/_release-notes.toml
```diff
@@ -1,29 +1,113 @@
 [[release_notes]]
-title = "Add support for KHR_texture_transform"
+title = "Meshlets (Experimental)"
```
JMS55, line 2: I'd like to call this virtual geometry. I'm probably going to rename the feature eventually, meshlets was a placeholder name to avoid getting stuck on naming things :P

alice-i-cecile (approved):

release-content/0.14/release-notes/_release-notes.toml
alice-i-cecile:
```suggestion
title = "Virtual Geometry (Experimental)"
```

PR #1467, opened 2024-06-27
Write pipeline diagnostics release notes

alice-i-cecile:
Fixes https://github.com/bevyengine/bevy-website/issues/1264.

This is deeper into the rendering stack than I usually delve, and I struggled quite a bit to piece things together. While the docs are clear, they don't cover usage very effectively, and the set of tracked-by-default diagnostics is very hard to uncover.

That said, I'm pretty happy with this: it should provide the bread crumbs needed for experts to start poking at and improving this feature.

### DIFF
--- a/release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
+++ b/release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
@@ -17,4 +17,5 @@ By default, this feature will track:
 - [Clipper invocations](http://gpa.helpmax.net/en/intel-graphics-performance-analyzers-help/metrics-descriptions/extended-metrics-description/rasterizer-metrics/clipper-invocations/)
 - [Clipper primitives](http://gpa.helpmax.net/en/intel-graphics-performance-analyzers-help/metrics-descriptions/extended-metrics-description/rasterizer-metrics/post-clip-primitives/)
 
-but as outlined in the documentation, you can use this feature to track additional information that you might care about.
+ You can also track individual render/compute passes, groups of passes (e.g. all shadow passes), and individual commands inside passes (like draw calls).
+ To do so, instrument them using methods from the [`RecordDiagnostics`](https://dev-docs.bevyengine.org/bevy/render/diagnostic/trait.RecordDiagnostics.html) trait.
--- a/release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
+++ b/release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
@@ -1,13 +1,19 @@
-When working on rendering features (in the engine and in projects), it's useful to know how expensive they are. 
-[Tracy](https://github.com/bevyengine/bevy/blob/main/docs/profiling.md) already lets us measure CPU time per system, but we didn't have a way of measuring how long the GPU spends on each render node.
+While [Tracy](https://github.com/bevyengine/bevy/blob/main/docs/profiling.md) already lets us measure CPU time per system, our GPU diagnostics are much weaker.
+In Bevy 0.14 we've added support for two classes of rendering-focused statistics via the [`RenderDiagnosticsPlugin`](https://dev-docs.bevyengine.org/bevy/render/diagnostic/struct.RenderDiagnosticsPlugin.html):
 
-The [`RenderDiagnosticsPlugin`](https://dev-docs.bevyengine.org/bevy/render/diagnostic/struct.RenderDiagnosticsPlugin.html) provides this information,
-collecting both information about the CPU/GPU spent per node and [pipeline](https://sotrh.github.io/learn-wgpu/beginner/tutorial3-pipeline/) statistics such as the number of triangles drawn.
-This data is stored in Bevy's central [`DiagnosticsStore`](https://dev-docs.bevyengine.org/bevy/diagnostic/struct.DiagnosticsStore.html),
-and can be accessed by calling `DiagnosticsStore::iter()` to get a list of all registered diagnostics.
-The [`DiagnosticPath`](https://dev-docs.bevyengine.org/bevy/diagnostic/struct.DiagnosticPath.html) contained in the [`Diagnostic`](https://dev-docs.bevyengine.org/bevy/diagnostic/struct.Diagnostic.html)s returned can also be used to fetch specific diagnostics for more focused investigation.
+1. **Timestamp queries:** how long did specific bits of work take on the GPU?
+2. **Pipeline statistics:** information about the quantity of work sent to the GPU.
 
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
@LeshaInc I really appreciate your work on this feature, and would love your feedback on what I got wrong :)

release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
```diff
@@ -1 +1,20 @@
...
+
+- Elapsed CPU time
+- Elapsed GPU time
+- [Vertex shader](https://www.khronos.org/opengl/wiki/Vertex_Shader) invocations
+- [Fragment shader](https://www.khronos.org/opengl/wiki/Fragment_Shader) invocations
+- [Compute shader](https://www.khronos.org/opengl/wiki/Compute_Shader) invocations
+- [Clipper invocations](http://gpa.helpmax.net/en/intel-graphics-performance-analyzers-help/metrics-descriptions/extended-metrics-description/rasterizer-metrics/clipper-invocations/)
+- [Clipper primitives](http://gpa.helpmax.net/en/intel-graphics-performance-analyzers-help/metrics-descriptions/extended-metrics-description/rasterizer-metrics/post-clip-primitives/)
+
+but as outlined in the documentation, you can use this feature to track additional information that you might care about.
```
LeshaInc, line 20: This feature actually doesn't allow you to record any additional information (not in the list).

You can still use DiagnosticsStore to add custom measurements, but that was always possible.

release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
```diff
@@ -1 +1,20 @@
...
+[Tracy](https://github.com/bevyengine/bevy/blob/main/docs/profiling.md) already lets us measure CPU time per system, but we didn't have a way of measuring how long the GPU spends on each render node.
+
+The [`RenderDiagnosticsPlugin`](https://dev-docs.bevyengine.org/bevy/render/diagnostic/struct.RenderDiagnosticsPlugin.html) provides this information,
+collecting both information about the CPU/GPU spent per node and [pipeline](https://sotrh.github.io/learn-wgpu/beginner/tutorial3-pipeline/) statistics such as the number of triangles drawn.
+This data is stored in Bevy's central [`DiagnosticsStore`](https://dev-docs.bevyengine.org/bevy/diagnostic/struct.DiagnosticsStore.html),
+and can be accessed by calling `DiagnosticsStore::iter()` to get a list of all registered diagnostics.
+The [`DiagnosticPath`](https://dev-docs.bevyengine.org/bevy/diagnostic/struct.DiagnosticPath.html) contained in the [`Diagnostic`](https://dev-docs.bevyengine.org/bevy/diagnostic/struct.Diagnostic.html)s returned can also be used to fetch specific diagnostics for more focused investigation.
+
+By default, this feature will track:
```
LeshaInc, line 10: I would also add that we can track individual render/compute passes, groups of passes (e.g. all shadow passes), and individual commands inside passes (like draw calls). For this to work, they have to be instrumented with calls to methods from `RecordDiagnostics` trait.

To summarize, render diagnostics aren't recorded automatically, nodes must explicitly record them. As far as I'm aware, not all nodes currently have this instrumentation.

LeshaInc (commented):
Overall looks good to me. I'd like to add that diagnostics aren't recorded automatically and require manual instrumentation (similar to tracing::span, but for gpu).

release-content/0.14/release-notes/9135_Add_pipeline_statistics.md
JMS55: Yeah, this has no integration with nodes atm. You have to specify the rendering work you want to record.

I also don't think the details about the types are useful. It would be better to show an example of getting the diagnostic recorder, instrumenting a span of rendering commands, and then showing how the LogDiagnosticsPlugin can print the results to the terminal.

JMS55 (changes requested):
So, many things:
1. This work covers both pipeline statistics (vertices clipped, fragments shaded, etc etc), and timestamp queries, which are the "how long did <X> take on the GPU"
2. Timestamp queries tend not to be stable across frames. GPUs will dynamically ramp up and down clock speed due to workload (idle gaps in GPU work, e.g., a bunch of consecutive barriers, or the tail end of a large dispatch) or heat (thermals). You probably want to look at mean/median/min/max, not just raw per-frame numbers.
3. Bevy has a few built-in timestamp spans as of 0.14, but we need to go through the engine and instrument everything properly. Users shouldn't expect comprehensive statistics atm.
4. Timestamp queries will tell you speed, but it will not tell you _why_ things are slow. For finding bottlenecks, you want to use a GPU profiler from your GPU vendor (Nvidia - NSight, AMD - RGP, Intel - GPA, Apple - XCode). They will give you much more detailed stats about cache hit rate, warp occupancy, etc. On the other hand they lock your GPU's clock to base speeds for stable results, so they won't give you a good indicator of real world performance.
5. We should talk about how GPUs work. On the CPU, you need to record CommandBuffers containing work you want the GPU to do. This is what the render nodes are doing. Then, you submit them on the RenderQueue, and they'll be executed asynchronously by the GPU at some point in the future. Both parts can/are expensive. Tracy/perf will give you insight into which part of command recording is expensive, while GPU timestamps/profiling tools will give you insight into the execution of those command buffers by the GPU. I'd like to emphasize that (excluding iGPUs) GPUs are _separate_ physical devices with their own processors, and any API calls you make from Bevy on the CPU are not actually performing any GPU work. When you say render_pass.draw(), the GPU is not doing anything at that point. It's just command encoding.

alice-i-cecile:
> We should talk about how GPUs work. On the CPU, you need to record CommandBuffers containing work you want the GPU to do. This is what the render nodes are doing. Then, you submit them on the RenderQueue, and they'll be executed asynchronously by the GPU at some point in the future. Both parts can/are expensive. Tracy/perf will give you insight into which part of command recording is expensive, while GPU timestamps/profiling tools will give you insight into the execution of those command buffers by the GPU. I'd like to emphasize that (excluding iGPUs) GPUs are separate physical devices with their own processors, and any API calls you make from Bevy on the CPU are not actually performing any GPU work. When you say render_pass.draw(), the GPU is not doing anything at that point. It's just command encoding.

I think that this is too in-depth for release notes, even though it is very useful information.

>  It would be better to show an example of getting the diagnostic recorder, instrumenting a span of rendering commands, and then showing how the LogDiagnosticsPlugin can print the results to the terminal.

I would love to have this but IMO this should be a link to an example, and I neither have the time nor the ability to write this right now.

LeshaInc (approved):

PR #1470, opened 2024-06-28
Skip empty migration guides

IceSentry:
Sometimes migration guides section are empty.

Skip them and log it so the person that ran the tool is aware of the potential false positive.


BD103 (approved):
Thank you! I don't think I'll be able to use this this release cycle, but it will be helpful to whoever does it for 0.15!

Issue #1352, opened 2024-06-04
Condense release notes on gizmos


alice-i-cecile:
There'll be a few PRs here: we may want to merge their sections together during editing.

alice-i-cecile:
Annulus needs to be added, and the rounded box gizmo needs to be condensed.

Issue #1188, opened 2024-05-22
Tracking issue: release Bevy 0.14


alice-i-cecile:
This is a tracking issue for our progress (and also useful for the public-drafts feature, see #1186).

You can preview the release notes locally at http://127.0.0.1:1111/news/draft-bevy-0-14

Before the milestone is empty:

- [x] Generate initial migration guides
- [x] Generate initial release notes
- [x] Complete a first draft for all sections of the release notes
- [x] Initial editing pass on release notes
- [x] Initial editing pass on migration guide
- [x] Complete or punt on all issues and PRs in https://github.com/bevyengine/bevy/milestone/20

After the milestone is ready:
- [x] Regenerate and revise release notes to check for missing sections
- [x] Regenerate and revise migration guide to check for missing sections
- [x] Prepare release candidate on its own branch
- [x] Ship release candidate
- [x] Announce release candidate in dev-announcements
- [x] Fix any discovered bugs, backporting to the release-candidate branch

After the release candidate has concluded:

- [x] Regenerate and revise release notes to check for missing sections
- [x] Regenerate and revise migration guide to check for missing sections
- [x] Generate changelog
- [x] Generate contributors list
- [x] Final editing pass on release notes
- [x] Final editing pass on migration guide
- [x] Select and add a splash image
- [x] Undraft the migration guide
- [x] Undraft the release notes
- [x] Bump `bevy_color` to 0.14.1 and other crates to 0.14.0 using: https://github.com/bevyengine/bevy/pull/14126
- [x] Release Bevy 0.14 on crates.io following https://github.com/bevyengine/bevy/blob/main/docs/release_checklist.md
- [x] Announce Bevy 0.14

alice-i-cecile:
Editing tasks:

- [ ] update all doc links to point to 0.14 on crates.io
- [ ] update all example links to point to the 0.14 tag on Github
- [ ] move all images into the correct news folder
- [ ] remove all dead sections
- [ ] edit titles
- [ ] double-check ordering of sections
- [ ] copyedit each section

PR #1485, opened 2024-07-01
Remove dead sections from 0.14 release notes

alice-i-cecile:
Part of #1188. All dead sections are removed, except for the gizmo ones, which I want to tackle separately as part of #1352.

I've discussed each of the removed PRs with the respective authors / contributors already: I don't feel bad about cutting any of these, even ignoring the desire to get these notes out the door.


alice-i-cecile:
Per #1326, @cart thinks the TAA work should stay so I've reverted that removal.

BD103 (approved):

PR #1506, opened 2024-07-03
Third 0.14 editorial pass

cart:


alice-i-cecile (approved):
Good, I like all of these changes. Splitting out the 3D rendering features is probably wise, even if there are a *ton* of them.

PR #1508, opened 2024-07-03
0.14: cover image

cart:


alice-i-cecile (approved):
I'm content with this. I think it's nice to show off the new features, and I like that there's a repo link. Thanks @IceSentry!

BD103 (approved):

BD103:
I think I like both the first and second forest shot. I'd prefer to not do foggy square, since it looks dull and less vibrant.

alice-i-cecile (changes requested):
image


This image by @Jondolf is substantially better: nicer exposure on the sky, better framing, less blinding and shows off the water shader. We should swap to it.

cart:
Yeah I think that one is much better. Currently in the wrong aspect ratio though. I'll see if cropping is reasonable.

PR #1513, opened 2024-07-03
Add images for SMAA

### LINKED ISSUE #1500: SMAA image is missing from release notes
Spotted by @cart.

alice-i-cecile:
Fixes #1500. I decided to compare this to no anti-aliasing, as it most clearly demonstrates what it's actually doing.

No AA:

image

MSAA:
image

FAA:
image

SMAA: 
image


alice-i-cecile:
I followed #1507 and used lossless jpgs, which were indeed meaningfully smaller than even maximally compressed PNGs.

cart (approved):

PR #1519, opened 2024-07-04
Remove hard-coded width and height from news page image

cart:
The hard coded image size (specifically the height) prevents it from following the aspect ratio of the image, resulting in it stretching when there isn't enough space. These were added in #1026 ... I don't think this is a good rule to enforce.


cart:
(merging now in the interest of this not holding up tomorrow morning's release)

PR #1547, opened 2024-07-09
Automatically run `oxipng` on example screenshots

### LINKED ISSUE #1534: Automatically run pngcrush on example screenshots
> It would be good to run `pngcrush` or another image optimization tool on these, before it gets merged.

I crushed it manually for this time, we could update the action to do it automatically for next.

_Originally posted by @mockersf in https://github.com/bevyengine/bevy-website/pull/1530#issuecomment-2209547149_

Chubercik:
- Fixes #1534.
- This PR adds an image compression step to the [`Update Screenshots`](https://github.com/bevyengine/bevy-website/actions/workflows/update-screenshots.yml) GitHub Action.
- The whole compression overhead (installing `oxipng` and running it with the maximum compression level) adds about a minute to the workflow runtime, which is minuscule when compared to the avg. ~3h completion time.

### DIFF
--- a/.github/workflows/update-screenshots.yml
+++ b/.github/workflows/update-screenshots.yml
@@ -58,7 +58,7 @@ jobs:
 
       - name: Install oxipng
         run: |
-          which oxipng || cargo install oxipng
+          cargo install oxipng
 
       - name: Take Screenshots
         run: |
@@ -66,7 +66,7 @@ jobs:
 
       - name: Optimize PNGs
         run: |
-          oxipng -o 4 -s -r screenshots
+          oxipng -o 6 -s -r screenshots
 
       - name: Upload Generated Files
         uses: actions/upload-artifact@v4
--- a/.github/workflows/update-screenshots.yml
+++ b/.github/workflows/update-screenshots.yml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Optimize PNGs
         run: |
-          oxipng -o 4 -s screenshots/*
+          oxipng -o 4 -s -r screenshots
 
       - name: Upload Generated Files
         uses: actions/upload-artifact@v4
--- a/.github/workflows/update-screenshots.yml
+++ b/.github/workflows/update-screenshots.yml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Optimize PNGs
         run: |
-          oxipng -o 4 -s screenshots
+          oxipng -o 4 -s screenshots/*.png
 
       - name: Upload Generated Files
         uses: actions/upload-artifact@v4
--- a/.github/workflows/update-screenshots.yml
+++ b/.github/workflows/update-screenshots.yml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Optimize PNGs
         run: |
-          oxipng -o 4 -s screenshots/
+          oxipng -o 4 -s screenshots/*
 
       - name: Upload Generated Files
         uses: actions/upload-artifact@v4
--- a/.github/workflows/update-screenshots.yml
+++ b/.github/workflows/update-screenshots.yml
@@ -58,7 +58,7 @@ jobs:
 
       - name: Install oxipng
         run: |
-          cargo install oxipng
+          cargo install oxipng --version ^9.1.1
 
       - name: Take Screenshots
         run: |
@@ -66,7 +66,7 @@ jobs:
 
       - name: Optimize PNGs
         run: |
-          oxipng -o 6 -s -r screenshots
+          oxipng --opt max --strip safe --recursive screenshots
 
       - name: Upload Generated Files
         uses: actions/upload-artifact@v4
--- a/.github/workflows/update-screenshots.yml
+++ b/.github/workflows/update-screenshots.yml
@@ -56,9 +56,12 @@ jobs:
           sudo apt-get update
           sudo apt install -y xvfb libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
 
+      - name: Install cargo-binstall
+        uses: cargo-bins/cargo-binstall@main
+
       - name: Install oxipng
         run: |
-          cargo install oxipng
+          cargo binstall oxipng
 
       - name: Take Screenshots
         run: |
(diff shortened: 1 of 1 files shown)


Chubercik:
By comparing the artifact sizes of [the last official GA workflow run](https://github.com/bevyengine/bevy-website/actions/runs/9797634883) and [the one run on this PR](https://github.com/Chubercik/bevy-website/actions/runs/9860601911), we can see noticeable reductions:

|     Artifact    |  Current | This PR | Size reduction |
| --------------- | -------- | ------- | -------------- |
|  screenshots-0  |   4.5 MB | 3.47 MB |     -22.9%     |
|  screenshots-1  |  7.05 MB |    -    |        -       |
|  screenshots-2  |  976 KB  |  565 KB |     -42.1%     |
|  screenshots-3  |  3.34 MB | 1.98 MB |     -40.7%     |
|  screenshots-4  |  1.28 MB |  749 KB |     -41.5%     |
|  screenshots-5  |   3 MB   | 1.93 MB |     -35.7%     |
|  screenshots-6  |  3.1 MB  | 1.99 MB |     -35.8%     |
|  screenshots-7  |  3.28 MB | 2.26 MB |     -31.1%     |
|  screenshots-8  |  3.63 MB | 2.57 MB |     -29.2%     |
|  screenshots-9  |  4.61 MB | 2.66 MB |     -42.3%     |
|  screenshots-10 |  1.77 MB | 1.21 MB |     -32.7%     |
|  screenshots-11 |  3.87 MB | 3.06 MB |     -20.9%     |
|  screenshots-12 |  1.95 MB | 1.53 MB |     -21.5%     |
| all screenshots | 42.36 MB |    -    |        -       |

It's worth noting that the sizes from the current workflow aren't the same as image sizes that would be used on the website, given the fact that @mockersf used `pngcrush` on them manually before merging https://github.com/bevyengine/bevy-website/pull/1530.

TrialDragon (approved):
LGTM.

.github/workflows/update-screenshots.yml
```diff
@@ -49,17 +49,25 @@ jobs:
...
+      - name: Install xvfb, llvmpipe and lavapipe
         run: |
           sudo apt-get update -y -qq
           sudo add-apt-repository ppa:kisak/turtle -y
           sudo apt-get update
           sudo apt install -y xvfb libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
 
+      - name: Install oxipng
+        run: |
+          cargo install oxipng
```
janhohenheim, line 61: I'd use https://github.com/marketplace/actions/install-cargo-binstall and then run `cargo binstall` to not have to compile `oxipng` by hand

janhohenheim (approved):
LGTM, just left a note for optimizing speed

Chubercik:
With @janhohenheim's suggestion applied, `oxipng`'s install time has been brought down to ~1 second, effectively cutting the overhead to around half its previous length ^^

.github/workflows/update-screenshots.yml
mockersf: not sure it's worth it, when the "taking the screenshots" part is already taking a few hours

.github/workflows/update-screenshots.yml
```diff
@@ -49,17 +49,28 @@ jobs:
...
-      - name: install xvfb, llvmpipe and lavapipe
+      - name: Install xvfb, llvmpipe and lavapipe
         run: |
           sudo apt-get update -y -qq
           sudo add-apt-repository ppa:kisak/turtle -y
           sudo apt-get update
           sudo apt install -y xvfb libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
 
+      - name: Install cargo-binstall
+        uses: cargo-bins/cargo-binstall@main
```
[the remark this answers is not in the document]
Chubercik: sure

.github/workflows/update-screenshots.yml
Chubercik: It comes at no cost to compression quality, so I don't see why not.

mockersf: It's additional complexity, maintenance cost and update management, for no benefit in compression quality

Chubercik: Fair enough, I guess more people could chime in on the issue.

mockersf: after checking with other Bevy repos and conversation, we decided generally against using `cargo-binstall`. I think it should be kept the same and not changed just in a single place

mockersf (changes requested):
Could you remove `cargo-install` usage?

.github/workflows/update-screenshots.yml
```diff
@@ -49,17 +49,25 @@ jobs:
...
+        run: |
+          cargo install oxipng
+
       - name: Take Screenshots
         run: |
           xvfb-run cargo run -p example-showcase -- --page ${{ matrix.page }} --per-page ${{ env.PER_PAGE }} run --screenshot --in-ci
 
+      - name: Optimize PNGs
+        run: |
+          oxipng -o 6 -s -r screenshots
```
BD103, line 69:
```suggestion
          oxipng --opt max --strip safe --recursive screenshots
```

I would prefer to use the long form of the options. It's more clear for people unfamiliar reading over the workflow, and it increases the likelihood of typos being caught. (`-o max` is the same as `-o 6`, but it is more clear on what behavior we want from the optimizer.)

.github/workflows/update-screenshots.yml
BD103, line 61: Could you please pin `oxipng` to be version 9? This prevents breaking changes from an eventual version 10.0.0 from hurting us later. I believe it would look like this:

```suggestion
        run: |
          cargo install oxipng --version ^9.1.1
```

I believe that will allow versions >= 9.1.1 and < 10.0.0, but please test it first!

BD103 (approved):
Looks good, thanks for doing this!

I have two notes for how to improve the workflow, but they're nits don't prevent this from being merged. :)

BD103:
@alice-i-cecile would you mind merging this, when you have the chance? I think it was missed in today's merge train. (Which is understandable, given how many PRs you had to merge today!)

PR #1550, opened 2024-07-12
Fix migration guide related to `App`

### LINKED ISSUE #1540: Migration guide corrections
The signature of the function you need to pass to `App::set_runner` has changed and this isn't mentioned at all in the migration guide.

Also, the section "Separate states from core ECS" states
> Projects that use `bevy_app` directly and use states will need to add the `bevy_state` feature.

but `bevy_app` has no such feature. I think maybe you meant "that use `bevy` directly?" (It has a feature with that name)

BD103:
Fixes #1540. This now mentions how `App::set_runner()` requires the function to return `AppExit`, and how `App`'s properties are now private.


TrialDragon (approved):
Outside of one inconsequential nit pick, LGTM.

alice-i-cecile (approved):

PR #1553, opened 2024-07-13
Update optimization suggestions based on benchmark

benfrankel:
Data thanks to @rparrett:

https://github.com/rparrett/bevy_wasm_bench/blob/main/results/d98b3a8_win11_i713700KF_4080super.csv

### DIFF
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -158,7 +158,7 @@ lto = "thin"
 # Default to release profile values.
 inherits = "release"
 # Optimize with size in mind (also try "z", sometimes it is better).
-# Slightly slows compile time, great improvements to file size and runtime performance.
+# Slightly slows compile times, great improvements to file size and runtime performance.
 opt-level = "s"
 # Strip all debugging information from the binary to slightly reduce file size.
 strip = "debuginfo"
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -172,7 +172,7 @@ When releasing for web, you can pass `--profile wasm-release` to `cargo` instead
 that provides a `wasm-opt` CLI tool for making `.wasm` files smaller and faster:
 
 ```sh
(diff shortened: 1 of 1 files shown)


benfrankel:
Some additional thoughts:

> Interestingly `codegen-units=1` doesn't seem to do much besides increase compile times if you already enabled `lto`
> Maybe it shouldn't be recommended at all
> Or that might be overfitting to the benchmark

A quick search actually found discussions suggesting the _opposite_ behavior, so I left `codegen-units=1` in.

benfrankel:
As for `opt-level`, `z` significantly reduces file size compared to `s`, but it also significantly reduces runtime performance. `s` should be a more balanced option to suggest by default.

tbillington:
With LTO & wasm-opt the size is already pretty good. The savings of s/z over 2/3 don't tend to be that meaningful. In a context where the user is already downloading a 15mb wasm blob, potentially having a few more mb is worth it for better performance, imho. We're already running in a resource constrained environment, so hamstringing optimisations is not desirable.

Just my 2c :)

rparrett:
Just making a note that I am not sure how much my data generalizes to
- other targets
- other hardware
- other browsers
- other workloads
- etc

And there wasn't a whole lot of scientific rigor involved in testing to begin with.

I gathered the data because I couldn't find any real information about how various file-size-focused optimizations affected runtime performance in the browser.

My takeaway for this particular setup was
- `opt-level = "z"` costs too much performance, I'd rather accept slightly larger files.
- `lto = "thin"` seems to give more than a "marginal" boost: ~10%. But the compile time cost is significant.

But there are various tradeoffs so people should just know that they might want to fiddle with this stuff and do their own measurements.

benfrankel:
> With LTO & wasm-opt the size is already pretty good. The savings of s/z over 2/3 don't tend to be that meaningful.

While I agree in concept, it should be noted that `-Os` isn't really losing runtime performance compared to `-O3`; in this benchmark at least.

mockersf (approved):

alice-i-cecile (approved):

PR #1561, opened 2024-07-16
Sort assets that use the latest Bevy version first

### LINKED ISSUE #760: Improved Sorting for Assets based on Bevy Version
Currently it's a bit difficult to see what bevy assets are the most relevant.
Multiple options come to mind:
* stars
* recent downloads
* forks
* most recent commit
* etc...

However as mentioned in discord by bestRanar:
> I think sorting by (bevy_version_is_semver_compatible_with_latest_release, random) would be a great step up from the current random. 
> - https://discord.com/channels/691052431525675048/695741366520512563/1159219167388500029

I think this would be a good uncontroversial improvement.

### Details
Crates that don't provide version information would be sorted to the bottom since I can't think of a much better mechanism.

### Implementation:
I already looked into how difficult it would be and it doesn't seem particularly difficult.
The strangest thing I found so far is that versions are currently stored as a list:
https://github.com/bevyengine/bevy-website/blob/main/generate-assets/src/lib.rs#L26
https://github.com/bevyengine/bevy-website/blob/main/generate-assets/src/lib.rs#L296

To make things simpler I would probably change things to just a single version for now since I don't know what the author (@Selene-Amanita) intended.

rparrett:
# Motivation

Help users find up-to-date assets quicker

Fixes #760
Fixes https://github.com/bevyengine/bevy-assets/issues/297

# Solution

Sorts assets by the following key:
```rust
(
    // Any existing order defined in the bevy-assets metadata.
    // e.g. for "official" templates and learning material
    existing_order,
    // Uses the semver crate to compare the "version requirement" from the asset with
    // the latest Bevy version
    is_semver_compatible_with_latest_bevy,
    // If all else is equal, randomize
    random
)
````

### DIFF
--- a/generate-assets/src/bin/generate.rs
+++ b/generate-assets/src/bin/generate.rs
@@ -74,21 +74,22 @@ fn sort_section(nodes: &mut [AssetNode], latest_bevy_version: &semver::Version) 
         let is_semver_compat = node_semver_compat_with(node, latest_bevy_version);
         let random: u32 = rand::random();
 
-        to_sort.push((node, is_semver_compat, random));
+        let existing_order = match node {
+            AssetNode::Asset(asset) => asset.order.unwrap_or(usize::MAX),
+            _ => continue,
+        };
+
+        to_sort.push((node, existing_order, !is_semver_compat, random));
     }
 
-    to_sort.sort_by_key(|sorts| (!sorts.1, sorts.2));
+    to_sort.sort_by_key(|sorts| (sorts.1, sorts.2, sorts.3));
 
-    for (i, (node, _, _)) in to_sort.into_iter().enumerate() {
+    for (i, (node, _, _, _)) in to_sort.into_iter().enumerate() {
         let AssetNode::Asset(asset) = node else {
             continue;
         };
 
-        if asset.order.is_none() {
-            // Reserve some space in the front for official assets that have been
-            // manually ordered.
-            asset.order = Some(100 + i);
-        }
+        asset.order = Some(i);
     }
 }
 
--- a/generate-assets/src/bin/generate.rs
+++ b/generate-assets/src/bin/generate.rs
@@ -72,13 +72,13 @@ fn sort_section(nodes: &mut [AssetNode], latest_bevy_version: &semver::Version) 
     let mut to_sort = vec![];
     for node in nodes {
         let is_semver_compat = node_semver_compat_with(node, latest_bevy_version);
-        let random: u32 = rand::random();
 
         let existing_order = match node {
             AssetNode::Asset(asset) => asset.order.unwrap_or(usize::MAX),
             _ => continue,
         };
 
+        let random: u32 = rand::random();
         to_sort.push((node, existing_order, !is_semver_compat, random));
     }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by janhohenheim — APPROVED** [review]

File: generate-assets/src/bin/generate.rs [hunk 1]
```diff
@@ -48,12 +48,67 @@ fn main() -> anyhow::Result<()> {
...
     Ok(())
 }
 
+/// Sort the assets in the section so that:
+/// - Assets that have been manually assigned an order in `bevy-assets` are first
+/// - Assets that are semver compatible with Bevy are next
+/// - If all else is equal, sort randomly
+fn sort_section(nodes: &mut [AssetNode], latest_bevy_version: &semver::Version) {
+    for node in nodes.iter_mut() {
+        if let AssetNode::Section(ref mut section) = node {
```
benfrankel, line 67:
```suggestion
    for node in nodes {
        if let AssetNode::Section(section) = node {
```
This seems to still compile.

generate-assets/src/bin/generate.rs
```diff
@@ -48,12 +48,67 @@ fn main() -> anyhow::Result<()> {
...
+            sort_section(&mut section.content, latest_bevy_version);
+        }
+    }
+
+    let mut to_sort = vec![];
+    for node in nodes.iter_mut() {
+        let is_semver_compat = node_semver_compat_with(node, latest_bevy_version);
+        let random: u32 = rand::random();
```
benfrankel, line 75: Would move this line down to where it's used, since we might `continue` before needing to generate a random number.

generate-assets/src/bin/generate.rs
```diff
@@ -48,12 +48,67 @@ fn main() -> anyhow::Result<()> {
...
+/// - If all else is equal, sort randomly
+fn sort_section(nodes: &mut [AssetNode], latest_bevy_version: &semver::Version) {
+    for node in nodes.iter_mut() {
+        if let AssetNode::Section(ref mut section) = node {
+            sort_section(&mut section.content, latest_bevy_version);
+        }
+    }
+
+    let mut to_sort = vec![];
+    for node in nodes.iter_mut() {
```
benfrankel, line 73:
```suggestion
    for node in nodes {
```

benfrankel (changes requested):
Small nits.

generate-assets/src/bin/generate.rs
rparrett: Are you sure? Does not compile on my end and I'm a bit confused about how it possibly could.

benfrankel: Hm. This compiled on the Rust playground specifically:
```rust
struct Section {
    content: Vec<AssetNode>,
}

enum AssetNode {
    Section(Section),
}

fn sort_section(nodes: &mut [AssetNode]) {
    for node in nodes {
        if let AssetNode::Section(section) = node {
            sort_section(&mut section.content);
        }
    }
}
```

rparrett: I am doing more stuff than that in the real code though.

benfrankel: Okay, so `IntoIter` on an `&mut [T]` will consume it apparently. This doesn't happen for `&[T]`...

Amended suggestion:

```suggestion
    for node in nodes.iter_mut() {
        if let AssetNode::Section(section) = node {
```

benfrankel: Oh. It's because `&T` is `Copy`, but `&mut T` is not.

benfrankel (approved):

PR #1562, opened 2024-07-16
Support `crate` field for Bevy Assets

benfrankel:
Context: https://github.com/bevyengine/bevy-assets/issues/458.

### DIFF
--- a/generate-assets/src/lib.rs
+++ b/generate-assets/src/lib.rs
@@ -6,6 +6,7 @@ use gitlab_client::GitlabClient;
 use serde::Deserialize;
 use std::cmp::Ordering;
 use std::{fs, path::PathBuf, str::FromStr};
+use url::Url;
 
 pub mod github_client;
 pub mod gitlab_client;
@@ -23,6 +24,8 @@ pub struct Asset {
     pub description: String,
     pub order: Option<usize>,
     pub image: Option<String>,
+    #[serde(rename = "crate")]
+    pub crate_name: Option<String>,
     pub licenses: Option<Vec<String>>,
     pub bevy_versions: Option<Vec<String>>,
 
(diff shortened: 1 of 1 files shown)


generate-assets/src/lib.rs
```diff
@@ -220,7 +225,10 @@ fn get_extra_metadata(
 ) -> anyhow::Result<()> {
     println!("Getting extra metadata for {}", asset.name);
 
-    let url = url::Url::parse(&asset.link)?;
+    let url = match &asset.crate_name {
+        Some(crate_name) => Url::parse(&format!("https://crates.io/crates/{crate_name}")),
+        None => Url::parse(&asset.link),
+    }?;
```
janhohenheim, line 231: I'd be a bit confused if I submitted a link to my assets entry only to find that it isn't actually used. Maybe it should be made explicit in https://github.com/bevyengine/bevy-assets/pull/459 that the link is only a fallback for when the crate name is missing.

generate-assets/src/lib.rs
```diff
@@ -23,6 +24,8 @@ pub struct Asset {
     pub description: String,
     pub order: Option<usize>,
     pub image: Option<String>,
+    #[serde(rename = "crate")]
+    pub crate_name: Option<String>,
```
janhohenheim, line 28: Doesn't this make the link de facto `Optional`? You only need at least one of `{ crate_name, link }`

janhohenheim (approved):
LGTM, got some minor inquiries that are not blocking.

generate-assets/src/lib.rs
benfrankel: Yes, it has to be optional at least for now, since many assets are not crates, and even some of the already-submitted crates don't have a crates.io upload. For example `bevy_doryen`, which has no crates.io upload, but there's a _different_ crate also named `bevy_doryen` that _does_ have a crates.io upload.

generate-assets/src/lib.rs
benfrankel: This is only the case when determining the latest supported Bevy version. The `link` field will still be exclusively used as the place you're directed to when you click on the asset on the website.

TrialDragon (approved):
This LGTM.

> [!NOTE]
> This PR has merge conflicts which must be resolved.

benfrankel:
Something got messed up with git on my end. Should be good now.

PR #1566, opened 2024-07-18
Added target folder size warning with Cranelift on Windows

FlooferLand:
When using Cranelift on Windows with debug symbols turned on for all dev packages, you can end up with a target folder up to 20 GBs for the hello world project and initial compile time can go up to 30 minutes.

image

This PR adds a warning to prevent users from bumping into this issue if they don't have much storage or time to spare.

Feel free to edit the message itself! I have done my best to write down the problem in as little characters as possible in order to not make a large warning note.

### DIFF
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -322,6 +322,11 @@ Notably, Wasm builds do not work yet.
 
 When shipping your game, you should still compile it with LLVM.
 
+{% callout(type="warning") %}
+With Cranelift on Windows, you can end up with very large `target/` folders if you have debug symbols turned on.
+Disable debug symbols by adding `debug = false` under the `[profile.dev.package."*"]` section in your `Cargo.toml`.
+{% end %}
+
 #### Generic Sharing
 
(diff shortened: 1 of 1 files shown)


SkiFire13:
Is this with all dependencies compiled with cranelift or just the final binary crate?

FlooferLand:
This was tested after doing a `cargo clean`, so all dependencies.
The final binary being compiled with Cranelift wouldn't make much of a difference. Most of the compile time was spent on the thousand dependencies.

FlooferLand:
After a few more problems, it appears to not be Cranelift..?
I'm not sure what causes it now but it seems to be one of the optimizations.
Using `debug = false` solves it

janhohenheim:
@FlooferLand huh, weird. I suppose we can close this PR then?

FlooferLand:
Adding some form of warning telling people to use `debug = false` if their initial build is taking too long might not be a bad idea. I'm not sure

janhohenheim:
If there is some additional investigation into this confirming that `debug = false` can speed up Windows builds for others as well, I'll gladly approve a new PR :)
In the mean time, I'll keep it in the back of my mind and direct people on Discord here if they report slow build times.

PR #1570, opened 2024-07-25
0.13 migration: Fix canvas size css

rparrett:
This seems to be necessary.

It's possible that it's not necessary when specifying a canvas by id, but I didn't test that. Either way, with Bevy's defaults, I believe this is needed.

note: `fit_canvas_to_parent` was re-added in Bevy 0.14, but seemingly only as a convenience for users without the ability to use css. So I am pretty sure this is still relevant.

### DIFF
--- a/content/learn/migration-guides/0.12-to-0.13.md
+++ b/content/learn/migration-guides/0.12-to-0.13.md
@@ -1473,8 +1473,8 @@ Remove uses of `Window::fit_canvas_to_parent` in favor of CSS properties, for ex
 
 ```css
 canvas {
-  width: 100%;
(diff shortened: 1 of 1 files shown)


rparrett:
Blocked by #1571

TrialDragon (approved):
LGTM. This seems to be the correct solution given the context. 

> [!NOTE]
> From what I can tell, the reason the PR that removed this was reverted was due to the functions we call from winit overriding user styling for these attributes, requiring this workaround. Also apparently it was problematic, but I literally couldn't find any info about what about it was problematic, prior to the winit issue, on GitHub or Discord).

PR #1574, opened 2024-07-28
Mention in migration guide 13317 that `Copy` trait is no longer implemented on `RenderLayers`

### LINKED ISSUE #1554: 0.13 to 0.14 Migration guide: Mention that RenderLayers is no longer Copy
This MR https://github.com/bevyengine/bevy/pull/13317 makes RenderLayers no longer Copy, requiring a Clone instead. The 0.13->0.14 migration docs should mention this, as it breaks existing programs.

```
layer: RenderLayers
----- move occurs because `layer` has type `RenderLayers`, which does not implement the `Copy` trait
```

TrialDragon:
Fixes #1554 

This was an undocumented breaking change in https://github.com/bevyengine/bevy/pull/13317 which missed initial migration guides writing.


BD103 (approved):
Thanks! (I kind of got burnt out on the migration work once I finished, so I appreciate you doing the follow up!)

BD103:
Marking this PR as trivial and ready-to-merge :)

PR #1577, opened 2024-07-28
Separate contributors from authors for release notes

### LINKED ISSUE #1468: Authors field for release notes has too many false positives
>  `@Alice` maybe we should default to Authors being populated by the original PR creator, then have a second (not actually rendered) list of "potential authors", populated by commit authors

- @cart

The problem here is that people who left tiny suggestions get credited by default. While people who submit fixes or do extensive design work should be credited, typo fixes don't count 😛

TrialDragon:
Fixes #1468 

These are being separated since contributors can include changes on the level of typo fixes which is not desired for the authors position. 

Tested with cargo check, clippy, and fmt.

### DIFF
--- a/generate-release/src/release_notes.rs
+++ b/generate-release/src/release_notes.rs
@@ -66,8 +66,6 @@ pub fn generate_release_notes(
 
         contributors.remove(&author);
 
-        // Make sure the author is always the first in the list
-        let authors = vec![author];
         // Separate the contributors from authors for manual
         // filtering since contributors may include typo fixes
         // and other minor changes unwanted for the author position.
@@ -75,7 +73,7 @@ pub fn generate_release_notes(
 
         notes_metadata.push(generate_metadata_block(
             &title,
-            &authors,
+            &author,
             &contributors,
             pr.number,
             &file_name,
@@ -114,7 +112,7 @@ pub fn generate_release_notes(
 
 fn generate_metadata_block(
     title: &str,
-    authors: &[String],
+    author: &str,
     contributors: &[String],
     pr_number: u64,
     file_name: &str,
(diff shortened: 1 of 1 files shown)


generate-release/src/release_notes.rs
```diff
@@ -67,11 +67,18 @@ pub fn generate_release_notes(
         contributors.remove(&author);
 
         // Make sure the author is always the first in the list
-        let mut authors = vec![author];
-        authors.extend(contributors);
+        let authors = vec![author];
```
IceSentry, line 70: does this still need to be a vec? Are there situations where you can have multiple authors per PR if you don't include the contributors?

TrialDragon: I don't think so; I'll simplify the code ASAP.

generate-release/src/release_notes.rs
```diff
@@ -107,21 +112,23 @@ pub fn generate_release_notes(
...
+    contributors: &[String],
     pr_number: u64,
     file_name: &str,
 ) -> String {
     // TODO should probably add some weight for sorting
     format!(
         r#"[[release_notes]]
 title = "{title}"
-authors = [{authors}]
+authors = ["{author}",]
```
IceSentry, line 124: Technically, this doesn't need to be an array anymore, but I'm not sure if it will break the generator tool if it's changed so we should probably just keep it like that. The trailing comma doesn't seem necessary though.

TrialDragon: It should still be an array since we will be manually adding back contributors to it depending on the level of contribution. And the release notes template then will still expect it to be an array, so it should be an array type.

As for the trailing comma, I left it there for ease of use when folks do add contributors back to the array on a case by case basis.

alice-i-cecile (approved):

IceSentry (approved):

Issue #265, opened 2022-01-11
Tracking issue: new book


alice-i-cecile:
As I'm sure the audience is aware, the current book is *very* short. Naively extending it wasn't going to lead to great outcomes, so this [RFC](https://github.com/bevyengine/rfcs/blob/main/rfcs/23-quick_start_book.md) for a new structure [was approved](https://github.com/bevyengine/rfcs/pull/23)!

So, how do we get there? So far, we've:
- created a `new-book` branch that we can safely work off of
- written (and reviewed!) some draft chapters

However, this process sucked, for several reasons:
1. Shortcodes are miserable to write, buggy, require authors to learn a custom DSL, have limited functionality, and make reading the raw code very hard. This is bad for both authors and reviewers.
2. Reviewing an entire chapter at a time means that the turn-around on PRs is super long, and reviewing is very hard.
3. Code samples would regularly break in dumb or subtle ways, as they weren't being compiled. This also made them a lot harder to write effectively.
4. We didn't have a proper style guide in any form.

Overall, this made the process slow, painful and inaccessible. I want to change that. On the technical / process side, I want to:

- [ ] Rebase the `new-book` branch.
- [ ] Replace short codes with links (#224).
- [ ] Make the doc links pretty, like our current links. Extend with #200?
- [ ] Make book PRs one page at a time, rather than one chapter at a time.
- [ ] Merge book PRs rapidly and aggressively. These don't need to be perfect: an iterative improvement is much more useful and open. We can and should do a final, nitpicky quality pass on the end result before public release.
- [ ] #207, for quality assurance
- [ ] #177, to improve reliability of links to docs.
- [ ] #83, to make sure our code actually works (and continues to work!).
- [ ] Write a basic style guide for the book, which links to the RFC.

In terms of content, I would like to:
- [ ] pick two or three games, and create high-quality tutorials to them that assume 0 Bevy knowledge
- [ ] merge and polish the ECS chapter
- [ ] finish the game logic chapter
- [ ] add a rendering chapter
- [ ] add an assets chapter
- [ ] add an input chapter

Everything else is non-essential (or, like audio and UI, too immature); if you want to work on another chapter feel free, but I don't think we should block the book release for it.

cart:
This all sounds good to me. #272 works way better than I expected (largely because I didn't know inline code block link aliases were a thing). That also solves #273, as it already formats them as "inline code block links".

[...]

PR #1578, opened 2024-07-29
Enhance learning material code examples validation and formatting

### LINKED ISSUE #279: Enforce formatting of code examples in Bevy book

TrialDragon:
# Objective
Part of #265 
Fixes #279

As it stands code snippets used in the learning materials are not formatted and are currently only checked / validated by `code-validation` which is a generally fragile, and at times frustrating, tool.

## Solution

This PR removes the `code-validation` tool and supplants it with `learning-code-examples` which is a Cargo project without a `src/` folder which stores the learning material code examples in the `examples/` directory.  From here we can run `cargo check --examples`, `cargo clippy --examples`, and most importantly `cargo fmt --check` to validate and format the learning material code examples. Tho for ease of use it also supplies a `validate_examples.sh` script which runs all three of those commands on the examples.

The examples within have `// ANCHOR: name`, `// ANCHOR_END: name` and `// HIDE` to create the code snippets used in the pages.

The `file_code_block` shortcode also got refactored to accommodate this and is now the primary way to use code block examples in the learning materials.

The Quick Start Guide has been converted to use this tool as well.

> [!NOTE]
> See the `learning-code-examples/README.md` for full usage details.

> [!NOTE]
> This tool may also supplant `write-rust-doc-lines`, but I am unsure if `write-rust-doc-lines` impacts only the learning material code blocks, or if we also use it in the migration guides and other places in the site's content.

## Testing

Run `zola serve` to see the results of the snippets being pulled in.
Run `./validate_examples.sh` to see the code examples be confirmed as running. (This actually helped spot that the plugin-development generic plugin example was non-functional; now it _should_ work).


alice-i-cecile (approved):
Delightful! Really pleased to see this working, and I think this is a reasonable approach.

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
 
       - name: Build & run doc tests
-        run: cd code-validation && cargo test
+        run: cd learning-code-examples && cargo check --examples && cargo clippy --examples && cargo fmt --check
```
BD103, line 43:
```suggestion
        run: cd learning-code-examples && ./validate_examples.sh
```

It might be worth replacing this long line with a call to `validate_examples.sh`.

CONTRIBUTING.md
```diff
@@ -20,11 +20,14 @@ To check out any local changes you've made:
...
 A local server should start and you should be able to access a local version of the website from there.
 
-## Testing changes to book code snippets
+## Testing changes to learning material code examples
 
-The code in the book is compiled and tested to make sure that the examples work for readers.
+The code in the learning materials (e.g., Bevy Book, Quick Start Guide, Advanced Examples, etc.) is compiled, formatted, and tested to make sure that the examples work for readers.
 
-To check your local changes, move into the `code-validation` folder of this project using `cd code-validation`, and then run `cargo test`. If you write a new section of the book, you have to add it in `code-validation/src/lib.rs` to enable code validation for it.
+To validate your code snippets either `./learning-code-examples/validate_examples.sh` (Recommended) or from the root of the project `cd learning-code-examples && cargo check --examples && cargo clippy --examples && cargo fmt --check`.
```
BD103, line 27: Should this also recommend running `validate_examples.sh` instead? It's not cross-platform, but it is shorter.

learning-code-examples/Cargo.toml
```diff
@@ -0,0 +1,62 @@
+[package]
+name = "learning-code-examples"
+version = "0.1.0"
+edition = "2021"
```
BD103, line 4:
```suggestion
name = "learning-code-examples"
edition = "2021"
publish = false
```

The version number can be removed, since version tracking doesn't really make sense in this scenario. I also like to add `publish = false` to protect against accidental `cargo publish` calls, though this mistake is quite rare.

Additionally, it may be worth adding [`autoexamples = false`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery). This will prevent example auto-discovery, protecting against confusing behavior where some examples are caught and some aren't.

learning-code-examples/README.md
```diff
@@ -0,0 +1,31 @@
+# Learning Code Examples
+
+This folder contains all the examples used in the official Bevy website learning materials, e.g., Bevy Book, Quick Start Guide, Advanced Examples, et cetera.
+
+## Usage
+
+1. Create an `*.rs` example in `./examples/{LEARNING_MATERIAL}` where `{LEARNING_MATERIAL}` is the official Bevy learning material the example will be used in, e.g., Bevy Book, Quick Start Guide, Advanced Examples, et cetera.
+  > [!TIP]
+  > In these Rust examples you use `// ANCHOR: name` and `// ANCHOR_END: name` to define the snippets that can be used in the pages. You also have `// HIDE` which you can append to the end of a line to hide that line from being parsed / rendered.
```
BD103, line 9: It may be worth adding an example of this, but that's probably for a follow-up PR :)

learning-code-examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+> [!NOTE]
+> The file in the `file_code_block` shortcode is already prefixed with `learning-code-examples/examples`!
+
+## Validating Code
+
+To validate code you can run the bash script contained in this directory:
+`./validate_examples.sh`
+
+>[!TIP]
+> The bash script can be called from any directory!
```
BD103, line 29: 🥳

BD103 (approved):
Yay, I'm so happy this is finally happening! Thank you for your hard work :D

CONTRIBUTING.md
TrialDragon: That is what I intended; the `(Recommended)` is intended to apply towards `./learning-code-examples/validate_examples.sh`. I'll experiment with the wording to try and make it less ambiguous.

PR #1579, opened 2024-07-29
Change generate-release to use a better path to release-content

### LINKED ISSUE #1465: `generate-release` must be run from the `generate-release` folder
`generate-release` assumes that it is run from the [`generate-release`](https://github.com/bevyengine/bevy-website/tree/45998c45f2b43fe770526bf268f68b7dd6f44632/generate-release) folder, so it always writes to `../release-content`. This is an issue when it is called from the root directory using `cargo run -p generate-release`, and tripped me up when I first started using the tool.

I propose that `generate-release` always switch into the correct directory, so it can keep these assumptions. If this were a shell script I'd do something like the following, but it may be more involved for Rust:

https://github.com/bevyengine/bevy-website/blob/45998c45f2b43fe770526bf268f68b7dd6f44632/generate-errors/generate_errors.sh#L3-L4

TrialDragon:
# Objective
Fixes #1465 

Currently `generate-release` uses the relative path of `../release-content`, but this is dependent on where you're calling from and as such is undesirable.

## Solution
This PR instead uses the `current_exe()` function to get the absolute path to the executable, which in our use cases will always be `{workspace}/target/{debug|release}/{executable}`. So if we get the workspace directory (equivalent to the root of the project, which is what we want), then we always have the correct directory regardless of if we call it in `generate-release` or `{workspace}`.

## Testing
It has not been tested fully; when I try to run it, it seems to require a `GITHUB_TOKEN` to make issues and PRs which is undesirable in testing imo.

### DIFF
--- a/generate-release/src/main.rs
+++ b/generate-release/src/main.rs
@@ -5,7 +5,7 @@ use contributors::generate_contributors;
 use github_client::BevyRepo;
 use migration_guides::generate_migration_guides;
 use release_notes::generate_release_notes;
-use std::path::PathBuf;
+use std::env::current_exe;
 
 mod changelog;
 mod contributors;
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Perfect, this has annoyed me in the past.

alice-i-cecile:
Use the dry-run flag to avoid creating issues during testing :)

TrialDragon:
> Use the dry-run flag to avoid creating issues during testing :)

It just says `error: unexpected argument '--dry-run' found`, or if I try it as a subcommand just in case `error: unrecognized subcommand 'dry-run'`

NthTensor (approved):

PR #1600, opened 2024-08-16
Add specific zola version currently being used

victorb:
Currently, the build process uses Zola `0.17.2` (ref https://discord.com/channels/691052431525675048/695741366520512563/1272241115558314037) so added that to the README so people don't try to use the latest Zola version (which is currently incompatible)

Also changed the link to go directly to the installation page, and changed the wording from `Download Zola` to `Install Zola`, small copyedits :)


alice-i-cecile (approved):
🥲

PR #1608, opened 2024-08-19
Update Zola to v0.19.2

### LINKED ISSUE #1601: Enable use of latest Zola version (0.19.2)
Currently, it seems the website is stuck with version `0.18.0` of Zola, and is incompatible with the latest version (0.19.2 at the time of writing).

Some of the config options in `config.toml` have changed slightly, leading to people not being able to use the Zola version available in their distributions, meaning people have to manually install an earlier version of Zola. 

https://github.com/bevyengine/bevy-website/pull/1600 is a temporary workaround but a proper fix would be to migrate the config to work with latest Zola version.

TrialDragon:
# Objective

Closes #1601 

Update Zola to v0.19.2 so that folks can just use the latest version without worry.

## Solution

Modify the the `config.toml` to account for the support for multiple types of feeds (RSS, and Atom), and remove an erring line on our part (`highlight_themes` seems to be only valid under the `[markdown]` category, and I don't know why this didn't error before).

Change the work flows to use v0.19.2 of Zola.

Also update the CONTRIBUTING.md and Contributing Guide to reflect that we've changed versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,7 +245,7 @@ jobs:
           path: content/community/people
 
       - name: "Build website"
-        uses: shalzz/zola-deploy-action@v0.18.0
+        uses: shalzz/zola-deploy-action@v0.19.2
         env:
           PAGES_BRANCH: gh-pages
           BUILD_DIR: .
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: "Build and deploy website"
         if: github.repository_owner == 'bevyengine'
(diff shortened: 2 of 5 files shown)


alice-i-cecile:
@cart do we need to update anything on the hosting side for this? I know that for leafwing's site we need to bump the Zola version with Netlify too.

BD103 (approved):
Tested locally and it works with no obvious issues :+1:

@alice-i-cecile since we use Github Pages to host the site, the only thing that needs to be changed is Github Actions. Trial already got that in this PR, though, so we're good! (Feel free to correct me if I'm wrong, though!)

cart (approved):
@alice-i-cecile yup this is all we need. This is good to go!

PR #1623, opened 2024-08-30
Update the latest version on the Getting Started pages.

Aityz:
This PR updates the latest version from 0.14 to 0.14.1.


ItsDoot:
If you specify `my_crate = X.Y` in Cargo, [it will actually grab the latest `my_crate = X.Y.Z`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), so this doesn't really change anything.

BD103:
Yeah, I'd prefer leaving it as-is because it behaves the same and means we don't have the update the website each time a patch version is released.

PR #1624, opened 2024-09-06
better documentation for patch releases

mockersf:
mostly adds a shell script so that I don't miss PRs when cherry picking a milestone


alice-i-cecile (approved):

BD103 (approved):
Looks good looking it over, though I haven't run the Bash script. I'm assuming you've run it before and know how to fix it if something's broken. :)

PR #1626, opened 2024-09-10
Fix internal link to birthday post

piedoom:
The current deploy 404s, this fixes it using Zola's newer [internal linking symbol](https://www.getzola.org/documentation/content/linking/) that grabs the specific file.


alice-i-cecile (approved):
Thank you!

PR #1643, opened 2024-10-08
Update .gitignore and templates to resize donor and news images

MyZeD:
Optimize the size of the donor logos as they are too large. (See first screenshot)

The total file size of all images is 2.3 MB - in the new version the resizing only occurs in the rendered view to the same maximum height.
The result: 704kB (see #2 screenshot).

Bildschirmfoto vom 2024-10-08 18-33-54
Bildschirmfoto vom 2024-10-08 18-34-25

I also adjusted the height and width of the news images to the maximum rendering size.

I also wrote the donor_logos in the gitignore so that you can't upload them accidentally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ content/community/people/
 content/community/donate/
 static/assets/examples/
 static/processed_images/
+static/assets/donor_logos/
 
 # Tools
 target/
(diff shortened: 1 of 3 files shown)


krunchington (approved):
Good use of the resize_image macro, and clamping to integers.

PR #1644, opened 2024-10-08
Write in a alternative logo-title if name is not given

MyZeD:
The change ensures that if a donor's name is not provided, the `alt` attribute for the donor's logo will use the filename (without the extension) as the alternative text.

* [`templates/sponsors.html`](diffhunk://#diff-d8ee107f524a55dc818ad29b30faf95d6f235424d7b12a83a874d24971fd1777R76-R77): Added a fallback for the `alt` attribute to use the donor's logo filename if the donor's name is not available.


TrialDragon (approved):

BD103:
@MyZeD could you update your branch to have the latest changes in `main`? There should be a nice button on this page that does it for you :)

The CI error was unrelated to your changes, so if you update we can then merge this!

alice-i-cecile:
I can press that button too :)

PR #1628, opened 2024-09-15
Record initial actions by the board without a meeting

alice-i-cecile:
Nothing earth-shattering here, but it's good to do for transparency and others may find it interesting / helpful.

Going forward, these will be updated much more promptly after each decision: it's much easier with a structure in place.

PR #1645, opened 2024-10-09
Fix warning on `actions-without-meeting` minutes

BD103:
This page was [missing the `date` field](https://github.com/bevyengine/bevy-website/actions/runs/11250795257/job/31280424360), as helpfully pointed out by @TrialDragon. I filled it out based off the merge date of #1628.

While I was there, I also added the author field, since it was missing.

This will not fix the website not deploying, since there's an issue with the donation page, but it will silence the only other warning.


BD103:
I also included a fix where the assets cache would not be found, but Bevy Assets would not be built. Because actions can only output strings, `!steps.restore-cached-assets.outputs.cache-hit` was always being resolved to `false`. (Because `'true'` and `'false'` are both non-empty strings.)

While what happened with previous CI runs has technically been fixed in https://github.com/actions/cache/pull/1467 with `actions/cache@v4.1.1`, this prevents further issues from happening in the future.

cart (approved):

PR #1743, opened 2024-10-25
Sort migrations by area and improve styling

doup:
- Change `migration_guides.rs` so that it generates sorted `_guides.toml` each time
- Add area sections in migration guides and improve styling

https://github.com/user-attachments/assets/c45dc82e-2360-4f3a-923d-18ac1b021945

### DIFF
--- a/generate-release/src/migration_guides.rs
+++ b/generate-release/src/migration_guides.rs
@@ -1,5 +1,4 @@
 use anyhow::Context;
-use rayon::vec;
 use serde::Deserialize;
 
 use crate::{
--- a/generate-release/src/migration_guides.rs
+++ b/generate-release/src/migration_guides.rs
@@ -59,9 +59,9 @@ pub fn generate_migration_guides(
             .transpose()?;
 
         eprintln!("metadata exists? {}", preexisting_metadata.is_some());
-// Populate the metadata to be written with the
-// preexisting metadata so that it is not lost,
-// or overwritten.
+        // Populate the metadata to be written with the
+        // preexisting metadata so that it is not lost,
+        // or overwritten.
         preexisting_metadata
             .map(|metadata| metadata.guides)
             .unwrap_or_default()
--- a/generate-release/src/migration_guides.rs
+++ b/generate-release/src/migration_guides.rs
@@ -57,7 +57,7 @@ pub fn generate_migration_guides(
             .as_deref()
             .map(toml::from_str)
             .transpose()?;
-    
+
         eprintln!("metadata exists? {}", preexisting_metadata.is_some());
 
         preexisting_metadata
@@ -125,7 +125,7 @@ pub fn generate_migration_guides(
                 let b_areas = b.areas.clone().into_iter().collect::<Vec<_>>().join(" ");
 
                 a_areas.cmp(&b_areas)
-            },
+            }
             (false, true) => std::cmp::Ordering::Less,
             (true, false) => std::cmp::Ordering::Greater,
             (true, true) => std::cmp::Ordering::Equal,
--- a/generate-release/src/migration_guides.rs
+++ b/generate-release/src/migration_guides.rs
@@ -59,7 +59,9 @@ pub fn generate_migration_guides(
             .transpose()?;
 
         eprintln!("metadata exists? {}", preexisting_metadata.is_some());
-
+// Populate the metadata to be written with the
+// preexisting metadata so that it is not lost,
+// or overwritten.
         preexisting_metadata
             .map(|metadata| metadata.guides)
             .unwrap_or_default()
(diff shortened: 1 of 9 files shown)


alice-i-cecile (approved):
I really like the effect. Thanks!

generate-release/src/migration_guides.rs
```diff
@@ -42,61 +42,46 @@ pub fn generate_migration_guides(
...
+        // and then transposes / swaps the internal result of that operation to external,
+        // and returns the error of that result if there is one,
+        // else we have our preexisting metadata, ready to use.
+        let preexisting_metadata: Option<MigrationGuides> = preexisting_metadata_file
+            .as_deref()
+            .map(toml::from_str)
+            .transpose()?;
+
+        eprintln!("metadata exists? {}", preexisting_metadata.is_some());
+
```
TrialDragon, line 62:
```suggestion
// Populate the metadata to be written with the
// preexisting metadata so that it is not lost,
// or overwritten.
```

generate-release/src/migration_guides.rs
```diff
@@ -112,45 +97,62 @@ pub fn generate_migration_guides(
...
+                    file_name,
+                });
             }
         }
     }
 
-    if !new_prs {
-        return Ok(());
-    }
+    // Sort by: Area ASC (empty areas at the end), Title ASC
```
TrialDragon, line 120:
```suggestion
    // Sort by: Area in ascending order (empty areas at the end), and Title in ascending order
```

generate-release/src/migration_guides.rs
```diff
@@ -112,45 +97,62 @@ pub fn generate_migration_guides(
...
+            }
+            (false, true) => std::cmp::Ordering::Less,
+            (true, false) => std::cmp::Ordering::Greater,
+            (true, true) => std::cmp::Ordering::Equal,
+        };
+
+        areas_cmp.then_with(|| a.title.cmp(&b.title))
+    });
+
+    // Replace and overwrite file.
```
TrialDragon, line 137:
```suggestion
    // Create the metadata file, and overwrite it if it already exists.
    //
    // Note:
    // The file, while overwritten,
    // may still contain the same underlying data gotten from
    // the preexisting metadata earlier, if overwrite_existing is false,
    // thus preserving the data even if the file itself is overwritten.
```

Just wanted to add a clarifying comment, so that no one thinks that this is also overwriting the data (unless overwrite_existing was true earlier), not just the file itself.

generate-release/src/migration_guides.rs
```diff
@@ -112,45 +97,62 @@ pub fn generate_migration_guides(
...
-        OpenOptions::new()
-            .append(true)
-            .create(true)
-            .open(path.join("_guides.toml"))
-            .context("Failed to create _guides.toml")?
-    };
     for metadata in guides_metadata {
-        writeln!(&mut guides_toml, "{metadata}")?;
+        // Generate the metadata block for this migration
+        // We always re-generate it because we need to keep the ordering if a new migration is added
```
TrialDragon, line 147:
```suggestion
        // Generate the metadata block for this migration.
        //
        // We always freshly generate and write this data to the file,
        // rather than appending to the end-of-file,
        // so that we can maintain proper ordering of the entries.
```

TrialDragon (changes requested):
Site works still, and any major concerns I had regarding how the Migration Guides generator would work from purely code review have been thoroughly debunked after testing this, so from a purely functional standpoint I feel comfortable with this PR.
However, I still have changes to request for consideration regarding clear documentation of what has changed in the Migration Guides generator.

(Also, does the migration guide generator still need to be run for the 0.15 guides? Those ones have Without Area at the top instead of bottom from what I can see).

And, finally, purely an observation, nitpick, and bikeshed that'd be best suited for a potential follow-up; perhaps we should more blatantly style the scope of each area? Like, some sort of border containing them, or lighter background behind them? Something to make it clear where one area ends and another begins.

doup:
> (Also, does the migration guide generator still need to be run for the 0.15 guides? Those ones have Without Area at the top instead of bottom from what I can see)

Yes, the command now always overwrites the file because if we pick a new migration guide we don't were it's going to be located (previously it appended); assuming the guides list is not sorted, we must run the command. But the operation is non-destructive as it'll read the file before and only add the entries that are missing.

> And, finally, purely an observation, nitpick, and bikeshed that'd be best suited for a potential follow-up; perhaps we should more blatantly style the scope of each area? Like, some sort of border containing them, or lighter background behind them? Something to make it clear where one area ends and another begins.

hahaha, ok, I'm not sure if I'll like wrapping the whole area section in some border or background… but I'm open to be convinced otherwise; maybe we can make the section header sticky so it's always present while scrolling? 🤔

TrialDragon (approved):
Lgtm now.

alice-i-cecile:
Once CI is green I'll merge this @doup :)

doup:
Ooops… @alice-i-cecile now is fixed.

PR #1754, opened 2024-10-29
Add 0.15 migration guide for GameActivity

richchurcher:
Attempted a migration guide based on docs in the PR.


richchurcher:
@Litttlefish I'm not well set up to test this kind of thing. Do you think you could do a quick run-through with the 0.15 release candidate and see if the guide here suffices? Anything you'd like to change?

Litttlefish:
> @Litttlefish I'm not well set up to test this kind of thing. Do you think you could do a quick run-through with the 0.15 release candidate and see if the guide here suffices? Anything you'd like to change?

Looks good to me, 2 suggestions:
for libc++_shared.so, you can mention that there is a solution in the mobile example(gradle projects)
and for native-activity, it is not simply open it, but needs disabling default and enable defaults with native instead of game. quite complex, I'm still thinking if there's better way🤔

richchurcher:
> Looks good to me, 2 suggestions: for libc++_shared.so, you can mention that there is a solution in the mobile example(gradle projects) and for native-activity, it is not simply open it, but needs disabling default and enable defaults with native instead of game. quite complex, I'm still thinking if there's better way🤔

First is easy done! I don't think I'm qualified to write the second one though. If you can come up with a bullet point list, I can summarise?

Litttlefish:
> > Looks good to me, 2 suggestions: for libc++_shared.so, you can mention that there is a solution in the mobile example(gradle projects) and for native-activity, it is not simply open it, but needs disabling default and enable defaults with native instead of game. quite complex, I'm still thinking if there's better way🤔
> 
> First is easy done! I don't think I'm qualified to write the second one though. If you can come up with a bullet point list, I can summarise?

to switch to native for now they should:
 - disable default features in Cargo.toml
 - enable all default features except game-activity
 - enable native-activity instead

as you can see, this is quite complicated, so I'm wondering if we can set both game and native to optional and let users pick one when compiling android apps

richchurcher:
I think I'd lean pretty heavily in favour of making `GameActivity` the default. Most people, especially for toy projects, are probably not going to care about `NativeActivity`, and if they have deeper understanding of the ecosystem hopefully they can follow the instructions to enable it. I think we want to make the most likely path the easiest.

Litttlefish:
> I think I'd lean pretty heavily in favour of making `GameActivity` the default. Most people, especially for toy projects, are probably not going to care about `NativeActivity`, and if they have deeper understanding of the ecosystem hopefully they can follow the instructions to enable it. I think we want to make the most likely path the easiest.

then probably the bulleted content will do the job

richchurcher:
No worries, I'll update the PR when I get a sec.

release-content/0.15/migration-guides/12095_Add_features_to_switch_NativeActivity_and_GameActivity_usa.md
```diff
@@ -1,2 +1,46 @@
...
+can be found under `app/build/outputs/apk`).
+
+Android Studio may also be used.
+
+Bevy may require the `libc++_shared.so` library to run on Android. This can be manually obtained
+from NDK source, or NDK describes a
+[`build.rs`](https://github.com/bbqsrc/cargo-ndk?tab=readme-ov-file#linking-against-and-copying-libc_sharedso-into-the-relevant-places-in-the-output-directory)
+approach. A suggested solution is also presented in the Bevy mobile example.
+
+Applications that still require `NativeActivity` should:
```
JMS55, line 43: There's no mention of NativeActivity being replaced by GameActivity in this file, so the sentence "Applications that still require `NativeActivity` should:" is confusing and missing context.

richchurcher: Ah right, I mistakenly assumed these get placed with a title describing the change. Will fix.

alice-i-cecile (approved):

PR #1770, opened 2024-11-04
Add 0.15 release notes for UI scrolling

Piefayth:
Luckily there's not too much API surface to discuss here. My first release note, so let me know what I need to change!

### DIFF
--- a/release-content/0.15/release-notes/_release-notes.toml
+++ b/release-content/0.15/release-notes/_release-notes.toml
@@ -429,3 +429,10 @@ authors = ["@cart"]
 contributors = ["@alice-i-cecile", "@killercup", "@BD103"]
 prs = [14791]
 file_name = "14791_Required_Components.md"
+
+[[release_notes]]
+title = "UI Scrolling"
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):
This is solid and appropriately brief. This might be a nice spot for a polished screenshot, but that can be added later.

release-content/0.15/release-notes/_release-notes.toml
```diff
@@ -429,3 +429,10 @@ authors = ["@cart"]
 contributors = ["@alice-i-cecile", "@killercup", "@BD103"]
 prs = [14791]
 file_name = "14791_Required_Components.md"
+
+[[release_notes]]
```
alice-i-cecile, line 433: Double check that we don't have a duplicate in here: the stub should already have been generated.

alice-i-cecile (changes requested):
Actually, one quick request: I think you duplicated the release note stub?

Piefayth:
Whoops. Fixed.

alice-i-cecile (approved):

PR #1771, opened 2024-11-05
Light mode in selected sections

doup:
This PR builds on top of #1603 and hence on the huge effort spearheaded by @lynnpepin.

There are some changes though:
- Synced with latest `main` commit.
- Changed how CSS variables are used to a more semantic approach. We have variables ("tokens") for different BEM components, and depending on light/dark theme we change the colors assigned to these. **We don't alter the color anymore**. So no more color redefinition: `--color-white: #000;`. **All BEM components and pages use token for colors now**.
- And **the most important change, light theme is only enabled on selected templates of the web page**. [Cart raised his concerns in Discord](https://discord.com/channels/691052431525675048/695741366520512563/1303109455751151726) about the quality difference between the light and dark themes so we opted for this more gradual approach. For now it's only applied to `docs.html` which means that the quickstart, book, contributing guide and migration guides will show the light theme **IF** the user prefers light themes.
    - To enable light mode in a template we only need to add the following line in any Tera template that extends `layouts/base.html`. This will add the `light-mode` class to the `<html/>` element.
    ```jinja
    {% block theme %}light-mode{% endblock theme %}
    ```

https://github.com/user-attachments/assets/cd8a166f-ca44-46f8-b3fe-32d412b491de


alice-i-cecile:
Marked as controversial, not because I think anyone has any objection, but because I know @cart will have opinions on the palette etc. I think it looks really nice :)

He's heads-down in BSN work right now though, so it may be a while until you get a review.

lynnpepin:
I really like this!! Definitely a big improvement haha :) 

I've rebuilt the docs from the latest commit and I'm hosting the latest version here, for anyone who would like to demo it: https://bevydocs.lynndotpy.dev/

lynnpepin:
> And the most important change, light theme is only enabled on selected templates of the web page. [Cart raised his concerns in Discord](https://discord.com/channels/691052431525675048/695741366520512563/1303109455751151726) about the quality difference between the light and dark themes so we opted for this more gradual approach. For now it's only applied to docs.html which means that the quickstart, book, contributing guide and migration guides will show the light theme IF the user prefers light themes.

I'd like to suggest that the code examples (everything under `examples/`) also gets in line for an optional light-theme. I think these are just as important as the rest of the documentation, but are effectively unusable on the Bevy site.

(Lower priority, since the "View on GitHub" button takes you to the source, and GitHub has good chops around light/dark preferences.)

Thank you again for taking on this effort :)

doup:
@cart this is blocked because it introduces a new light mode design and [a couple of color palettes](https://github.com/bevyengine/bevy-website/blob/4567dad1d8204398801bbc43aa0188351fff86da/sass/_vars.scss#L44-L71), sometimes with subtly different colors than the ones we're using now [1]. I'm not very happy with these palettes, it's an attempt to introduce color tokens but this PR still uses a lot of custom colors… it's half-way there. Ideally all colors should come from palettes/tokens.

So, **I'm thinking that I could revert these palettes and use the same colors that we're using currently… and leave a design system/tokens effort for a later PR, would that help with moving this PR forward?** You only would need to review how the light pages look like.

1: I don't think this is a problem, as it should look good on it's own, it doesn't need to match 100% with the old colors.

doup:
Alternatively, if the greys palette had more in-betweens (150, 250, 350…) it would reduce any big differences with `main`.

doup:
I've done a side by side comparison of this PR and `main` and it looks good as it is (with the original palettes, without in-betweens). I've fixed some minor inconsistencies and replaced all color references to [use CSS vars](https://github.com/bevyengine/bevy-website/pull/1771/commits/b4bc487a2103e9e0b6fcaf11b25f38775d2e8a15).

cart (commented):
In general this looks good to me. Nice work!

> So, I'm thinking that I could revert these palettes and use the same colors that we're using currently… and leave a design system/tokens effort for a later PR, would that help with moving this PR forward? You only would need to review how the light pages look like.

> I've done a side by side comparison of this PR and main and it looks good as it is (with the original palettes, without in-betweens)

I'd prefer it if you reverted a couple of the "big" dark mode color changes:

1. There are a number of instances of "blue-ification" of black backgrounds. It would be nice if we could get those as close as possible to the original colors

doup:
@cart dark-mode now should look the same as `main`.

Light-mode will be enabled only for: Getting Started, Migration Guides, Contributing Guide and Errors. News light-mode was started, but must be finished & activated in another PR.

templates/404.html
```diff
@@ -6,6 +6,7 @@ <h1>404</h1>
     <img src="/assets/bevy_icon_dark.svg"
           alt="Bevy logo"
           width="176"
-          height="130">
+          height="130"
+          class="inverted">
```
krunchington, line 10: Suggestion- rename this class to `invertable` as the actual state of whether it's inverted or not depends on a media query.  Super nitpick though.

sass/_vars.scss
```diff
@@ -58,14 +38,457 @@ $border-radius: 10px;
...
+    --tree-menu-label-active-bg-color: var(--color-neutral-88);
+    --tree-menu-label-active-color: var(--color-black);
+    --tree-menu-label-color: var(--color-neutral-33);
+    --tree-menu-toc-bg-color: var(--color-neutral-93);
+    --tree-menu-toggle-hover-color: var(--color-neutral-75);
+
+    --warning-color: rgb(179, 158, 24);
+
+    .inverted {
+      filter: invert(90%) hue-rotate(180deg);
```
krunchington, line 467: See previous comment in content markdown
> Suggestion- rename this class to invertable as the actual state of whether it's inverted or not depends on a media query. Super nitpick though.

sass/components/_themed-picture.scss
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.light-mode {
+  .themed-picture {
+    &__light {
+      display: initial;
+    }
+
+    &__dark {
+      display: none !important;
```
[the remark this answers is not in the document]
krunchington: (only on pages that have the light-theme class I mean)

templates/docs.html
```diff
@@ -93,9 +95,6 @@
       {% endfor %}
       <h1>
         {{ section_or_page.extra.long_title | default(value=section_or_page.title) | safe }}
-        {% if section_or_page.extra.subtitle %}
-          <span class="docs-page-subtitle">{{ section_or_page.extra.subtitle }}</span>
```
krunchington, line 97: is this no longer used?

krunchington:
Left a few comments but overall the restructuring totally makes sense.  The stepwise approach with the intersection of class and media queries is also great.

cart:
Yup the palette looks good to me now! You've made inline code blocks "squarer". I prefer the "rounder" style / I think it fits our vibes more.  Can you revert that?

Additionally, the text in the headerbar is still pure white. Can you switch back to the slightly off-white color?

templates/docs.html
doup: I couldn't find any instance of this being used, I think it's a feature no one knows about.

sass/components/_themed-picture.scss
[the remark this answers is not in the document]
doup: Good catch, this is being used in the homepage, but since it's out of the initial light-mode pages I didn't see it. Btw, there were three issues… 😅 
1. The image didn't show because the dark-mode ("light display none") had preference
2. After fixing that it was showing the light mode unconditionally (fixed with the media query)
3. center alignment with margin wasn't working because it's using `display: initial`

doup:
I fixed the reported issues and found a couple of extra more. I think now it's done. 🤞

krunchington (approved):
Leaving my approval, the css changes look well scoped to not cause issues with the current experience, and allow the gradual rollout of the media query based preference. I'd still defer to actual maintainers for approval of the color/style values.

cart:
Haha ok _one more_ change and then we're good to merge: the cards on the Donate page no longer use the standard card background color. Can you switch that back?

doup:
# 🙈🙈🙈🙈🙈🙈

I just pushed `final-light-mode.v4.definitive-NOW-FOR-REAL.pr`… maybe it's ready, maybe not. Schrodinger PR.

---

A new bg-color rule slipped for the donate cards, now removed. I've also fixed a couple of issues with the link colors in light-mode.

cart (approved):
We're good to go now. Thanks for bearing with me!

lynnpepin:
Yay!! This might technically be cosmetic but this is my most highly anticipated Bevy feature :D Thank you so much for all this hard work, I'm excited to see this on the Bevy website

doup:
No problem @cart, the PR had out of scope changes.

PR #1774, opened 2024-11-05
Release notes for additive blending

mweatherley:
Unfortunately, I think the main thing that would improve this would be something like a video of additive blending in action, but:
1. the animated fox example model isn't particularly good for demonstrating additive animations;
2. none of our examples use additive animation blending as far as I know.


alice-i-cecile (approved):
Lovely. We can add a video in a follow-up PR.

Issue #1730, opened 2024-10-20
Write Future Work section for 0.15 release notes


alice-i-cecile:
This should cover at least:

- ghost nodes and UI reactivity
- bevy_cli
- Interaction-yeet
- relations plans
- *maybe* editor prototypes
- more text plans

I'll tackle the first draft of this, and then others can add it to it.

Jondolf:
Also constructs, BSN, and other scene improvements

viridia:
Keyboard focus

PR #1777, opened 2024-11-05
Add stub for Bevy 0.15 release blog post

alice-i-cecile:
Apparently this stub didn't exist yet! I needed this for #1730, and it's cleaner to ship this as its own PR.

When reviewing, please be *very* sure that I didn't screw up the draft status of the post.

Sections marked TODO are, you know, still to be done. I've used a consistent symbol so we can easily double check before publication.

### DIFF
--- a/content/news/2024-??-??-bevy-0.15/index.md
+++ b/content/news/2024-??-??-bevy-0.15/index.md
@@ -1,0 +1,40 @@
++++
+title = "Bevy 0.15"
+date = 2024-12-31 # TODO, fix date
+draft = true
+[extra]
+image = "cover.jpg" # TODO, add cover image
+show_image = true
+image_subtitle = "TODO"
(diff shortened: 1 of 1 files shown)


TrialDragon:
Is this intended to be a public draft or a Zola draft? The latter, which is the draft it is currently using, means this page won't be rendered by Zola at all unless you pass the drafts flag when building or serving, iirc.

If you meant to use our custom public draft system, then this needs
```toml
[extra]
public_draft = ISSUENUMBER
status = 'hidden'
```
Instead of
```toml
draft = true
```

mweatherley:
I don't know anything about the draft status but the content of the post looks fine to me. :)

alice-i-cecile:
I'm aiming for Zola draft for now to be absolutely sure that search engines don't pick this up.

PR #1786, opened 2024-11-10
0.15 Release Notes: Retained Rendering

Trashtalk217:
This truly is a terrible first draft, but I had fun writing it and I will probably improve it in the future.

### DIFF
--- a/release-content/0.15/migration-guides/15320_Retained_Rendering.md
+++ b/release-content/0.15/migration-guides/15320_Retained_Rendering.md
@@ -34,7 +34,7 @@ pub fn extract_clusters(
     }
 }
 ```
-An extract query in the render world queries for entities and components in the main world. Here `entity` is a main world entity and `get_or_spawn(main_world_entity).insert(...)` potentially inserts components on the wrong entity. Remember, there is no longer a one-to-one correspondance between the main and render world entities. Moreover `get_or_spawn` has been deprecated.
+An extract query in the render world queries for entities and components in the main world. Here `entity` is a main world entity and `get_or_spawn(main_world_entity).insert(...)` potentially inserts components on the wrong entity. Remember, there is no longer a one-to-one correspondence between the main and render world entities. Moreover `get_or_spawn` has been deprecated.
 
 In 0.15, you should use `RenderEntity` in place of `Entity` to get the correct entity in the render world. For entities to have a `RenderEntity` they need to be synced first. This can be done either via `WorldSyncPlugin` or adding the `SyncToRenderWorld` to the main world entity.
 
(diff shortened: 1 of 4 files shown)


cart (approved):
I think you hit all of the right notes here, but there are some issues with this first draft:

1. Too wordy: I think we can express the same things in about half the space
2. "Too much personality" / told from your perspective: I think your wording and tone is fun, but I also think it comes at the cost of clarity. A few too many "in jokes" and implied context. And in one place you mention "yours truly". In general, the release notes are written by "us" not "me"
3. "Too much history":  theres too much coverage of the details of how this was implemented (how PRs played out, bugs fixed after the initial prs, etc).

I've reworded it to address these issues. Feel free to tweak further, provided it doesn't regress on that front.

Trashtalk217:
@cart I never claimed it was any good. Thank you for fixing it up. I've since added the migration guide. It's probably a bit long (the second 'Extract components with `ExtractComponentPlugin`' is technically superfluous), but I think it covers everything.

Also, I'd like to credit the authors of all the original prs: @kristoff3r, @re0132 and @tychedelia and I'm not quite sure where that goes.

alice-i-cecile:
> Also, I'd like to credit the authors of all the original prs: @kristoff3r, @re0312 and @tychedelia and I'm not quite sure where that goes.

This goes in the `release-notes.md` at the top of that folder :) There's a section for this metadata for each release note.

release-content/0.15/migration-guides/15320_Retained_Rendering.md
```diff
@@ -0,0 +1,96 @@
...
+) {
+    for (render_entity, clusters, camera) in &views {
+        // some code
+        // After the sync step, all main world entities with a &RenderEntity have a corresponding (empty) render world entity. This should never panic.
+        commands.entity(render_entity).insert(...);
+    }
+}
+
+// in main world, when spawning
+world.spawn(Clusters::default(), Camera::default(), SyncToRenderWorld)
```
JMS55, line 57:
```suggestion
world.spawn((Clusters::default(), Camera::default(), SyncToRenderWorld))
```

release-content/0.15/migration-guides/15320_Retained_Rendering.md
```diff
@@ -0,0 +1,96 @@
...
+) {
+    for (main_entity, clusters, camera in &views) {
+        // do something
+    }
+}
+```
+
+### General advice for working with main and render world entities
+
+When working with entities from both worlds it can be confusing. If you are every in a scenario where this isn't entirely clear (for example, when working on custom extraction code in the render world), we advice that you use `RenderEntity` and `MainEntity` as simple wrappers around `Entity`. Mixing these up can become a real headache and lead to some non-obvious errors.
```
JMS55, line 77:
```suggestion
When working with entities from both worlds it can be confusing. If you are every in a scenario where this isn't entirely clear (for example, when working on custom extraction code in the render world), we advise that you use `RenderEntity` and `MainEntity` as simple wrappers around `Entity`. Mixing these up can become a real headache and lead to some non-obvious errors.
```

alice-i-cecile (approved):

PR #1799, opened 2024-11-13
Revert to old content font color

cart:
#1743 changed the default content font color in a way that significantly reduced contrast. The old color was a better default content color.


cart:
^ @doup

alice-i-cecile (approved):

PR #1801, opened 2024-11-13
Add Bluesky

cart:
It feels like its time


alice-i-cecile (approved):

alice-i-cecile:
Good, I'm excited for this :) The game dev BlueSky community really seems to be picking up steam.

ItsDoot:
Decided not to go with a `@bevyengine.org` handle? That's what Godot does, for example: https://bsky.app/profile/godotengine.org

cart:
> Decided not to go with a @bevyengine.org handle? That's what Godot does, for example: https://bsky.app/profile/godotengine.org

Was planning on doing that soon. We can change the handle retroactively.

PR #1805, opened 2024-11-14
0.15 Release Notes: `Set` Reflection

RobWalt:
Write the section mentioned in #1665

### DIFF
--- a/release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
+++ b/release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
@@ -1,27 +1,39 @@
 <!-- Dedicated `Reflect` implementation for `Set`-like things -->
 <!-- https://github.com/bevyengine/bevy/pull/13014 -->
 
-Up until version 0.14, our `std::collections::HashSet` and `hashbrown::HashSet` didn't enjoy the same level of reflectability as, for example, `std::collections::HashMap`. Now, with version 0.15, these set-like types are fully reflectable!
+Up until version 0.14, our `std::collections::HashSet` and `hashbrown::HashSet`
+didn't enjoy the same level of reflectability as, for example,
+`std::collections::HashMap`. Now, with version 0.15, these set-like types are
+fully reflectable!
 
 ### Why does this matter?
 
-If you're a fan of the excellent `bevy-inspector-egui` plugin crate, you'll now be able to see the contents of your `HashSet`s after updating both the crate to the version compatible with bevy 0.15. This enhancement allows for deeper introspection and easier debugging in your game development workflow.
+If you're a fan of the excellent `bevy-inspector-egui` plugin crate, you might
+have noticed that the contents of sets are not inspectable. This change in the
+bevy crate unblocks improvements to the plugin crate, so you'll finally be able
+to see the contents of your `HashSet`s. This enhancement allows for deeper
+introspection and easier debugging in your game development workflow.
 
-Here is an example of entity connections where this improvement has a good impact:
+Here is an example of entity connections where this improvement will be
+helpful:
 
 ```rust 
+!// previously inspecting this would only show 
+!// 
+!// > "HashSet<Entity, EntityHash> is #[reflect_value], but has no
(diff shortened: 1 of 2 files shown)


release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
```diff
@@ -1,4 +1,58 @@
...
+!// to work
+!// 
+!// After the bevy-inspector-egui update it should just show the entities in a
+!// similar way compared to `Vec`s
+
+#[derive(Component)]
+pub struct Person;
+
+#[derive(Component, Reflect)]
+pub struct Friendships(EntityHashSet<Entity>);
```
cart, line 38: EntityHashSet is not generic. Perhaps it would be better to make this `HashSet<Entity>`, rather than using our built in / special cased `EntityHashSet`, just for simplicity?

release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
```diff
@@ -1,4 +1,58 @@
...
+
+### Why does this matter?
+
+If you're a fan of the excellent `bevy-inspector-egui` plugin crate, you might
+have noticed that the contents of sets are not inspectable. This change in the
+bevy crate unblocks improvements to the plugin crate, so you'll finally be able
+to see the contents of your `HashSet`s. This enhancement allows for deeper
+introspection and easier debugging in your game development workflow.
+
+Here is an example of entity connections where this improvement will be
```
MrGVSV, line 17: I'm not sure how others feel about this but I feel like we should be highlighting the new features provided by `Set` rather than explain how it will affect a single crate (not that we _can't_ mention 3rd party crates, just that I don't think it should be the main focus)

release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
```diff
@@ -1,4 +1,58 @@
...
+helpful:
+
+```rust
```
MrGVSV, line 20: Nit: This example might be better served showcasing how `Set` works, maybe even comparing it to `Map`.

Or perhaps if we do want to use bevy-inspector-egui here, maybe we could show a screenshot of this in action?

alice-i-cecile:
@RobWalt are you planning to get back to this in the next couple of days?

release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
RobWalt: I'm actually not really sure about further implications since I'm not super familiar with bevy reflect. My main motivation for this change was to unblock bevy-inspector-egui after I had a deep dive into the issue.

release-content/0.15/release-notes/13014_Dedicated_Reflect_implementation_for_Setlike_things.md
RobWalt: Re: Screenshots: I had the same thought and wanted to do before/after screenshots but `Set` also has to be implemented from the bevy-inspector-egui side first. I wanted to create a pre-release PR there, but there are a lot of bigger structural changes and not just adjustments because of the new `Set` variant, so I kind of stopped because I don't feel competent enough.

RobWalt:
@alice-i-cecile We might as well skip this section in the release notes if it blocks anything, no hard feelings. I actually didn't think that the PR was big enough for something like notes in the first place. It only was supposed to unblock bevy_inspector_egui and I didn't really think about other use cases yet.

alice-i-cecile:
@SkiFire13 @MrGVSV I've added you both as contributors due to your incredibly valuable help during review. @RobWalt I've shortened this substantially to keep this simple and clear. The people who care about this will be very excited but most users will go "hmm that's kinda neat I guess". If you have a moment, take a look and give me your LGTM?

MrGVSV (approved):

cart (approved):

PR #1814, opened 2024-11-18
Rewrite Text migration guide for 0.15

### LINKED ISSUE #1782: Migration Guide for 0.15 mentions UiTextWriter but it is not found when used.
The Migration Guide mentions UiTextWriter but if you try to use it anywhere it gives this error:
cannot find type `UiTextWriter` in this scope

There is nothing to import.

alice-i-cecile:
Fixes #1782.

This PR leans heavily on the migration guide provide by @MiniaczQ in https://github.com/bevyengine/bevy/pull/15887 <3

### DIFF
--- a/release-content/0.15/migration-guides/15591_Text_rework.md
+++ b/release-content/0.15/migration-guides/15591_Text_rework.md
@@ -1,8 +1,28 @@
-TODO: very breaking
+The `Text` API in Bevy has been overhauled in several ways as part of Bevy 0.15.
+There are several major changes to consider:
 
-__Accessing text spans by index__
+- `ab_glyph` has been replaced with `cosmic-text`. These changes are mostly internal and the majority of users will not interact with either text backend directly.
+- each text section is now stored as a distinct entity within the standard hierarchy, rather than as a `Vec<TextSection>` on the `Text` component. Children of `Text`/`Text2d` entities with `Text` `Text2d` components will act as additional text sections.
+- like other aspects of Bevy's API, required components have replaced bundles
 
-Text sections are now text sections on different entities in a hierarchy, Use the new `TextReader` and `TextWriter` system parameters to access spans by index.
+## `TextBundle` and text styling
+
+`TextBundle` has been removed.
+Add the `Text` component to set the string displayed.
+
+`TextLayout`, `TextFont` and `TextColor` are required components for `Text`,
+and are automatically added whenever `Text` is.
+Set those values to change the text section's style.
+
+Like elsewhere in Bevy, there is no style inheritance. Consider writing your own abstraction for this if this is something you'd like to use.
+
+To control the layout of a `Text` section, modify the properties of its `Node`.
+
+## Accessing text spans by index
+
+Previously, text sections were elements of a vector stored within `Text`.
+Now, they are stored as distinct entities under the same `Parent`.
+You can use the new `TextUiReader` and `TextUiWriter` system parameters to conveniently access text spans by index.
 
 Before:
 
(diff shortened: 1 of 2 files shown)


richchurcher (approved):

release-content/0.15/migration-guides/15591_Text_rework.md
```diff
@@ -1,8 +1,28 @@
-TODO: very breaking
+The `Text` API in Bevy has been overhauled in several ways as part of Bevy 0.15.
+There are several major changes to consider:
 
-__Accessing text spans by index__
+- `ab_glyph` has been replaced with `cosmic-text`. These changes are mostly internal and the majority of users will not interact with either text backend directly.
+- each text section is now stored as a distinct entity within the standard hierarchy, rather than as a `Vec<TextSection>` on the `Text` component. Children of `Text`/`Text2d` entities with `Text` `Text2d` components will act as additional text sections.
```
viridia, line 5: Perhaps not a migration issue, but somewhere I think it useful to mention that since text sections are entities, you can now insert components on them, which is useful for things like:
* Dynamically-computed text
* Picking behavior
* Animated text styles.

release-content/0.15/migration-guides/15591_Text_rework.md
```diff
@@ -1,8 +1,28 @@
...
+## `TextBundle` and text styling
+
+`TextBundle` has been removed.
+Add the `Text` component to set the string displayed.
+
+`TextLayout`, `TextFont` and `TextColor` are required components for `Text`,
+and are automatically added whenever `Text` is.
+Set those values to change the text section's style.
+
+Like elsewhere in Bevy, there is no style inheritance. Consider writing your own abstraction for this if this is something you'd like to use.
```
viridia, line 17: I can provide examples of this if interested.

rparrett (approved):
Much improved. Couple notes: 

There are so many changes to text that I think we could use a few quick before/afters:

- Spawning some text
- Spawning some text with multiple sections

The guide for 15368 also needs adjustment after 15857. It's content would also be better communicated with a diff, I think.

I wonder if we should be linking off to the required components blog post section for context in related migration guides.

release-content/0.15/migration-guides/15591_Text_rework.md
MiniaczQ, line 5:
```suggestion
- each text section is now stored as a distinct entity within the standard hierarchy, rather than as a `Vec<TextSection>` on the `Text` component. Children of `Text`/`Text2d` entities with `TextSection` components will act as additional text sections.
```
Only root is different, sections are shared by all AFAIR

release-content/0.15/migration-guides/15591_Text_rework.md
alice-i-cecile: A link here would be good.

release-content/0.15/migration-guides/15591_Text_rework.md
alice-i-cecile: The `TextSection` component doesn't exist.

alice-i-cecile: Yeah, this belongs in the release notes, not the migration guide.

rparrett: Right, the Component shared by both for sections is `TextSpan`.

alice-i-cecile:
```suggestion
- each text section is now stored as a distinct entity within the standard hierarchy, rather than as a `Vec<TextSection>` on the `Text` component. Children of `Text`/`Text2d` entities with `TextSpan` components will act as additional text sections.
```

release-content/0.15/migration-guides/15591_Text_rework.md
viridia: https://github.com/viridia/thorium_ui/blob/main/crates/thorium_ui_controls/src/text_styles.rs

alice-i-cecile:
```suggestion
Like elsewhere in Bevy, there is no style inheritance. Consider [writing your own abstraction for this](https://github.com/viridia/thorium_ui/blob/main/crates/thorium_ui_controls/src/text_styles.rs) if this is something you'd like to use.
```

PR #1823, opened 2024-11-19
Bubble and trouble: release notes for bubbling observers

alice-i-cecile:
Fixes https://github.com/bevyengine/bevy-website/issues/1657.

### DIFF
--- a/release-content/0.15/release-notes/13991_Minimal_Bubbling_Observers.md
+++ b/release-content/0.15/release-notes/13991_Minimal_Bubbling_Observers.md
@@ -25,7 +25,7 @@ impl <E> Event for Pointer<E>{
 
 This means that, unless you call [`Trigger::propagate(false)`], pointer events will be replicated up the hierarchy (accessing the `Entity` stored in the [`Parent`] component) until it reaches the entity root.
 
-Any type that implements the [`Traversal`] trait can be used as the associated type, and can access arbitrary read-only query data from the world.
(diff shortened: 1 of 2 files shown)


release-content/0.15/release-notes/13991_Minimal_Bubbling_Observers.md
```diff
@@ -1,4 +1,38 @@
...
+```rust
+impl <E> Event for Pointer<E>{
+    type Traversal = &Parent;
+    const AUTO_PROPAGATE: bool = true;
+}
+```
+
+This means that, unless you call [`Trigger::propagate(false)`], pointer events will be replicated up the hierarchy (accessing the `Entity` stored in the [`Parent`] component) until it reaches the entity root.
+
+Any type that implements the [`Traversal`] trait can be used as the associated type, and can access arbitrary read-only query data from the world.
```
BenjaminBrienen, line 28:
```suggestion
Any type that implements the [`Traversal`] trait can be used as the associated type and can access arbitrary read-only query data from the world.
```

"and" is a coordinating conjunction here

BenjaminBrienen (changes requested):
Really good!

BenjaminBrienen (approved):

aevyrie (approved):
I think this is missing credit for @NthTensor?.I seem to recall he drove most of this work, though I might be misremembering.

alice-i-cecile:
Nth is listed as the author here, which is the primary credit :D

PR #1824, opened 2024-11-20
Add domain verification file for Bluesky/AT Proto

coreh:
This should allow changing the handle from `@bevy.bsky.social` to `@bevyengine.org`, matching the official website domain. Assuming that GitHub pages doesn't mess up the deploy/serving of `.well-known`.

### DIFF
--- a/static/.well-known/atproto-did
+++ b/static/.well-known/atproto-did
@@ -1,0 +1,1 @@
+did:plc:rurf32i2ytudjhvyytliz7ct


alice-i-cecile:
Neat! This was on our list of "things it would be nice to do".

cart:
We generally do this sort of thing from the domain registrar itself, which feels cleaner / more direct to me. I've been holding off on making this switch for "secret domain-related reasons" (secrecy intentional to not build hype before things happen)

cart:
I do like this idea, but I'd like to defer for another week-ish, and use a DNS record when we're ready.

PR #1834, opened 2024-11-21
Improve release note styling

cart:
This improves release note style and functionality in a number of ways:

1. We now link to author github profiles when in the form `@X` 
1. Prefers single-line formatting. This was particularly heinous with new lines for each PR
2. Harmonize "pr" and "author" formatting
2. Group prs under a single "PRs" section, rather than labeling each one `PR #1, PR #2, etc`
3. When breaking lines to run out of room, break by PR/Author section first
4. Makes section headings `##` instead of `###`. They were "too small" in their current form / the wrong section level
5. Reduces prominence of PR / Author sections by reducing font size and replacing blue link color with white
6. Reduced the space between the section heading and the author/pr section. This helps logically group them together / helps the reader easily distinguish between "header" information and content. This is the same approach used for the main title + authors.
7. Fixed a number of "over prominent" subheaders in 0.15 articles

### Before
image

### After
image

### DIFF
--- a/release-content/0.15/release-notes/14920_Add_cached_run_system_API.md
+++ b/release-content/0.15/release-notes/14920_Add_cached_run_system_API.md
@@ -31,10 +31,10 @@ let id2 = world.register_system_cached(quux);
 assert!(id1 == id2);
 ```
 
-## Comparison to `run_system_once`
+### Comparison to `run_system_once`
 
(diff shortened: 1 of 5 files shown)


BenjaminBrienen (approved):

doup (approved):
Looks good 👌 

We should probably share these styles with the Migration Guides, but maybe it's for another PR.

Issue #1831, opened 2024-11-21
Add missing migration guides for 0.15


alice-i-cecile:
> https://github.com/bevyengine/bevy/pull/14082 still seems missing. 

 _Originally posted by @rparrett in [#1724](https://github.com/bevyengine/bevy-website/issues/1724#issuecomment-2491185955)_

To check manually:

- [x] https://github.com/bevyengine/bevy/pull/6042: non-breaking
- [x] https://github.com/bevyengine/bevy/pull/14917: non-breaking
- [ ] https://github.com/bevyengine/bevy/pull/15090: missing
- [x] bevy_reflect: Add `Reflectable` trait sha: 69541462c5ef70914fd719fcd1f503045c76a453: non-breaking
- [ ]  https://github.com/bevyengine/bevy/pull/14082: missing
- [x] The Cooler 'Retain Rendering World' sha: 56f8e526dde49b4e4ad62efb7ad27bfe0bd6f617

rparrett:
There are likely other totally missing sections -- I think that this one and perhaps a few others are missing due to the way we do filtering by date in the generation tool. Some PRs that were breaking and created during the 0.14 RC period would fall into the date range of `0.14.0..whatever`.

alice-i-cecile:
Fixing #1518 and re-running the tool is probably the least painful way to resolve this.

alice-i-cecile:
Even after removing the filter on time closed, I'm still getting the following output.
```
PR not found for Allow mix of hdr and non-hdr cameras to same render target sha: 027f8e21ecedec4844ff0e93c077d166ce259891
PR not found for rename bevy_state_macros to bevy_state_macros_official sha: 401234a5fb6865ea9ddd5740d47e13d022959fef
PR not found for Update crate metadata for bevy state sha: 98bc5ff8c4731e8b8fec5babc815c8111173a4b5
PR not found for Document the Release Candidate process sha: ecdd1624f302c5f71aaed95b0984cbbecf8880b7
PR not found for rename the crate bevy_state_macros_official back to its original name sha: b17292f9d11cf3d3fb4a2fb3e3324fb80afd8c88
PR not found for improved the error message by insert_boxed (issue #13646) (again) sha: d45bcfd0434702c0b97de1727e5d5a9b2055dad3
PR not found for Make the component types of the new animation players clonable. sha: 651f3d08d75077c4b12f519d3b22ecc047a9670f
[...]

PR #1846, opened 2024-11-25
Add some missed migration guides for 0.15

alice-i-cecile:
This is part of #1831. This is only the new ones: it was generated by running the migration-guide tool and culling any faulty migration guides.

Some of these new migration guides aren't great, but I'd prefer to merge them as is and then do a clean up pass independently.


release-content/0.15/migration-guides/16233_Use_Name_component_for_gamepad.md
```diff
@@ -0,0 +1,4 @@
+- `GamepadInfo` no longer exists:
+  - Name now accesible via `Name` component.
```
BD103, line 2:
```suggestion
  - Name now accessible via `Name` component.
```

This should fix CI!

BD103 (approved):
Looks good from a brief pass, though the actual content will definitely need to be cleaned up later.

PR #1850, opened 2024-11-26
Add some missing migration guides from older PRs

### LINKED ISSUE #1518: Changelog and contributors tools can fail to find PRs
>  IMO: all of the missed PRs look like they got excluded because they were originally opened before the "from" date for the release

- @ChristopherBiscardi 

Discovered in #1516 / #1517.

alice-i-cecile:
Part of #1831. Fixes #1518. While this didn't hit *all* of the missed PRs, the remaining PRs seem to be renames, and are few enough that they can be manually handled.

Like before, I'm aiming to merge these so-so migration guides as is, and then fix up their content in other PRs. There are still PRs that were absolutely missed: I don't see a better way than manually adding the remaining breaking changes for this cycle.


generate-release/src/github_client.rs
```diff
@@ -198,19 +198,6 @@ impl GithubClient {
 
             prs.append(&mut prs_in_page);
             page += 1;
-            if let Some(pr) = prs.last() {
-                if let Some(datetime_utc) = datetime_utc {
-                    if let Some(closed_at) = pr.closed_at {
-                        if closed_at < datetime_utc {
```
alice-i-cecile, line 204: This data seems to be unreliable, and early filtering here doesn't actually do us any good.

release-content/0.15/migration-guides/10193_Cosmic_text.md
```diff
@@ -0,0 +1,5 @@
+- `Text2dBounds` has been replaced with `TextBounds`, and it now accepts `Option`s to the bounds, instead of using `f32::INFINITY` to inidicate lack of bounds
```
alice-i-cecile, line 1:
```suggestion
- `Text2dBounds` has been replaced with `TextBounds`, and it now accepts `Option`s to the bounds, instead of using `f32::INFINITY` to indicate lack of bounds
```

release-content/0.15/migration-guides/_guides.toml
```diff
@@ -610,18 +694,36 @@ prs = [15735]
...
 prs = [16473]
 areas = ["Picking"]
 file_name = "16473_Add_flags_to_SpritePlugin_and_UiPlugin_to_allow_disabling_.md"
 
+[[guides]]
+title = "Fix `bevy_picking` plugin suffixes"
+prs = [16082]
+areas = ["Picking"]
+file_name = "16082_Fix_bevy_picking_plugin_suffixes.md"
+
```
alice-i-cecile, line 726:
```suggestion
```

release-content/0.15/migration-guides/_guides.toml
```diff
@@ -370,6 +430,18 @@ prs = [14354]
 areas = ["ECS"]
 file_name = "14354_Update_trigger_observers_to_operate_over_slices_of_data.md"
 
+[[guides]]
+title = "Use the fully qualified name for `Component` in the `require` attribute"
+prs = [16378]
+areas = ["ECS"]
+file_name = "16378_Use_the_fully_qualified_name_for_Component_in_the_require_.md"
+
```
alice-i-cecile, line 438:
```suggestion
``

File: generate-release/src/github_client.rs (hunk 1 above)
Comment (rparrett) [reply generate-release/src/github_client.rs:204]: I agree that this filtering seems pointless

File: generate-release/src/helpers.rs [hunk 5]
```diff
@@ -19,21 +19,16 @@ pub fn get_merged_prs(
...
+    println!("Getting list of all merged PRs with label {label:?}");
 
     let base_commit = client.get_commit(from, BevyRepo::Bevy)?;
     let base_commit_date = &base_commit.commit.committer.date[0..10];
 
     // We also get the list of merged PRs in batches instead of getting them separately for each commit
-    let prs = client.get_issues_and_prs(
-        BevyRepo::Bevy,
-        IssueState::Merged,
-        Some(base_commit_date),
```
rparrett, line 31: It's not totally clear to me why this helps, based on GH's [documentation](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues) about the `since` param in their API, but it does seem to cause more stuff to get generated.

rparrett (approved):
This seems to make sense. Included PRs are already filtered by the list of commits. Filtering PRs is just an optimization and it seems broken.

I guess i the future we would want to eventually collect a list of PR numbers and retrieve those specific PRs via the graphql API if possible.

PR #1852, opened 2024-11-26
Use PR number for matching commits to PRs in release/guide generator

### LINKED ISSUE #1851: Match commits to PRs by using the extracted PR number
As seen in #1518, matching PRs by names is unreliable due to both time windows and renames. The fundamental strategy here is dumb though: we should be able to extract the PR number / link responsible from the commit message and then look it up directly. No inexact matching or expensive "grab all PRs" required.

rparrett:
Partially `fixes` #1851

I think there still may be a bit of a mismatch because we only fetch PRs from a particular date range, or whatever. But this should at least help with the renamed PRs.

### DIFF
--- a/generate-release/src/helpers.rs
+++ b/generate-release/src/helpers.rs
@@ -40,12 +40,12 @@ pub fn get_merged_prs(
 
     let mut out = vec![];
     for commit in &commits {
-        let Some(title) = get_pr_title_from_commit(commit) else {
+        let Some((title, number)) = get_title_parts_from_commit(commit) else {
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I've removed the date range filter for the PRs in #1850. We may be able to do a smarter lookup than "fetch all of the PRs", but these two PRs together should be reliable.

alice-i-cecile:
Merging in concert with #1850. There's no user-facing impact here, and this is clearly better.

PR #1860, opened 2024-11-28
Fix up egregiously bad migration guides

alice-i-cecile:
While the migration guides as a whole aren't super polished this release, some of them were *unacceptably* bad. I've gone through every migration guide and removed all of the TODO, not needed and otherwise completely defective migration guides.

The replacements generally aren't as polished as I would like, but they're much better than absolutely nothing or just noise.


release-content/0.15/migration-guides/14082_Optimize_common_usages_of_AssetReader.md
```diff
@@ -1,38 +0,0 @@
-The trait method `bevy_asset::io::AssetReader::read` (and `read_meta`) now return an opaque type instead of a boxed trait object. Implementors of these methods should change the type signatures appropriately
```
djeedai, line 1: Why is that release note deleted?

djeedai (approved):
Aside from the deleted things which I assume have a reason to not ship (were cut?), looks all good.

release-content/0.15/migration-guides/14082_Optimize_common_usages_of_AssetReader.md
alice-i-cecile: I duplicated it earlier by mistake.

PR #1861, opened 2024-11-28
Update bluesky link to new handle

Niashi24:
Link wasn't updated to the new handle so it linked to the old one.


alice-i-cecile (approved):
Good catch, thanks :)

PR #1863, opened 2024-11-29
Increase footer gap

### LINKED ISSUE #1862: Increase the social footer links gap slightly
It's a wee bit close together, but it shouldn't be widened too much.

Just need to change this value:

https://github.com/bevyengine/bevy-website/blob/main/sass%2Fcomponents%2F_footer.scss#L6

bienjensu:
Increases the footer gap from 10px to 25px.

Fixes #1862

### DIFF
--- a/sass/components/_footer.scss
+++ b/sass/components/_footer.scss
@@ -3,7 +3,7 @@
     display: flex;
     flex-direction: row;
     justify-content: center;
-    gap: 10px;
+    gap: 25px;
     padding: 20px;
   }
 


alice-i-cecile:
Can you include a screenshot of what this looks like now for reviewers?

TrialDragon:
The PR result should look like this I think:

after

Compared to before:

before

TrialDragon:
Mobile after:
image
Mobile before:
image

cart:
I'm cool with the new spacing. But I think it has increased the prominence slightly and it was already too prominent. I'm thinking maybe we should make them smaller and a bit gray-er. But that can happen later.!

cart (approved):

PR #1877, opened 2024-11-29
0.15 release notes: Fix a few minor formatting issues

rparrett:

PR #1881, opened 2024-11-29
0.15 migration: Fix other guides getting swallowed by broken formatting

rparrett:
Mentioned by Nikl here: https://github.com/bevyengine/bevy-website/pull/1877#issuecomment-2508729122

Had to go back to the original PR to figure out what was going on here. https://github.com/bevyengine/bevy/pull/15482

I think that most of this content was not intended to be in the migration guide.


rparrett:
#1880 does it better

doup:
I was going to comment that there are more issues 😅👍

PR #1891, opened 2024-12-01
Various migration guide fixes

rparrett:
This is just barely scratching the surface to fix some low hanging fruit.

- 13859: The formatting in this one was really messed up
- 14416: No breaking changes, removed.
- 14842: Used callout shortcode that doesn't work here
- 15281: I think this is outdated, and is covered by general warning at top of page, removed.
- 15666: `bevy_render` -> `bevy::render` for user code. Hopefully a little less confusing now too


alice-i-cecile:
Thanks! These were not very polished this cycle so I appreciate your efforts here.

PR #1897, opened 2024-12-02
Process shortcodes in migration guides

rparrett:
Makes a similar change to #1448, but for migration guides.

I am not totally sure this is correct, but it seems to work.

(See http://localhost:1111/learn/migration-guides/0-14-to-0-15/#replace-the-wgpu-trace-feature-with-a-field-in-bevy-render-settings-wgpusettings) after building the site locally.

Also not making any particular judgment here about whether we want this particular `callout` in the migration guide or if we want those in the migration guides in general.

### DIFF
--- a/templates/shortcodes/callout.html
+++ b/templates/shortcodes/callout.html
@@ -1,3 +1,3 @@
 <aside class="callout callout--{{ type | default(value="info") }}">
-  {{ body | markdown | safe }}
+  {{ body | safe }}
 </aside>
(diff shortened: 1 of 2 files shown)


rparrett:
This shouldn't be merged as-is. It breaks the callouts in other areas in the site by not parsing their markdown.

http://localhost:1111/learn/quick-start/introduction/

rparrett:
Not motivated enough to keep dig deeper.

PR #1901, opened 2024-12-03
Hide assets/examples filters to non-JS users

doup:
Since the beginning the website has favored non-JS solutions when possible. But there are features than can only be implemented with JS. For those users that have JS disabled in their browser we shouldn't show these features.

**This PR hides assets/examples filters to non-JS users.**

To try this in Firefox:
- Go to: `about:config`
- Set `javascript.enabled` to `false`
- Navigate to assets/examples, you shouldn't see the filters

image

image

### DIFF
--- a/templates/assets.html
+++ b/templates/assets.html
@@ -6,17 +6,19 @@
 {% block page_content %}
   <div class="assets">
     {{ assets_macros::init_svg() }}
-    <div class="assets-search">
-      <input class="assets-search__input"
-              type="text"
-              id="assets-search"
-              placeholder="Search (ie: 0.11 MIT)">
-    </div>
-    <div class="assets-filters">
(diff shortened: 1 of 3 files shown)


templates/assets.html
```diff
@@ -6,17 +6,19 @@
...
+                type="text"
+                id="assets-search"
+                placeholder="Search (ie: 0.15 MIT)">
+      </div>
+      <div class="assets-filters">
+        <label>Bevy last supported version</label>
+        <select class="asset-version-select" id="assets-filter">
+          <option value="all_versions">All</option>
+        </select>
+      </div>
```
doup, line 21: I have just wrapped the old code in `div.hidden [data-assets-filters]`.

cart (approved):
Very nice!

PR #1902, opened 2024-12-03
Remove donate JS code in favor of CSS-only solution

doup:
Replace individual and corporate membership options toggle with a CSS only solution.

https://github.com/user-attachments/assets/260a0486-420e-4696-8248-41991796ea17


sass/pages/_donate.scss
```diff
@@ -237,3 +237,21 @@ div.donate-sponsor {
...
+
+#membership {
+  display: block;
+}
+
+#corporate-sponsorship {
+  display: none;
+}
+
+@include state-checked("donate-view") {
```
rparrett, line 249: For other reviewers, this is defined over here: https://github.com/rparrett/bevy-website/blob/76bd33f/sass/_mixins.scss#L43

rparrett:
I find the toggling to be quite awkward with the toggle button below the list of tiers, and the size of the list changing when toggling. But that's not an issue with this PR.

rparrett (approved):

cart (approved):
Removing JS is definitely a win! I do agree with @rparrett that the toggling behavior could use a rethink, but thats a separate convo.

PR #1906, opened 2024-12-05
Hide heading anchors by default

doup:
Hide heading anchors (`#`) by default to reduce noise:

https://github.com/user-attachments/assets/c1b080a6-c53b-429d-9675-b7b94afa6db4


rparrett (approved):
I like this a lot

cart (approved):
Lovely. This was bothering me :)

PR #1907, opened 2024-12-05
Use `<hr>` instead of `border-top` in Migration headings

doup:
Fixes a minor detail 😅. Right now Migration guides H3 headings have the `border-top` set. The problem is that when the user clicks on the sidebar links (or `#`) the scrolls aligns with the horizontal line because that's part of the heading.

The solution is to remove the `border-top` from the H3 heading and use `<hr>` instead, this way the scroll aligns with the text instead of the horizontal line.

Demo:
- First old behaviour
- Then this PR behaviour

https://github.com/user-attachments/assets/380a8c4d-6e74-4dee-bb06-0dcd0d07e563

### DIFF
--- a/sass/pages/_migration_guide.scss
+++ b/sass/pages/_migration_guide.scss
@@ -19,9 +19,18 @@
   }
 
   h3 {
-    margin-block: 2rem .5rem;
-    padding-top: 2rem;
-    border-top: solid rgba(#fff, 0.05) 1px;
+    margin-block: 2rem 0.5rem;
+  }
+
+  hr {
+    $color: rgba(#fff, 0.05);
+
+    height: 1px;
+    color: $color;
+    background: $color;
+    font-size: 0;
+    border: 0;
+    margin-block: 2rem;
   }
 }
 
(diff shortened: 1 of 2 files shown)


rparrett (approved):
Do we not have a GH issue documenting the known wonkiness of the sidebar?

( video over here: https://discord.com/channels/691052431525675048/695741366520512563/1298740373102792726 )

Anyhow, still seeing that issue on my end (chrome/macos), but I like this little detail.

doup:
I saw that mentioned in Discord… but no idea if there is an GH issue.

I just checked, and it's working fine, but since we have a fixed header it feels weird (?). It probably gets fixed if we change the top part in `rootMargin` in the intersection observer. Probably needs the header height and that's it, which can be accessed with `getComputedStyle(document.body).getPropertyValue('--header-height')`.

https://github.com/user-attachments/assets/dee9248b-eafe-4263-aa00-089d10a0b635

rparrett:
Cool, thanks for looking into that. Documented in separate issue.

BenjaminBrienen (approved):

cart (approved):

PR #1909, opened 2024-12-05
Fix sidebar section highlighter

### LINKED ISSUE #1908: Table of contents in sidebar feels wonky
> I saw that mentioned in Discord… but no idea if there is an GH issue.
>
> I just checked, and it's working fine, but since we have a fixed header it feels weird (?). It probably gets fixed if we change the top part in `rootMargin` in the intersection observer. Probably needs the header height and that's it, which can be accessed with `getComputedStyle(document.body).getPropertyValue('--header-height')`.
>
> https://github.com/user-attachments/assets/dee9248b-eafe-4263-aa00-089d10a0b635

_Originally posted by @doup in https://github.com/bevyengine/bevy-website/issues/1907#issuecomment-2520945939_

doup:
Apply the header offset to `on-this-page.js` intersection observer. This way the sidebar will highlight the next section as soon as the old section gets below the header. [Previously it waited until exiting the viewport](https://github.com/bevyengine/bevy-website/pull/1907#issuecomment-2520945939).

Fixes #1908.

New behavior (header opacity reduced to 0.5 and yellow header **just for visualization**)

https://github.com/user-attachments/assets/e5b597e3-d13c-4609-a16f-82daa216aeb6

### DIFF
--- a/static/on-this-page.js
+++ b/static/on-this-page.js
@@ -18,13 +18,8 @@ let otp_state = new Map();
  * @param {string | HTMLElement} id_or_node 
  */
 function otp_set_active(id_or_node){
-  let id = "";
-  if(typeof id_or_node == "object"){
-    id = id_or_node.getAttribute("id");
-  } else {
-    id = id_or_node;
-  }
-  id = "#" + id;
+  let id = `#${id_or_node instanceof HTMLElement ? id_or_node.getAttribute("id") : id_or_node}`;
+
   document.querySelectorAll(".on-this-page a").forEach(a => {
     a.setAttribute("data-active", a.getAttribute("href") == id);
   });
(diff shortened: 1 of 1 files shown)


doup:
I couldn't resist, an easy fix. Hopefully improves the feeling. 🙈

rparrett:
I wonder if something else is going on as well -- this is what I'm seeing on my end with this PR:

https://github.com/user-attachments/assets/929fc3d7-47c6-4b02-99ed-8eea956f47bd

doup:
Hahaha, OK, that's really broken. It's not because of the header offset _feeling_.
My videos where in Firefox, but now I've checked in Chrome and Safari and I see the same issue.

doup:
@rparrett I think now is fixed. The sorting of the visible headings was broken, which caused to highlight the last visible heading, not the first one.

This is in Safari:

https://github.com/user-attachments/assets/d020b1fa-2a63-48c2-8b39-53d5ed1807f2

rparrett (approved):
Nice! Working well in my browsers.

cart (approved):
Works in firefox and chromium for me on linux. Big win. Thanks!

PR #1910, opened 2024-12-05
Enable `highlight-intrapage-link` sidebar highlight in News

doup:
Enable sidebar menu highlight on News. This uses the same mechanism as in the docs sidebar. Renamed JS file name to `highlight-intrapage-link` to better express what it does and differentiate from the BEM class (since now the JS part is used by two different BEM classes).

Note that the current implementation is still a little bit buggy but I think it's out of the scope of this PR. See [Page TOC sidebar highlight doesn't work when the section content is very long #1943](https://github.com/bevyengine/bevy-website/issues/1943).

Demo:

https://github.com/user-attachments/assets/ccac010b-dbe5-4b29-9ce5-f8890136e555


static/on-this-page.js
```diff
@@ -20,8 +20,8 @@ let otp_state = new Map();
 function otp_set_active(id_or_node){
   let id = `#${id_or_node instanceof HTMLElement ? id_or_node.getAttribute("id") : id_or_node}`;
 
-  document.querySelectorAll(".on-this-page a").forEach(a => {
-    a.setAttribute("data-active", a.getAttribute("href") == id);
+  document.querySelectorAll("[data-on-this-page] a").forEach(a => {
+    a.setAttribute("data-active", a.getAttribute("href").includes(id));
```
krunchington, line 24: This feels like it could have unintended side effects, if an `id` is a superset of another `id`

static/on-this-page.js
```diff
@@ -20,8 +20,8 @@ let otp_state = new Map();
 function otp_set_active(id_or_node){
   let id = `#${id_or_node instanceof HTMLElement ? id_or_node.getAttribute("id") : id_or_node}`;
 
-  document.querySelectorAll(".on-this-page a").forEach(a => {
-    a.setAttribute("data-active", a.getAttribute("href") == id);
+  document.querySelectorAll("[data-on-this-page] a").forEach(a => {
```
krunchington, line 23: I like the `data-on-this-page` switch to be able to enable this on multiple components.  However still being tied to the `a` tag is a little brittle.  It might be worth considering a `data-on-this-page-target` attribute to the links themselves?  Or maybe this is overkill abstraction.  Not highly convicted a change like that would be worth it but could be worth the conversation.

sass/components/_tree-menu.scss
```diff
@@ -57,6 +57,10 @@
         &, &:focus, &:active, &:hover, &:link, &:visited {
             color: var(--tree-menu-label-color);
         }
+
+        &[data-active=true] {
```
krunchington, line 61: Suggestion: it's a little more `js` but this feels like a good opportunity for a BEM modifier, e.g. `&--otp-active`

krunchington:
`id.includes` is the only thing I'd want to actually debate.  Other comments are nitpicks not blockers.

sass/components/_tree-menu.scss
doup: I think the way it is makes sense, in the sense that this is not specific to `on-this-page` BEM component (it's also used by `tree-menu` BEM) so it's OK that it doesn't use a CSS class to track the active state.

The problem is that we're calling the same way the `on-this-page` JS code and the BEM class, but these are two different things. I mean, we've two BEM components (`.tree-menu` & `.on-this-page`) that use the `on-this-page` JS.

I think I'll:
- rename `.on-this-page` BEM to `.toc`
- rename `on-this-page` JS to… `highlight-active-heading-link`?

static/on-this-page.js
doup: Makes sense, it's not that hard to change it.

static/on-this-page.js
doup: Uhmm… true, an edge case but I'll review it.

static/on-this-page.js
doup: Change of mind I don't see a scenario (that makes sense) where this would break.

sass/components/_tree-menu.scss
doup: I've only renamed the JS file: `highlight-intrapage-link.js`

static/on-this-page.js
doup: Fixed.

cart (approved):
I dig it!

PR #1912, opened 2024-12-05
:pencil2: Fix "the this" typo

homersimpsons:
Thank you for this amazing project. The blog is really interesting to read!

I found other typos earlier but didn't found this repository before and I cannot recall where they were.


rparrett (approved):
Looks like there's an opportunity to fix two more of the same `the this` typo over in https://github.com/bevyengine/bevy if you're motivated :)

mockersf (approved):

PR #1914, opened 2024-12-10
Use tabs for examples code

doup:
- Adds tabs in the example page (they also work with keyboard once focused)
- Adds `tabs` BEM _component_
- Adds `.wgsl` syntax highlighter. It's using https://github.com/relrelb/sublime-wgsl/tree/main which is an [outdated version of this](https://github.com/PolyMeilex/vscode-wgsl/blob/master/syntaxes/wgsl.tmLanguage.yml). If someone is willing to convert the VSCode version to the Sublime format apparently there is a way of doing it within Sublime, but I couldn't make it work.

(ignore the changing cursor, it's because of my screen recorder)

https://github.com/user-attachments/assets/0f91e339-8b5d-4a76-87fb-60c44ec559a1

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -26,5 +26,8 @@ taxonomies = [
 highlight_code = true
 highlight_theme = "css"
 
+# Load extra syntaxes (e.g. WGSL) for syntax highlighting
+extra_syntaxes_and_themes = ["syntaxes"]
+
 [extra]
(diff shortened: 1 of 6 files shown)


cart (approved):
Love it. Solid improvements all around.

PR #1918, opened 2024-12-11
Improve styling of the example/s graphics API callout

doup:
### Examples
- Use `callout` _component_ to style the web graphics API warning.
image
image

### Example
- Use `callout` _component_ to style the web graphics API warning.
- Add `callout--plain` modifier that hides the callout top border.
- Change link to it goes directly to the example and not the listing.

image
image

### DIFF
--- a/templates/examples-webgpu.html
+++ b/templates/examples-webgpu.html
@@ -5,8 +5,10 @@ These examples demonstrate how to use Bevy's features in a minimal, easy to unde
 Click an example below to run it in your browser (using WASM + WebGPU) and view the source code.
 You can also view these examples (and others) in the <a href="https://github.com/bevyengine/bevy/tree/main/examples#examples">Bevy repo</a>.
 <aside class="callout callout--warning">
-<h3 class="warning">Support Warning</h3>
-<p>WebGPU is currently only supported on Chrome starting with version 113, and only on desktop.
-If they don't work on your configuration, you can check the WebGL2 examples <a href="/examples">here</a>.</p>
+  <h3 class="warning">Support Warning</h3>
+  <p>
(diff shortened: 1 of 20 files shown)


mockersf:
if you're touching this callout, something that I would like is a way to switch between webgl2/webgpu on the same example without going through the index page

doup:
Makes sense, I've updated the example page link.

rparrett (approved):
Looks good. Love the new links.

templates/examples-webgpu.html
```diff
@@ -4,7 +4,11 @@
...
 Click an example below to run it in your browser (using WASM + WebGPU) and view the source code.
 You can also view these examples (and others) in the <a href="https://github.com/bevyengine/bevy/tree/main/examples#examples">Bevy repo</a>.
-<h2 class="warning">Support Warning</h2>
-WebGPU is currently only supported on Chrome starting with version 113, and only on desktop.
-If they don't work on your configuration, you can check the WebGL2 examples <a href="/examples">here</a>.
+<aside class="callout callout--warning">
+  <h3 class="warning">Support Warning</h3>
+  <p>
+    WebGPU is currently only supported on Chrome starting with version 113, and only on desktop.
+    If they don't work on your configuration, you can check the WebGL2 examples <a href="/examples">here</a>.
```
mockersf, line 11: for a followup PR: this is no longer true, WebGPU also works on some mobile browser, and on SafarI TP, and Firefox Nightly
We should probably link to something like https://caniuse.com/webgpu for a list or https://browserleaks.com/webgpu so that user can test their browser instead of keeping a list ourselves

mockersf (approved):

PR #1931, opened 2025-02-11
docs: call out `cargo run -p ci` in contributing section

JeanMertz:


mockersf:
Not a fan of having commands copy pasted in several places, and this is already mentioned here, with more informations: https://github.com/bevyengine/bevy-website/blob/7b1e16281ec8747a49dcd0605a27dda8ccdb360d/content/learn/contribute/helping-out/opening-pull-requests.md?plain=1#L49

And I don't really like "make sure", this is definitely not mandatory to run locally

JeanMertz:
> Not a fan of having commands copy pasted in several places, and this is already mentioned here, with more informations:
> 
> https://github.com/bevyengine/bevy-website/blob/7b1e16281ec8747a49dcd0605a27dda8ccdb360d/content/learn/contribute/helping-out/opening-pull-requests.md?plain=1#L49
> 
> And I don't really like "make sure", this is definitely not mandatory to run locally

Ah! I hadn't seen that part. To be honest, I got to https://bevyengine.org/learn/contribute/helping-out/how-you-can-help/#contributing-code, saw almost no text but a link to "opening pull request", and hadn't considered that would contain more details on how to contribute. In hindsight it makes more sense, but perhaps that could be made more clear, somehow?

As to the "make sure", it references to the fact that it helps you catch errors without having to wait for CI to do the validation for you. But I'm not a native English speaker, so happy to take any alternative suggestions (assuming we don't morph this into improving the link name, or closing this PR entirely).

alice-i-cecile:
Ah, I see: it's already in there :) Let's just close this out then.

PR #1933, opened 2025-02-18
Navigate release workflow branch links

guspan-tanadi:


mockersf:
this is not the correct fix, the script doesn't exist anymore and the link should be removed, and the place where it's used should be rephrased

PR #1934, opened 2025-02-18
Tidy up "past sponsors" markup

rparrett:
# Objective

Tidy up this section of the tera template.

The output should be identical, except without the space between names and the following comma.

## Solution

- Replace manual array filtering with `filter` filter
- Replace manual array index tracking with `loop`
- Get rid of pesky space between names and comma
- Replace spaces in names with non-breaking spaces so they don't get split onto two lines.

## Before
Screenshot 2025-02-18 at 7 02 21 AM

## After

Screenshot 2025-02-18 at 7 01 32 AM

### DIFF
--- a/templates/sponsors.html
+++ b/templates/sponsors.html
@@ -96,7 +96,8 @@
         <h2 class="past-donors-title">Past Donors</h2>
         <div class="past-donor">
         {% for donor in past_donors %}
-            {% if loop.last %}{{ donor.name }}{% else %}{{ donor.name }},{% endif %}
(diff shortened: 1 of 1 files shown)


axlitEels:
Could also be great to somehow replace all spaces within each sponsor's name with NBSPs

axlitEels:
`| replace(from=" ", to=" ")` after both `donor.name`s should do the job.
Github will wreck it, right?

mockersf (approved):

doup (approved):

PR #1204, opened 2024-05-27
The One True Search™

SIGSTACKFAULT:
Just barely works!

https://github.com/bevyengine/bevy-website/assets/14184826/a26e8dbf-9c9b-4b63-80c2-ce871d9832bc

in hindsight i have no clue why i left the inspector open.

**TODO:**

- [ ] assimilate assets search
- [ ] assimilate examples search
- [x] add keyboard shortcut
- [ ] show preview snippet in results
- [ ] prioritize the section of the site that you're currently in
- [ ] show something in the empty search dialog
- [ ] think of some other tips
- [x] make it look good in the mobile menu
  - [x] minor rework of mobile menu styles; now uses a vertical flexbox.
- [x] add close button

I will accept MRs on the SIGSTACKFAULT:one-true-search branch.

PR #1935, opened 2025-02-26
Add search functionality using Pagefind

doup:
This PR adds search functionality using [Pagefind](https://pagefind.app) Rust based library. The search index built by Pagefind is generated based on the final website HTML files. Because of that I had to change the deploy action; now building with Zola, creating the search index and deploying are different steps.

### ⚠️⚠️ Reviewer, please read this ⚠️⚠️
You'll need to **generate the Pagefind index,** follow `Search Index > Local Development` instructions:
https://github.com/doup/bevy-website/tree/pagefind-search?tab=readme-ov-file#local-development

Closes #1204.

image

image

image

image

image

---

## Deploy Settings

I had to use these settings to deploy with the new modified `deploy` action. Fine-grained PAT Permissions:

image

Also, had to add my branch `pagefind-search` to the repo `Settings > Environments > Configure "github-pages" > Deployment branches and tags`. So, **`main` probably needs to be there**.


reindex.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
krunchington, line 1: recommend adding `set -e` to the top of any shell script used as part of ci, so that silent failures don't end up with false positive builds.

sass/components/_search-category.scss
```diff
@@ -0,0 +1,67 @@
...
+    --search-category-bg-color: #953562;
+  }
+
+  &--news {
+    --search-category-bg-color: #2c5580;
+  }
+
+  &--contribute {
+    --search-category-bg-color: #78527e;
+  }
```
krunchington, line 41: If these are intended to be [BEM](https://getbem.com/) I think the convention would be `&__quickstart` since these would qualify as elements within the `.search-category` block.  The `&--active` below though, seems to follow the correct convention for modifiers.

krunchington: Hm, I see you've done the element convention on other stuff so maybe I'm misunderstanding the purpose of these classes and they are modifiers.

static/search.mjs
```diff
@@ -0,0 +1,651 @@
...
+        event.preventDefault();
+      }
+
+      if (event.code === "ArrowUp" && this.results) {
+        this.focusPrevious();
+        event.preventDefault();
+      }
+
+      // Open with `S`
+      if (event.code === "KeyS" && this.inputEl !== document.activeElement) {
```
krunchington, line 114: Are there any other text inputs on any of the pages this might break?

static/search.mjs
```diff
@@ -0,0 +1,651 @@
...
+
+    if (termEl) {
+      termEl.innerHTML = term;
+    }
+  }
+}
+
+class SearchCategories {
+  /** @private @readonly */
+  STORAGE_KEY = "bevy-search-categories";
```
krunchington, line 349: You can make this a private class property
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

Personally I don't love the syntax but it is widely supported these days by browsers.

reindex.sh
doup: This is for local dev, I'll delete it when the PR is ready. Which makes me remember that I need to document how to setup in local. I'll add few extra tasks in the PR description.

sass/components/_search-category.scss
doup: I don't agree, these are modifiers at the block level. Furthermore, the structure doesn't even have nesting, so using elements doesn't make much sense:

```html
<div class="search-category">Category Name</div>
```

It's similar to the button example in the introduction: https://getbem.com/introduction/#under-the-hood

static/search.mjs
doup: I forgot this was a thing, thanks! :-)

doup:
Thanks for the review @krunchington 🙏
I'll ask for a new one once I fix the issues and finish the CI part.

static/search.mjs
doup: I think I won't change this… :-( JSDoc/TS is not happy:

image

It already was verbose and it forces me to do this:

```js
class Example {
  /** @readonly @type {Pagefind} */
  #pagefind;

  constructor(
    /** @type {Pagefind} */ pagefind,
  ) {
    this.#pagefind = pagefind;
  }
}
```

krunchington: Ah bummer.  I get the value proposition of JSDoc but what an annoyance :(

templates/search.html
```diff
@@ -9,7 +9,7 @@
             <i class="icon icon--times"></i>
           </div>
         </div>
-        <div class="search__close button" data-search-close>Close</div>
+        <a class="search__close button" data-search-close href="">Close</a>
```
krunchington, line 12: Can you do this with a `tabindex` instead of an empty `href` ? feels more semantic that way I think.

krunchington: (this is a nitpick, I don't really think it's that important unless someone with stronger a11y sensibilities has an opinion here)

krunchington:
New changes look good!

krunchington (approved):

static/search.mjs
doup: Fixed in https://github.com/bevyengine/bevy-website/pull/1935/commits/286359c415b68280a392b77f724c8df9fdbac323

templates/search.html
doup: I've changed it to a `button`: https://github.com/bevyengine/bevy-website/pull/1935/commits/8fa9c536909fc20fc8b96afe11a939afcca7e115. Should be better, no? 🙃

krunchington: Yep, didn't even think of that but definitely better!

doup:
@BenjaminBrienen have you generated the Pagefind index? You'll need to follow these steps:
https://github.com/doup/bevy-website/tree/pagefind-search?tab=readme-ov-file#local-development

Sorry, my bad, I forgot to mention it… 😅

doup:
@alice-i-cecile can you add the "for adoption" label? I think this is still interesting specially now that new docs are almost ready. But I'm not willing to push this to the finish line.

static/search.mjs
```diff
@@ -0,0 +1,106 @@
...
+    subResultTplEl instanceof HTMLTemplateElement
+  ) {
+    try {
+      // @ts-ignore
+      const pagefind = await import("/pagefind/pagefind.js");
+      const filters = await pagefind.filters();
+      const categories = new SearchCategories(
+        Object.keys(filters.category),
+        [
+          "Quick Start",
```
cookie1170, line 66: i feel like quick start should be replaced by the whole "learn" category, which would also include the book once that's out

cookie1170:
image

code snippets in the results don't show up very nicely since they aren't in a monospace font and the newlines are stripped

fallible-algebra (approved):
The "download pagefind and run it in the directory" part is a bit annoying but the functionality is great + the structure / styling is reasonable. It's not within scope to expect this PR to to also introduce a buildscript for local development. LGTM.

NthTensor (approved):
The install process is gross but it works.

doup:
Oh, that was unexpected. Thanks for the reviews 🚀

static/search.mjs
doup: Once the book is ready is a matter of updating [the categories](https://github.com/bevyengine/bevy-website/blob/main/templates/macros/pagefind.html) and `search.mjs`

fallible-algebra:
> Oh, that was unexpected. Thanks for the reviews 🚀

Sorry for taking so long! Bevy Unconf power :)

PR #1942, opened 2025-03-07
Add S-Ready-For-Implementation to triage docs

krunchington:
While adding the S-Nominated-To-Close tag in another PR, I noticed that S-Ready-For-Implementation was one of the most common labels.  While it's probably pretty straightforward without explanation, I felt this might be a good call to action (which is why I added it to the top of the list).


alice-i-cecile (approved):
Also merging as trivial :)

PR #1952, opened 2025-03-18
Add Asset title text outline

### LINKED ISSUE #1951: Asset title unreadable on light background
Image

hankjordan:
Screenshot_2025-03-17_23-56-12

Closes #1951

### DIFF
--- a/sass/components/_asset-card.scss
+++ b/sass/components/_asset-card.scss
@@ -44,6 +44,7 @@ $asset-card-padding: 0.4rem;
     font-size: 1.2rem;
     background-color: var(--asset-card-title-bg-color);
     color: var(--asset-card-title-color);
+    text-shadow: -1px 0 black, 0 1px black, 1px 0 black, 0 -1px black;
   }
 
   &__description {


rparrett:
I think we should figure out why the 50% opacity black background for the title div is not working.

doup:
I think that this PR should be considered on its own as a way to improve assets legibility (I personally don't think it's necessary, but maybe other people think otherwise) not as a solution for #1951. #1951 has a wider impact than the issue we've noticed in the assets and the correct solution is #1953.

sass/components/_asset-card.scss
```diff
@@ -44,6 +44,7 @@ $asset-card-padding: 0.4rem;
     font-size: 1.2rem;
     background-color: var(--asset-card-title-bg-color);
     color: var(--asset-card-title-color);
+    text-shadow: -1px 0 black, 0 1px black, 1px 0 black, 0 -1px black;
```
doup, line 47: In general, colors shouldn't be used directly, define a `--asset-cart-title-text-shadow-color` (verbose… 🤮) in `_vars.scss` in its corresponding section. The reason is that this will allow to change the shadow color for light mode.

@rparrett FYI, in case you see a similar issue in other PRs.

doup: Anyway, no need to do the change unless we agree that we want the text shadow.

PR #1953, opened 2025-03-18
Remove CSS Relative Colors usage

### LINKED ISSUE #1937: Remove `unquote` calls from SASS files
#1771 & #1935 use [relative colors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Relative_colors) with CSS vars… but a workaround is needed to appease the SASS compiler. Remove the `unquote` calls once Zola allows to use these without `unquote`. See: https://github.com/connorskees/grass/issues/107

### LINKED ISSUE #1951: Asset title unreadable on light background

doup:
#1771 introduced the usage of [CSS Relative Colors](https://caniuse.com/css-relative-colors) but this feature is not **widely** available yet (even if it's part of the [2024 Baseline](https://web-platform-dx.github.io/web-features/)). This PR moves the relative colors computation from CSS to SASS.

Once CSS Relative Colors has **wide support** this PR should be reverted.

Fixes #1951
Fixes #1937

### DIFF
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -1,3 +1,5 @@
+@use 'sass:color';
+
 // Typography
 $size-body: 16px; // Usual browser default size
 $size-body-mobile: 14px; // Size for mobile
@@ -39,8 +41,12 @@ $border-radius: 10px;
 // the `sass:color` package to derive new colors. If a color is not going to be
 // used to derive a new color, then CSS vars should be used instead.
 //
-// See PR:
+// See PR: https://github.com/bevyengine/bevy-website/pull/1953
 // See Relative Colors support: https://caniuse.com/css-relative-colors
+$color-neutral-17: #2c2c2d;
+$color-neutral-93: #ececec;
+$color-black: #000;
+$color-white: #fff;
 
 // CSS Vars
 :root {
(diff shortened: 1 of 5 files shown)


sass/_vars.scss
```diff
@@ -32,6 +32,23 @@ $mobile-menu-width: 300px;
 $content-top-margin: 30px;
 $border-radius: 10px;
 
+// Colors
+// If you're reading this in late 2025 (or later), check the PR because these
```
alice-i-cecile, line 36:
```suggestion
// If you're reading this in late 2025 (or later), check PR #1953 because these
```

kristoff3r (approved):
Tested on a modern browser and the asset page works, I don't have an easy way of testing if it actually fixes the issue.

doup:
image
(couldn't resist, I've played Titanfall 2 recently…)

Anyway, I didn't test it on an older browser, I just double checked that the SASS equivalent `rgba()` calls were using the correct color and alpha. I think that's enough as we're quite sure from the user that reported the issue that _CSS Relative Colors_ is the problem.

bushrat011899 (approved):
Makes sense. Hopefully it can be reverted _eventually_.

PR #2006, opened 2025-03-25
Remove empty migration guides

ElliottjPierce:
Lots of migration guides just contained "n/a". Those have been removed now. We should consider checking for `M-needs-migration-guide` before creating a migration guide for it in the generator.


alice-i-cecile (approved):
😔 Thankfully this is the last time we're doing this.

PR #2052, opened 2025-04-01
Write migration guide for 18451

ElliottjPierce:
[18451](https://github.com/bevyengine/bevy/pull/18451) was (hopefully) the last breaking change of 0.16!

It was created after most migration guides were auto-generated, but this PR adds the migration guide for it.


release-content/0.16/migration-guides/18451_Get_names_of_queued_components.md
```diff
@@ -0,0 +1,3 @@
+`Components::get_name` now returns `Option<Cow<'_, str>` instead of `Option<&str>`. This is because it now returns results for queued components. If that behavior is not desired, or you know the component is not queued, you can use `components.get_info().map(ComponentInfo::name)` instead.
```
BD103, line 1: > This is because it now returns results for queued components.

(Nit) I'm not sure I understand what this means. Could you elaborate slightly on what queued components are, and why it is beneficial for them to be supported in `get_name()`?

BD103 (approved):

release-content/0.16/migration-guides/18451_Get_names_of_queued_components.md
ElliottjPierce: You're right. This is a new thing, so we need to communicate that to users. New version is up now.

BD103: Looks good now, thank you!

PR #2058, opened 2025-04-01
Improve clarity of tracing log macros migration

### LINKED ISSUE #2055: Mention that the `bevy_log` feature is needed for libraries to call `debug!` and friends
The [relevant section](https://bevyengine.org/learn/migration-guides/0-15-to-0-16/#utils) already technically mentions this, but it's a bit obscure. I'd wager that quite a few plugin authors use `debug!`, `warn!`, `info!`, etc., and will search for those when encountering compiler errors. The guide however talks about `tracing` in general.

alice-i-cecile:
Fixes #2055.


bushrat011899 (approved):
Makes sense!

janhohenheim (approved):

PR #2065, opened 2025-04-02
HashMap/HashSet migration guide

### LINKED ISSUE #2045: Mention that users need to change `HashMap` and `HashSet` initializations in 0.16
[Discord discussion](https://discordapp.com/channels/691052431525675048/692572690833473578/1355925049890177144)

Methods that are affected:
- `new` -> `default`
- `from` -> `from_iter`
- possibly others? Note from @Victoronz: "To find every DefaultHashBuilder specific impl, the docs of hashbrown should be easily searchable"

## Examples

```rust
use bevy::platform_support::collections::HashMap;

struct SomeHashMap(HashMap<u32, u32>);

fn main() {
    // This fails to compile (same for HashSet):
    let some_hash_map = SomeHashMap(HashMap::new());

    // This works fine (same for HashSet):
    let some_hash_map = SomeHashMap(HashMap::default());
}
```
And
```rust
use bevy::platform_support::collections::HashSet;

struct SomeHashSet(HashSet<u32>);

fn main() {
    // This fails to compile:
    let some_hash_set = SomeHashSet(HashSet::from([1, 2, 3]));

    // This works fine:
    let some_hash_set = SomeHashSet(HashSet::from_iter([1, 2, 3]));
}
```

alice-i-cecile:
The PR that changed this upstream was https://github.com/rust-lang/hashbrown/pull/563, which seems like a sensible choice for them.

We need to:

a) file an issue with `hashbrown` requesting changes to make it more feasible to use an alternate default hasher
b) file an issue on `bevyengine/bevy` to decide what to do for 0.16 and going forward

Closes #2045.

### DIFF
--- a/release-content/0.16/migration-guides/hash_explicit_types.md
+++ b/release-content/0.16/migration-guides/hash_explicit_types.md
@@ -1,0 +1,25 @@
+`hashbrown`'s default hasher [has changed in `hashbrown 0.15`](https://github.com/rust-lang/hashbrown/pull/563), resulting in serious regressions in type inference and method availability for Bevy users of `HashSet` and `HashMap`.
+We have updated versions to avoid compiling duplicate copies of the crate, as Bevy's own dependencies have already updated.
+
+Bevy uses a different (faster on average, but slower in the face of malicious inputs) than `hashbrown`'s new default hasher.
+Unfortunately, many methods are only implemented for the slower `HashMap` and `HashSet` types with the slower `RandomState` type as a generic argument, rather than the Bevy's preferred `FixedHasher` type.
+
+For example, the `HashMap::new` method will produce a type that uses the incorrect hasher, leading to errors like:
+
+```sh
+error[E0308]: mismatched types
+   --> crates/bevy_plugin/src/line_provider/asset_provider.rs:160:14
+    |
+160 |         Self(HashMap::new())
+    |         ---- ^^^^^^^^^^^^^^ expected `HashMap<&str, UntypedHandle, FixedHasher>`, found `HashMap<_, _>`
+    |         |
+    |         arguments to this function are incorrect
+    |
+    = note: expected struct `hashbrown::map::HashMap<&'static str, bevy::prelude::UntypedHandle, FixedHasher>`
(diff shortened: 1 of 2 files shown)


release-content/0.16/migration-guides/hash_explicit_types.md
```diff
@@ -0,0 +1,25 @@
+`hashbrown`'s default hasher [has changed in `hashbrown 0.15`](https://github.com/rust-lang/hashbrown/pull/563), resulting in serious regressions in type inference and method availability for Bevy users of `HashSet` and `HashMap`.
+We have updated versions to avoid compiling duplicate copies of the crate, as Bevy's own dependencies have already updated.
+
+Bevy uses a different (faster on average, but slower in the face of malicious inputs) than `hashbrown`'s new default hasher.
```
tbillington, line 4: Looks like a word or two is missing after different

release-content/0.16/migration-guides/hash_explicit_types.md
```diff
@@ -0,0 +1,25 @@
...
+    |         arguments to this function are incorrect
+    |
+    = note: expected struct `hashbrown::map::HashMap<&'static str, bevy::prelude::UntypedHandle, FixedHasher>`
+               found struct `hashbrown::map::HashMap<_, _, bevy::bevy_platform_support::hash::RandomState>`
+```
+
+`HashMap::new` can be replaced with `HashMap::default`, which is flexible in the correct way.
+Other methods, such as `with_capacity` may be replaced with the `with_capacity_and_hasher` method.
+
+We are understand your frustrations with this regression, and are actively discussing solutions to improve this going forward.
```
tbillington, line 25: Grammer is a bit funny.

release-content/0.16/migration-guides/hash_explicit_types.md
alice-i-cecile:
```suggestion
Bevy uses a different hasher (faster on average, but slower in the face of malicious inputs) than `hashbrown`'s new default hasher.
```

alice-i-cecile:
FYI, I don't think we should merge this, and should instead solve the regression per the linked issue. I decided to write this up to lay out how gnarly the problem is and to have a fallback in case we decide to 🚀 anyways.

PR #2070, opened 2025-04-03
Replace deprecated `Parent` with `ChildOf` in relationships migration guide

StudioLE:
The `Parent` component has been renamed to `ChildOf` but the migration guide `17398_Relationships_nonfragmenting_onetomany.md` still refers to `Parent`.

Relevant: https://github.com/bevyengine/bevy-website/blob/bf214ec9efda4b5e73fd69f6e81f26c62ea60c89/release-content/0.16/migration-guides/17427_Parent__ChildOf.md?plain=1#L1


alice-i-cecile (changes requested):
This is back to a tuple struct again!

alice-i-cecile (approved):

PR #2072, opened 2025-04-04
Write release notes for occlusion culling

### LINKED ISSUE #2056: Write release notes for occlusion culling
@JMS55 mentioned that they thought this was worth calling out seperately from the other rendering perf improvements, and I agree! This is an optimization that makes sense to non-rendering experts and is relatively high impact for some users.

https://github.com/bevyengine/bevy/pull/17934 handles this for deferred, but apparently another PR (https://github.com/bevyengine/bevy/pull/17413) did it already for non-deferred. https://github.com/bevyengine/bevy/pull/12898 did this for meshlets.

https://github.com/bevyengine/bevy/pull/12889 was precursor work that might have useful context.

alice-i-cecile:
Fixes #2056.

I'm at the edge of my rendering knowledge here: this was mostly written based off of @pcwalton's lovely PR description and some brief research to understand what that meant. Please double-check technical details and don't hesitate to correct me!

### DIFF
--- a/release-content/0.16/release-notes/_release-notes.toml
+++ b/release-content/0.16/release-notes/_release-notes.toml
@@ -73,9 +73,9 @@ file_name = "16427_Use_multidraw_for_opaque_meshes_when_GPU_culling_is_in_use.md
 
 [[release_notes]]
 title = "Occlusion culling"
-authors = ["@pcwalton"]
+authors = ["@JMS55", "@pcwalton"]
 contributors = []
-prs = [17413, 17934]
+prs = [12898, 17413, 17934]
 file_name = "occlusion_culling.md"
 
 [[release_notes]]
--- a/release-content/0.16/release-notes/_release-notes.toml
+++ b/release-content/0.16/release-notes/_release-notes.toml
@@ -144,13 +144,6 @@ prs = [17559]
 file_name = "17559_Basic_UI_text_shadows.md"
 
 [[release_notes]]
-title = "Multiple box shadow support"
-authors = ["@ickshonpe"]
-contributors = ["@alice-i-cecile"]
-prs = [16502]
-file_name = "16502_Multiple_box_shadow_support.md"
-
-[[release_notes]]
 title = "Add \"bevy_input_focus\" crate."
 authors = ["@viridia", "@alice-i-cecile"]
 contributors = []
--- a/release-content/0.16/release-notes/_release-notes.toml
+++ b/release-content/0.16/release-notes/_release-notes.toml
@@ -72,6 +72,13 @@ prs = [16427]
 file_name = "16427_Use_multidraw_for_opaque_meshes_when_GPU_culling_is_in_use.md"
 
 [[release_notes]]
+title = "Occlusion culling"
+authors = ["@pcwalton"]
+contributors = []
+prs = [17413, 17934]
+file_name = "occlusion_culling.md"
+
+[[release_notes]]
 title = "Add a bindless mode to `AsBindGroup`."
 authors = ["@pcwalton"]
 contributors = []
--- a/release-content/0.16/release-notes/_release-notes.toml
+++ b/release-content/0.16/release-notes/_release-notes.toml
@@ -73,7 +73,7 @@ file_name = "16427_Use_multidraw_for_opaque_meshes_when_GPU_culling_is_in_use.md
 
 [[release_notes]]
 title = "Occlusion culling"
-authors = ["@JMS55", "@pcwalton"]
+authors = ["@pcwalton", "@JMS55"]
 contributors = []
 prs = [12755, 17413, 17934, 17951]
 file_name = "occlusion_culling.md"
--- a/release-content/0.16/release-notes/_release-notes.toml
+++ b/release-content/0.16/release-notes/_release-notes.toml
@@ -75,7 +75,7 @@ file_name = "16427_Use_multidraw_for_opaque_meshes_when_GPU_culling_is_in_use.md
 title = "Occlusion culling"
 authors = ["@JMS55", "@pcwalton"]
 contributors = []
-prs = [12898, 17413, 17934, 17951]
+prs = [12755, 17413, 17934, 17951]
 file_name = "occlusion_culling.md"
 
 [[release_notes]]
--- a/release-content/0.16/release-notes/_release-notes.toml
+++ b/release-content/0.16/release-notes/_release-notes.toml
@@ -75,7 +75,7 @@ file_name = "16427_Use_multidraw_for_opaque_meshes_when_GPU_culling_is_in_use.md
 title = "Occlusion culling"
 authors = ["@JMS55", "@pcwalton"]
 contributors = []
-prs = [12898, 17413, 17934]
+prs = [12898, 17413, 17934, 17951]
 file_name = "occlusion_culling.md"
 
 [[release_notes]]
--- a/release-content/0.16/release-notes/occlusion_culling.md
+++ b/release-content/0.16/release-notes/occlusion_culling.md
@@ -1,4 +1,4 @@
-When it comes to pushing hardware to support larger levels and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
+When it comes to pushing hardware to support larger scenes and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
 These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to draw the thing despite its pixels later being overwritten completely.
 
 **Occlusion culling** is the idea that we don't need to draw something that's completely blocked by other opaque objects,
@@ -6,12 +6,12 @@ from the perspective of the camera.
 That makes sense! We don't need to draw a car parked on the other side of a building, even if it's within the range used for fustrum culling.
 
 Before you worry too much: we're already checking for this category of wasted work in some form, via the use of a **depth prepass**.
-A depth prepass renders a simplified version of the scene, recording only the depth (distance from the camera) of each object into the 2-dimensional depth buffer.
+A depth prepass renders the scene with a simplified shader, recording only the depth (distance from the camera) of each object into the 2-dimensional depth buffer.
 Then, during the more expensive main pass, objects that ended up behind something else during this simpler calculation can be skipped.
 Because of this approach, we can avoid most fragment shading costs (dominated by textures and lighting) for occluded objects, but the vertex shading overhead is still present, in addition to the inherent cost of this fragment testing process.
 
 We can do better than that.
-As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering, because it works well with the GPU-driven rendering architecture (cold specialization, retained bins) that we've established during this cycle!
+As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering, because it works well with the GPU-driven rendering architecture that we've established during this cycle!
 
 For now, this feature is marked as experimental, due to known precision issues that can mark meshes as occluded even when they're not.
 In practice, we're not convinced that this is a serious concern, so please let us know how it goes!
--- a/release-content/0.16/release-notes/occlusion_culling.md
+++ b/release-content/0.16/release-notes/occlusion_culling.md
@@ -1,9 +1,9 @@
 When it comes to pushing hardware to support larger scenes and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
-These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to draw the thing despite its pixels later being overwritten completely.
+These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to actually perform the draw.
 
 **Occlusion culling** is the idea that we don't need to draw something that's completely blocked by other opaque objects,
 from the perspective of the camera.
-That makes sense! We don't need to draw a car parked on the other side of a building, even if it's within the range used for fustrum culling.
+That makes sense! We don't need to draw a person hidden behind a wall, even if they're within the range used for frustrum culling.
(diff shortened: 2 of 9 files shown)


JMS55:
I would appreciate a small mention that this was already written for / is already part of virtual geometry, and that the new part is that it's now also integrated into the standard mesh renderer.

release-content/0.16/release-notes/_release-notes.toml
```diff
@@ -71,6 +71,13 @@ contributors = []
 prs = [16427]
 file_name = "16427_Use_multidraw_for_opaque_meshes_when_GPU_culling_is_in_use.md"
 
+[[release_notes]]
+title = "Occlusion culling"
+authors = ["@JMS55", "@pcwalton"]
```
alice-i-cecile, line 76: Listed in chronological order here.

release-content/0.16/release-notes/_release-notes.toml
```diff
@@ -71,6 +71,13 @@ contributors = []
...
+contributors = []
+prs = [12898, 17413, 17934, 17951]
```
alice-i-cecile, line 78: 12898 is not new this cycle, but is IMO a vital part of the story here. Ordinarily I wouldn't cite previous work that laid foundations, but this is an adaption of an existing in-Bevy design to a new domain.

Both of those are hard, useful and worth celebrating, and occlusion culling for meshlets was not previously explicitly called out as an accomplishment.

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+
+Before you worry too much: we're already checking for this category of wasted work in some form, via the use of a **depth prepass**.
+A depth prepass renders a simplified version of the scene, recording only the depth (distance from the camera) of each object into the 2-dimensional depth buffer.
+We can check the resulting depth buffer, count how many pixels are associated with each object, and then discard objects whose pixels were completely overwritten.
+As a result, we can avoid most fragment shading costs (dominated by textures and lighting) for occluded objects, but the vertex shading overhead is still present, in addition to the inherent cost of this fragment testing process.
+
+We can do better than that.
+As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering, because it works well with the GPU-driven rendering architecture (cold specialization, retained bins) that we've established during this cycle!
```
alice-i-cecile, line 14: I've cited 17413 here because the PR description does a great job explaining the gory technical details for curious rendering engineers.

release-content/0.16/release-notes/_release-notes.toml
JMS55: So actually occlusion culling was first added in https://github.com/bevyengine/bevy/pull/12755 (along with the rest of the initial virtual geometry renderer, it was a very very large PR). And then there were a couple of bugfix PRs later on, and then 12898 which switched the method we used for keeping track of what was visible last frame or not.

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
+When it comes to pushing hardware to support larger levels and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
+These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to draw the thing despite its pixels later being overwritten completely.
+
+**Occlusion culling** is the idea that we don't need to draw something that's completely blocked by other opaque objects,
+from the perspective of the camera.
+That makes sense! We don't need to draw a car parked on the other side of a building (even if it's within the range used for fustrum culling), and we *never* need to draw all of the teeth inside of a character's closed mouth.
```
IceSentry, line 6: I don't know if the teeth mentioned here is the best example. Most of the time, they will just be part of the character mesh and unless we are talking about meshlets, it's generally the entire character mesh that is completely visible or not.

Maybe something about vegetation on the other side of a mountain? Or trees occluding other trees in a dense forest?

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+
+Before you worry too much: we're already checking for this category of wasted work in some form, via the use of a **depth prepass**.
+A depth prepass renders a simplified version of the scene, recording only the depth (distance from the camera) of each object into the 2-dimensional depth buffer.
```
IceSentry, line 9: Saying "a simplified version of the scene" kinda implies to me that it uses simplified meshes, but that's not what's happening. I think it would be more accurate and less misleading to only mention the depth only part without saying it's a simplified scene. Maybe saying it's a simplified shader?

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+We can check the resulting depth buffer, count how many pixels are associated with each object, and then discard objects whose pixels were completely overwritten.
```
IceSentry, line 10: That's not really what the depth prepass is doing. We don't do any kind of pixel counting (unless I'm misunderstanding what you mean by that). What it does is record the top most fragment for each fragment on the screen, then during the main pass we can use that information and avoid any work that would be behind that top level fragment.

release-content/0.16/release-notes/occlusion_culling.md
IceSentry, line 14: Cold specialization isn't related to GPU driven. It's an important part optimization for the renderer but it's relevant to both the cpu driven and gpu driven path. Same for retained bin actually. I would simply remove that parenthesis to be honest.

IceSentry:
```suggestion
As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering, because it works well with the GPU-driven rendering architecture that we've established during this cycle!
```

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+As a result, we can avoid most fragment shading costs (dominated by textures and lighting) for occluded objects, but the vertex shading overhead is still present, in addition to the inherent cost of this fragment testing process.
+
+We can do better than that.
+As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering, because it works well with the GPU-driven rendering architecture (cold specialization, retained bins) that we've established during this cycle!
+
+For now, this feature is marked as experimental, due to known precision issues that can mark meshes as occluded even when they're not.
```
IceSentry, line 16: Maybe mention that this could result in pop in. It might be easier to understand the tradeoff for a lot of people.

IceSentry (changes requested):
Overall this looks good, I added a few nitpicks for some things that I think should be clarified.

release-content/0.16/release-notes/occlusion_culling.md
JMS55: It's not really pop in, it'll be meshes not rendering when they're supposed to.

release-content/0.16/release-notes/occlusion_culling.md
superdump: I don’t think of it as the ’topmost’ but rather the closest opaque fragment to the camera along the ray through the centre of that fragment.

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
+When it comes to pushing hardware to support larger levels and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
```
superdump, line 1: Swap levels for scenes, given that’s what we call them?

release-content/0.16/release-notes/occlusion_culling.md
IceSentry: Sure, but that will just last a few frames and it will eventually get rendered again which will cause pop in right?

release-content/0.16/release-notes/occlusion_culling.md
alice-i-cecile:
```suggestion
A depth prepass renders the scene with a simplified shader, recording only the depth (distance from the camera) of each object into the 2-dimensional depth buffer.
```

alice-i-cecile:
Thanks for the feedback: it's important to get the details right. I've done my best to incorporate your explanations, but please take another look before we merge this in.

release-content/0.16/release-notes/_release-notes.toml
JMS55: pcwalton should go first imo, I didn't work on any of the actual parts of this in 0.16.

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
+When it comes to pushing hardware to support larger scenes and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
+These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to draw the thing despite its pixels later being overwritten completely.
```
JMS55, line 2:
```suggestion
These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to actually perform the draw.
```

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Occlusion culling** is the idea that we don't need to draw something that's completely blocked by other opaque objects,
+from the perspective of the camera.
+That makes sense! We don't need to draw a car parked on the other side of a building, even if it's within the range used for fustrum culling.
```
JMS55, line 6:
```suggestion
That makes sense! We don't need to draw a person hidden standing a wall, even if they're within the range used for frustum culling.
```

release-content/0.16/release-notes/occlusion_culling.md
JMS55: Not entirely sure, but don't think so. This is about over-agressive culling. If you shift the camera it might render again, though.

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+
+For now, this feature is marked as experimental, due to known precision issues that can mark meshes as occluded even when they're not.
+In practice, we're not convinced that this is a serious concern, so please let us know how it goes!
+To try out the new mesh occlusion culling, add the [`DepthPrepass`] and [`OcclusionCulling`] components to your camera.
+
+As we said above, the check has to be faster than simply drawing the thing: occlusion culling won't be faster on all scenes.
+Small scenes, or those using simpler non-PBR rendering are particularly likely to be slower with occlusion culling turned on.
+Like always: you need to measure your performance to improve it.
+
+If you're a rendering engineer who'd like to help us resolve these precision issues and stabilize this feature, we're looking to learn from Hans-Kristian Arntzen design in [Granite].
```
JMS55, line 24:
```suggestion
If you're a rendering engineer who'd like to help us resolve these precision issues and stabilize this feature, we're looking to borrow from Hans-Kristian Arntzen's design in [Granite].
```

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
...
+from the perspective of the camera.
+That makes sense! We don't need to draw a car parked on the other side of a building, even if it's within the range used for fustrum culling.
+
+Before you worry too much: we're already checking for this category of wasted work in some form, via the use of a **depth prepass**.
+A depth prepass renders the scene with a simplified shader, recording only the depth (distance from the camera) of each object into the 2-dimensional depth buffer.
+Then, during the more expensive main pass, objects that ended up behind something else during this simpler calculation can be skipped.
+Because of this approach, we can avoid most fragment shading costs (dominated by textures and lighting) for occluded objects, but the vertex shading overhead is still present, in addition to the inherent cost of this fragment testing process.
+
+We can do better than that.
+As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering, because it works well with the GPU-driven rendering architecture that we've established during this cycle!
```
JMS55, line 14:
```suggestion
As [PR #17413] lays out, we're adopting the modern two-phase occlusion culling (in contrast to a traditional potentially visible sets design) already used by our virtual geometry rendering system, because it works well with the GPU-driven rendering architecture that we've established during this cycle!
```

JMS55 (changes requested):

release-content/0.16/release-notes/occlusion_culling.md
JMS55: person hidden behind a wall*

release-content/0.16/release-notes/_release-notes.toml
alice-i-cecile, line 76:
```suggestion
authors = ["@pcwalton", "@JMS55"]
```

release-content/0.16/release-notes/occlusion_culling.md
alice-i-cecile:
```suggestion
That makes sense! We don't need to draw a person hidden behind a wall, even if it's within the range used for frustrum culling.
```

release-content/0.16/release-notes/occlusion_culling.md
```diff
@@ -0,0 +1,32 @@
+When it comes to pushing hardware to support larger scenes and more detailed meshes, "simply draw less stuff" is a tried-and-true strategy.
+These techniques are collectively referred to as "culling": we can save work whenever we can determine "this doesn't need to be drawn" for cheaper than it would cost to actually perform the draw.
+
+**Occlusion culling** is the idea that we don't need to draw something that's completely blocked by other opaque objects,
+from the perspective of the camera.
+That makes sense! We don't need to draw a person hidden behind a wall, even if they're within the range used for frustrum culling.
```
alice-i-cecile, line 6:
```suggestion
That makes sense! We don't need to draw a person hidden behind a wall, even if they're within the range used for frustum culling.
```

IceSentry (approved):

PR #2093, opened 2025-04-09
Add no_std to migration guide highlights

alice-i-cecile:
After chewing on it, this stood out to me as one of the more important / common breaking changes of the 0.16 cycle, so I figured we can declare it upfront.


bushrat011899 (approved):
I agree, relocation's are a pain for compiler errors.

PR #2098, opened 2025-04-10
Show example explanation

doup:
Starting in 0.16 the [example code documentation will be exposed as markdown content](https://github.com/bevyengine/bevy/pull/18191). We need to update the example template to show that content.

Notes:
- This PR shows the content conditionally to be compatible with 0.15, once 0.16 is released we can remove the conditionals.
- Eventually this page could use all the space available and have a better layout for desktop. E.g. maybe the explanation could be on a sidebar.

image

---

Alternatively I can change this PR to put the content above the tabs, like so:

image


alice-i-cecile (approved):

cart:
I think we should show the example code by default / display the explanation above the code, although I think we need to adjust the styling, as the line height feels too big and the alignment feels off in the screenshot above.

doup:
I've changed it so the explanation is above the code. There is some horizontal padding so different texts align.

I won't tweak the line-height since the explanation is using the same `media-content` style as in the docs/news and I think those look _good enough_. The thing is that in this case these are two paragraphs, and that's why it looks a little bit weird. As we extend/improve the content it will look much better.

image

janhohenheim (approved):

PR #2102, opened 2025-04-14
Update and rename EntityHashSet 0.16 migration section

guyguy2001:
Solves https://github.com/bevyengine/bevy/issues/18841

I also renamed the md file, since the name seems like a mistake to me.

I didn't test the md - should I compile the docs locally? Or is a review enough?

The 0.16 doesn't get any rust analyzer errors and go to definition works for it, and the 0.15 version compiles. (I didn't have a compiling 0.16 projects at hand)

### DIFF
--- a/release-content/0.16/migration-guides/17447_Dont_reexport_EntityHashSet_and_EntityHashMap.md
+++ b/release-content/0.16/migration-guides/17447_Dont_reexport_EntityHashSet_and_EntityHashMap.md
@@ -2,8 +2,8 @@
 
 ```rust
 // 0.15
-use bevy::ecs::entity::{HashSet, HashMap};
+use bevy::ecs::entity::{EntityHashSet, EntityHashMap};
 
 // 0.16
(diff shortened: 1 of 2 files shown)


guyguy2001:
Also, I never used these structs before (I'm toying with upgrading bevy_tnua to 0.16, but I never touched it before) - so I may have missed important context in this migration tip, which might make my change wrong (although the new code I'm suggesting doesn't get any rust-analyzer error, and go to definition works on it, so I'm assuming it's right).

JaySpruce:
This correction is correct for bevyengine/bevy#17447, but it's out of date for 0.16 because the re-exports were re-introduced by bevyengine/bevy#18413. There's no longer any change from a user's perspective going from 0.15.3 to 0.16, so this migration guide ought to just be removed

guyguy2001:
Updated, and tested locally that the paragraph was indeed missing - was surprisingly simple.

And the 0.15 example indeed compiles for me on 0.16.0 rc 5.

JaySpruce (approved):

PR #2085, opened 2025-04-08
Release notes for decals

alice-i-cecile:
Fixes #1980. No image yet: none of the existing ones properly show off this feature.

PR #2106, opened 2025-04-22
Adopted Decals Release Notes 0.16

ChristopherBiscardi:
adopted version of #2085 that includes images and some adjustments based on feedback.


doup (approved):
LGTM, just be aware of the 1.6mb `decals.png` image, I would change it to jpg/webp.

doup:
AVIF sounds great, amazing compression ratio.

Be aware though that [Baseline](https://web-platform-dx.github.io/web-features/) discerns between newly and widely available. AVIF is still considered "newly" available… although it's been like that for a while: since January 2024. I mention this because we had a [regression report](https://github.com/bevyengine/bevy-website/issues/1951) because I used a "newly available" [CSS feature](https://caniuse.com/css-relative-colors) and had to [change it later on](https://github.com/bevyengine/bevy-website/pull/1953). I wasn't aware of this distinction and bite me later on… and it's just a 4% user support difference (although AVIF has been available for muuuch longer time). 😩

Buuut, if you're already using it on _This Week in Bevy_ and no one has complained… I would consider it seriously. It's 53% widely available, half-way there. 🙃

ChristopherBiscardi:
> AVIF is still considered "newly" available

Its worth examining the browser support behind this. The reason avif is "newly" available is that Edge gained support in January 2024. Other evergreen browsers have supported avif still images since 2020, 2021, and 2022. So its a pretty safe bet at this point. The set of "unsupported users" is basically "Edge users who haven't upgraded in over a year".

> It's 53% widely available, half-way there

I'm not sure where you're getting 53% from. Is that the 1 year left for Edge to trigger the "widely available" 2.5 year mark? If so, the 53% number is pretty misleading. All other browsers already pass the 2.5 year mark. Also worth noting that 2.5 years is like, enterprise software levels of browser support in my experience. I actually don't know of any consumer websites that use that length of time for their browser support goals.

> Buuut, if you're already using it on This Week in Bevy and no one has complained… I would consider it seriously.

Yeah I think the support here is well beyond usable tbh. One of the most well supported new features that exists at the moment.

---

This might be getting a little off topic but it also seems we haven't decided on a browser support policy so IMO: I do think supporting current-version evergreen browsers is a perfectly fine target for an open-source volunteer-heavy project.

So when it comes to web-related issues like https://github.com/bevyengine/bevy-website/issues/1951 , no changes should be made to solve an issue without knowing what version a user is using ("Linux, Firefox" is not good enough to justify a change). Especially since Firefox is an evergreen browser, which means the user has to have turned off updates intentionally (or kept firefox open without restarting for 4+ months) and thus should be aware they've turned off updates. "Linux, Firefox" could mean a Firefox version from 5 years ago, and we *definitely* should be telling users to upgrade if they have turned off updates for 5 years.

If someone wants to use an out of date browser, they can take advantage of user styles to fix the websites that will be broken as a result (especially if they are a technically-savvy linux user choosing to not upgrade). Bevy/maintainers/contributors shouldn't have to take on the maintenance burden for supporting that usage.

Personally I would say the bar for support should be something like "current evergreen browsers". "evergreen - 3 months" or something would also be acceptable, but 2.5 years is way too much and would prevent taking advantage of CSS improvements for half the time bevy has existed 😓 .

> I wasn't aware of this distinction and bite me later on

I actually don't think this bit you. The user upgraded their browser, as they should, and we probably shouldn't have driven changes based on supporting an out of date software configuration that the user who filed the issue isn't even using anymore.

doup:
Thanks for the reply, I really liked it. You gave me some new perspective and made me realize how a poor and lazy analysis I made. 🤣😅

Anyway, I would go for "last 2/3 versions" or "last 3 months". I'll share the discussion in `#website-dev` and let's see if there is a consensus.

---

53% (16 months since full support / 30 months to be considered "widely supported")… again, lazy. ;-)
It should be weighted by browser and usage… 🙃

alice-i-cecile:
@ChristopherBiscardi I'm ready to merge this once CI is passing.

PR #2126, opened 2025-04-28
Speed up generate-assets with database indices

kristoff3r:
It turns out the asset generation is extremely CPU bound in rustqlite queries, which can be avoided by adding some database indices. On my machine this takes every query from 5 seconds to 50ms in release mode.

### DIFF
--- a/generate-assets/src/lib.rs
+++ b/generate-assets/src/lib.rs
@@ -591,11 +591,23 @@ pub fn prepare_crates_db() -> anyhow::Result<CratesIoDb> {
         println!("Downloading crates.io data dump");
     }
 
-    Ok(CratesIODumpLoader::default()
+    let db = CratesIODumpLoader::default()
         .tables(&["crates", "dependencies", "versions"])
         .preload(true)
         .update()?
-        .open_db()?)
+        .open_db()?;
+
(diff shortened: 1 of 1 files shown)


kristoff3r:
Apparently this cuts 30 minutes off CI :smile: 

I was planning on doing async and stuff for the concurrency, but it looks like that's not really needed after this.

rparrett (approved):
I've been idly wondering why this (crates.io metadata specifically) was so slow. Cool.

An index on `crates(id)` would probably also be utilized by our code but that doesn't seem nearly as impactful. (I did not test very scientifically, but it seems like a slightly positive change)

kristoff3r:
> I've been idly wondering why this (crates.io metadata specifically) was so slow. Cool.
> 
> An index on `crates(id)` would probably also be utilized by our code but that doesn't seem nearly as impactful. (I did not test very scientifically, but it seems like a slightly positive change to add that one as well)

On my machine it makes no difference, every query is ~125ms +/- 5ms in both cases. But my approach wasn't very scientific either, so maybe some other indices might lower it even more.

rparrett:
`crates(name)` may also be useful

PR #2129, opened 2025-05-09
Add S-Needs-Reproduction to Contributing Guide

alice-i-cecile:


BenjaminBrienen (approved):

BenjaminBrienen:
Might call it "a reproducer", as in, code that produces the effect of the problem. It helps clarify between having that available to bug fixers and an issue that "needs some confirmation that it actually happens as described".

content/learn/contribute/reference/triage.md
```diff
@@ -32,6 +32,7 @@ Labels are our primary tool for organizing work. You can find a complete list wi
 - **S**: Status. The most common include:
   - `S-Ready-For-Implementation`: this issue is ready for someone to pick it up and open a PR!
   - `S-Needs-Triage`: this issue needs to be labeled.
+  - `S-Needs-Reproduction`: this issue needs a reproduction to allow fixes to be tested.
```
andriyDev, line 35: I sort of agree with @BenjaminBrienen, but I don't think we need to rename the tag. Maybe just adjusting the description here is enough. 

```suggestion
  - `S-Needs-Reproduction`: this issue needs a reproducible example to allow fixes to be tested.
```

andriyDev (approved):

PR #1949, opened 2025-03-17
Add an examples explanation contribution page

fallible-algebra:
This is a style guide for writing explanations for examples, designed following conversations[^1][^2] in the discord and the development of [doup's docs extraction](https://github.com/bevyengine/bevy/pull/18191#issue-2902754589).

This is an edited version of [the hackMD document (which featured 3 examples)](https://hackmd.io/@8b_tZaukRTafBLVz3oQH8w/SkH4y7vV1e) that I've posted to the discord a few times, to make it a bit closer to the surrounding contribution guide.

This PR needs some more feedback, there's divergence from the writing docs guide in terms of "who is looking at this" that I'm not sure is necessary, and some things may have been cut from the hackMD doc that shouldn't have been.

[^1]: (original conversation) https://discord.com/channels/691052431525675048/1316318586247315487/1316384953390600223
[^2]: (more recent feedback) https://discord.com/channels/691052431525675048/743559241461399582/1331017971862208590

PR #2131, opened 2025-05-09
Create "usage examples" category of first-party Bevy examples

alice-i-cecile:
As Bevy grows and matures, the need for better documentation grows with it. While we've gotten our API-focused documentation to a fairly good state with our existing examples + rustdoc, that only answers the "what does this API do?" question.

Users, especially experienced game developers, will commonly ask "how do I a specific task in Bevy", and will come up empty-handed. Instead, they're told to read or community-maintained resources, ask questions in our community spaces,or just "figure it out". While these can be excellent strategies, they're inefficient, commonly outdated, and are a missed opportunity to showcase how simple and effective a lot of these common tasks are in Bevy.

I've opted to build on the excellent work by @fallible-algebra over in #1949 to propose a new category of example aimed at filling this goal, while cleaning up and clarifying some guidance about our existing examples and example-like code.

From the perspective of the engine developer, this category of example represents a non-trivial maintenance burden, especially, as @IceSentry has pointed out, in terms of rendering code due to weaker abstractions and higher difficulty during refactoring. That said, I feel that this is ultimately valuable to engine devs as well, because these examples:

1. Focus attention on important, realistic workflows, allowing us to ensure that they are both straightforward and functional.
2. Present a larger test bed to catch and easily reproduce regressions in functionality, especially in concert with the excellent screenshot comparison tools built out by @mockersf.

This PR is only the first step: the existing examples and tooling will need to be refactored to match this proposal (e.g. the split screen example is an existing usage example), and if this proposal is accepted, I would expect to see a substantial number of new added examples demonstrating common tasks.

I'm opening this PR now, in the hopes that it can be polished, discussed and accepted in time for May 16th, the Bevy Unconf at RustWeek 2025: this refactoring and example writing effort would be a great fit for a large group of in-person contributors of all skill levels :)

### DIFF
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -76,7 +76,7 @@ Please adhere to the following guidelines when creating or updating an example.
 ### Stylistic preferences
 
 1. Use simple, descriptive variable names.
-   1. Avoid names like `MyComponent` in favor of more descriptive terms like `Events`.
+   1. Avoid names like `MyComponent` in favor of more descriptive or evocative terms like `SparseComponent` or `Life`.
    2. Prefer single letter differentiators like `EventsA` and `EventsB` to nonsense words like `EventsFoo` and `EventsBar`.
    3. Avoid repeating the type of variables in their name where possible. For example, `Color` should be preferred to `ColorComponent`.
 2. Prefer glob imports of `bevy::prelude::*` and `bevy::sub_crate::*` over granular imports (for terseness).
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -73,6 +73,7 @@ Please adhere to the following guidelines when creating or updating an example.
 5. Try to structure app / plugin construction in the same fashion as the actual code.
 6. Examples should typically not have tests, as those are not directly reusable by the Bevy user.
 7. Examples should not share common "utility" logic: if you feel this is warranted, improve the engine instead!
+   1. Similarly, try to avoid incidental complexity and abstractions. Find a simpler way for now, and work to create these tools and resolve this tension in the engine itself.
 
 ### Dependencies
 
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -98,6 +98,15 @@ As a result, you should currently reuse assets for your examples whenever possib
 If you *need* an asset to demonstrate a newly added feature, make it as small as possible.
 If you want to add a new example that requires a new asset to properly teach a new concept, it will almost certainly be rejected until the issue linked above is resolved.
 
+## Platform support
+
+By default, every example must work on our desktop platforms: Linux, Windows and MacOS.
+
+It is acceptable (but not ideal) if certain examples do not work on web, or low end hardware.
+However, this should be prominently called out and explained inside of the examples.
+
+Examples for mobile or embedded platforms require special setup, and are generally split out, focusing on the specific quirks of those platforms.
+
 ### Dependencies
 
 1. Examples should not take on new dependencies unless they are required for core functionality of that example.
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -63,7 +63,8 @@ Please adhere to the following guidelines when creating or updating an example.
    3. Game examples live in `examples/games`, with each game example getting its own dedicated subfolder.
    4. Stress tests live in the root level `stress-tests` folder.
    5. Testbeds live in the root level `tests` folder, subdivided by domain (e.g. `3d`).
-2. Examples should be a single file if possible.
+2. Each example should consist of exactly one `.rs` file, except where assets or shaders are required.
+   1. Game examples are an exception to this, and should be structured as self-contained small projects with a realistic file and module structure.
 3. Assets live in `/assets`. Try to avoid adding new assets unless strictly necessary to keep the repository small. Don't add "large" asset files.
 4. Each example should try to follow this order:
    1. Imports
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -72,6 +72,7 @@ Please adhere to the following guidelines when creating or updating an example.
    3. Example logic
 5. Try to structure app / plugin construction in the same fashion as the actual code.
 6. Examples should typically not have tests, as those are not directly reusable by the Bevy user.
+7. Examples should not share common "utility" logic: if you feel this is warranted, improve the engine instead!
 
 ### Dependencies
 
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -53,6 +53,11 @@ Please adhere to the following guidelines when creating or updating an example.
 ### Organization
 
 1. Examples should live in an appropriate subfolder of `/examples`.
+   1. API examples live in `examples/api`, then subdivided by engine category.
+   2. Usage examples live in `examples/usage`, then subdivided by domain.
+   3. Game examples live in `examples/games`, with each game example getting its own dedicated subfolder.
+   4. Stress tests live in the root level `stress-tests` folder.
+   5. Testbeds live in the root level `tests` folder, subdivided by domain (e.g. `3d`).
 2. Examples should be a single file if possible.
 3. Assets live in `/assets`. Try to avoid adding new assets unless strictly necessary to keep the repository small. Don't add "large" asset files.
 4. Each example should try to follow this order:
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -5,8 +5,21 @@ insert_anchor_links = "right"
 weight = 5
 +++
 
-Most [examples in Bevy](https://github.com/bevyengine/bevy/tree/main/examples) aim to clearly demonstrate a single feature, group of closely related small features, or show how to accomplish a particular task (such as asset loading, creating a custom shader, or testing your app).
-In rare cases, creating new "game" examples is justified in order to demonstrate new features that open a complex class of functionality in a way that's hard to demonstrate in isolation or requires additional integration testing.
+[Examples in Bevy](https://github.com/bevyengine/bevy/tree/main/examples) in Bevy can be categorized into three distinct categories,
+each of which have their own goals and subfolder:
+
+1. API examples: clearly demonstrating a single feature or group of closely related small features.
+2. Usage examples: teach users opinionated patterns for accomplishing common game development tasks.
+3. Small games: focused on integrating these patterns into a cohesive whole while demonstrating how to structure larger projects.
+
+In addition, we have two additional categories which are treated as examples by `cargo`, but are not intended to teach users:
+
+1. Stress tests: uses extreme entity counts or otherwise unrealistic settings to allow us to readily detect performance changes in particular performance-sensitive hot paths.
+2. Testbeds: designed to test features and scenes that rely on graphical rendering or interactivity, in both an automated and manual fashion.
+
+Unlike examples, stress tests and test beds should be configurable, allowing Bevy contriubtors (and curious users) to see the impact of changes.
+
+## Writing good examples
 
 Examples in Bevy should be:
 
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -67,6 +67,12 @@ Please adhere to the following guidelines when creating or updating an example.
 5. Try to structure app / plugin construction in the same fashion as the actual code.
 6. Examples should typically not have tests, as those are not directly reusable by the Bevy user.
 
+### Dependencies
+
+1. Examples should not take on new dependencies unless they are required for core functionality of that example.
+2. Examples can never take on dependencies to Bevy ecosystem crates, to avoid circular dependencies and maintenance risk.
+3. Any dependencies needed for examples should be listed as `dev-dependencies`.
+
 ### Stylistic preferences
 
 1. Use simple, descriptive variable names.
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -78,7 +78,7 @@ Please adhere to the following guidelines when creating or updating an example.
    5. Testbeds live in the root level `tests` folder, subdivided by domain (e.g. `3d`).
 2. Each example should consist of exactly one `.rs` file, except where assets or shaders are required.
    1. Game examples are an exception to this, and should be structured as self-contained small projects with a realistic file and module structure.
-3. Assets live in `/assets`. Try to avoid adding new assets unless strictly necessary to keep the repository small. Don't add "large" asset files.
+3. Assets live in `/assets`, organized by kind, rather than by example.
 4. Each example should try to follow this order:
    1. Imports
    2. A `fn main()` block
@@ -87,6 +87,16 @@ Please adhere to the following guidelines when creating or updating an example.
 6. Examples should typically not have tests, as those are not directly reusable by the Bevy user.
 7. Examples should not share common "utility" logic: if you feel this is warranted, improve the engine instead!
    1. Similarly, try to avoid incidental complexity and abstractions. Find a simpler way for now, and work to create these tools and resolve this tension in the engine itself.
+
+### Assets
+
+While we're [currently trying to figure out](https://github.com/bevyengine/bevy/issues/13875) how to best use larger and more realistic assets inside of Bevy's learning material, the current status quo is that we are extremely constrained on asset usage for examples.
+
+Adding new assets permanently increases the size of the Bevy repository, causing us problems with hosting, and resulting in pain for Bevy contributors when they clone or work with the repository.
+
+As a result, you should currently reuse assets for your examples whenever possible.
+If you *need* an asset to demonstrate a newly added feature, make it as small as possible.
+If you want to add a new example that requires a new asset to properly teach a new concept, it will almost certainly be rejected until the issue linked above is resolved.
 
 ### Dependencies
 
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -19,6 +19,19 @@ In addition, we have two additional categories which are treated as examples by 
 
 Unlike examples, stress tests and test beds should be configurable, allowing Bevy contriubtors (and curious users) to see the impact of changes.
 
+## What makes a good example?
+
+Each of our example categories has its own goals, and are best considered separately.
+
+A good API example is complementary to good API documentation, and should only be added to promote the discovery of important features, show API usage in context, or show off graphical, audible or interactive functionality that cannot be appropriately conveyed by written documentation alone.
+
+Usage examples tackle common tasks that might be encountered by veterans to game development (or CAD or...) that are trying to translate existing concepts or workflows into Bevy's APIs.
+These should be focused on realism, building out the minimum required to demonstrate that they actually solve the problem at hand.
+
+Game examples are much more expensive to maintain, and should be used sparingly. Each game example should demonstrate Bevy's capabilities in a new genre or style of game, while being simple enough for new Bevy users to follow and get excited about.
+
+Full game templates or more complex demonstrations of functionality are also welcome, but to manage maintenance burden these should be submitted to [`bevy-assets`](https://github.com/bevyengine/bevy-assets).
+
 ## Writing good examples
 
 Examples in Bevy should be:
@@ -32,8 +45,6 @@ When you add a new example, be sure to update `examples/README.md` with the new 
 Run `cargo run -p build-templated-pages -- build-example-page` to do this automatically.
 Use a generous sprinkling of keywords in your description: these are commonly used to search for a specific example.
 See the [example style guide](#style-guide) to help make sure the style of your example matches what we're already using.
-
-More complex demonstrations of functionality are also welcome, but these should be submitted to [`bevy-assets`](https://github.com/bevyengine/bevy-assets).
 
 ## Style guide
 
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -5,6 +5,20 @@ insert_anchor_links = "right"
 weight = 5
 +++
 
+Examples in Bevy should be:
+
+1. **Working:** They must compile and run, and any intentionally introduced errors in them should be obvious (through tests, simple results, or clearly displayed behavior).
+2. **Clear:** They must use descriptive variable names, be formatted, and be appropriately commented. Try your best to showcase best practices when it doesn't obscure the point of the example.
+3. **Relevant:** They should explain, through comments or variable names, what they do and how this can be useful to a game developer.
+4. **Minimal:** They should be no larger or complex than is needed to meet the goals of the example.
+
+When you add a new example, be sure to update `examples/README.md` with the new example and add it to the root `Cargo.toml` file.
+Run `cargo run -p build-templated-pages -- build-example-page` to do this automatically.
+Use a generous sprinkling of keywords in your description: these are commonly used to search for a specific example.
+See the [example style guide](#style-guide) to help make sure the style of your example matches what we're already using.
+
+## Types of examples
+
 [Examples in Bevy](https://github.com/bevyengine/bevy/tree/main/examples) in Bevy can be categorized into three distinct categories,
 each of which have their own goals and subfolder:
 
@@ -31,20 +45,6 @@ These should be focused on realism, building out the minimum required to demonst
 Game examples are much more expensive to maintain, and should be used sparingly. Each game example should demonstrate Bevy's capabilities in a new genre or style of game, while being simple enough for new Bevy users to follow and get excited about.
 
 Full game templates or more complex demonstrations of functionality are also welcome, but to manage maintenance burden these should be submitted to [`bevy-assets`](https://github.com/bevyengine/bevy-assets).
-
-## Writing good examples
-
-Examples in Bevy should be:
-
-1. **Working:** They must compile and run, and any intentionally introduced errors in them should be obvious (through tests, simple results, or clearly displayed behavior).
-2. **Clear:** They must use descriptive variable names, be formatted, and be appropriately commented. Try your best to showcase best practices when it doesn't obscure the point of the example.
-3. **Relevant:** They should explain, through comments or variable names, what they do and how this can be useful to a game developer.
-4. **Minimal:** They should be no larger or complex than is needed to meet the goals of the example.
-
-When you add a new example, be sure to update `examples/README.md` with the new example and add it to the root `Cargo.toml` file.
-Run `cargo run -p build-templated-pages -- build-example-page` to do this automatically.
-Use a generous sprinkling of keywords in your description: these are commonly used to search for a specific example.
-See the [example style guide](#style-guide) to help make sure the style of your example matches what we're already using.
 
 ## Style guide
 
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -5,12 +5,13 @@ insert_anchor_links = "right"
 weight = 5
 +++
 
-Examples in Bevy should be:
+Each example in Bevy must be:
 
-1. **Working:** They must compile and run, and any intentionally introduced errors in them should be obvious (through tests, simple results, or clearly displayed behavior).
-2. **Clear:** They must use descriptive variable names, be formatted, and be appropriately commented. Try your best to showcase best practices when it doesn't obscure the point of the example.
-3. **Relevant:** They should explain, through comments or variable names, what they do and how this can be useful to a game developer.
-4. **Minimal:** They should be no larger or complex than is needed to meet the goals of the example.
+1. **Goal-driven:** Each example must have a single clearly stated learning-focused goal.
+2. **Working:** Every example must compile and run, and any intentionally introduced errors in them should be obvious (through tests, simple results, or clearly displayed behavior).
+3. **Clear:** Examples should use descriptive variable names, be formatted, and be appropriately commented. Try your best to showcase best practices when it doesn't obscure the point of the example.
+4. **Relevant:** Examples should explain, through comments or variable names, what they do and how this can be useful to a game developer.
+5. **Minimal:** Examples should be no larger or complex than is needed to meet the goals of the example.
 
 When you add a new example, be sure to update `examples/README.md` with the new example and add it to the root `Cargo.toml` file.
 Run `cargo run -p build-templated-pages -- build-example-page` to do this automatically.
--- a/content/learn/contribute/helping-out/creating-examples.md
+++ b/content/learn/contribute/helping-out/creating-examples.md
@@ -49,6 +49,19 @@ Usage examples can be more opinionated, and may demonstrate multiple possible ap
 Game examples are much more expensive to maintain, and should be used sparingly. Each game example should demonstrate Bevy's capabilities in a new genre or style of game, while being simple enough for new Bevy users to follow and get excited about.
 Polish is important here: these are often a new user's first impression of the engine, although it must be balanced against added complexity.
 
+Each game example should be a few hundred to a few thousand lines at most.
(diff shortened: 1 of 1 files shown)


content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -5,22 +5,51 @@ insert_anchor_links = "right"
...
 See the [example style guide](#style-guide) to help make sure the style of your example matches what we're already using.
 
-More complex demonstrations of functionality are also welcome, but these should be submitted to [`bevy-assets`](https://github.com/bevyengine/bevy-assets).
+## Types of examples
+
+[Examples in Bevy](https://github.com/bevyengine/bevy/tree/main/examples) in Bevy can be categorized into three distinct categories,
+each of which have their own goals and subfolder:
+
+1. API examples: clearly demonstrating a single feature or group of closely related small features.
+2. Usage examples: teach users opinionated patterns for accomplishing common game development tasks.
```
janhohenheim, line 27: YES please to more opinionated stuff! I think Bevy has in the past shied away from teaching very useful stuff in fear of being too opinionated, so addressing this is the right step imo

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -29,7 +58,13 @@ Please adhere to the following guidelines when creating or updating an example.
 ### Organization
 
 1. Examples should live in an appropriate subfolder of `/examples`.
-2. Examples should be a single file if possible.
+   1. API examples live in `examples/api`, then subdivided by engine category.
+   2. Usage examples live in `examples/usage`, then subdivided by domain.
```
janhohenheim, line 62: Not sure the distinction between these two is clear to newcomers, maybe some better names can be found?

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -29,7 +58,13 @@ Please adhere to the following guidelines when creating or updating an example.
...
+   3. Game examples live in `examples/games`, with each game example getting its own dedicated subfolder.
+   4. Stress tests live in the root level `stress-tests` folder.
+   5. Testbeds live in the root level `tests` folder, subdivided by domain (e.g. `3d`).
+2. Each example should consist of exactly one `.rs` file, except where assets or shaders are required.
+   1. Game examples are an exception to this, and should be structured as self-contained small projects with a realistic file and module structure.
```
janhohenheim, line 67: This was a huge reason for why I didn't seriously consider contributing a tiny game before, so this is *very* welcome :)

janhohenheim (commented):
In risk of opening up a big bikeshedding party, I suggest adding an extra constraint to game examples: use the [plugin organization pattern](https://github.com/TheBevyFlock/bevy_new_2d/blob/main/docs/design.md#pattern)

This is by now tried and tested by many members of the community. It's not really thaaaat useful for frameworks or crates, so you tend to not see it often when perusing your dependencies, but it's really, *really* helpful in organizing actual games. 
I say this because I know that if I'm upstreaming a tiny game example, I'm definitely going to try and use this pattern, and it may be nice to just start using it for all game examples for consistency. I've also helped many learners on Discord that had problems with figuring out how to organize themselves beyond "put everything into one or two files", and quite a few of them wished that Bevy itself had taught them this pattern sooner, perhaps in examples.

Again, I know this is very opinionated, I know others disagree on the usefulness of this, and we can also let different game examples show different styles of organization, that's fine too :)

Just wanted to open this potential can of worms rather sooner than later.

content/learn/contribute/helping-out/creating-examples.md
janhohenheim, line 67: Are game examples allowed to run only well on native, or should they all be usable on Wasm?
Asking because on Wasm
- playing music is generally choppy, and it's hard to add polish there without a lot of hacks
- locking and unlocking the cursor behaves very intuitively, meaning that any first person examples would have to include hacks for that

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -5,22 +5,51 @@ insert_anchor_links = "right"
...
+When possible, prefer documentation tests and module documentation over API examples: these are easier to keep up to date, more clearly scoped and better connected to the APIs in question.
+
+Usage examples tackle common tasks that might be encountered by veterans to game development (or CAD or...) that are trying to translate existing concepts or workflows into Bevy's APIs.
+These should be focused on realism, building out the minimum required to demonstrate that they actually solve the problem at hand.
+Usage examples can be more opinionated, and may demonstrate multiple possible approaches to the same problem, explaining the tradeoffs involved.
+
+Game examples are much more expensive to maintain, and should be used sparingly. Each game example should demonstrate Bevy's capabilities in a new genre or style of game, while being simple enough for new Bevy users to follow and get excited about.
+Polish is important here: these are often a new user's first impression of the engine, although it must be balanced against added complexity.
```
janhohenheim, line 50: Are game examples allowed to run only well on native, or should they all be usable on Wasm?
Asking because on Wasm

- playing music is generally very quickly very choppy, so it's hard to add polish there without some advanced hacks
- locking and unlocking the cursor behaves very unintuitively, meaning that any first person examples would have to include weird hacks for that (I can elaborate if needed)

alice-i-cecile: WASM support is a nice-to-have. Ideally we should fix both of those long-term.

content/learn/contribute/helping-out/creating-examples.md
alice-i-cecile: Yeah, bikeshed welcome on this!

alice-i-cecile:
> In risk of opening up a big bikeshedding party, I suggest adding an extra constraint to game examples: use the [plugin organization pattern](https://github.com/TheBevyFlock/bevy_new_2d/blob/main/docs/design.md#pattern)

Definitely agree with using plugins, neutral on fn plugins vs dedicated structs. That can wait for follow-up IMO: "realistic structure" covers "games should use plugins".

janhohenheim:
Then that's good enough for me, as one is just a shorthand for the other :)

janhohenheim:
Another point: what is the guideline on starting the example document with `//!` blocks?
I'm asking because I have a huuuge block of information in the [fixed timestep](https://github.com/bevyengine/bevy/blob/main/examples/movement/physics_in_fixed_timestep.rs) example, but the [online version](https://bevyengine.org/examples/movement/physics-in-fixed-timestep/) completely omits that part of the example.
Not sure if that's a bug or a mistake on my part.

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -5,22 +5,51 @@ insert_anchor_links = "right"
...
+When possible, prefer documentation tests and module documentation over API examples: these are easier to keep up to date, more clearly scoped and better connected to the APIs in question.
+
+Usage examples tackle common tasks that might be encountered by veterans to game development (or CAD or...) that are trying to translate existing concepts or workflows into Bevy's APIs.
+These should be focused on realism, building out the minimum required to demonstrate that they actually solve the problem at hand.
+Usage examples can be more opinionated, and may demonstrate multiple possible approaches to the same problem, explaining the tradeoffs involved.
+
+Game examples are much more expensive to maintain, and should be used sparingly. Each game example should demonstrate Bevy's capabilities in a new genre or style of game, while being simple enough for new Bevy users to follow and get excited about.
```
janhohenheim, line 49: How much should game examples repeat important information that is already provided elsewhere? For example, every first person game should imo have a [separate view model and world model](https://bevyengine.org/examples/camera/first-person-view-model/) for accessibility reasons, since Bevy cares a lot about accessibility.
Is it enough to add a comment on that part of a game example that says "see this example an explanation of why we are doing this"? Or should it try to be entirely self-contained?

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -29,7 +58,13 @@ Please adhere to the following guidelines when creating or updating an example.
...
 1. Examples should live in an appropriate subfolder of `/examples`.
-2. Examples should be a single file if possible.
+   1. API examples live in `examples/api`, then subdivided by engine category.
+   2. Usage examples live in `examples/usage`, then subdivided by domain.
+   3. Game examples live in `examples/games`, with each game example getting its own dedicated subfolder.
+   4. Stress tests live in the root level `stress-tests` folder.
+   5. Testbeds live in the root level `tests` folder, subdivided by domain (e.g. `3d`).
+2. Each example should consist of exactly one `.rs` file, except where assets or shaders are required.
+   1. Game examples are an exception to this, and should be structured as self-contained small projects with a realistic file and module structure.
 3. Assets live in `/assets`. Try to avoid adding new assets unless strictly necessary to keep the repository small. Don't add "large" asset files.
```
janhohenheim, line 68: How big are game examples allowed to get? Fox example, textures and normal maps are not *strictly* necessary, but are part of polish. Is there some file size we should aim for per example, e.g. every example game is allowed to bring in up to 5 MiB of new assets?

tbillington:
> In risk of opening up a big bikeshedding party, I suggest adding an extra constraint to game examples: use the [plugin organization pattern](https://github.com/TheBevyFlock/bevy_new_2d/blob/main/docs/design.md#pattern)

I think there might be a little more nuance to it than plugins vs no plugins. Imo where the plugins are located, and what they're abstracting away is important to the discussion.

Semi structured thoughts :)

- Plugins are great, and for anything but a basic single file prototype or `bevy_ecs` only use we should encourage users to use them in their projects, though most of the games in bevy repo examples don't use them because their small enough which is fine. Right tool for the job etc. Adding plugins for the sake of it would detract from the smaller examples.
- Plugins in the same file like [`game_menu`](https://github.com/bevyengine/bevy/blob/main/examples/games/game_menu.rs) are easy for a reader to jump between. Extra files to reference are a much bigger pain especially when reading examples on a browser or phone as opposed to having an IDE open where you can split pane, jump to definition etc.
- Plugins to help compartmentalise and break up larger actual implementation logic of menu/game/splash like [`game_menu`](https://github.com/bevyengine/bevy/blob/main/examples/games/game_menu.rs) make sense and I'm in favour of, if the example is big enough to require them, because it encourages a good practice the user will benefit from, and makes the example easier to understand by keeping it in digestible chunks.
- Putting "common"/"utility" things that will be used by all examples like input handling or state transitions etc is a double edged sword and what I dislike when I open a projects examples. As a maintainer yes it's annoying to duplicate stuff, but as reader trying to implement the example in my own project I usually still need to know what that helper is doing to understand the full context of the example. Eg in [`quinn`](https://github.com/quinn-rs/quinn/blob/main/quinn/examples/common/mod.rs) it has a common file for basic connection stuff, but as a user I still need to know what's in all those helpers to actually implement the example..
[`configure_server`](https://github.com/quinn-rs/quinn/blob/main/quinn/examples/common/mod.rs#L58) is a good example of what I'd consider a bad helper. It makes it looks so simple when you're reading the actual example, just "configure the server", but it's actually doing a lot of nuanced things like generating TLS cert, encryption, configuring transports and limits etc. Very "just draw the rest of the owl".
**⚠️ If we are hiding complexity behind util helpers we should be asking why that is necessary, it's a pain the user will feel when using Bevy, can we improve the situation instead of hiding it?**
- On bigger projects I've seen helper/utilties grow into their own crate with their own dependencies and a bunch of files in them. As a user trying to learn from an example that is like the worst thing to encounter. I can't even just copy the helper code into my project because often they pull in extra libraries in the helpers too 😅 so I need to open Cargo.toml, look at what dependencies they have, see which I actually need or can I rewrite the helper code to not need them.. etc. It's just a lot of friction and indirection for a new user struggling to gain context of what's happening.

janhohenheim:
- Fully agreed on disliking "common" / "util" plugins.
- Also fully agreed on having weird mega helper crates

For the rest of your points I agree in the case of regular examples. My comment is specifically about example games, which I would assume are fine to optimize for IDE usage and are less "how do I do this specific thing" and more "how do all of these separate things come together to form an entire game?"

tbillington:
Yea I guess the size and necessary complexity of the "example" will dictate what makes sense where. I'm still a little fuzzy on what would qualify for the "usage examples" or "game examples", and if this PR is only for the game examples, and where do we draw the line? Most practical examples won't need a full game to demonstrate them, split screen is a great example.

How big would these game examples be? Breakout is 400 lines, though I can easily imagine a few thousand lines for larger examples, especially if there's a lot of comments. It could even make sense to have those big example be their own crate, so they can specify their bevy features etc directly and you could just build the example or copy it's whole directory out of the bevy examples and run it standalone (aka as a starting point). Though I haven't thought that through too much. It would be the "most realistic" version.

I'm wary of discussing hypothetical scenarios, or those from different contexts like the quinn example much more. Would much prefer if we had short descriptions of a handful of these game-examples so we could consider these points in the context of how they would be used.

janhohenheim:
Good point! I personally had in mind potentially coding a first person shooting gallery with targets moving around, where you try to hit as many as accurately as possible in a time window, after which you get a score board. That example would include a nicely designed level, so some vegetation, some buildings, etc. as a backdrop, as well as animations and sound effects. There also needs to be a pause menu that allows you to change the mouse sensitivity and FOV for accessibility.

This is just the first idea that came into mind. @alice-i-cecile is this about the right scope? Or is it already wayyy to much?

ChristopherBiscardi (approved):
I think the examples directory is a particular strength of Bevy's ecosystem and I'm very happy to see interest in expanding it. I constantly tell people the examples directory should be their first stop.

> [!NOTE]  
> I feel like I have to disclose that I do make income from educational content, some of which is Bevy related. I think that's mostly generally known but just in case someone reading is unaware.

That said, there is a significant cost in terms of review and maintenance. I've have seen real contributors express explicit sentiment that they aren't pursuing a PR because they don't want to update the examples.

Its also worth noting that even the current examples aren't kept up to date with new APIs. I say this mostly to air the maintenance/review burden, not to say the examples are bad in any way. A good example right now is that the UI examples specifically have largely not been updated to use the children![] APIs, which is one of the more useful locations to teach that API. At the current pace of PR review, updating the directory will take months.

I fear that "opinionated" code organization/style conversations will devolve into bikeshed conversations quickly. We've already started off the very first comment in this PR with a self-described bikeshed topic that is probably getting as much or more discussion than the PR contents.

That is all to say, I feel like maintaining these examples is basically going to turn into a full time job for someone. Creation, maintenance, review, decisions around what actually fits or doesn't, and seemingly pushing into "how is this serialized as useful a learning path" from the description here.

---

To add some context/color for maintenance for "full game" examples, I'm currently in the middle of scripting workshops for about 5 games for 0.16 (2048, block breaker, snake, minesweeper, flappy bird). These end up being 500-1000 lines of code if you don't include anything strictly unecessary (In my experience people often skip/don't want to see UI/menus/etc for example). These examples should also probably skip menu systems unless a menu system is the point of the example.

Games that include significant non-bevy-related code should probably also be excluded. ex: 2048 requires writing the board-shift logic generically (or in 4 directions) and that's probably not useful learning material for bevy use cases when compared to the maintenance.

The most common thing I run into is a missing bevy feature that requires an external dependency. For example, Asteroids is almost completely doable in Bevy without dependencies, but as a new user you have to make a choice between switching to Avian or not just for a triangle collider for the ship. (IntersectsVolume supports circle/sphere and rect/cube afaik). Bevy relies on its ecosystem heavily currently (physics, particles, etc), so exluding ecosystem crates speaks to the limited scope for the examples.

It would be nice if (broadly speaking) every time there was a helper/util or similar created for one of these games, if an issue was filed to figure out how to eliminate/upstream that helper. Constructing custom abstractions (ex: SpawnObserver) shouldn't be a focus of these examples.

---

Rendering use case examples are often verbose but I think that's widely known. I've stopped trying to PR new rendering use-case examples because they've been denied in the past. I wonder when/which Rendering examples would be ok to PR to this new directory.

---

Overall, what's written here seems great. I wonder how we'll maintain it over the long term (it seems like a job role in the foundation one day) and where we expect wider ecosystem learning material to fit. I try to make sure the material I create outside of the repo complements what exists in Bevy's repo/docs rather than repeat it so its helpful to have these kinds of goals written down so I can plan that work.

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -5,22 +5,51 @@ insert_anchor_links = "right"
...
+
+Usage examples tackle common tasks that might be encountered by veterans to game development (or CAD or...) that are trying to translate existing concepts or workflows into Bevy's APIs.
+These should be focused on realism, building out the minimum required to demonstrate that they actually solve the problem at hand.
+Usage examples can be more opinionated, and may demonstrate multiple possible approaches to the same problem, explaining the tradeoffs involved.
+
+Game examples are much more expensive to maintain, and should be used sparingly. Each game example should demonstrate Bevy's capabilities in a new genre or style of game, while being simple enough for new Bevy users to follow and get excited about.
+Polish is important here: these are often a new user's first impression of the engine, although it must be balanced against added complexity.
+
+Full game templates or more complex demonstrations of functionality are also welcome, but to manage maintenance burden these should be submitted to [`bevy-assets`](https://github.com/bevyengine/bevy-assets), as part of our semi-curated collection of community-maintained learning resources.
```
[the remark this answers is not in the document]
janhohenheim: (Not for me to decide, but personally I think Chris should definitely be allowed to put his paid content there, especially since we want to transition the asset page to an asset store at some point in the future anyways)

tbillington:
Some great points. 

Bevy's examples are probably the best official examples of any game engine I know. I've seen I'm not the only one who appreciates that quite a bit. That positive reinforcement likely leads to prioritisation on keeping them compiling, up to date, and covering more areas.

Bevy is still quite feature-lite compared to what one might get from a more established engine. I won't get into examples but it's not hard to imagine how a feature that is "just" a node/component/checkbox in another engine would require a decent amount of custom work and often domain knowledge, therefore leading people to ask for examples of how to achieve it in Bevy.

And Bevy still has a relatively small (but passionate!) educator community (youtube, blogs, tutorials), which traditionally has filled in the majority of examples and sample projects in other engines. Not to mention API evolution makes older material less usable and often contradictory to current "best practices" (see constant questions about EventReader v Observers in discord 🤣 ).

Together all of these points put pressure on more first party examples in the repo and keeping them fresh.

But as you said we have to keep in mind the burden it puts on engine contributors and maintainers. There must be more we can do to alleviate the burden. Perhaps examples could be tiered, and non-tier 1 examples aren't required to be updated to merge a PR, instead becoming a "good first issue" style thing to encourage contributors and allow other people to share the work? But I'm just spitballing, I'm sure there are other ideas.

---

> That is all to say, I feel like maintaining these examples is basically going to turn into a full time job for someone.

I definitely agree with the sentiment, though (maybe naively) as Bevy matures the rate of API change should slow down as patterns emerge and we start to build upon the foundations more than changing them. Some examples may also be removed and absorbed into the engine as features when they make sense, eg ui/button hover styles.

---

It's hard to state how nice it is to quickly flick to the examples to see such comprehensive and thorough use of the APIs. Other engines often have their flagship first-party new-feature demos no longer compiling 6 months later which is pretty sad which speaks to their priorities and motivations, and not in a good way 😓 . I'm glad the Bevy contributors put the amount of effort they do into the examples, it's one of the ways I feel taken care of as a user of Bevy.

content/learn/contribute/helping-out/creating-examples.md
alice-i-cecile: Cross-linking in various forms is fine :)

content/learn/contribute/helping-out/creating-examples.md
IceSentry: This is one of my biggest concern with allowing that style of example. Almost every PR I've seen that wanted to add something like that came with adding a new texture or asset. This isn't sustainable long term.

IceSentry:
I think all the points raised by @ChristopherBiscardi are very valid and it pretty much echoes my own opinions.

Personally, I think we should consider usage example to be in a separate repo and maybe even only track the last released version of bevy. This could allow different standards for review and it won't block PR authors trying to work on a new feature. Once we are close to the release, it would also make for a good benchmark of the migration guide to update all those usage examples. It will also help with keeping asset bloat separated from the main repo. We still need a long term solution for that, but this would help in the short term.

content/learn/contribute/helping-out/creating-examples.md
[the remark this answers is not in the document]
alice-i-cecile: Yes, please feel free to put paid content on Bevy Assets! Supporting a broader community of people making a living off of the ecosystem is part of our long-term vision here.

alice-i-cecile:
> The most common thing I run into is a missing bevy feature that requires an external dependency. For example, Asteroids is almost completely doable in Bevy without dependencies, but as a new user you have to make a choice between switching to Avian or not just for a triangle collider for the ship. (IntersectsVolume supports circle/sphere and rect/cube afaik). Bevy relies on its ecosystem heavily currently (physics, particles, etc), so exluding ecosystem crates speaks to the limited scope for the examples.

Yep, pushing this sort of problem to the forefront of the minds of maintainers and contributors is part of why I want to move in this direction. Bevy cannot and should not do everything, but in the long-term third-party dependencies should be for saving work, advanced users or replacing an existing module, not required for core functionality (particles, physics, action management, UI widgets...).

> That is all to say, I feel like maintaining these examples is basically going to turn into a full time job for someone. Creation, maintenance, review, decisions around what actually fits or doesn't, and seemingly pushing into "how is this serialized as useful a learning path" from the description here.

Agreed: my kingdom for a budget to hire someone to do docs full-time. That said, I think this is a critical next step for adoption: we're getting there in terms of features, but our perceived quality and learning curve is still too low.

> Personally, I think we should consider usage example to be in a separate repo and maybe even only track the last released version of bevy. This could allow different standards for review and it won't block PR authors trying to work on a new feature.

I'm strongly against this. I understand the frustration and concerns here, but we a) need to be able to spot migration and regression problems immediately and b) we can't afford to pile on end-of-cycle work. End-of-cycle work is extremely hard to distribute and painful for maintainers in general, and leads to near-stoppages of productivity for the entire rest of the engine.

It would be particularly painful here, as updating the Bevy version depended on by these examples would either need to be done in one mega-PR, or merge code with examples in broken states.

If the rendering SMEs are united in "we can't afford to do this yet", I can agree to that as a temporary exemption while we address technical debt and boilerplate. But other more gameplay focused examples are much easier to keep up to date, and other areas like assets are sorely in need of more realistic examples for both learning and testing.

>  It will also help with keeping asset bloat separated from the main repo. We still need a long term solution for that, but this would help in the short term.

I agree that we need a real solution for this. We should avoid adding new assets to our examples until this is resolved.

alice-i-cecile:
I've added a dedicated section on asset usage inside of examples in https://github.com/bevyengine/bevy-website/pull/2131/commits/6dcac00048bcb575564cdaa3176f249a16b13a82.

While I am very much discontent with that as a long-term solution, it's good to communicate the current rules, and also clarify for reviewers of *this* PR that I am not currently proposing changing them.

I expect that many valuable usage examples (especially in audio, rendering and animation) will largely be blocked by https://github.com/bevyengine/bevy/issues/13875, but that's another can of worms to solve, and something that I know @mockersf has plans for.

alice-i-cecile:
> Perhaps examples could be tiered, and non-tier 1 examples aren't required to be updated to merge a PR, instead becoming a "good first issue" style thing to encourage contributors and allow other people to share the work? But I'm just spitballing, I'm sure there are other ideas.

My view here is that the standard for updating examples in PRs is "all examples must continue to work", which is critically distinct from "all examples must be idiomatic and fully up-to-date" :) We need to ensure that viable migration paths exist, code paths don't quietly rot, and that the examples remain usable learning material.

content/learn/contribute/helping-out/creating-examples.md
alice-i-cecile: Strongly agree, and I've added clarification to this effect in a later commit.

alice-i-cecile:
> Not sure if that's a bug or a mistake on my part.

This is a serious bug: can you check the issues and PRs on this repo please?

fallible-algebra:
> Not sure if that's a bug or a mistake on my part.

It might have something to do with https://github.com/bevyengine/bevy-website/pull/2098 not yet being merged, but I'm not certain.

content/learn/contribute/helping-out/creating-examples.md
laundmo: I also think "patterns" works for smaller examples showcasing a single concept. Maybe even "snippets" or "parts". I also think that for these small examples, somehow linking them in the module-level docs of what they're shocasing would be neat, but that is probably out of scope for this. well, except maybe that api/pattern/snippet examples could be placed in folders matching the module hierarchy?

laundmo:
Either along this effort, or as a follow-up, i think it would be good to re-order how examples are categorized on the homepage. Specifically, i think in some way the most complete, best looking examples should come first. Be that just by ordering the "game" examples above others, or by introducing "featured" or "favourited" examples which are hoisted above all others, it's just not very good optics having to scroll past pages of default-grey background examples showing a single sprite, a basic scene, etc.

Its a thought i've had for a while, but its not something i wanted to make an issue for without knowing the bigger picture, but i can understand if this should be a follow-up issue as it may not be too relevant for the considerations in this PR.

content/learn/contribute/helping-out/creating-examples.md
analytik: Not a contributor so sorry for spamming, but: Yes to opinionated examples! Just as a user/beginner I think it's important to be told _when_ something is opinionated, and if it's a footgun, or if it's just a good practice with other, potentially good alternatives.

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -5,22 +5,51 @@ insert_anchor_links = "right"
...
-In rare cases, creating new "game" examples is justified in order to demonstrate new features that open a complex class of functionality in a way that's hard to demonstrate in isolation or requires additional integration testing.
+Each example in Bevy must be:
 
-Examples in Bevy should be:
-
-1. **Working:** They must compile and run, and any intentionally introduced errors in them should be obvious (through tests, simple results, or clearly displayed behavior).
-2. **Clear:** They must use descriptive variable names, be formatted, and be appropriately commented. Try your best to showcase best practices when it doesn't obscure the point of the example.
-3. **Relevant:** They should explain, through comments or variable names, what they do and how this can be useful to a game developer.
-4. **Minimal:** They should be no larger or complex than is needed to meet the goals of the example.
+1. **Goal-driven:** Each example must have a single clearly stated learning-focused goal.
```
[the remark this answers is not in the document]
NthTensor: Reading through and dumping my thoughts as I go. I worry it may be difficult to get contributors to think like this. We might end up in a (a) well motivated, (b) polished, (c) complete; pick two situation.

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -5,22 +5,64 @@ insert_anchor_links = "right"
...
 Use a generous sprinkling of keywords in your description: these are commonly used to search for a specific example.
 See the [example style guide](#style-guide) to help make sure the style of your example matches what we're already using.
 
-More complex demonstrations of functionality are also welcome, but these should be submitted to [`bevy-assets`](https://github.com/bevyengine/bevy-assets).
+## Types of examples
+
+[Examples in Bevy](https://github.com/bevyengine/bevy/tree/main/examples) in Bevy can be categorized into three distinct categories,
+each of which have their own goals and subfolder:
+
+1. API examples: clearly demonstrating a single feature or group of closely related small features.
```
NthTensor, line 26: I think we'd agree these are objectively the easiest to write. How are we going to encourage people to do the harder two, short of getting them all in a room at a con?

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -29,19 +71,52 @@ Please adhere to the following guidelines when creating or updating an example.
...
-2. Examples should be a single file if possible.
-3. Assets live in `/assets`. Try to avoid adding new assets unless strictly necessary to keep the repository small. Don't add "large" asset files.
+   1. API examples live in `examples/api`, then subdivided by engine category.
+   2. Usage examples live in `examples/usage`, then subdivided by domain.
+   3. Game examples live in `examples/games`, with each game example getting its own dedicated subfolder.
+   4. Stress tests live in the root level `stress-tests` folder.
+   5. Testbeds live in the root level `tests` folder, subdivided by domain (e.g. `3d`).
+2. Each example should consist of exactly one `.rs` file, except where assets or shaders are required.
+   1. Game examples are an exception to this, and should be structured as self-contained small projects with a realistic file and module structure.
+3. Assets live in `/assets`, organized by kind, rather than by example.
```
NthTensor, line 81: Are we at all concerned about the asset needs of the small game examples stretching the abilities of git?

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -29,19 +71,52 @@ Please adhere to the following guidelines when creating or updating an example.
...
 4. Each example should try to follow this order:
    1. Imports
    2. A `fn main()` block
    3. Example logic
 5. Try to structure app / plugin construction in the same fashion as the actual code.
 6. Examples should typically not have tests, as those are not directly reusable by the Bevy user.
+7. Examples should not share common "utility" logic: if you feel this is warranted, improve the engine instead!
+   1. Similarly, try to avoid incidental complexity and abstractions. Find a simpler way for now, and work to create these tools and resolve this tension in the engine itself.
```
NthTensor, line 89: I'd want to contribs to file issues for pain points like these which they encounter during example development.

content/learn/contribute/helping-out/creating-examples.md
NthTensor: Ah, I see this is addressed later.

content/learn/contribute/helping-out/creating-examples.md
```diff
@@ -29,19 +71,52 @@ Please adhere to the following guidelines when creating or updating an example.
...
+
+Adding new assets permanently increases the size of the Bevy repository, causing us problems with hosting, and resulting in pain for Bevy contributors when they clone or work with the repository.
+
+As a result, you should currently reuse assets for your examples whenever possible.
+If you *need* an asset to demonstrate a newly added feature, make it as small as possible.
+If you want to add a new example that requires a new asset to properly teach a new concept, it will almost certainly be rejected until the issue linked above is resolved.
+
+## Platform support
+
+By default, every example must work on our desktop platforms: Linux, Windows and MacOS.
```
NthTensor, line 103: Both X and Wayland?

NthTensor (approved):
Strongly needed. User-driven examples like this would be a huge step forward. Broadly in favor. Have questions about how we get from here to there.

Getting people to write more usage examples and fewer API examples seems like a cultural shift that will take some effort to effect. I guess the plan is to sort of start small and hope it catches on? 

Actually I think it may be a bit tricky to identify interesting and useful usage examples. But once identified, I imagine many examples could be implemented by junior devs (with some guidence). So maybe we should start collecting a list of potential candidates for people to flesh out? An inventory system UI with various item types maybe (people often seem to ask about that).

Pretty much all the existing examples will become API examples, yes?

IceSentry:
> Getting people to write more usage examples and fewer API examples seems like a cultural shift that will take some effort to effect. I guess the plan is to sort of start small and hope it catches on?

I'm not sure I agree with the "fewer API examples" part. We should still encourage api examples for new apis because they are still super valuable. The one thing this could allow is to make the api examples leaner and remove things that were done to make the example look better but maybe less straight forward, but not write less of them.

content/learn/contribute/helping-out/creating-examples.md
IceSentry: I don't think we have to encourage that? We already have people that want to submit usage examples and this makes it more structured, but I don't see why that implies prioritizing those.

fallible-algebra:
Exciting to see this! (Just had the time to sit down and get through all the discussion)

On maintenance burden: examples are not load-bearing to the engine, and I think it makes sense to considering at least archiving an example if it is causing problems for maintenance.

A future pitfall with expanding / changing the scope of examples might be imagining that examples should become more iterative (walking through changes in problem solving in a Jupyter notebook style way) than descriptive (we're doing it this way, and here's how and a bit of reasoning for why)[^footnote1].

> I'm not sure I agree with the "fewer API examples" part.

Seconded. I think some can be considered superfluous, but not that many. A lot of the examples make for a really good "W3schools" style stripped-down intro to bevy in terms of showing off API in-practice. But not to contradict my first point, it might be wise to think about when it makes sense to retire an example, even if we're not over-eager to.

[^footnote1]: I think this mode should exist, but that moving towards examples being purely that would be a mis-identification of the pedagogical value of examples (not to say they're perfect, but that they have been praised for occupying a rare spot in the rust ecosystem). This may be me inventing someone in my head to disagree with, mind.

alice-i-cecile:
This has been thoroughly reviewed, and discussed by dozens of contributors, both online and at RustWeek 2025 :) There has been broad consensus that something along these lines is desirable and important to prioritize.

I'm going to go ahead and merge this now, and unblock the work of our contributors here at the Bevy Unconf.

PR #2137, opened 2025-05-17
iOS migration guide update

Leinnan:
Few changes:
- update the code to restore, it was not complete
- restructure solution part a bit- first suggest preffered approach, then add information about how to restore old behaviour
- add possible ways for performing transition into preferred approach


BD103 (approved):
Looks good, thank you!

PR #2152, opened 2025-06-18
Restructure Bevy Foundation Actions Without a Meeting and add new actions

alice-i-cecile:
## Problem

The existing actions-without-a-meeting ended up crammed into a single item. This would be fine, except that almost all of our actions are and likely will be done without a meeting.

## Solution

1. Give each of these their own entry.
2. Add the seven actions that the Bevy Foundation Board has since taken :)


content/foundation/meeting-minutes/buying-bevy-org-domain/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+So lets vote:
+
+Pay $5,000 for the domain,
+Take the domain for free now and enter a contract stating we'll pay $6,000 two years from now, or give it back,
+Do nothing / give up on the domain,
+
+My personal vote is (1). I think it is worth it, and while it isn't an inconsequential amount of money, it also isn't going to set us back meaningfully.
+
+**Proposed by:** Carter Anderson
```
cart: I think putting "proposed by" right after the Proposal (or combining it with that line), and then putting the proposal in "quote syntax" makes sense, in order to make who is speaking clear (given that you are the author / speaker by default in this context).

cart (approved):
This looks correct to me. Left one comment about format. Once thats resolved I'm on board!

content/foundation/meeting-minutes/budget-reduction-strategy/index.md
mockersf, line 1: we actually decided to not enact this one for now. while we agree on the intent, we wanted to take more time on the wording

mockersf (approved):

PR #2153, opened 2025-07-03
New book chapters from Unconf

theotherphil:
This is the branch that was worked on during rust week, but reverting the changes to `learn/book/_index.md` so that the book is hidden again.


theotherphil:
The CI failure is in unrelated files and is caused by the `uninlined_format_args` lint being switched from `Pedantic` to `Style` in Clippy 1.88.

alice-i-cecile:
@theotherphil what do you want to do before taking this out of draft? :)

viridia (approved):
I'm not going to read the whole thing, but the changes I made appear intact :)

PR #2155, opened 2025-07-04
Intro edits

viridia:
* Converted some technical asides into callouts.
* Got rid of the mysterious unexplained metavariable 'R'
* Minor rewordings.


alice-i-cecile (approved):
Really nice: this is much improved, and I like the use of callouts.

content/learn/book/intro/the-three-letters.md
```diff
@@ -35,7 +37,9 @@ In the "in-memory database" model, entities are the row keys in our database, wi
 While entities are conceptually similar to Objects in object-oriented engines, they are distinctly different bececause they **do not store any behavior**.
 This is handled by [systems](#the-s-systems).
 
+{% callout(type="note") %}
```
TrialDragon, line 40: A bit late of me, but fyi there is no "note" type for callouts, so this is defaulting to "info" type.

viridia: There are other uses of "note" already - I copied and pasted from those.

PR #2169, opened 2025-07-07
Improve book page on schedules

alice-i-cecile:
Tackling more sections of the book. I wasn't happy with the framing of the last section, and it didn't talk about the "run schedule on demand" pattern that's so useful, so I added a section!


NthTensor (approved):
Big improvement.

PR #2170, opened 2025-07-07
Update setup guide Rust edition to 2024

lwitzman:
Regarding `resolver` in virtual workspaces: Bevy with default plugins runs for me regardless of value, but cargo will warn and default to `resolver = "1"` if none is specified.

```
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2024 which implies `resolver = "3"`
note: to use the edition 2024 resolver, specify `workspace.resolver = "3"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
```

### DIFF
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -130,20 +130,6 @@ cargo add bevy
 
 Make sure to use the latest `bevy` crate version ([![Crates.io](https://img.shields.io/crates/v/bevy.svg)](https://crates.io/crates/bevy)).
 
-<details>
-  <summary>
-
-  ### Cargo Workspaces
-  </summary>
-
-  If you are using [Cargo Workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html), you will also need to add the resolver to your Cargo.toml file in the root directory:
-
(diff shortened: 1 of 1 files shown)


content/learn/quick-start/getting-started/setup.md
```diff
@@ -140,7 +140,7 @@ Make sure to use the latest `bevy` crate version ([![Crates.io](https://img.shie
 
   ```toml
   [workspace]
-  resolver = "2" # Important! wgpu/Bevy needs this!
+  resolver = "3"
```
alice-i-cecile: Do we still need to set this at all? I haven't been bothering.

alice-i-cecile: Ah, I see your note in the PR description.

alice-i-cecile (approved):

lwitzman:
Removed the workspace section as there are no Bevy-specific instructions.

janhohenheim (approved):
Yep, we don't need the resolver stuff. The default for 2021 is already 2, and the default for 2024 is 3. There's no reason why the version 3 should not work for us, speaking from experience :)

PR #2173, opened 2025-07-07
Rewrite Local page

alice-i-cecile:
This page wasn't particularly helpful in explaining why these exist, and its information about defaults was wrong. I've opted for a complete (but still brief) rewrite.

### DIFF
--- a/content/learn/book/storing-data/local-system-param.md
+++ b/content/learn/book/storing-data/local-system-param.md
@@ -10,7 +10,7 @@ There's one more place where the ECS can store data: on the systems themselves.
 With the help of the [`Local<T>`] system parameter, you can cache private data from one run of a system to the next.
 
 In most cases, a [resource] is going to be more flexible, as you can read and mutate it from other systems.
-But from time-to-time, you might want to:
+However, from time-to-time, you might want to:
 
 - store some system-specific state about the last value
 - maintain a cache of intermediate results to improve efficiency of computation
--- a/content/learn/book/storing-data/local-system-param.md
+++ b/content/learn/book/storing-data/local-system-param.md
@@ -7,7 +7,7 @@ status = 'hidden'
 +++
 
 There's one more place where the ECS can store data: on the systems themselves.
-With the help of the [`Local<T>`] system parameter, you can cache private data from one run of a system to the next.
+With the help of the [`Local<T>`] system parameter, you can cache data from one run of a system to the next.
 
 In most cases, a [resource] is going to be more flexible, as you can read and mutate it from other systems.
 However, from time to time, you might want to:
--- a/content/learn/book/storing-data/local-system-param.md
+++ b/content/learn/book/storing-data/local-system-param.md
@@ -1,15 +1,51 @@
 +++
-title = "Local System Parameters"
+title = "System-Local Data Storage"
 insert_anchor_links = "right"
 [extra]
 weight = 5
 status = 'hidden'
 +++
 
-Sometimes you may need to preserve some state between executions of a [system].
-For this, we have the [`Local<T>`] system parameter.
+There's one more place where the ECS can store data: on the systems themselves.
+With the help of the [`Local<T>`] system parameter, you can cache private data from one run of a system to the next.
 
-Any type that implements `Default` can be stored in a local. `Default` is required so that the value can be initialized before the system runs. If your type does not implement `Default`, you can use `Option<T>` instead.
+In most cases, a [resource] is going to be more flexible, as you can read and mutate it from other systems.
+But from time-to-time, you might want to:
 
-[system]: /learn/book/control-flow/systems
+- store some system-specific state about the last value
+- maintain a cache of intermediate results to improve efficiency of computation
+- avoid allocating a new large data structure each frame
+
+Bevy itself uses `Local` system params in two prominent places:
+
+1. As part of the [`EventReader`] abstraction, keeping track of which events each system has read.
+2. In run conditions like [`on_timer`], to track how much time has run.
+
+## Starting values
+
+`Local` system params are always initialized with a default value.
+This value is set by either the [`FromWorld`] trait or the [`Default`] trait.
+The [`FromWorld`] trait allows you to access arbitrary data from the world,
+allowing more complex initialization.
+
+If you can't figure out what a good default value might be, remember that [`Option<T>`]
+implements `Default`, even if `T` does not!
+
+```rust,hide_lines=1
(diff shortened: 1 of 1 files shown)


content/learn/book/storing-data/local-system-param.md
```diff
@@ -1,15 +1,51 @@
...
 +++
 
-Sometimes you may need to preserve some state between executions of a [system].
-For this, we have the [`Local<T>`] system parameter.
+There's one more place where the ECS can store data: on the systems themselves.
+With the help of the [`Local<T>`] system parameter, you can cache private data from one run of a system to the next.
 
-Any type that implements `Default` can be stored in a local. `Default` is required so that the value can be initialized before the system runs. If your type does not implement `Default`, you can use `Option<T>` instead.
+In most cases, a [resource] is going to be more flexible, as you can read and mutate it from other systems.
+But from time-to-time, you might want to:
```
BenjaminBrienen:
```suggestion
However, from time-to-time, you might want to:
```

BenjaminBrienen (approved):

content/learn/book/storing-data/local-system-param.md
```diff
@@ -1,15 +1,51 @@
...
 insert_anchor_links = "right"
 [extra]
 weight = 5
 status = 'hidden'
 +++
 
-Sometimes you may need to preserve some state between executions of a [system].
-For this, we have the [`Local<T>`] system parameter.
+There's one more place where the ECS can store data: on the systems themselves.
+With the help of the [`Local<T>`] system parameter, you can cache private data from one run of a system to the next.
```
janhohenheim:
```suggestion
With the help of the [`Local<T>`] system parameter, you can cache data from one run of a system to the next.
```
I don't like the word `private` here, as it may be conflated with visibility modifiers.

content/learn/book/storing-data/local-system-param.md
```diff
@@ -1,15 +1,51 @@
...
 
-Any type that implements `Default` can be stored in a local. `Default` is required so that the value can be initialized before the system runs. If your type does not implement `Default`, you can use `Option<T>` instead.
+In most cases, a [resource] is going to be more flexible, as you can read and mutate it from other systems.
+However, from time-to-time, you might want to:
```
janhohenheim:
```suggestion
However, from time to time, you might want to:
```

content/learn/book/storing-data/local-system-param.md
```diff
@@ -1,15 +1,51 @@
...
+# use bevy::prelude::*;
+
+struct NoGoodDefaultValue(u8);
+
+fn increment_local_system_data(mut local: Local<Option<NoGoodDefaultValue>>){
+    if local.is_none() {
+        *local = Some(NoGoodDefaultValue(0));
+    }
+    
+    local.unwrap().0 += 1;
```
janhohenheim, line 44:
```suggestion
fn increment_local_system_data(mut local: Local<Option<NoGoodDefaultValue>>){
    let local = local.get_or_insert_with(|| {
        NoGoodDefaultValue(0)
    });
    
    local.0 += 1;
```
Death to `unwrap`s!

janhohenheim (approved):
Minor suggestions, feel free to ignore :)

content/learn/book/storing-data/local-system-param.md
alice-i-cecile: This is better code, but less clear for learners :) The simple if statement does a really good job explaining the pattern, especially to Rust novices.

content/learn/book/storing-data/local-system-param.md
```diff
@@ -1,15 +1,51 @@
...
+}
```
gmorenz, line 45: This example doesn't compile for me, using either bevy 0.16.1 or the latest bevy from git.

It fails with

```
error[E0507]: cannot move out of dereference of `bevy::prelude::Local<'_, std::option::Option<NoGoodDefaultValue>>`
    --> src/lib.rs:9:5
     |
9    |     local.unwrap().0 += 1;
     |     ^^^^^ -------- value moved due to this method call
     |     |
     |     help: consider calling `.as_ref()` or `.as_mut()` to borrow the type's contents
     |     move occurs because value has type `std::option::Option<NoGoodDefaultValue>`, which does not implement the `Copy` trait
     |
note: `std::option::Option::<T>::unwrap` takes ownership of the receiver `self`, which moves value
    --> /home/greg/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1002:25
     |
1002 |     pub const fn unwrap(self) -> T {
     |                         ^^^^
help: you could `clone` the value and consume it, if the `NoGoodDefaultValue: Clone` trait bound could be satisfied
     |
9    |     <std::option::Option<NoGoodDefaultValue> as Clone>::clone(&local).unwrap().0 += 1;
     |     +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++     +
help: consider annotating `NoGoodDefaultValue` with `#[derive(Clone)]`
     |
3    + #[derive(Clone)]
4    | struct NoGoodDefaultValue(u8);
     |
```

This can be fixed via

1. Replacing `local.unwrap().0 += 1;` with `local.as_mut().unwrap().0 += 1;`
2. Taking janhohenheim's suggestion (this is a case where proper initialization forces the right pattern!)

The compiler error, but not the code, can be fixed via `#[derive(Clone, Copy)]`  on `NoGoodDefaultValue`. But then what the code does is *copy* the value out of local and add one to the copy, not increment the value in local. I.e. this test "passes"

```rust
use bevy::prelude::*;

#[derive(Copy, Clone)]
struct NoGoodDefaultValue(u8);
fn increment_local_system_data(mut local: Local<Option<NoGoodDefaultValue>>) -> u8 {
    if local.is_none() {
        *local = Some(NoGoodDefaultValue(0));
    }

    local.unwrap().0 += 1;
    local.unwrap().0
}

#[test]
fn test_lib() {
    let mut world = World::new();
    let one = world
        .run_system_cached(increment_local_system_data)
        .unwrap();
    let two = world
        .run_system_cached(increment_local_system_data)
        .unwrap();

    // We're just returning the initialized value unmodified.
    assert_eq!([one, two], [0, 0]);
}
```

content/learn/book/storing-data/local-system-param.md
Shatur: I don't know, I'd say the suggested `get_or_insert_with` is clearer - it's right there in the name :thinking:

Issue #2161, opened 2025-07-06
Book critique notes


viridia:
Reading through the whole text of The Book, and noting things as I go:

General:

* Missing sections for Graphics, Audio, User Interface, etc.

In `intro/what-is-bevy/`:

* "Bevy is a modern, ECS-first game engine written in Rust". This comes before the explanation of what ECS is. How about we just drop that bit and say "Bevy is a modern game engine written in Rust"

In `intro/the-three-letters/`:
* The "database metaphor" is mentioned in several chapters, including this `intro/the-three-letters/`, `storing-data/entities-components/` and `storing-data/world/`. But there are problems:
  * Each chapter talks about the concept as if it were being introduced for the first time, despite the fact that we've already talked about it in previous chapters.
  * There's a lot of redundancy and repetition in the different explanations.
  * The explanation in `storing-data/entities-components/` is the clearest of the three, however the one in `storing-data/world/` introduces the concept of archetypes.
  * We should trim down these sections to include only the new ideas and not re-explain the concept every time; we can refer to the database metaphor via link to previous sections.
  * There are also a couple of other places where the database metaphor is referenced, but are clearly written in a way that implies that the concept has already been introduced in previous sections. These are OK but should have links.

In `intro/the-three-letters/`
* Consider adding a final section, "Why ECS?". This can talk about the benefits of performance, and reuse via composability. We should probably also note that the maximum benefit of ECS can only be achieved by adopting an ECS mindset.

In `storing-data/local-system-param/`
* I would put this section last in `storing-data`, because this feature is so rarely used; at the very least it should come after resources and relations

In `control-flow/systems/`
* Lots of bad links
* Needs more examples, but fewer inline examples

In `control-flow/run-conditions/`
* Needs lots more examples

In `control-flow/commands/`
* Examples please

In `the-game-loop/app/`
* Show example of plugin registration

In `architecture/`
* I'd like to find a better word than "architecture" here, or at least call it "High-level architecture".

In `architecture/states/`
* I think this section would be better placed under "Control Flow" or "The Game Loop"

In `best-practices/`
* This section isn't really so much about best practices (how to achieve engineering quality), it's about developer happiness. Maybe rename to "Productive Development" or "Development Tips"?

@alice-i-cecile 
@theotherphil

viridia:
Thinking about this some more, I would rename the "architecture" section to "modular design", and move the "states" section under "control flow".

alice-i-cecile:
[...]

PR #2179, opened 2025-07-07
Write the book page for plugins

alice-i-cecile:
Tackles some of the criticism in #2161.

There are some serious limitations and footguns in our plugin architecture, but I figured we should actually teach new users that they exist / how to work around them, rather than letting them stumble over them on their own.

If it's embarrassing enough, maybe it will motivate us to actually fix them ;)

### DIFF
--- a/content/learn/book/modular-architecture/plugins.md
+++ b/content/learn/book/modular-architecture/plugins.md
@@ -73,6 +73,28 @@ The choice is mostly stylistic: the "fn plugin" style is a bit terser,
 but the "struct plugin" style is easier to change if you find that your plugin needs
 extra features in the future.
 
+## Configuring plugins
+
+As alluded to above, you can use the `&self` argument in [`Plugin::build`] to change
+the behavior of the plugin based on the passed in configuration.
+But *should* you use this pattern?
+
+Whenever possible, you should prefer using resources for configuration,
+alerting users to their presence via documentation,
+and responding dynamically to changes in their value.
+
+This is inherently more flexible, allowing your user to tweak the settings at runtime
+to account for changing needs.
+It *also* plays nicer with various tooling, allowing users
+to easily inspect and dynamically modify the values.
+
+But sometimes, the value passed in *only* make sense
+to be provided once, at the time of plugin definition.
+A relatively common example of this is allowing the user
+to configure which schedule your plugins' systems should be added to.
+In that case, using a struct with fields as your `Plugin` type
+is the best and only option.
+
 ## Plugin groups
 
 You may have noticed that [`App::add_plugins`] is a method that takes `&mut App`.
--- a/content/learn/book/modular-architecture/plugins.md
+++ b/content/learn/book/modular-architecture/plugins.md
@@ -37,7 +37,7 @@ Let's break that down:
 
 Let's see how this works in practice:
 
-```rust
+```rust,hide_lines=1-4
 # use bevy::prelude::*;
 #
 # fn check_if_player_is_dead() {}
@@ -57,7 +57,7 @@ App::new().add_plugins(PlayerPlugin);
 Because our plugin is so simple (it doesn't take any config, and only uses the `build` method),
 we could instead choose to use [the blanket for all functions that take a &mut App] instead:
 
-```rust
+```rust,hide_lines=1-4
 # use bevy::prelude::*;
 #
 # fn check_if_player_is_dead() {}
--- a/content/learn/book/modular-architecture/plugins.md
+++ b/content/learn/book/modular-architecture/plugins.md
@@ -67,6 +67,38 @@ extra features in the future.
 
 ## Plugin groups
 
+You may have noticed that [`App::add_plugins`] is a method that takes `&mut App`.
+Does that mean you can add plugins via other plugins?
+Yes, yes it does.
+
+You probably shouldn't do this though: it can make it harder to follow what's going on,
+and it becomes challenging to selectively disable plugins for testing or things like "making a server build".
+
+Instead, you probably want to use a [`PluginGroup`].
+You've likely already encountered these: [`DefaultPlugins`] is a [`PluginGroup`]!
+
+[`PluginGroup`] comes with a few extra niceties over simply recursively adding plugins:
+
+1. You can write dedicated docs for the collection of plugins.
+2. You can overwrite the values of contained plugins via [`PluginGroup::set`], changing the default config.
+3. With the help of [`PluginGroupBuilder`] you can enable and disable contained plugins cleanly.
+
+## Plugin ordering and dependencies
+
+When working with multiple plugins, be mindful that they're effectively just functions that immediately mutate the [`App`].
+As a result, plugins are [evaluated in the order that they are added to the `App`].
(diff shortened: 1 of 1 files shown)


content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,122 @@ insert_anchor_links = "right"
...
+```rust
+trait Plugin {
+    fn build(&self, app: &mut App) {}
+}
+```
+
+Let's break that down:
+
+- this is a trait, so we need to implement it for a user-defined type
+- `build` takes `&self`, allowing you to change behavior based on the value of our type
```
BenjaminBrienen: "our" type doesn't make sense to me here. Also, this whole bit seems confusing. The value of the type?

```suggestion
- `build` takes `&self`, allowing you to change the behavior based on the value.
```

BenjaminBrienen:
Looks good except I'm not quite grasping what that one line is trying to say.

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,122 @@ insert_anchor_links = "right"
...
+extra features in the future.
+
+## Plugin groups
+
+You may have noticed that [`App::add_plugins`] is a method that takes `&mut App`.
+Does that mean you can add plugins via other plugins?
+Yes, yes it does.
+
+You probably shouldn't do this though: it can make it harder to follow what's going on,
+and it becomes challenging to selectively disable plugins for testing or things like "making a server build".
```
janhohenheim: *Heavily* disagree with this. It's completely true in library code, but extremely useful and by now quite widespread in actual game code. It's also what the 2D template and various other community style guides explicitly recommend you to do for game code.

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,122 @@ insert_anchor_links = "right"
...
+is the least bad solution for now.
+
+Similarly, to avoid tricky bugs, adding the same plugin to your app multiple times will panic by default.
+This behavior can be overridden by overwriting the default [`Plugin::is_unique`] method.
+The same duck-typing solution can be used to check if the plugin already exists,
+and avoid re-adding it if another dependency has already pulled it in.
+
+## The `Plugin` lifecycle
+
+When a plugin is added though [`App::add_plugins`], the app calls `Plugin::build`, and the plugin typically accesses and configures the world.  Then, when the app is run, a few other plugin life-cycle functions are called, and finally we enter the run loop:
```
janhohenheim:
```suggestion
When a plugin is added though [`App::add_plugins`], the app calls `Plugin::build`, and the plugin typically accesses and configures the world.  
Then, when the app is run, a few other plugin life-cycle functions are called, and finally we enter the run loop:
```
I assume that's why the two spaces were in there ;)

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,122 @@ insert_anchor_links = "right"
...
+
+- The app polls `Plugin::finished` until all the added plugins return `true`.
+- The app calls `Plugin::finish` on all plugins.
```
janhohenheim, line 142: Maybe worth mentioning that `finish` is the only way for how you can get stuff out of the render world?

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,122 @@ insert_anchor_links = "right"
...
+    fn build(&self, app: &mut App) {}
+}
+```
+
+Let's break that down:
+
+- this is a trait, so we need to implement it for a user-defined type
+- `build` takes `&self`, allowing you to change behavior based on the value of our type
+- it takes a`&mut App` reference, allowing you to mutate the [`App`] state freely
+  - adding systems via [`App::add_systems`] is the most common and important method
```
janhohenheim, line 34: Maybe worth mentioning observers as well? I'd say they are equally as important in game code (not library code)

janhohenheim (changes requested):
Looks good other than one disagreement. The other comments are not blocking :)
In general: great work you're doing with these book PRs!

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,122 @@ insert_anchor_links = "right"
...
+
+Because our plugin is so simple (it doesn't take any config, and only uses the `build` method),
+we could instead choose to use [the blanket for all functions that take a &mut App] instead:
+
+```rust
+# use bevy::prelude::*;
+#
+# fn check_if_player_is_dead() {}
+#
+fn player_plugin(app: &mut App){
```
janhohenheim, line 65: thanks for pointing out my beloved syntax :D

viridia:
I'd like to elaborate on what plugins actually do:

---

Bevy Plugins only run during app initialization. Their job is to set up all the things: the ECS systems, resources, observers, hooks, and other miscellanea needed to support a particular library or feature. These things, once they have been registered and installed, will outlive the Plugin which created them.

content/learn/book/modular-architecture/plugins.md
alice-i-cecile: I'd like to avoid pulling in new concepts here if possible.

content/learn/book/modular-architecture/plugins.md
alice-i-cecile: That's a problem for the rendering chapter IMO.

viridia (approved):

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,157 @@ insert_anchor_links = "right"
...
-When a plugin is added though `App::add_plugins`, the app calls `Plugin::build`, and the plugin typically accesses and configures the world.  Then, when the app is run, a few other plugin life-cycle functions are called, and finally we enter the run loop:
-+ The app polls `Plugin::finished` until all the added plugins return `true`.
-+ The app calls `Plugin::finish` on all plugins.
-+ The app calls `Plugin::cleanup` an all plugins. 
-+ The app calls the run loop function on itself.
+Plugins are best used to organize your code into functional units:
+`WorldGenPlugin`, `InventoryPlugin` or `AudioPlugin` are all plausible.
+Inside of each plugin, you can initialize resources,
+add systems, setup observers, register types and generally handle the setup needed
+to make your subsystem function.
```
janhohenheim, line 16: Nit: a beginner might think that "subsystem" is related to ECS systems.
Even worse: they might read "function" as a noun here!

janhohenheim (approved):
Looks fantastic now!

content/learn/book/modular-architecture/plugins.md
```diff
@@ -5,10 +5,157 @@ insert_anchor_links = "right"
...
+## Plugin groups
+
+You may have noticed that [`App::add_plugins`] is a method that takes `&mut App`.
+Does that mean you can add plugins via other plugins?
+Yes, yes it does.
+
+For game code, this can be really convenient,
+allowing you to readily group functionality into a nice hierarchy.
+
+If you're writing a library though, you probably shouldn't do nest plugins like this.
```
viridia, line 107: "shouldn't do nest plugins" - extra word in there.

PR #2187, opened 2025-07-11
Add book page on time, fixed time and timers

alice-i-cecile:
Some notes for reviewers:

- I've opted not to try and explain how to do delayed actions, waiting for a fix via https://github.com/bevyengine/bevy/issues/15129
- the lack of first-party interpolation seriously sucks (see https://github.com/bevyengine/bevy/issues/1259), but it's important to call out the problem. I've opted to lay out the broad shape of the solution, and hint "go look for `bevy_transform_interpolation`" without drawing controversy by straight-up linking it

### DIFF
--- a/content/learn/book/the-game-loop/time-and-timers.md
+++ b/content/learn/book/the-game-loop/time-and-timers.md
@@ -102,33 +102,59 @@ Alternatively, [states] and [run conditions] can be used to skip systems while t
 
 ## Fixing your timestep
 
-FIX YOUR TIMESTEP ARTICLE.
-When thinking about fixed time, it's important to clearly distinguish **frames** and **ticks**.
+Compensating for fluctating frame times using delta time is a great start.
+But as the timeless [fix your timestep article] explains for projects that require a higher level of stability and reproducibility,
+it's better to simply always advance time by a fixed amount.
+This is particularly important for physics and
 
-To understand how to work with fixed time, we need to understand a little bit about how [`Time`] actually works under the hood.
+To understand how to work with fixed time in Bevy, we need to first learn a little bit about how [`Time`] actually works under the hood.
 As the excellent docs on [`Time`] explain, there's actually *three* distinct types of time being measured:
 
 - real time: the actual wall time
   - use this for things like UI animations that you don't want to be affected by pausing
 - virtual time: the "in-game time"
   - if you're using fixed time, this is useful for graphical effects
-  - otherwise, this is used for all of your gameplay logic
+  - otherwise, this is used for all of your gameplay logic as well
 - fixed time
-  - advances by a [`Time::timestep`] every time [`FixedMain`] runs
-  - [`FixedMain`] will run repeatedly until it has caught up with the elapsed virtual time
-  - if not enough time has elapsed, [`FixedMain`] will not run at all for the frame
-  - please read the excellent docs on [`Fixed`] for more details!
+  - all of your gameplay logic should go here if you're using a fixed timestep approach
 
-Simply requesting [`Time`] in your systems will get you the correct flavor 90% of the time: virtual time in the main schedule,
-and fixed time in the fixed main schedule.
+When thinking about fixed time, it's important to clearly distinguish **frames** and **ticks**.
+A "frame" is one pass of the [`Main`] schedule, and corresponds to a single rendered frames.
+A "tick" is one pass of the [`FixedMain`] schedule, and corresponds to one iteration of gameplay logic.
+The ratio between frames elapsed and ticks elapsed is not constant: they may be faster or slower, depending
+on the rendering performance and your game's simulation needs.
+
+{% callout(type="info") %}
+Simply requesting [`Time`] in your systems will get you the correct flavor 90% of the time: virtual time in the [`Main`] schedule,
+and fixed time in the [`FixedMain`] schedule.
 To request a specific variation, change the implicit generic in [`Time`] from `()`
 to [`Real`], [`Virtual`] or [`Fixed`].
+{% end %}
 
-[`Time::timestep`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html#method.timestep
+Now that we have the required vocabulary, let's go over exactly how the fixed timestep logic works in Bevy:
+
+- each frame, during the [`RunFixedMainLoop`] schedule, we determine how many times we should run the [`FixedMain`] schedule
+  - we add the the elapsed virtual time for the frame to a running time buffer
+  - if that time buffer is greater than or equal to [`Time<Fixed>::timestep`], we run the [`FixedMain`] schedule once, and subtract that timestep from our buffer
+- when the [`FixedMain`] schedule is evaluated, the various subschedules are iterated through, running [`FixedPreUpdate`], [`FixedUpdate`] and so on in order
+- within your fixed time systems, use the exact same delta time techniques to properly account for changes in your desired timestep during development
+
+This means that we may have 0, 1, 2 or more ticks per frame,
+with our fixed timestep logic running repeatedly until it's caught back up.
+For even more detail, check out the excellent documentation on [`Fixed`].
+
+Note that Bevy's "fixed timesteps" are not the right mechanism to use for gameplay logic like "every 5 seconds update this building".
+Bevy only supports a single fixed timestep across your entire project, and its use is completely optional.
+Please see our [timers and cooldowns] section below for the tools available to model this type of behavior.
+
+[`Time<Fixed>::timestep`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html#method.timestep
+[`Main`]: https://docs.rs/bevy/latest/bevy/app/struct.Main.html
 [`FixedMain`]: https://docs.rs/bevy/latest/bevy/app/struct.FixedMain.html
 [`Real`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Real.html
 [`Virtual`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Virtual.html
 [`Fixed`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Fixed.html
+[fix your timestep article]: https://gafferongames.com/post/fix_your_timestep/
+[`RunFixedMainLoop`]: https://dev-docs.bevy.org/bevy/app/struct.RunFixedMainLoop.html
 
 ## Interpolation and smooth movement
 
--- a/content/learn/book/the-game-loop/time-and-timers.md
+++ b/content/learn/book/the-game-loop/time-and-timers.md
@@ -1,0 +1,18 @@
++++
+title = "Time and Timers"
+insert_anchor_links = "right"
+[extra]
+weight = 5
++++
+
+## Timers
+
+## Cooldowns
+
+## Delayed actions
+
+## Delta time
+
+## Fixed time
+
+## Interpolation best practices
--- a/content/learn/book/the-game-loop/time-and-timers.md
+++ b/content/learn/book/the-game-loop/time-and-timers.md
@@ -171,7 +171,7 @@ The core strategy is simple enough, but the devil is in the details:
 
 1. Create your own custom `GamePosition` type, holding the translation/rotation/scale information that your project needs.
 2. Modify and read this `GamePosition` type for all of your game logic and physics.
-3. Interpolate/extrapolate between the previous [`GlobalTransform`] and the `GamePosition`, based on the elapsed [`Fixed`] time.
+3. Interpolate between the previous [`GlobalTransform`] and the `GamePosition`, based on the elapsed [`Fixed`] time.
    1. This should occur after the final fixed update for the frame has run, but before rendering occurs.
    2. You ultimately want to set the [`GlobalTransform`], but doing this correctly in the presence of hierarchies is hard.
    3. As a result, modifying the local [`Transform`] and relying on transform propagation can be the least bad solution.
--- a/content/learn/book/the-game-loop/time-and-timers.md
+++ b/content/learn/book/the-game-loop/time-and-timers.md
@@ -5,14 +5,108 @@ insert_anchor_links = "right"
 weight = 5
 +++
 
-## Timers
+Responding to the passage of time is essential for gameplay, animations and audio.
+But doing so correctly is surprisingly subtle.
 
-## Cooldowns
+The [`Time`] resource in Bevy is the source of truth
+for the "current in-game time".
+Unlike [`Instant::now()`], the [`Time`] is fixed for the entire frame,
+rather than reflecting the precise wall clock time at the time of system evaluation.
+This value is set at the start of the frame based on the time supplied by the renderer,
+during the [`TimeSystem`] system set in the [`First`] schedule.
+
+This is helpful for performance reasons, but more critically,
+it ensures consistency of behavior across all of the various bits of game logic.
+
+[`Time`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html
+[`Instant::now()`]: https://doc.rust-lang.org/std/time/struct.Instant.html#method.now
+[`TimeSystem`]: https://docs.rs/bevy/latest/bevy/time/struct.TimeSystem.html
+[`First`]: https://docs.rs/bevy/latest/bevy/app/struct.First.html
+
+## Frame-rate independence and delta time
+
+Suppose we want to move our player to the right:
+
+```rust
+use bevy::prelude::*;
+
+#[derive(Component)]
+struct Player;
+
+fn move_player(mut player_transform: Single<&mut Transform, With<Player>>){
+    const PLAYER_MOVEMENT: f32 = 1.;
+    
+    player_transform.translation.x += PLAYER_MOVEMENT;
+}
+```
+
+Every frame, our player will move 1 unit to the right. Great!
+But what happens when our game stutters, and the frame rate drops?
+Suddenly, rather than moving 60 units per second at our target 60 frames per second,
+our player is moving at an unsteady 20-30 units per second. Oh no!
+
+Instead, we can compensate for this effect by fixing our *speed* (or other rates of change per second),
+and then multiplying by the elapsed time.
+
+```rust, hide_lines=1-4
+# use bevy::prelude::*;
+# 
+# #[derive(Component)]
+# struct Player;
+#
+fn move_player(mut player_transform: Single<&mut Transform, With<Player>>, time: Res<Time>){
+    // At 60 FPS, this will be the same as before
+    const PLAYER_SPEED: f32 = 60.;
+    
+    player_transform.translation.x += PLAYER_SPEED * time.delta_secs_f32();
+}
+```
+
+This technique is commonly known as "delta time" among game devs, because physicists use "delta" to mean "a change in a quantity".
+
+## Pausing and time control
+
+With all of our time-dependent logic driven by the delta time,
+we can start playing tricks with the value of [`Time`] to implement features like pausing and slowing down the game.
+
+```rust
+use bevy::prelude::*;
+
+
+```
+
+## Fixing your timestep
+
+FIX YOUR TIMESTEP ARTICLE.
+When thinking about fixed time, it's important to clearly distinguish **frames** and **ticks**.
+
+To understand how to work with fixed time, we need to understand a little bit about how [`Time`] actually works under the hood.
+As the excellent docs on [`Time`] explain, there's actually *three* distinct types of time being measured:
+
+- real time: the actual wall time
+  - use this for things like UI animations that you don't want to be affected by pausing
+- virtual time: the "in-game time"
+  - if you're using fixed time, this is useful for graphical effects
+  - otherwise, this is used for all of your gameplay logic
+- fixed time
+  - advances by a [`Time::timestep`] every time [`FixedMain`] runs
+  - [`FixedMain`] will run repeatedly until it has caught up with the elapsed virtual time
+  - if not enough time has elapsed, [`FixedMain`] will not run at all for the frame
+  - please read the excellent docs on [`Fixed`] for more details!
+
+Simply requesting [`Time`] in your systems will get you the correct flavor 90% of the time: virtual time in the main schedule,
+and fixed time in the fixed main schedule.
+To request a specific variation, change the implicit generic in [`Time`] from `()`
+to [`Real`], [`Virtual`] or [`Fixed`].
+
+[`Time::timestep`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html#method.timestep
+[`FixedMain`]: https://docs.rs/bevy/latest/bevy/app/struct.FixedMain.html
+[`Real`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Real.html
+[`Virtual`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Virtual.html
+[`Fixed`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Fixed.html
+
+## Interpolation and smooth movement
+
+## Timers and cooldowns
 
 ## Delayed actions
-
-## Delta time
-
-## Fixed time
-
-## Interpolation best practices
--- a/content/learn/book/the-game-loop/time-and-timers.md
+++ b/content/learn/book/the-game-loop/time-and-timers.md
@@ -169,9 +169,9 @@ games need to handle it.
 
 The core strategy is simple enough, but the devil is in the details:
 
-1. Create your own custom `GamePosition` type, holding the translation/rotation/scale information that your project needs.
-2. Modify and read this `GamePosition` type for all of your game logic and physics.
-3. Interpolate between the previous [`GlobalTransform`] and the `GamePosition`, based on the elapsed [`Fixed`] time.
+1. Create your own custom `GameTransform` type, holding the translation/rotation/scale information that your project needs.
+2. Modify and read this `GameTransform` type for all of your game logic and physics.
+3. Interpolate between the previous [`GlobalTransform`] and the `GameTransform`, based on the elapsed [`Fixed`] time.
    1. This should occur after the final fixed update for the frame has run, but before rendering occurs.
    2. You ultimately want to set the [`GlobalTransform`], but doing this correctly in the presence of hierarchies is hard.
    3. As a result, modifying the local [`Transform`] and relying on transform propagation can be the least bad solution.
--- a/content/learn/book/the-game-loop/time-and-timers.md
+++ b/content/learn/book/the-game-loop/time-and-timers.md
@@ -62,18 +62,43 @@ fn move_player(mut player_transform: Single<&mut Transform, With<Player>>, time:
 }
 ```
 
-This technique is commonly known as "delta time" among game devs, because physicists use "delta" to mean "a change in a quantity".
+This technique is commonly known as ["delta time"] among game devs, because physicists use "delta" to mean "a change in a quantity".
+
+["delta time"]: https://en.wikipedia.org/wiki/Delta_timing
 
 ## Pausing and time control
 
 With all of our time-dependent logic driven by the delta time,
 we can start playing tricks with the value of [`Time`] to implement features like pausing and slowing down the game.
+To do this, we need to change how we account for [`Virtual`] (in-game) time:
 
 ```rust
 use bevy::prelude::*;
 
+fn toggle_pause(mut time: ResMut<Time<Virtual>>) {
+    if time.is_paused() {
+        time.unpause();
+    } else {
+        time.pause();
+    }
+}
 
+#[derive(Event)]
+struct SetGameSpeed(f32);
+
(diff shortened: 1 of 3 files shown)


content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+  - we add the the elapsed virtual time for the frame to a running time buffer
+  - if that time buffer is greater than or equal to [`Time<Fixed>::timestep`], we run the [`FixedMain`] schedule once, and subtract that timestep from our buffer
+- when the [`FixedMain`] schedule is evaluated, the various subschedules are iterated through, running [`FixedPreUpdate`], [`FixedUpdate`] and so on in order
+- within your fixed time systems, use the exact same delta time techniques to properly account for changes in your desired timestep during development
+
+This means that we may have 0, 1, 2 or more ticks per frame,
+with our fixed timestep logic running repeatedly until it's caught back up.
+For even more detail, check out the excellent documentation on [`Fixed`].
+
+Note that Bevy's "fixed timesteps" are not the right mechanism to use for gameplay logic like "every 5 seconds update this building".
```
mgi388, line 150: Is it possible to explain the “why not right” here? And what to do instead? Related to this are the timer examples at the end. But you don’t say what schedule to put those systems in which update timers.

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+struct SetGameSpeed(f32);
+
+fn set_game_speed(mut time: ResMut<Time<Virtual>>, events: EventReader::<SetGameSpeed>){
+  if let Some(new_speed) = events.iter().last(){
+      time.set_relative_speed(new_speed.0);
+  }
+}
+```
+
+If your systems uniformly rely on [`Time`], this will effect your entire game:
```
mgi388:
```suggestion
If your systems uniformly rely on [`Time`], this will affect your entire game:
```

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+halting, slowing or speeding up animations, movement, projectiles, physics and so on.
+Alternatively, [states] and [run conditions] can be used to skip systems while the game is paused.
+
+[states]: ../control-flow/states.md
+[run conditions]: ../control-flow/run-conditions.md
+
+## Fixing your timestep
+
+Compensating for fluctating frame times using delta time is a great start.
```
onbjerg:
```suggestion
Compensating for fluctuating frame times using delta time is a great start.
```

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+But as the timeless [fix your timestep article] explains for projects that require a higher level of stability and reproducibility,
```
onbjerg: this is a stylistic nit, but I think as a reader I am used to book names, articles and so on that are referred to by name to be written as the title is, maybe italicized:


```suggestion
But as the timeless [*Fix Your Timestep!*] article explains for projects that require a higher level of stability and reproducibility,
```

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+it's better to simply always advance time by a fixed amount.
+This is particularly important for physics and
```
onbjerg: sentence stops midway

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+[`Transform`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Transform.html
+
+## Timers and cooldowns
+
+Delta time is great for physics and animations,
+but what if we want to implement other time-driven gameplay logic?
+We might want to automatically generate cookie clicks every 5 seconds
+or add a 3 second cooldown for our fireball ability.
+
+The simplest of these tools is the [`Timer`].
```
onbjerg, line 207: I've started diving into Bevy again recently and one thing that wasn't really clear to me when using `Timer`s is whether or not I am responsible for handling the case where a repeating timer elapses multiple times per tick, just how the fixed time system set might run multiple times to catch up. I'm not sure if it is appropriate to cover in this book section, but something to think about.

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+Please see our [timers and cooldowns] section below for the tools available to model this type of behavior.
+
+[`Time<Fixed>::timestep`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html#method.timestep
+[`Main`]: https://docs.rs/bevy/latest/bevy/app/struct.Main.html
+[`FixedMain`]: https://docs.rs/bevy/latest/bevy/app/struct.FixedMain.html
+[`Real`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Real.html
+[`Virtual`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Virtual.html
+[`Fixed`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Fixed.html
+[fix your timestep article]: https://gafferongames.com/post/fix_your_timestep/
+[`RunFixedMainLoop`]: https://dev-docs.bevy.org/bevy/app/struct.RunFixedMainLoop.html
```
janhohenheim: You probably wanted docs.rs ;)

janhohenheim (changes requested):
A lot of the information here is also present in the [physics in fixed timestep example](https://bevy.org/examples/movement/physics-in-fixed-timestep/) on our example site. Could we cross link these? I want to make sure that people stumbling on one know that the other explanation exists. Since this book is still WIP, I wouldn't link to it from the example yet, but we can link to the example in here.

Also, I *think* we use the same wording and instructions in both, but we should double check so that readers don't get mixed signals.

Update after reading this in detail: I don't think the sections for the fixed timestep and interpolation are actionable for people who don't already know how to do this. Which is fine, there is a good practical example already, just link it aggressively :)

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+
+```rust, hide_lines=1-4
+# use bevy::prelude::*;
+# 
+# #[derive(Component)]
+# struct Player;
+#
+fn move_player(mut player_transform: Single<&mut Transform, With<Player>>, time: Res<Time>){
+    // At 60 FPS, this will be the same as before
+    const PLAYER_SPEED: f32 = 60.;
```
janhohenheim, line 63: This has an intuitive unit in 2D: pixels per second. In 3D it's meters per second, not generic units, as our renderer assumes SI units.
Could we make this more explicit? I believe that would teach better how game devs think about this in practice when creating these otherwise magic numbers.

As in, when I decide on an initial speed I wonder "how many meters should this zombie traverse in one second?" and then iterate on that. In contrast, the approach used here implicitly teaches "grab a speed value out of a hat and multiply it by 60"

Bonus: the average walking speed for humans is about 1.5 m/s, which is useful to know in 3D, but prooooobably too off topic for this section.

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+
+Compensating for fluctating frame times using delta time is a great start.
+But as the timeless [fix your timestep article] explains for projects that require a higher level of stability and reproducibility,
+it's better to simply always advance time by a fixed amount.
+This is particularly important for physics and
+
+To understand how to work with fixed time in Bevy, we need to first learn a little bit about how [`Time`] actually works under the hood.
+As the excellent docs on [`Time`] explain, there's actually *three* distinct types of time being measured:
+
+- real time: the actual wall time
```
janhohenheim: Wall?

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+  - use this for things like UI animations that you don't want to be affected by pausing
+- virtual time: the "in-game time"
+  - if you're using fixed time, this is useful for graphical effects
+  - otherwise, this is used for all of your gameplay logic as well
+- fixed time
+  - all of your gameplay logic should go here if you're using a fixed timestep approach
```
janhohenheim, line 123: No mention of the camera scheduling? Most 3D devs and some 2D devs will run into this *very* quickly, and right now we are setting them up for failure by either letting them put camera code in fixed updates (= extremely choppy) or in Update (= rotation delayed by a frame, which is really bad for players at 30 FPS or lower).

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+2. Modify and read this `GamePosition` type for all of your game logic and physics.
+3. Interpolate between the previous [`GlobalTransform`] and the `GamePosition`, based on the elapsed [`Fixed`] time.
+   1. This should occur after the final fixed update for the frame has run, but before rendering occurs.
+   2. You ultimately want to set the [`GlobalTransform`], but doing this correctly in the presence of hierarchies is hard.
+   3. As a result, modifying the local [`Transform`] and relying on transform propagation can be the least bad solution.
+
+Bevy does not currently offer any built-in functionality for this form of interpolation,
+but open source [ecosystem crates] are available to use and learn from.
+
+[ecosystem crates]: https://bevy.org/assets/
```
janhohenheim, line 194: I would personally directly link to the crate we are all thinking about. Don't think this is controversial. The only other community crate for interpolation that was in use was mine, and I intentionally shelved it in favor of Jondolfs :p

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Fixing your timestep
+
+Compensating for fluctating frame times using delta time is a great start.
+But as the timeless [fix your timestep article] explains for projects that require a higher level of stability and reproducibility,
+it's better to simply always advance time by a fixed amount.
+This is particularly important for physics and
+
+To understand how to work with fixed time in Bevy, we need to first learn a little bit about how [`Time`] actually works under the hood.
+As the excellent docs on [`Time`] explain, there's actually *three* distinct types of time being measured:
```
janhohenheim: From the perspective of a reader, both of these documents were written by "Bevy", so this is like that meme of Obama handing himself a medal lol

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+This will lead to visible jittering, lagginess and tiny speedups.
+
+To account for this, we need to distinguish between the logical and rendered position (and rotation, and sometimes scale) of
+game objects.
+This is a common problem and pattern for multiplayer games, but using a fixed timestep means that even single player
+games need to handle it.
+
+The core strategy is simple enough, but the devil is in the details:
+
+1. Create your own custom `GamePosition` type, holding the translation/rotation/scale information that your project needs.
```
janhohenheim: I wouldn't call this `Position`, as that is often used for translation (see Avian), but we also care about rotation.

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+2. Modify and read this `GamePosition` type for all of your game logic and physics.
+3. Interpolate between the previous [`GlobalTransform`] and the `GamePosition`, based on the elapsed [`Fixed`] time.
+   1. This should occur after the final fixed update for the frame has run, but before rendering occurs.
```
janhohenheim, line 183: Link to the correct schedule here, see the fixed timestep example I mentioned earlier. Jondolf's crate uses the same schedule. 

(Not linking it myself because I'm on my phone and browsing code is hard, sry! It's called something like `RunFixedMainLoop::AfterFixedMainLoop`)

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+If your systems uniformly rely on [`Time`], this will effect your entire game:
+halting, slowing or speeding up animations, movement, projectiles, physics and so on.
+Alternatively, [states] and [run conditions] can be used to skip systems while the game is paused.
+
+[states]: ../control-flow/states.md
+[run conditions]: ../control-flow/run-conditions.md
+
+## Fixing your timestep
```
janhohenheim, line 107: In practice, most users will probably run into fixed timesteps when they first use a physics engine. Maybe add a disclaimer that when using a physics engine, they will use fixed timesteps by default?

NB: Avian uses the Bevy schedules as expected, Rapier uses their own custom fixed timestep implementation in… PostUpdate, I believe?

content/learn/book/the-game-loop/time-and-timers.md
Jondolf: Probably meant "wall clock time"

content/learn/book/the-game-loop/time-and-timers.md
Jondolf: Rapier uses a variable time step in `PostUpdate` [by default](https://docs.rs/bevy_rapier3d/latest/src/bevy_rapier3d/plugin/configuration.rs.html#53), they just cap the maximum delta time to something like `1.0 / 60.0` seconds to avoid huge jumps in time. I don't think they have any custom fixed time step thing, you can opt in to running in `FixedUpdate` with `RapierPhysicsPlugin::<()>::default().in_fixed_schedule()`

Jondolf: But yes Avian runs in `FixedPostUpdate` by default, and interpolation should Just Work™ with the `TransformInterpolation` component provided by `bevy_transform_interpolation`. In Rapier you need to do a bit more setup there

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+# use bevy::prelude::*;
+# 
+# #[derive(Component)]
+# struct Player;
+#
+fn move_player(mut player_transform: Single<&mut Transform, With<Player>>, time: Res<Time>){
+    // At 60 FPS, this will be the same as before
+    const PLAYER_SPEED: f32 = 60.;
+    
+    player_transform.translation.x += PLAYER_SPEED * time.delta_secs_f32();
```
Jondolf: `delta_secs_f32` doesn't exist, there's [`delta_secs`](https://dev-docs.bevy.org/bevy/prelude/struct.Time.html#method.delta_secs) and [`delta_secs_f64`](https://dev-docs.bevy.org/bevy/prelude/struct.Time.html#method.delta_secs_f64)

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+    } else {
+        time.pause();
+    }
+}
+
+#[derive(Event)]
+struct SetGameSpeed(f32);
+
+fn set_game_speed(mut time: ResMut<Time<Virtual>>, events: EventReader::<SetGameSpeed>){
+  if let Some(new_speed) = events.iter().last(){
```
Jondolf, line 90: `jondolf fmt`
```suggestion
  if let Some(new_speed) = events.iter().last() {
```

content/learn/book/the-game-loop/time-and-timers.md
Jondolf: Personally, I'd also explicitly credit the author when referencing their article in a public book-like context like this

```suggestion
But as the timeless [*Fix Your Timestep!*] article by Glenn Fiedler explains, for projects that require a higher level of stability and reproducibility,
```

content/learn/book/the-game-loop/time-and-timers.md
Jondolf: Not super important to mention, but you can also define your own clocks (though it will require some manual shenanigans). For example, Avian uses `Time<Physics>` within the `PhysicsSchedule`, allowing physics to be advanced independently of the rest of the application.

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+
+- each frame, during the [`RunFixedMainLoop`] schedule, we determine how many times we should run the [`FixedMain`] schedule
+  - we add the the elapsed virtual time for the frame to a running time buffer
+  - if that time buffer is greater than or equal to [`Time<Fixed>::timestep`], we run the [`FixedMain`] schedule once, and subtract that timestep from our buffer
+- when the [`FixedMain`] schedule is evaluated, the various subschedules are iterated through, running [`FixedPreUpdate`], [`FixedUpdate`] and so on in order
+- within your fixed time systems, use the exact same delta time techniques to properly account for changes in your desired timestep during development
+
+This means that we may have 0, 1, 2 or more ticks per frame,
+with our fixed timestep logic running repeatedly until it's caught back up.
+For even more detail, check out the excellent documentation on [`Fixed`].
```
Jondolf: Same as what Jan said earlier, Bevy referring to its own documentation as "excellent" in its own book feels quite weird. Something like "extensive" or "comprehensive" would be more fitting, or we could even just drop the adjective altogether.

content/learn/book/the-game-loop/time-and-timers.md
Jondolf: (this is responding to Jan's message, GitHub decided to copy this to be an out-of-context review comment outside the thread)

Depends on which version these docs are targeting. It's called [`RunFixedMainLoopSystem`](https://docs.rs/bevy/latest/bevy/app/enum.RunFixedMainLoopSystem.html) in 0.16, and `RunFixedMainLoop` on the main branch / in 0.17.

---

Side tangent, these links use `docs.rs/bevy/latest`; we'll want to switch the book's links to use a specific version like 0.16 at some point so that links aren't broken for old builds of the book, especially if we make the book versioned. Maybe we could even have a CI test that denies docs.rs links with `/latest/` to avoid links like that accidentally slipping through?

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+){
+  let delta_time = time.delta();
+
+  for mut cookie_clicker in query.iter_mut(){
+    cookie_clicker.timer.tick(delta_time);
+    if cookie_clicker.timer.just_finished(){
+      cookies.0 += cookie_clicker.resources_gained;
+    }
+  }
+}
```
Jondolf: `jondolf fmt`
```suggestion
#[derive(Component)]
struct AutomaticCookieClick {
    // This should be initialized as a repeating timer
    // ensuring it automatically resets
    timer: Timer,
    resources_gained: u64,
}

fn automatically_click_cookies(
    mut cookies: ResMut<Cookies>, 
    mut query: Query<&mut AutomaticCookieClick>,
    time: Res<Time>,
) {
    let delta_time = time.delta();

    for mut cookie_clicker in query.iter_mut() {
        cookie_clicker.timer.tick(delta_time);
        if cookie_clicker.timer.just_finished() {
            cookies.0 += cookie_clicker.resources_gained;
        }
    }
}
```
Side note, I bet we could get CI to automatically check formatting for code snippets

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+
+fn update_cooldowns(time: Res<Time>, mut cooldowns: Query<&mut Cooldown>) {
+  let delta_time = time.delta();
+  for mut cooldown in cooldowns.iter_mut(){
+    // We never want our remaining time to become negative
+    cooldown.remaining.saturating_sub(delta_time);
+  }
+}
+```
```
Jondolf: `jondolf fmt`

```suggestion
# use bevy::prelude::*;
#
#[derive(Relationship)]
#[relationship(relationship_target = Abilities)]
struct AbilityOf(Entity);

#[derive(RelationshipTarget)]
#[relationship_target(relationship = AbilityOf)]
struct Abilities(Vec<Entity>);

#[derive(Component)]
struct Cooldown {
    duration: Duration,
    remaining: Duration,
}

impl Cooldown {
    fn expend(&mut self) {
        self.remaining = self.duration;
    }
  
    fn is_ready(&self) -> bool {
        self.remaining == Duration::ZERO;
    }
}

fn update_cooldowns(time: Res<Time>, mut cooldowns: Query<&mut Cooldown>) {
    let delta_time = time.delta();
    for mut cooldown in cooldowns.iter_mut(){
        // We never want our remaining time to become negative
        cooldown.remaining.saturating_sub(delta_time);
    }
}
```
```

File: content/learn/book/the-game-loop/time-and-timers.md [hunk 21]
```diff
@@ -0,0 +1,277 @@
...
+
+Responding to the passage of time is essential for gameplay, animations and audio.
+But doing so correctly is surprisingly subtle.
+
+The [`Time`] resource in Bevy is the source of truth
+for the "current in-game time".
+Unlike [`Instant::now()`], the [`Time`] is fixed for the entire frame,
+rather than reflecting the precise wall clock time at the time of system evaluation.
+This value is set at the start of the frame based on the time supplied by the renderer,
+during the [`TimeSystem`] system set in the [`First`] schedule.
```
Jondolf: Not sure how I'd reword this, but `Time` isn't exactly fixed for the entire frame, since it uses a different clock depending on the schedule, like is described later on. I understand what the intent here is though

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+}
+
+#[derive(Event)]
+struct SetGameSpeed(f32);
+
+fn set_game_speed(mut time: ResMut<Time<Virtual>>, events: EventReader::<SetGameSpeed>){
+  if let Some(new_speed) = events.iter().last(){
+      time.set_relative_speed(new_speed.0);
+  }
+}
```
Jondolf: `jondolf fmt`

```suggestion
fn set_game_speed(mut time: ResMut<Time<Virtual>>, events: EventReader<SetGameSpeed>) {
    if let Some(new_speed) = events.iter().last() {
        time.set_relative_speed(new_speed.0);
    }
}
```

content/learn/book/the-game-loop/time-and-timers.md
janhohenheim: image

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,277 @@
...
+  - use this for things like UI animations that you don't want to be affected by pausing
+- virtual time: the "in-game time"
+  - if you're using fixed time, this is useful for graphical effects
+  - otherwise, this is used for all of your gameplay logic as well
+- fixed time
+  - all of your gameplay logic should go here if you're using a fixed timestep approach
+
+When thinking about fixed time, it's important to clearly distinguish **frames** and **ticks**.
+A "frame" is one pass of the [`Main`] schedule, and corresponds to a single rendered frames.
+A "tick" is one pass of the [`FixedMain`] schedule, and corresponds to one iteration of gameplay logic.
```
viridia, line 127: In my reactive frameworks, I advance the tick counter more often than this. Quill's "run to convergence" algorithm uses a loop which may advance the tick more than once in a single system execution (because I'm looking to detect fine-grained changes).

janhohenheim: Is that not the same? A single frame can run as many ticks as it needs to "catch up" in Bevy's fixed schedule.

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile:
```suggestion
    player_transform.translation.x += PLAYER_SPEED * time.delta_secs();
```

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile:
```suggestion
This is particularly important for physics and networking.
```

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile:
```suggestion
As the docs on [`Time`] explain, there's actually *three* distinct types of time being measured:
```

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile:
```suggestion
- real time: the actual wall clock time
```

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile:
```suggestion
For even more detail, check out the documentation on [`Fixed`].
```

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile: That's a good note for the [`Timer`] docs.

content/learn/book/the-game-loop/time-and-timers.md
viridia: How about changing the comment to:
```
// Distance the player will move in one second.
// At 60 FPS, this will be the same effective speed as before, as `delta_secs` will be 1/60th of a second.
```

content/learn/book/the-game-loop/time-and-timers.md
viridia: What I was trying to express is that the phrase "corresponds to one iteration" may not be accurate. In most games, this is indeed true, but nothing prevents a game from advancing the tick counter more than once in this phase.

content/learn/book/the-game-loop/time-and-timers.md
viridia: How about "Time is derived from a clock whose value is fixed for the entire frame", while not mentioning (yet) that we use different clocks in different contexts.

(Side note: I ran into a similar problem many years ago while working at Google, in which a server could produce a bad result if it's feature configuration settings were changed in mid-request.)

content/learn/book/the-game-loop/time-and-timers.md
janhohenheim: I would be happy if we could explicitly mention pixels and/or meters :)

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,276 @@
...
+One key problem with using a fixed timestep is that your game logic (including physics!) will have an uneven
+number of updates between each frame.
+This will lead to visible jittering, lagginess and tiny speedups.
+
+To account for this, we need to distinguish between the logical and rendered position (and rotation, and sometimes scale) of
+game objects.
+This is a common problem and pattern for multiplayer games, but using a fixed timestep means that even single player
+games need to handle it.
+
+The core strategy is simple enough, but the devil is in the details:
```
viridia: I feel like there is too much of a logical leap here: we start by describing the problem in a broad way, and then jump to an implementation strategy, without giving a high-level description of the solution.

Something like:

Because frames and ticks are sampled at different rates, the most recent tick may correspond to a clock time that is different than time of the current frame. For this reason, we need to mathematically adjust the rendered position of our game objects to account for this difference. This requires maintaining a source of truth for the object's position (and possibly rotation, or any other dynamic parameters that affect rendering) that is separate from the position information used by the renderer.

content/learn/book/the-game-loop/time-and-timers.md
onbjerg: I opened up an issue here: https://github.com/bevyengine/bevy/issues/20132

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile: This is something that we've resisted in the past, and it makes these docs a bit more brittle. I don't want to do this in the current PR.

content/learn/book/the-game-loop/time-and-timers.md
alice-i-cecile: Yep, I'm glad this exists, but I think this is advanced and niche enough to leave out of the book.

janhohenheim (approved):

viridia (approved):

content/learn/book/the-game-loop/time-and-timers.md
```diff
@@ -0,0 +1,324 @@
...
+Responding to the passage of time is essential for gameplay, animations and audio.
+But doing so correctly is surprisingly subtle.
+
+The [`Time`] resource in Bevy is the source of truth
+for the "current in-game time".
+Unlike [`Instant::now()`], the [`Time`] is fixed for the entire frame,
+rather than reflecting the precise wall clock time at the time of system evaluation.
+Time is derived from a clock whose value is fixed for the entire frame,
+set at the start of the frame based on the time supplied by the renderer
+during the [`TimeSystem`] system set in the [`First`] schedule.
```
Jondolf, line 17: "is fixed for the entire frame" is repeated here in consecutive sentences. Maybe the sentence with `Instant::now()` could be just removed? Or reworded
```suggestion
The [`Time`] resource in Bevy is the source of truth for the "current in-game time".
It is derived from a clock whose value is fixed for the entire frame, set at the start
of the frame based on the time supplied by the renderer during the [`TimeSystem`]
system set in the [`First`] schedule.
```

PR #2188, opened 2025-07-11
Change detection page for new book

viridia:
@alice-i-cecile

### DIFF
--- a/content/learn/book/control-flow/change-detection.md
+++ b/content/learn/book/control-flow/change-detection.md
@@ -187,5 +187,5 @@ of the component with the tick count of the last time the system was run.
 {% end %}
 
 One thing to beware of is potential 1-frame delay, if the system that is causing the change
-runs after the system that is checking for changes. You may need to pay attention to [run schedules]
+runs after the system that is checking for changes. You may need to pay attention to how you [run schedules]
 or use [explicit system ordering] to prevent this.


content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,170 @@ insert_anchor_links = "right"
 weight = 8
 status = 'hidden'
 +++
+
+Change detection is an inherent feature of Bevy's Entity Component System (ECS). Each time a
+[`Component`] or [`Resource`] is modified, Bevy marks the item as changed. This mechanism helps
+optimize performance by avoiding unnecessary system executions when no relevant changes have
```
Jondolf: I don't think it avoids system executions, it just filters out items that haven't changed while you're iterating over the query

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,170 @@ insert_anchor_links = "right"
...
+occurred. Change detection can also be used to trigger actions in response to changes, such as
+synchronizing data between two contexts.
+
+In this section, we'll explore how Bevy tracks changes and how you can leverage this
+feature in your game development workflow.
+
+## Filtering
+
+You can configure queries to only iterate over entities where certain components have been modified.
```
Jondolf: `Changed<T>` and `Added<T>` are not archetype filters. This means that it still needs to iterate over all entities that otherwise match the query, it just only retains items that were changed or added. Excluding possible inlining differences, I believe these should be roughly equivalent:

```rust
fn system1(query: Query<(&MyComponent, &Transform), Changed<Transform>>) {
    for (my_component, transform) in &query {
        todo!()
    }
}

fn system2(query: Query<(&MyComponent, Ref<Transform>)>) {
    for (my_component, transform) in &query {
        if transform.is_changed() {
            todo!()
        }
    }
}
```

IMO this is a pretty important performance detail for users to be aware of. I've seen Bevy's built-in change detection be seriously sub-optimal when you have a small number of changes but a lot of matched entities.

I would like to see this worded slightly differently, and to perhaps have an aside that briefly talks about archetype filters (similar to what Bevy's [docs have](https://dev-docs.bevy.org/bevy/ecs/prelude/struct.Changed.html#time-complexity)). It could also mention `OnAdd` as an (often preferable IME) alternative to `Added<T>`, though we probably haven't covered observers yet in the book, so it can also be left for later.

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,170 @@ insert_anchor_links = "right"
...
+
+To detect when components are removed, you can use the [`RemovedComponents`] event:
+
+```rust
+fn detect_removed_position(mut removed: RemovedComponents<Position>) {
+    for entity in removed.iter() {
+        println!("Entity {:?} just lost its Position component", entity);
+    }
+}
+```
```
Jondolf: IME using `RemovedComponents<T>` is an anti-pattern for the  majority of cases. It can miss component removals when used in `FixedUpdate`, and doesn't let you access the component values. Using `OnRemove` or the `on_remove` hook tends to be preferable, excluding specific cases where you want to quickly process a very large number of removals at once, and don't need the component values.

So I would *at least* link to `OnRemove` here, though again I don't think we have a section on observers yet...

content/learn/book/control-flow/change-detection.md
viridia: Changed "system executions" to "calculations".

content/learn/book/control-flow/change-detection.md
viridia: OK let me think about how to word this.

content/learn/book/control-flow/change-detection.md
viridia: Unfortunately, I have found cases where `RemovedComponents` is inescapable, particularly if I am reacting to more than one removed component (because using `OnRemove` means (a) I get called once for each removed component, and (b) before the removal actually happens).

Jondolf (approved):

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,186 @@ insert_anchor_links = "right"
 weight = 8
 status = 'hidden'
 +++
+
+Change detection is an inherent feature of Bevy's Entity Component System (ECS). Each time a
```
alice-i-cecile:
```suggestion
Change detection is built into Bevy's Entity Component System (ECS). Each time a
```

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,186 @@ insert_anchor_links = "right"
...
+optimize performance by avoiding unnecessary calculations when no relevant changes have
+occurred. Change detection can also be used to trigger actions in response to changes, such as
+synchronizing data between two contexts.
+
+In this section, we'll explore how Bevy tracks changes and how you can leverage this
+feature in your game development workflow.
+
+## Filtering
+
+You can configure queries to only iterate over entities where certain components have been modified.
```
alice-i-cecile:
```suggestion
You can configure queries to filter out entities unless certain components have been modified.
```

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,186 @@ insert_anchor_links = "right"
...
+
+In this section, we'll explore how Bevy tracks changes and how you can leverage this
+feature in your game development workflow.
+
+## Filtering
+
+You can configure queries to filter out entities unless certain components have been modified.
+
+The [`Added<T>`] query filter detects new component instances, either if the component was added to
+an existing entity, or a new entity with that component was spawned.
```
alice-i-cecile, line 23:
```suggestion
an existing entity, or a new entity with that component was spawned.
This is also triggered if a component is reinserted on an entity that already had it.
```

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,187 @@ insert_anchor_links = "right"
...
+- [`RemovedComponents`] can miss component removals when used in `FixedUpdate`.
+
+{% end %}
+
+Bevy does not provide any API for detecting when resources are removed.
+
+## What gets detected?
+
+Change detection occurs when a component or resource is mutably dereferenced. [`Mut<T>`] and
+[`ResMut<T>`] implement `DerefMut`, which marks the item as changed.
```
alice-i-cecile:
```suggestion
[`ResMut<T>`] implement `DerefMut`, with an implementation that marks the item as changed.
```

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,187 @@ insert_anchor_links = "right"
...
+- You get access to the component values being removed.
+- [`RemovedComponents`] can miss component removals when used in `FixedUpdate`.
+
+{% end %}
+
+Bevy does not provide any API for detecting when resources are removed.
+
+## What gets detected?
+
+Change detection occurs when a component or resource is mutably dereferenced. [`Mut<T>`] and
```
alice-i-cecile:
```suggestion
Change detection is triggered when a component or resource is mutably dereferenced. [`Mut<T>`] and
```

content/learn/book/control-flow/change-detection.md
```diff
@@ -5,3 +5,187 @@ insert_anchor_links = "right"
...
+using [states] or [run conditions]), so you do not need to worry about "missing" changes.
+
+{% callout(type="info") %}
+Internally, Bevy stores an "engine tick count" with every component and resource that marks the
+last time that it was updated. When a system calls [`.is_changed()`], it compares the tick count
+of the component with the tick count of the last time the system was run.
+{% end %}
+
+One thing to beware of is potential 1-frame delay, if the system that is causing the change
+runs after the system that is checking for changes. You may need to pay attention to [run schedules]
```
alice-i-cecile:
```suggestion
runs after the system that is checking for changes. You may need to pay attention to how you [run schedules]
```

alice-i-cecile (approved):
Good! This needs tie-ins to observers, but I'll do that later :) Very solid and helpful: merging.

PR #2190, opened 2025-07-14
Write resources chapter.

viridia:


content/learn/book/storing-data/resources.md
```diff
@@ -6,6 +6,157 @@ weight = 5
...
 
-- Resources are singletons
-- Working with resources
-- Singleton entities vs resources
+Resources are used to represent global, shared data that exists independently of any specific entity
+and which can be accessed by any system in your application. Unlike components, which are attached
+to individual entities, resources provide a way to store singleton data such as game settings, asset
+handles, time information, or any other state that needs to be accessible across multiple systems.
+
+To create a new resource type, simply derive a Rust `struct` or `enum`, and derive the [`Resource`]
```
alice-i-cecile:
```suggestion
To create a new resource type, simply create a Rust `struct` or `enum`, and derive the [`Resource`]
```

content/learn/book/storing-data/resources.md
```diff
@@ -6,6 +6,157 @@ weight = 5
...
+#[derive(Resource)]
+struct AudioSettings {
+    music_volume: f32,
+    speech_volume: f32,
+    effects_volume: f32,
+}
+```
+
+For each unique resource type `T`, there can only be one instance of that type in the Bevy [`World`].
+It you might need multiple instances, then consider using [entities and components] instead.
```
alice-i-cecile:
```suggestion
If you might need multiple instances, then consider using [entities and components] instead.
```

content/learn/book/storing-data/resources.md
```diff
@@ -6,6 +6,157 @@ weight = 5
...
+
+Bevy uses resources for many of the built-in features of the engine. For example, Bevy's
+[`AssetServer`] is a resource.
+
+## Accessing Resources
+
+To access the value of a resource use [`Res`] for immutable access, and [`ResMut`] for mutable access:
```
alice-i-cecile:
```suggestion
To access the value of a resource in a system, use [`Res`] for immutable access, and [`ResMut`] for mutable access:
```

alice-i-cecile (approved):
Good stuff: I quite like the "resources vs singleton entities" section.

PR #2195, opened 2025-07-25
Rendering book

JMS55:
17-year old me struggled to get a game made in Godot to look good. I had no idea how to light a scene or setup materials.

We frequently get questions from users on how to use rendering internals, as they're not documented.

Let's fix that.


content/learn/book/rendering/standard-material.md
```diff
@@ -1,11 +1,41 @@
...
 weight = 1
 status = 'hidden'
 +++
 
-Explain PBR, why consistent lighting models are useful
-Explain What a BRDF/material is
-Explain and showcase each StandardMaterial property
+In Bevy, the main `Material` the engine provides is the `StandardMaterial`.
+
+StandardMaterial, like most of the functionality in `bevy_pbr`, implements an idea called "physically based rendering", or PBR.
```
viridia, line 11: Use this link: https://en.wikipedia.org/wiki/Physically_based_rendering

JMS55: Tbh not my favorite wikipedia page. It's kinda generic and not very informative. I'll see if I can find something better.

content/learn/book/rendering/standard-material.md
```diff
@@ -0,0 +1,41 @@
...
+
+While this process worked fine for smaller scenes, as larger movies and games got created, assets became harder to reuse. A coin that looked the correct shade of yellow, with a certain shininess in one scene, might look completely wrong when reused in another scene under different lighting conditions.
+
+With PBR, to answer "what color should this object be?", you instead reference values from a database of real-world measurements like [Physically Based](https://physicallybased.info). Assets are more scene-independent, and behave consistently in a variety of different lighting conditions. Even if you're making a stylized game, PBR is still important.
+
+## Theory
+
+While `Material`s in Bevy implement fragment shaders (TODO: reference rendering stuff and the custom material chapter) that run arbitrary functions for computing the color of each pixel, `StandardMaterial` and PBR materials in general are based on the concept of a bidirectional reflectance distribution function (BRDF).
```
piedoom: Rendering noob commentary: I got a little confused here as it sounds like PBR materials do not implement fragment shaders. Is that true? I thought everything basically had a shader. I assume that BRDF is how colors are calculated? Or am I incorrect in my assumptions here

piedoom: Oh I see later down this is clarified.

content/learn/book/rendering/standard-material.md
```diff
@@ -0,0 +1,41 @@
...
+
+In the real world, when light hits a surface, a portion of the energy is absorbed, and a portion is reflected and bounces in a different direction. Given an incoming ray of light, a BRDF is a formula that outputs the possible directions the ray can bounce, and the amount of energy reflected.
```
piedoom: Making a quick real world example connection up front might help my understanding while in the flow of reading, though I see its mentioned later on. It kind of sounds like the BRDF is a description of the roughness of a surface

content/learn/book/rendering/standard-material.md
JMS55: Thank you for the feedback. 

Your assumptions were correct - all materials have fragment shaders (well unless you're rendering depth-only but that's out of scope... anyways). You can do whatever you want in a fragment shader. Hardcode the result to blue, return a value based on the current time, etc. StandardMaterial's fragment shader implements physically based lighting. It loops over the list of lights in the scene, and multiplies the light contribution by the BRDF. The sum of these values is the final pixel color. 

I'll reword the section.

content/learn/book/rendering/standard-material.md
JMS55: Somewhat. A BRDF just describes how light bounces off a surface. 

StandardMaterial roughness is one property (among many) that go into building the final BRDF that StandardMaterial uses.

Tbh I debated between including the BRDF stuff or not. Is it useful? Idk.

piedoom: I can see the usefulness in building an intuition on the building blocks of PBR and breaking it down like this taught me something I didn't know that seems important to understanding the pipeline, so I think it is useful to include it in some capacity!

PR #2200, opened 2025-08-07
Calmer callout colors

cart:
The current callout colors are very "loud". This calms them down a bit / harmonizes them more with their surroundings.

## Before

image

## After

image

### DIFF
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -168,13 +168,13 @@ $color-white: #fff;
   --button-square-bg-color: #{rgba($color-neutral-93, 0.05)};
 
   --callout-caution-accent-color: #d95e7a;
-  --callout-caution-bg-color: #4b373b;
+  --callout-caution-bg-color: #52363c;
   --callout-caution-color: #f1dfe3;
   --callout-info-accent-color: #817aaf;
-  --callout-info-bg-color: #36333d;
+  --callout-info-bg-color: #3b344c;
   --callout-info-color: var(--color-neutral-96);
   --callout-warning-accent-color: #ccb875;
-  --callout-warning-bg-color: #443f35;
+  --callout-warning-bg-color: #5a4d34;
   --callout-warning-color: #f9f9ee;
 
   --card-bg-color: var(--color-neutral-18);
@@ -221,7 +221,8 @@ $color-white: #fff;
   --code-yyy: #d19a66;
   --code-zzz: #d1af8f;
 
-  --details-bg-color: #{rgba($color-neutral-93, 0.1)};
+  --details-bg-color: #{rgba($color-neutral-93, 0.01)};
+  --details-border-color: #444;
   --details-text-shadow-color: var(--color-neutral-93);
 
   --docs-whats-a-bevy-filter: none;


cart:
I'd argue that this kind of "washed out" is part of the "Bevy aesthetic". See the various colors here, for example:
image

cart:
#2201 builds on this, and provides additional context for justifying this change

cart:
I think they harmonize nicely with the syntax theme as well:
image
image

cart:
I tried it on a couple more devices and I do think its slightly washed out. I bumped it up slightly:

## Before

image

## After

image

janhohenheim (approved):
Much much much better! Thanks for tweaking the colors :)

sass/_vars.scss
```diff
@@ -94,12 +94,12 @@ $color-white: #fff;
   --color-white: #{$color-white};
 
   // Blues by lightness
-  --color-blue-21: #2f2745;
+  --color-blue-21: #332e41;
```
TimJentzsch, line 97: Does it make sense to change these definitions directly? It appears to be changing the lightness, which is what the numbers represent

piedoom:
I personally think the readability (especially with purple links & yellow) is a bit better with the "washed out" colors (which I think look fine, the goal is to draw attn and they definitely will do that). But not to bikeshed, both are really similar and look much better than before  :) just thought i would give an unsolicited opinion from a prod designer.

alice-i-cecile (approved):

sass/_vars.scss
dontgetfoundout: The current colors are not just varying in lightness so this would not work without changing to color palette, but if you wanted to make a change like this you could use the hsl() function to define a blue color ramp that varies only in lightness.  Note that you may want to create separate color ramps for different themes like dark and light mode.

```css
--blue-hue: 211;
--blue-saturation: 26%;

// Blue color ramp
--color-blue-50:  hsl(var(--blue-hue), var(--blue-saturation), 97%);
--color-blue-100: hsl(var(--blue-hue), var(--blue-saturation), 92%);
--color-blue-200: hsl(var(--blue-hue), var(--blue-saturation), 85%);
--color-blue-300: hsl(var(--blue-hue), var(--blue-saturation), 75%);
--color-blue-400: hsl(var(--blue-hue), var(--blue-saturation), 65%);
--color-blue-500: hsl(var(--blue-hue), var(--blue-saturation), 55%);
--color-blue-600: hsl(var(--blue-hue), var(--blue-saturation), 47%);
--color-blue-700: hsl(var(--blue-hue), var(--blue-saturation), 35%);
--color-blue-800: hsl(var(--blue-hue), var(--blue-saturation), 25%);
--color-blue-900: hsl(var(--blue-hue), var(--blue-saturation), 15%);
--color-blue-950: hsl(var(--blue-hue), var(--blue-saturation), 8%);
```

cart: Oops this wasn't intended to be included. The themes in question aren't using this variable anymore. I'll turn it back in the other pr.

janhohenheim:
@piedoom good point, the purple links really are a bit hard to read, especially on the yellow background

PR #2201, opened 2025-08-07
Calmer Details

cart:
The "details" sections are very loud at the moment. This calms them down a bit. This builds on #2200.
## Before (without #2200)

image

## After (with #2200)

image


cart:
Just fixed an incorrect h4 weight and integrated changes from #2200 

## Before

image

## After

image

alice-i-cecile (approved):

PR #2204, opened 2025-08-12
Add the Bevy Org's AI Policy

james7132:
[RENDERED](https://hackmd.io/@bevy/ai_policy/view)

Added a new Policies section to the Contributing Guide and incorporated the above draft under that section. The triage reference has also been updated to make guidance for `S-Nominated-to-Close` aligned with the new AI policy.

### DIFF
--- a/content/learn/contribute/policies/ai.md
+++ b/content/learn/contribute/policies/ai.md
@@ -1,0 +1,79 @@
++++
+title = "AI Policy"
+insert_anchor_links = "right"
+[extra]
+weight = 2
++++
+
+In recent times, there have been a growing number of contributions that are fully or 
+partially produced by generative AI (e.g. LLMs and friends) which exhibit 
+characteristics that result in undue extra work for other contributors and maintainers. 
+While we've seen PRs and issues with these characteristics produced entirely by 
+humans, generative AI tools have significantly lowered the level of effort required to produce 
+"plausibly-worthwhile" contributions that are otherwise entirely unmergable or
+incorrectly report bugs, and so have become a major source of burdensome PRs and
+issues.
+
+Whether AI generated code are copyrightable works is also a hot-button legal topic that is 
+still being openly debated and litigated. How this impacts the legal aspects of maintaining
+a FOSS project is currently a unresolved question.
+
+This policy is established as a response targeted at the problem of an 
+increasing frequency of burdensome PRs/issues and to address the potential legal issues
+currently surrounding the intersection of AI generated code and the FOSS contribution model.
+
+## AI Generated Communications
+The unsolicited use of automated systems to communicate issues, bugs, or security vulnerabilites
+about Bevy Organization projects under the guise of a human is considered unacceptable
+and a Code of Conduct violation. Any individual contributor, operator of automated systems,
+or company they may represent may be barred from future contributions and banned from regular
+communication channels if these communications were found to be submitted in bad faith.
+
+This policy applies to all regular channels of communication used by members of the 
+Bevy Organization, including but not limited to GitHub Issues, GitHub Pull Requests, Discord, 
+other social media platforms, etc.
+
+We recognize that English may not be the primarily language for all contributors and that 
+machine translation is an indispensable tool for proper collaboration, and thus not subject to
+the above policy, though it is recommended by the community to use non-LLM machine translation
+options, as they tend to be less verbose while still getting the point across.
+
+## Copyrightable AI Generated Contributions
+At the current time of writing (August 11th, 2025), the US Copyright Office has
+[stated publicly][us-copyright-office-response] that "human authorship is a
+pre-requisite to copyright protection". A [more recent report][us-copyright-office-report] 
+from the same institution shows a much more contested legal space, both within the US and 
+internationally. In the case that AI generated works are not copyrightable, those same works 
+cannot be licensed to others, and thus AI-generated code and assets would not legally be 
+licensed under MIT or Apache.
+
+Erring on the side of caution in light of a openly debated legal topic, 
+all[^1] forms of AI-generated contributions cannot be merged into repos maintained 
+by the Bevy Organization. This includes both code and non-code game assets (e.g. textures,
+audio, etc).
+
+Any triage team member suspecting a pull request to be made primarily through the use of 
+LLMs or other generative tools should mark the PR as `S-Nominated-to-Close`, upon which a 
+maintainer can then review the PR for closure. To help identify these cases, 
+pull requests subject to this policy have characteristics such as (but not limited to): 
+
+ * Needlessly or overly verbose descriptions or responses.
+ * Not internally coherent or even self-contradictory.
+ * Demonstrates misunderstanding of important aspects of what the code is doing 
+   or the purpose of the change.
+
+Any contributor, operator of automated systems, or company they may represent found to
+have repeatedly submitted contributions with majority AI-generated code or assets may be 
+subject to:
+
+ * Blanket rejection of all future contributions to Bevy Organization projects.
+ * Retroactive removal of any potentially suspect AI-generated code and asset contributions.
+ * Further Code of Conduct actions if these contributions were found to be submitted in bad faith.
+
+This policy may be revisited when the legal debate has settled.
+
+[us-copyright-office-response]: https://www.copyright.gov/rulings-filings/review-board/docs/a-recent-entrance-to-paradise.pdf
+[us-copyright-office-report]: https://www.copyright.gov/ai/Copyright-and-Artificial-Intelligence-Part-2-Copyrightability-Report.pdf
+
+[^1]: Trivial LLM generated content such as variable renames or autocompleted function calls, often branded "predictions" or "suggestions", that is indistinguishable from traditional methods such as a regex search/replace or an LSP autocompletion is by definition not detectable and can be treated like other regular IDE tools such as Intellisense. 
+This is does not include cases where the prediction generates things like entire function blocks. 
--- a/content/learn/contribute/reference/triage.md
+++ b/content/learn/contribute/reference/triage.md
@@ -91,6 +91,10 @@ Some reasons to apply `D-Trivial` include:
 2. Moving a file from one location to another.
 3. Small changes to documentation or error messages.
 
+Some reasons to apply `S-Nominated-to-Close`  include:
+
+ 1. The PR shows inconsistencies and verbosity to suggest that the code was [AI generated][ai-policy].
+
 Remember that difficulty labels are for expertise required to either solve an issue or review a pull request. By labeling an issue as `D-Trivial`, you are marking it as a good first issue for new contributors to the Bevy project.
 
 ## Closing PRs and Issues
@@ -118,6 +122,7 @@ There are several paths for PRs to be closed:
 6. In the case of PRs where some members of the community, other than the author, are in favor, and some are opposed, any two relevant SMEs or Maintainers may act in concert to close the PR.
 7. For a PR that has been sitting for a while and became bitrotten, check with the original author if they intend to continue working on it. If not, or without a response, the PR can be labeled with `S-Adopt-Me`, and closed. Tracking adoption progress will happen in a linked issue.
 8. Inactive `X-Controversial` can be closed if relevant SMEs or Maintainers have decided there's no more interest for it. If it's still interesting and controversial, a decision must be made.
+9. Majority AI-Generated PRs cannot be merged as per the Bevy org's [AI policy][ai-policy]. Please mark these PRs as `S-Nominated-to-Close` for maintainers to make a final decision.
 
 When closing a PR, check if it has an issue linked. If it does not, you should strongly consider creating an issue and linking the now-closed PR to help make sure the previous work can be discovered and credited.
 
@@ -167,3 +172,4 @@ Members of the Triage Team within the Bevy organization have permissions to labe
 If that applies to you, then feel free to ask a Maintainer on [Discord] or email <support@bevy.org>. Everyone is welcome to do this. We generally accept membership requests, so don't hesitate if you are interested!
 
 [Discord]: https://discord.gg/bevy
+[ai-policy]: /learn/contribute/policies/ai
--- a/content/learn/contribute/policies/_index.md
+++ b/content/learn/contribute/policies/_index.md
@@ -1,0 +1,7 @@
++++
+title = "Policies"
+template = "docs.html"
+redirect_to = "/learn/contribute/policies/ai"
+[extra]
+weight = 4
++++
--- a/content/learn/contribute/reference/_index.md
+++ b/content/learn/contribute/reference/_index.md
@@ -3,5 +3,5 @@ title = "Reference"
 template = "docs.html"
 redirect_to = "/learn/contribute/reference/triage"
 [extra]
-weight = 4
+weight = 5
 +++


alice-i-cecile (approved):
Good. I think this lays out a clear position, and explains why we feel this is necessary without excessive moralizing or encouraging us to police problems that cannot be detected.

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,79 @@
...
+characteristics that result in undue extra work for other contributors and maintainers. 
+While we've seen PRs and issues with these characteristics produced entirely by 
+humans, generative AI tools have significantly lowered the level of effort required to produce 
+"plausibly-worthwhile" contributions that are otherwise entirely unmergable or
+incorrectly report bugs, and so have become a major source of burdensome PRs and
+issues.
+
+Whether AI generated code are copyrightable works is also a hot-button legal topic that is 
+still being openly debated and litigated. How this impacts the legal aspects of maintaining
+a FOSS project is currently a unresolved question.
```
mockersf:
```suggestion
a FOSS project is currently an unresolved question.
```

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,79 @@
...
+Whether AI generated code are copyrightable works is also a hot-button legal topic that is 
+still being openly debated and litigated. How this impacts the legal aspects of maintaining
+a FOSS project is currently an unresolved question.
+
+This policy is established as a response targeted at the problem of an 
+increasing frequency of burdensome PRs/issues and to address the potential legal issues
+currently surrounding the intersection of AI generated code and the FOSS contribution model.
+
+## AI Generated Communications
+The unsolicited use of automated systems to communicate issues, bugs, or security vulnerabilites
```
mockersf:
```suggestion
The unsolicited use of automated systems to communicate issues, bugs, or security vulnerabilities
```

content/learn/contribute/reference/triage.md
```diff
@@ -91,6 +91,10 @@ Some reasons to apply `D-Trivial` include:
 2. Moving a file from one location to another.
 3. Small changes to documentation or error messages.
 
+Some reasons to apply `S-Nominated-to-Close`  include:
+
+ 1. The PR shows inconsistencies and verbosity to suggest that the code was [AI generated][ai-policy].
```
mockersf: not a fan of numbered lists with 1 item, we should fine another one to add 🙂

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,79 @@
...
+
+Whether AI generated code are copyrightable works is also a hot-button legal topic that is 
+still being openly debated and litigated. How this impacts the legal aspects of maintaining
+a FOSS project is currently an unresolved question.
+
+This policy is established as a response targeted at the problem of an 
+increasing frequency of burdensome PRs/issues and to address the potential legal issues
+currently surrounding the intersection of AI generated code and the FOSS contribution model.
+
+## AI Generated Communications
```
mockersf, line 28:
```suggestion
## AI Generated Communications

```

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,79 @@
...
+This policy applies to all regular channels of communication used by members of the 
+Bevy Organization, including but not limited to GitHub Issues, GitHub Pull Requests, Discord, 
+other social media platforms, etc.
+
+We recognize that English may not be the primarily language for all contributors and that 
+machine translation is an indispensable tool for proper collaboration, and thus not subject to
+the above policy, though it is recommended by the community to use non-LLM machine translation
+options, as they tend to be less verbose while still getting the point across.
+
+## Copyrightable AI Generated Contributions
```
mockersf:
```suggestion
## Copyrightable AI Generated Contributions

```

content/learn/contribute/reference/triage.md
alice-i-cecile:
```suggestion
 1. The PR shows inconsistencies and verbosity to suggest that the code was [AI generated][ai-policy].
 2. The PR has serious flaws and has not attracted any support from reviewers.
 3. The PR is no longer needed.
```

james7132:
I did try a render of this locally and the footnote rendering we have on docs pages leaves quite a bit to be desired. Might want to follow this PR up with a style update to make it look nicer.

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,81 @@
...
+
+This policy is established as a response targeted at the problem of an 
+increasing frequency of burdensome PRs/issues and to address the potential legal issues
+currently surrounding the intersection of AI generated code and the FOSS contribution model.
+
+## AI Generated Communications
+
+The unsolicited use of automated systems to communicate issues, bugs, or security vulnerabilities
+about Bevy Organization projects under the guise of a human is considered unacceptable
+and a Code of Conduct violation. Any individual contributor, operator of automated systems,
```
jyn514: does this apply to communication that *is* clearly marked as AI-generated?

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,81 @@
...
+still being openly debated and litigated. How this impacts the legal aspects of maintaining
+a FOSS project is currently an unresolved question.
+
+This policy is established as a response targeted at the problem of an 
+increasing frequency of burdensome PRs/issues and to address the potential legal issues
+currently surrounding the intersection of AI generated code and the FOSS contribution model.
+
+## AI Generated Communications
+
+The unsolicited use of automated systems to communicate issues, bugs, or security vulnerabilities
```
jyn514: "automated systems" is not a good phrase here IMO, it excludes things like fuzzers and linters. i would explicitly say "generative AI" or "LLMs" throughout.

(the other two places you say "automated systems" are further qualified and probably don't need changes, but this one is unqualified.)

alice-i-cecile: Interestingly I think that automated systems is fine here: the key term is "unsolicited".

content/learn/contribute/policies/ai.md
alice-i-cecile: If unsolicited, yes. I think we should cut "under the guise of".

content/learn/contribute/policies/ai.md
jyn514: oh interesting - so you wouldn't want e.g. an academic lab to post the output of a fancy dynamic fuzzer they built? that's reasonable i suppose.

content/learn/contribute/policies/ai.md
james7132: The "under the guise of" was explicitly included to avoid capturing automated systems that do not pretend to be humans. There are systems like GitGuardian that does a useful service but makes zero effort to hide that it's a crawler bot.

Under the proposed situation, so long as the fuzzer results are presented either by a human, or by a clearly demarked bot account, it shouldn't be an issue. It's only when it files issues or PRs as if it were a human, and then subsequently wastes valuable volunteer time with the expectations that said bot would continue to engage as if it were a human where it becomes a major problem.

content/learn/contribute/policies/ai.md
alice-i-cecile: Not without asking us first, no :) We'd say yes, but asking first is much more polite!

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Copyrightable AI Generated Contributions
+
+At the current time of writing (August 11th, 2025), the US Copyright Office has
+[stated publicly][us-copyright-office-response] that "human authorship is a
+pre-requisite to copyright protection". A [more recent report][us-copyright-office-report] 
+from the same institution shows a much more contested legal space, both within the US and 
+internationally. In the case that AI generated works are not copyrightable, those same works 
+cannot be licensed to others, and thus AI-generated code and assets would not legally be 
+licensed under MIT or Apache.
```
chorman0773: The way this is worded appears wrong to me. A work that isn't copyrightable (and not an infringement of copyright) is public domain. I'd assume public domain code contributions would not generally be excluded from being contributed to bevy (there is code that would be public domain, typically code that has an explicit copyright waiver, or with 0BSD/CC0). Yes, you cannot apply a license to them, but you also do not need to apply a license to them to distribute them. 
If the answer to the question is "No they aren't copyrightable (and aren't derivative of the input dataset)", then there should be no issue. The issue arrises if they are copyrightable (they'd probably be copyright of the creator of the model), or they are derivative of the input set (which would infringe 3rd party copyright). I think this should be better worded to reflect those cases rather than calling out the one case that would be legally fine. 
Note that I am not taking a stance on what legal outcome it is, and I agree with erring on the side of caution, but I think it should refer to the actual legally problematic case, and I don't think this does.

alice-i-cecile: Very fair: we can be clearer here. The concern here is not that they are public domain (which is compatible with our licensing). The concern is that they are infringing on the works the model was trained on.

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,81 @@
...
+or company they may represent may be barred from future contributions and banned from regular
+communication channels if these communications were found to be submitted in bad faith.
+
+This policy applies to all regular channels of communication used by members of the 
+Bevy Organization, including but not limited to GitHub Issues, GitHub Pull Requests, Discord, 
+other social media platforms, etc.
+
+We recognize that English may not be the primarily language for all contributors and that 
+machine translation is an indispensable tool for proper collaboration, and thus not subject to
+the above policy, though it is recommended by the community to use non-LLM machine translation
```
janhohenheim: After asking my Chinese friends who rely on LLM translators for GitHub related things, they would much rather instruct the LLM to be concise than fiddle with non-LLM solutions, as those produce entire wrong translations as they don’t know that the context is GitHub. This is very relevant for Chinese, as each word has about a dozen possible translations, and non-LLM tools struggle to pick the correct one.
Can we amend or change this to recommend setting the LLM to output concise results? Chinese users are apparently aware of the verbosity otherwise and are happy to comply with that :)

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,81 @@
...
+ * Retroactive removal of any potentially suspect AI-generated code and asset contributions.
+ * Further Code of Conduct actions if these contributions were found to be submitted in bad faith.
+
+This policy may be revisited when the legal debate has settled.
+
+[us-copyright-office-response]: https://www.copyright.gov/rulings-filings/review-board/docs/a-recent-entrance-to-paradise.pdf
+[us-copyright-office-report]: https://www.copyright.gov/ai/Copyright-and-Artificial-Intelligence-Part-2-Copyrightability-Report.pdf
+
+[^1]: Trivial LLM generated content such as variable renames or autocompleted function calls, often branded "predictions" or "suggestions", that is indistinguishable from traditional methods such as a regex search/replace or an LSP autocompletion is by definition not detectable and can be treated like other regular IDE tools such as Intellisense. 
+This is does not include cases where the prediction generates things like entire function blocks.
```
NiklasEi:
```suggestion
This does not include cases where the prediction generates things like entire function blocks. 
```

content/learn/contribute/policies/ai.md
alice-i-cecile:
```suggestion
the above policy, though it is recommended by the community to use concise or non-LLM machine translation
```

janhohenheim: Sounds a bit weird in the context of the whole sentence imo. I suggest
> "the above policy. It is recommended by the community to instruct the LLM to produce a concise output or use non-LLM machine translation"

content/learn/contribute/policies/ai.md
janhohenheim: This reasoning clashes with what @james7132 says above:
> The "under the guise of" was explicitly included to avoid capturing automated systems that do not pretend to be humans. There are systems like GitGuardian that does a useful service but makes zero effort to hide that it's a crawler bot.

> Under the proposed situation, so long as the fuzzer results are presented either by a human, or by a clearly demarked bot account, it shouldn't be an issue. It's only when it files issues or PRs as if it were a human, and then subsequently wastes valuable volunteer time with the expectations that said bot would continue to engage as if it were a human where it becomes a major problem.

looks like James would find it okay to have a lab post the results of their academic fuzzer without asking as long as they clearly state what's happening?

content/learn/contribute/policies/ai.md
alice-i-cecile:
```suggestion
the above policy. The community recommends that you instruct the LLM to produce a concise output or use non-LLM machine translation
```

janhohenheim: Sounds great, thanks :)

ariofrio:
> _Disclaimer: I work day-to-day with the assistance of Gen AI tools to write software that uses Gen AI._

### Policy is Well-Scoped

I recently read [Asahi Linux's Generative AI policy](https://asahilinux.org/docs/project/policies/slop/), and I think this policy does a good job of avoiding the pitfalls they fell into:

- It doesn't attempt to predict or assess the limitations of Gen AI models as they develop into the future.
- It doesn't make unsupported philosophical claims about what "thought" or "reasoning" means.
- As @alice-i-cecile mentioned, it avoids excessive moralizing.

### Reasoning Seems Incorrect

I very much agree with @cart's concerns about the current draft's reasoning for forbidding non-trivial AI-generated output from contributions.

- If the output is not copyrightable, then it is under the public domain ([source](https://www.copyright.gov/comp3/chap300/ch300-copyrightable-authorship.pdf#page=38)) and therefore can be freely included in any work. The scenario mentioned in the current draft is not actually a problem.
- The second report seems to rule out copyrightability by the model's developer, "Copyright does not extend to purely AI-generated material" ([source](https://www.copyright.gov/ai/Copyright-and-Artificial-Intelligence-Part-2-Copyrightability-Report.pdf#page=8)). LLM providers like [OpenAI](https://openai.com/policies/row-terms-of-use/#content) and [Anthropic](https://www.anthropic.com/legal/consumer-terms#4-inputs-outputs-actions-and-materials) make doubly sure by assigning any rights they might have in the output to the user.
- The actual problem only occurs if the output is copyrightable, but under a pre-existing copyright—that of the original author(s).

### Comparing Risk Levels

One thing that's interesting about the comparison with Asahi's policy is that they make the (compelling) case that because their problem space is so esoteric and highly specific, Gen AI models may be more likely to reproduce the scarce, copyrighted training data that relates to their problem space.

FWIW, Asahi's reasoning seems to apply somewhat less readily to Bevy. Still, the three reports by the US Copyright Office do indeed raise concerns for any non-trivial use of LLMs to author copyrighted content, concerns that they do not resolve.

### Unclear Cases

I think the current policy is a bit unclear on whether the following use cases would be forbidden or allowed:

- Write most of a feature, all function signatures, and have Gen AI fill in the gaps and resolve compiler errors by running the compiler iteratively ("compiler-driven development").
- Have Gen AI write a first draft of unit tests, and then review and rewrite as needed.

These feel like they fall somewhere between "autocomplete" and "generating entire function blocks". They're not equivalent to pre-LLM functionality, but neither are they Jesus-take-the-wheel-style vibe coding. I _think_ the intention was to forbid these cases, though my own (perhaps less risk-averse) assessment would lean the other way, so it might be worth being a bit clearer. These cases would also be really hard to detect.

### Futility?

Finally, I'd be surprised if non-trivial LLM-generated code hasn't already made it into Bevy or any widely-contributed-to open source project. Not sure what to do about that, but I guess if the intent is to err on the side of caution, a good-faith attempt to forbid the practice would be all that's practical to do.

content/learn/contribute/policies/ai.md
alice-i-cecile: Yeah, I slightly disagree with James here, but it's not critical. Automated bots that are clearly marked as such and are trying to be helpful are both rare and not very obnoxious.

alice-i-cecile (approved):

alice-i-cecile:
> I very much agree with @cart's concerns about the current draft's reasoning for forbidding non-trivial AI-generated output from contributions.

I don't remember such comments? I agree with your concern, but did you mean @chorman0773?

content/learn/contribute/policies/ai.md
```diff
@@ -0,0 +1,83 @@
...
+or use non-LLM machine translation options, as they tend to be less verbose while still getting
+the point across.
+
+## AI Generated Contributions and Copyright
+
+At the current time of writing (August 11th, 2025), the US Copyright Office has
+[stated publicly][us-copyright-office-response] that "human authorship is a
+pre-requisite to copyright protection". A [more recent report][us-copyright-office-report]
+from the same institution shows a much more contested legal space, both within the US and
+internationally. In the case that AI generated works are protected under copyright, those works
```
janhohenheim: IANAL, but I believe this suggests there is no problem with an non-derivative ruling, which I don't think is true. 
The output could also be copyrighted and not derivative.
In that case, if the copyright is with the creator of the LLM, it cannot be open sourced unless the creator themselves waives the copyright.
If the copyright is with the authors in the training data, there's no meaningful way to license the output (AFAIK?).
It could also be copyrighted by the user writing the prompt and not derivative, in which case there is no legal issue at all.

Maybe reword this to say that it would be bad if it was copyright by someone else than the user and/or it was derivative?

janhohenheim (changes requested):

mockersf:
By using GitHub, our contributors are already agreeing to:
- [GitHub terms of service](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license), which mentions "Whenever you add Content to a repository containing notice of a license, you license that Content under the same terms, and you agree that you have the right to license that Content under those terms"
- [GitHub code of conduct mentions of ai generated content](https://github.com/community/community/blob/2b92f07c4184ed05ee95fdffd63a614423c2ccb9/CODE_OF_CONDUCT.md#reasonable-use-of-ai-generated-content)

Those two covers already most of what I expect, not sure an additional document is needed?

james7132:
> Those two covers already most of what I expect, not sure an additional document is needed?

I think we do need specific guidance on what contributors, org members, and maintainers are expected to do in the case we do see both contributions and issues from them, in addition to the policy and guidelines GitHub provides. Specifically providing the guidance to add `S-Nominated-to-Close` and how should maintainers should generally approach such situations.

janhohenheim (approved):
I have some nits with the exact legalese (see my unresolved comments), but I believe this is justified, reasonable, well-written, and useful :)

Edit: yay, my nits were addressed!

content/learn/contribute/policies/ai.md
james7132: This was primarily provided as an example of what might be an issue. I reworded the entire section and added a bunch of other open legal questions. Please take another look.

cart (approved):

PR #2207, opened 2025-08-14
Extend patron and corporate diamond tiers

mockersf:
add rewards about link/logo in readme in the thanks section

### DIFF
--- a/content/donate/options/index.md
+++ b/content/donate/options/index.md
@@ -55,6 +55,10 @@ Please only use names, logos, and links that you have the right and permission t
 
 Please wait a little. It might take up to eight hours for the automatic update to add you to the credits.
 
+## Where in the Project's Readme will my name / logo / link be added?
+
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Background context, which I posted [on Discord](https://discord.com/channels/691052431525675048/745355529777315850/1405691568899358821):

> We were approached by a potential sponsor who strongly preferred a link in the README, rather than the Sponsors page. <@&1064695155975803020> discussed how we felt about that, and decided that it would be fine in limited form. But rather than offering a one-off deal, we feel it is better to codify it explicitly

alice-i-cecile (approved):
Yeah, I'm fine with this if sponsors care. This is not fundamentally any different than putting them in the main list on our site.

content/donate/options/index.md
```diff
@@ -55,6 +55,10 @@ Please only use names, logos, and links that you have the right and permission t
 
 Please wait a little. It might take up to eight hours for the automatic update to add you to the credits.
 
+## Where in the Project's Readme will my name / logo / link be added?
```
cart, line 58: I think this needs some clarity on what tiers will receive this benefit. Something like "I backed at a tier that adds my logo / link to the project's README: where will it be added?".

PR #2209, opened 2025-08-26
Added searchable marker for unfinished chapters.

viridia:


alice-i-cecile:
Let me know when merge conflicts are fixed and I'll get this in. I think this is a helpful idea.

PR #2213, opened 2025-09-04
Add chapter on error handling

alice-i-cecile:
I've opted to cover this from a fairly elementary level: focusing on best practices within both Rust and programming for handling errors.

Bevy has some specific tools for helping with error handling, but in practice, most new Bevy users need to be told about and master the fundamentals before those are of much use.

## To do

- [ ] error handling for systems: the `Error` type, system piping, the default error handler
- [ ] error handling for commands: why this is hard, `queue_handled` and the `try_` variants
- [ ] configuring the default error handler


content/learn/book/control-flow/handling-errors.md
```diff
@@ -6,4 +6,226 @@ weight = 6
...
+    // Query::single returns a Result, which is an Err if there are 0 or 2+ matching entities
+    // Most of the time, we expect there to be a single player, so we can get away with just calling .unwrap()
+    let player_name = player_query.single().unwrap();
+    // This is one of Bevy's logging macros!
+    // We go into them in more detail below
+    info!("The player's name is {player_name}.");
+}
+```
+
+## Defensive programming in Rust
```
alice-i-cecile, line 58: This needs a section where we actually explain what defensive programming is and why it's good.

PR #2216, opened 2025-09-07
Configure Dependabot

BD103:
The configuration is almost identical to [the `dependabot.yml` in the engine repo](https://github.com/bevyengine/bevy/blob/eab4ad1e2058d49062a611c157f976a487323756/.github/dependabot.yml). Every week, Dependabot will run and check for updates for our Cargo and Github Actions dependencies. PRs that it opens will automatically be labelled as https://github.com/bevyengine/bevy-website/labels/C-Dependencies.

Dependabot has been setup for this repo for a while, but previously it didn't update Github Actions and used the wrong labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directories:
+      - /
+    schedule:
+      interval: weekly
+    labels:
+      - C-Dependencies
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    labels:
+      - C-Dependencies


alice-i-cecile (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directories:
+      - /
+    schedule:
+      interval: weekly
+    labels:
+      - C-Dependencies
```
alice-i-cecile, line 9: Does this need to be quoted?

BD103: Nope! It's inferred as a string here.

mockersf (approved):

PR #2241, opened 2025-09-24
Remove bold and vague benchmark claims

alice-i-cecile:
This is uniquely prone to provoking fights, comes off as bragging, and is too vague to be useful.

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -1,334 +1,269 @@
 {% extends "layouts/base.html" %}
 {% block head_extensions %}
-  <meta name="description"
-        content="A refreshingly simple data-driven game engine built in Rust. Free and Open Source Forever!" />
-  <link href="https://mastodon.social/@bevy" rel="me" />
+<meta name="description"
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi (approved):

cart (approved):
A long time coming (5 years!). Never should have been there in the first place.
(my bad 😄 )

templates/index.html
```diff
@@ -37,8 +37,7 @@ <h2 class="feature-title">Data Driven</h2>
             All engine and game logic uses Bevy ECS, a custom Entity Component System
             <ul class="feature-sublist">
               <li>
-                <b>Fast:</b> Massively Parallel and Cache-Friendly. The fastest ECS according to some
-                benchmarks
+                <b>Fast:</b> Massively Parallel and Cache-Friendly.
```
james7132, line 40: Can be addressed in a different PR, but what remains here might also need some rewording. Cache-friendly might be a bit niche for those not familiar with data oriented programming.

james7132 (approved):

PR #2256, opened 2025-09-28
Remove callbacks from release notes

alice-i-cecile:
#21247 wants to cut these. If we do, we will need to edit the release notes accordingly.

I've also made the call to cut the section on state management completely. I think that this is too academic and long for release notes, and trying to condense it will lose important nuance. I would be interested in seeing a blog post or dedicated documentation on this topic however. I do think it's interesting and valuable!


content/news/2025-09-24-bevy-0.17/index.md
```diff
@@ -451,38 +451,8 @@ expansion toggle).
 
 ### Widget Notifications
 
-Applications need a way to be notified when the user interacts with a widget. Currently in `bevy_ui_widgets`
-we're experimenting with a `Callback` system, which uses "one shot systems" under the hood. We're also heavily
-considering using Events / Observers for this. This is one of the primary reasons for the "experimental" label!
-
-### State Management
```
viridia, line 423: Awwww...

My concern here is that novices are going to ask "why doesn't my slider 'just work' like it does on the web?" Why do I have to do the extra work of managing the state?

alice-i-cecile: I think that this is a good question, and well-suited to the module docs of `bevy_ui_widgets` :) IMO this is an architectural question that we should answer in a durable form, and then at most link to here.

cart: I was planning on cutting most / all of this section too. I think its a biiiit too dense for release notes, especially now that we've simplified the story.

I agree that we need a good "state management" section in our docs (and/or the book)

cart (approved):

PR #2260, opened 2025-09-29
Add deco_buttons example to 0.17 release notes

janhohenheim:

PR #2267, opened 2025-09-30
News: Bevy 0.17

cart:
- Publish Bevy 0.17 release notes
- Fill in metrics
- Add contributors


janhohenheim:
Missing https://github.com/bevyengine/bevy-website/pull/2260
Is that intentional?

content/news/2025-09-30-bevy-0.17/index.md
janhohenheim, line 412:
```suggestion
These modulate the intensity of light cast upon surfaces for various artistic effects. See the [`light_textures`](https://github.com/bevyengine/bevy/blob/release-0.17.0/examples/3d/light_textures.rs) example for usage.

That said, platform availability for light textures is still a bit sparse: Only native Windows, Linux, and Android are supported at the moment.
```

janhohenheim: Suggested originally by @ChristopherBiscardi, confirmed by @robtfm 
Since most of Bevy's features at least work on all native platforms and *usually* also WebGPU, I think we should be very explicit that this feature doesn't follow suit.

These limitations are inherited from clustered decals, but there it's not that big of a deal since users can just pick forward decals for platform compatibility. For light textures, there's not really an alternative API right now.

Companion PR to add this to the Bevy docs too: https://github.com/bevyengine/bevy/pull/21296

janhohenheim: Update: we may be able to remove this limitation in 0.17.1, so feel free to ignore this suggestion :)
See https://github.com/bevyengine/bevy/pull/21297

PR #2274, opened 2025-10-01
Add fullscreen button to  examples

laundmo:
Sometimes, especially on smaller screens, it can be useful to Fullscreen an example to view it properly. This PR adds a simple Fullscreen button to the examples.

image

I've deliberately decided against making it hover over the example in a corner, as that would cover parts of the example and potentially interfere with UI placed in the same location. 

I've made the icon myself, using https://graphite.rs. Should I also add the project file, as it allows parametric control to adjust the icon?

### DIFF
--- a/sass/components/_icon.scss
+++ b/sass/components/_icon.scss
@@ -21,7 +21,7 @@
     ("github", 24),
     ("pencil", 19),
     ("times", 16),
-    ("fullscreen", 16),
+    ("fullscreen", 24),
   );
 
 @each $icon in $icons {
--- a/sass/components/_icon.scss
+++ b/sass/components/_icon.scss
@@ -21,6 +21,7 @@
     ("github", 24),
     ("pencil", 19),
     ("times", 16),
+    ("fullscreen", 16),
   );
 
 @each $icon in $icons {
--- a/templates/layouts/example.html
+++ b/templates/layouts/example.html
@@ -6,9 +6,10 @@
       {% set parent_idx = total_ancestors - 1 %}
       {% set category = get_section(path=page.ancestors | nth(n=parent_idx)) %}
       <h2 class="example__title">{{ category.title }} / {{ page.title }}</h2>
-      <a class="example__back" href="{% block examples_url %}{% endblock examples_url %}"><i class="icon icon--chevron-left"></i> Back to examples</a>
-      <a class="example__github"
-          href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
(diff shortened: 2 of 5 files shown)


templates/layouts/example.html
```diff
@@ -6,9 +6,10 @@
       {% set parent_idx = total_ancestors - 1 %}
       {% set category = get_section(path=page.ancestors | nth(n=parent_idx)) %}
       <h2 class="example__title">{{ category.title }} / {{ page.title }}</h2>
-      <a class="example__back" href="{% block examples_url %}{% endblock examples_url %}"><i class="icon icon--chevron-left"></i> Back to examples</a>
-      <a class="example__github"
-          href="https://github.com/bevyengine/bevy/blob/latest/{{ page.extra.github_code_path }}">
+      <a class="example__back" href="{% block examples_url %}{% endblock examples_url %}"><i
+          class="icon icon--chevron-left"></i> Back to examples</a>
+      <a id="fullscreen-button" class="example__fullscreen"><i class="icon icon--fullscreen"></i></a>
```
TrialDragon, line 11: Immediate concern is that this is an improper use of the anchor element. Anchors are meant for navigation to other places, whether they be to files, other web pages, another section in the same page, or even to someones email address. For an action like initializing fullscreen we should be using a button element in an anchor elements stead. Why this matters is accessibility and semantics as text to speech tools will have little context with this anchor that is not an actual anchor, and so any accessibility functions with regards to ARIA and such you get with using the semantic element properly is lost.

- Should use `<button>` and not `<a>`
- Should have a name to give text to speech readers a description, "Enter fullscreen" ` […] </i>Enter fullscreen</a>`. This can be hidden via CSS (immediate thought is div around text applying the CSS tricks mentioned in the article): https://www.a11yproject.com/posts/how-to-hide-content/ . 

See this for more on button accessibility: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#accessibility

sass/components/_icon.scss
```diff
@@ -21,18 +21,19 @@
     ("github", 24),
     ("pencil", 19),
     ("times", 16),
+    ("fullscreen", 24),
```
TrialDragon, line 24: <details>
<summary>Portrait Mode Phone Screenshot</summary>
Screen Shot 2025-10-01 at 20 03 30
</details>
<details>
<summary>Widescreen Computer Screenshot</summary>
image
</details>

In the screenshots above, the size of the fullscreen icon seems way to big on mobile, and maybe in an awkward place and or too big as well on PC. Not sure the exact solution, but it did pop out at me.

PR #2275, opened 2025-10-01
Allow building examples locally using helper script and zola config

laundmo:
The intention behind this PR is to allow contributors to easily build wasm examples locally, useful for example to include debug info.

For contributors, that means they now have access to:
- A script to build a number of examples, and generate the examples pages only for those built.
- A separate Zola config for serving the local assets, to be used after the above script.

There are a number of things about this (and the current setup as a whole) that I feel are not ideal:
- Mixture of bash scripts and multiple small rust CLIs, spread across both repos. This means theres no central place for some kinds of modifications to go, and some changes I would think are useful will require PRs against both repos. Thats enough of a hurdle I don't want to attempt those.
- The new script can only limit the amount of examples built, theres no way to choose a sepecific example. This is a limitation of tools/example-showcase in the bevy repo, which I did not want to touche. See first point.
- I had to write code to compare the different structures for example betweeen the paths `example-showcase build-wasm-examples` generates: `2D Rendering/Bloom 2D` and the ones `example-showcase build-website-list` generates: `2d-rendering/bloom-2d`

### DIFF
--- a/generate-wasm-examples/build_wasm_examples_debug.sh
+++ b/generate-wasm-examples/build_wasm_examples_debug.sh
@@ -19,11 +19,13 @@ fi
 # Switch to script's directory, letting it be called from any folder.
 cd $(dirname $0)
 
-# generate_wasm_examples will clone the bevy repo to bevy-website/generate-wasm-examples/bevy by default.
-# If you want to use your fork or local changes, clone, symlink or copy your own bevy repo/folder:
-# 1. $ git clone https://github.com/<username>/bevy bevy
-# 2. Make relevant changes to your local bevy repository
-# 3. $ cp ~/projects/bevy ~/projects/bevy-website/generate-wasm-examples/bevy
+# generate_wasm_examples will clone the bevy repo to ./generate-wasm-examples/bevy by default.
+# If you want to use your fork or local changes, clone, symlink or copy your own bevy repo/folder
+# to ./generate-wasm-examples/bevy:
+# 1. $ cd ./generate-wasm-examples
+# 2. $ git clone https://github.com/<username>/bevy
+# OR
+# 1. $ cp ~/projects/my-bevy-fork ./generate-wasm-examples/bevy
 ./generate_wasm_examples.sh --no-pull
 
 cd bevy


config.local.toml
```diff
@@ -0,0 +1,34 @@
+# Zola config for compiling examples locally
```
rparrett, line 1: IMO we shouldn't maintain a separate nearly identical copy of this file just for this purpose. I would move the local versions of `wasm_webgl2_base_url` to comments in `config.toml`.

generate-wasm-examples/build_wasm_examples_debug.sh
```diff
@@ -0,0 +1,48 @@
...
+# If you want to use your fork or local changes, clone, symlink or copy your own bevy repo/folder:
+# git clone https://github.com/<username>/bevy bevy
+# cp ~/projects/bevy ./bevy
+./generate_wasm_examples.sh --no-pull
+
+cd bevy
+
+git reset HEAD --hard
+CARGO_PROFILE_RELEASE_OPT_LEVEL='z' CARGO_PROFILE_RELEASE_DEBUG="true" cargo run -p example-showcase -- --per-page $1 --page 0 build-wasm-examples --content-folder ../../static/assets/examples/wasm_webgl2 --api webgl2 --website-hacks
+CARGO_PROFILE_RELEASE_OPT_LEVEL='z' CARGO_PROFILE_RELEASE_DEBUG="true" cargo run -p example-showcase -- --per-page $1 --page 0 build-wasm-examples --content-folder ../../static/assets/examples/wasm_webgpu --api webgpu
```
rparrett, line 35: IMO, it's a bit unfortunate to repeat these invocations here in this new script. Would prefer that `generate_wasm_example.sh` handle the limiting.

Also, I think I or someone may have added a `--debug` option to `example-showcase`.

README.md
```diff
@@ -21,3 +21,5 @@ A local server should start and you should be able to access a local version of
 ### Assets, Errors, and Examples pages
 
 These pages need to be generated in a separate step by running the shell scripts in the `generate-assets`, `generate-errors`, and `generate-wasm-examples` directories. On Windows, you can use [WSL](https://learn.microsoft.com/en-us/windows/wsl/install) or [git bash](https://gitforwindows.org/).
+
+The Examples page uses the same WASM binaries as https://bevy.org/examples by default. To build examples locally, have a look at `generate-wasm-examples/build_wasm_examples_debug.sh`. After running it, start zola with `zola --config config.local.toml serve` to use the locally built examples.
```
rparrett:
```suggestion
The Examples page uses the same Wasm binaries as https://bevy.org/examples by default. To build examples locally, have a look at `generate-wasm-examples/build_wasm_examples_debug.sh`. After running it, start zola with `zola --config config.local.toml serve` to use the locally built examples.
```

rparrett: The first sentence is *slightly* confusing for me. I would suggest something like "uses the same pre-built Wasm binaries" to clarify a tiny bit.

config.local.toml
laundmo: Personally, i'd love to not need either, but Zola is limited in this regard. Maybe comments can work, but my goal was that there won't be a need to edit a git tracked file for at least the basic usecases.

generate-wasm-examples/build_wasm_examples_debug.sh
laundmo: but generate_wasm_examples.sh never calls `example-showcase build-wasm-example`? only the CI yaml has another version of this.

laundmo: > Also, I think I or someone may have added a --debug option to example-showcase.

Just tried this: the option does exist, but its not implemented for the `build-wasm-example` command and silently gets ignored. Its only applied for the `run` command as far as I can tell: https://github.com/bevyengine/bevy/blob/0b8844a3a060b2a5b7c3302bd951e39fbf6aa1c3/tools/example-showcase/src/main.rs#L723

README.md
laundmo: I've just pushed a re-wording

generate-wasm-examples/build_wasm_examples_debug.sh
rparrett: ah right, sorry!

And yeah, looks like the new debug option isn't piped through from `example-showcase`.

config.local.toml
Rodhlann: Was just working with these changes myself to help troubleshoot another issue, and having the config predefined absolutely helped smooth out the process. 

I don't think it's particularly uncommon to have a local config file for this sort of thing, and it doesn't look like Zola supports extending configs. 

How often do these config files change? Will having an additional config file meaningfully increase maintenance costs?

PR #2285, opened 2025-10-07
Make code blocks more readable

### LINKED ISSUE #2175: Inline code blocks: too little contrast, too much padding
In the book, the CSS styles for inline code blocks have a couple problems:

* The background color of the inline code block is nearly indistinguishable from the page background.
* The left and right padding are large enough that when followed by punctuation such as a comma, it looks like there's an extra space.

Image

TrialDragon:
## Why
Folks found that the inline code blocks looked weird and lacked contrast making them hard or annoying to read.

## What
Fixes #2175 

Uses darker colors for darkmode code block background and lighter on light mode (also made light mode background darker to let the contrast show through; this probably needs more work in a proper lightmode redesign PR). Also decreases the empty space next in the blocks so that it doesn't look like there is a space character when their isn't, especially when there is punctuation next to a code block. Also lowered the border radius and made it relative so that it isn't cutting into the letters in an ugly way.

## How
Changed color variables (and linked the two separate code background colors that were the same), changed paddings and margins of content lists, and changed the global border radius variable.

## Try
1. Switch into your bevy website directory
1.opt Add my repository for the site if you haven't before: `git remote add trialdragon git@github.com:TrialDragon/bevy-website.git`
2. Fetch my repository for the site: `git fetch trialdragon`
3. Switch into the branch for this PR: `git switch trialdragon/2175_improve_readability_of_code_blocks`
4. Run the site: `zola serve --open`
4.opt If you need to install Zola, see https://www.getzola.org/documentation/getting-started/installation/
5. Explore the site, use it, see how you feel about the styling changes.

### DIFF
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -185,7 +185,7 @@ $color-white: #fff;
   --card-title-color: var(--color-neutral-93);
   --card-sponsor-icon-color: #b13376;
 
-  --code-bg-color: color-mix(in srgb, var(--color-neutral-18) 47%, black);
+  --code-bg-color: color-mix(in srgb, var(--color-neutral-18) 55%, black);
   --code-incorrect-bg-color: #481e21;
   --code-incorrect-border-color: darkred;
   --code-incorrect-border-hover-color: red;
@@ -436,7 +436,7 @@ $color-white: #fff;
     --html-bg-color: var(--color-neutral-82);
     --html-text-color: var(--color-neutral-13);
 
-    --layout-content-bg-color: color-mix(in srgb, var(--color-neutral-97) 96%, black);
+    --layout-content-bg-color: color-mix(in srgb, darkgray 15%, white);
     --layout-header-border-color: var(--color-neutral-45);
 
     @media #{$bp-tablet-landscape-down} {


BD103:
Here's some before and after screenshots, for other reviewers:

image

image

BD103 (approved):
Overall I like the changes!

As a side note--the color for links doesn't seem to have enough contrast with the background when in light mode:

image<br />

image

I'm going to open a follow-up issue for that.

cart:
Actually I object to the light mode changes too. Its trading the default text contrast for the code contrast, which feels wrong.

viridia:
> I object to how dark the dark mode background is

The problem is that without this change, there's just not enough contrast. The inline code sections don't stand out against the page background, they are nearly invisible. And you can't make them lighter, since then you would have a hard time reading the text, unless you plan to lighten the text too.

TrialDragon:
@cart, how does this look?
<details>
<summary>Light Mode</summary>
  <details>
  <summary>Before</summary>
  Screenshot From 2025-10-21 17-28-02
  
  </details>
  <details>
  <summary>After</summary>
  Screenshot From 2025-10-21 17-28-09
  
  </details>
</details>

<details>
<summary>Dark Mode</summary>
  <details>
  <summary>Before</summary>
  Screenshot From 2025-10-21 17-45-25
  
  </details>
  <details>
  <summary>After</summary>
  Screenshot From 2025-10-21 17-45-32
  
  </details>
</details>

cart:
> The inline code sections don't stand out against the page background, they are nearly invisible

This was actually the point. The difference between a "code section" and a "text section" isn't so important that it needs to yell "I AM A CODE SECTION". There is enough of a distinction to indicate that something is different, while still occupying roughly the same amount of mental space (and color scheme) as the background.

viridia:
> > The inline code sections don't stand out against the page background, they are nearly invisible
> 
> This was actually the point. The difference between a "code section" and a "text section" isn't so important that it needs to yell "I AM A CODE SECTION". There is enough of a distinction to indicate that something is different, while still occupying roughly the same amount of mental space (and color scheme) as the background.

I'm not sure what you are advocating here. Are you saying that things are fine as is, and we don't need to change anything? Look at the screenshot I posted in https://github.com/bevyengine/bevy-website/issues/2175 - the distinction between code and text is so subtle as to be nearly invisible, even to someone with reasonably good eyesight.

This complaint is mainly about inline code sections, not block code sections. The block sections are large enough that the distinction, even though subtle, can be seen relatively easily. But the inline code sections, which use the same color scheme, are really hard to see. If I turn my head and look for a moment at a bright light or window, I can't make them out at all until I give my eyes time to adjust.

TrialDragon:
> I do agree that inline code blocks could use a legibility boost. I'd like to explore solutions that don't throw out the current "character" of the larger code blocks.

So try to split up the color definitions so that larger code blocks have the old / current color and inline code blocks have a newer more contrasting color?

viridia:
> 1. Increase the size of the surrounding block to make it more prominent

Please don't do this - there's already too much padding in inline code blocks, look at any code followed by punctuation, it looks weird and disconcerting.

PR #2290, opened 2025-10-15
Remove extraneous period from index page

mat-1:
Critical fix for an issue introduced in https://github.com/bevyengine/bevy-website/pull/2241

None of the other lines on this page end in periods and this kept bothering me every time I looked at the website. Meow :3


BD103 (approved):
Woah, I recognize you from back when I used to use Replit! Cool stuff, thanks for the drive-by fix :)

mockersf (approved):

PR #2307, opened 2025-11-16
add "copy" button to every example code blocks

1ace:
(This part of) the Clipboard API is supported (https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) by Safari since 2009, by Chrome since 2011, and by Firefox since 2013, but just in case I included a detection code to make sure the "copy to clipboard" button is not shown if we can't copy to the clipboard :)

`clone.svg` and `check.svg` from https://dazzleui.gumroad.com/l/dazzleiconsfree; picked because they seem to go well with the rest of the design, but I'm happy to replace them with anything else :)

---

### Dark mode

first code block hovered:
image
after clicking on the button:
image

### Light mode

first code block hovered:
image
after clicking on the button:
image

### DIFF
--- a/sass/pages/_docs.scss
+++ b/sass/pages/_docs.scss
@@ -32,6 +32,30 @@
   p {
     margin-top: 5px;
   }
+
+  pre:has(code) {
+    position: relative;
+
+    button.copy {
+      position: absolute;
+      top: 0;
+      right: 0;
+      margin: 0.5rem;
+      width: 2rem;
+      height: 2rem;
+      border-color: transparent;
+      background-color: transparent;
+      opacity: 0;
+      transition: opacity 0.25s ease-in-out;
+    }
+
+    /* :hover for mouse & touchscreen,
+     * :focus for keyboard navigation */
+    &:hover button.copy,
+    button.copy:focus {
+      opacity: 1;
+    }
+  }
 }
 
 .docs-whats-a-bevy {
--- a/static/clipboard.js
+++ b/static/clipboard.js
@@ -1,0 +1,41 @@
+(() => {
+    'use strict';
+    document.addEventListener('DOMContentLoaded', () => {
+        // Don't add the "copy to clipboard" button if we can't copy to clipboard :)
+        if (!navigator.clipboard || !navigator.clipboard.writeText) {
+            return;
+        }
+
+        // Downloaded from https://dazzleui.gumroad.com/l/dazzleiconsfree
+        // Author: Dazzle UI
+        // License: CC-BY
+        // Modifications:
+        // - Removing `width` & `height` to adapt to the container size.
+        // - Replacing `stroke="black"` with `stroke="currentColor"` to respect dark/light modes.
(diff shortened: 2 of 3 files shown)


sass/pages/_docs.scss
```diff
@@ -32,6 +32,29 @@
   p {
     margin-top: 5px;
   }
+
+  pre:has(code) {
+    position: relative;
+
+    button.copy {
+      position: absolute;
+      top: .25rem;
```
Rodhlann: This seems to work better for single line code blocks if `top` is set to `.5rem`, currently the button sits too high in single line code blocks

### Single Line
image

### Multi-Line
image

Rodhlann: Messing with it a bit more these options seem to work better for both Desktop and Mobile, where the icons seem to get pretty small: 
```
  top: 0.3rem;
  right: 0.3rem;
  width: 2.5rem;
  height: 2.5rem;
```
image
image

EDIT: top / right value to match for visual cohesion and to better center the icon in single line code blocks

1ace: Thanks for the suggestions! I'm happy to tweak the size and position, but I think we should keep them paired for better visual consistency, ie. top == right and width == height.

I'll try these changes when I'm back on my computer, and update the PR :+1:

static/clipboard.js
```diff
@@ -0,0 +1,31 @@
...
+        const svg_check = '<svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M4 12.6111L8.92308 17.5L20 6.5" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>';
+
+        // Add button to every `<pre><code>`
+        [].forEach.call(document.querySelectorAll('pre:has(code)'), function(pre) {
+            const btn = document.createElement('button');
```
Rodhlann, line 30: We could add an aria-label or hidden label of "copy", and "copied" once clicked, to this button for [accessibility](https://www.sitelint.com/docs/accessibility/buttons-must-have-visible-accessible-content-that-describes-their-action-empty-or-unreachable-buttons-hinder-accessibility#use_aria_labels_for_icon-only_buttons). 

(Not sure if there's any existing accessibility standards for the Bevy website)

sass/pages/_docs.scss
1ace: `width`/`height` of `2.5rem` looks a bit too big to me:
image

I'm keeping it at `2rem` for now, but I have adjusted the `top`/`right` (which I have folded into `margin` to have a single value to change) to `0.4rem`:
image
image

How do this look to you?

static/clipboard.js
1ace: Good call, thank you! I'm sighted and always forget about these :(

sass/pages/_docs.scss
Rodhlann: 0.4rem still has the icon sitting a bit high in Desktop resolution, 0.5rem works a bit better here. 

But the Mobile resolution is still a concern: 
image

When in mobile resolution the icon is quite small, and quite a bit more noticeably offset, even at 0.5rem. 

I think the CSS I had suggested previously worked better to solve this without having to introduce media queries or similar. 

If we want to split the difference between a too-large icon in Desktop resolution with a too small icon in Mobile resolution something like this could work: 
```css
  top: 0.45rem;
  right: 0.45rem;
  width: 2.25rem;
  height: 2.25rem;
```

### Desktop
image

### Mobile
image

static/clipboard.js
Rodhlann: This looks great!

sass/pages/_docs.scss
1ace: I just tested in Chrome and it does look much smaller there than on Firefox; I'm guessing you're on Chrome and that's why you want it bigger? 😄

(BTW I'm not showing the mobile screenshots because on firefox they're the same size as desktop; I do see that in chrome they get even smaller on mobile, but I don't know why or how to work around that)

Since we're on github and it has a copy button as well *(which btw they show all the time, but IMO that's too crowded and I prefer to only show it if the user puts their cursor on the code block, which is their natural action to copy it already)*. I just had a look at its css, and it's `2rem` big and `0.5rem` away from the edges, so let's just go with that?

It's centered on single-line code blocks on chrome:
image
and doesn't look too bad on firefox:
image

Making it `2.25rem` tall doesn't look great on firefox:
image
though it does look better on chrome:
image

Given that firefox is my main browser, I'd prefer to keep it 2rem, but in the interest of getting this merged and not stuck on details like this, I've changed it to `2.25rem` in the PR 😇

Rodhlann: I'm actually using LibreWolf, which is a fork of Firefox, though I think it would be good for this to be tested in the major browsers! 

I do think copying GitHub's CSS here is a good call though, I believe they could be considered an authority here! 

Your latest changes look good on my end in various browsers, I'll go ahead and approve!

1ace: > I do think copying GitHub's CSS here is a good call though, I believe they could be considered an authority here!

Ack, done :+1:

PR #2313, opened 2025-11-29
Add banner to warn user of required feature flags for examples

Sigma-dev:
As a user, it kinda sucks copying a website example only to see it not working and with unhelpful rust errors.

Some examples do mention it in their description, but it's not universal and can be forgotten. It's also easy to miss.

My solution is to add a warning banner that lists out the optional feature flags that are required for the example to work (no banner if no required features).

The information is collected from the required_features specified in the root `Cargo.toml` of the bevy engine.
This data is not yet available to the website so I made another PR in bevy to make it available: https://github.com/bevyengine/bevy/pull/21975.

EDIT: The required change has now been merged, this is ready to go

Feel free to give feedback on the wording or anything else.

This is how it would look: 
image

### DIFF
--- a/templates/layouts/example.html
+++ b/templates/layouts/example.html
@@ -30,7 +30,7 @@
     </aside>
     {% if page.extra.required_features %}
     <aside class="callout callout--warning">
-      <h3 class="warning">Optional feature flag</h3>
+      <h3 class="warning">Feature flag required</h3>
       <p>
         The code in this example requires {% if page.extra.required_features | length > 1
         %}these feature flags{% else %}this feature flag{% endif %} to be enabled in order to function:


templates/layouts/example.html
```diff
@@ -1,138 +1,150 @@
...
+        <div class="bevy-instance__progress-bar" data-progress-bar></div>
       </div>
+    </div>
+    <canvas class="bevy-instance__canvas" id="bevy" width="1280" height="720"></canvas>
+      <aside class="example__callout media-content callout callout--plain callout--{% block callout_modifier %}{% endblock callout_modifier %}">
+      {% block callout %}{% endblock callout %}
+    </aside>
+    {% if page.extra.required_features %}
+    <aside class="callout callout--warning">
+      <h3 class="warning">Optional feature flag</h3>
```
Rodhlann: I think having this header text as "Optional feature flag" is confusing. If I am understanding this issue correctly the feature flag is *not* optional for this example, which is described with added clarity below.

Maybe "Feature flag required for example" or "Feature flag details" provide more clarity?

Sigma-dev: I agree, I think "Feature flag required" is a better name, I will make the change

Rodhlann (approved):

mockersf (approved):

PR #2314, opened 2025-12-06
Fix add_systems for hello_world in getting_started_v2.rs. Update -> Startup

kuropatka-earth:
The text in https://bevy.org/learn/quick-start/getting-started/ecs/ page tells "You should see hello world! printed once in your terminal." So the hello_world system should be added with Startup system instead of Update


alice-i-cecile (approved):

kfc35 (changes requested):
Actually, sorry to take back my approval, but I just realized that:

In the section of ## Your First System above your changes in `ecs.md`, there are still references to `Update` that should be changed to `Startup`

I think the `Update` link can be kept / moved down further since that eventually gets referenced in a later section, but `Startup` probably should get a link when specified in “Your First System"

Issue #2193, opened 2025-07-22
MVP content for new book


alice-i-cecile:
Before we ship the current draft of the new book, I would like to have a) a complete replacement for all of the existing book content b) an overview of the major ECS features. The existing chapter structure is solid, but it is convenient to track the outstanding writing to be done here:

Data storage:

- [ ] [World](https://bevy.org/learn/book/storing-data/world/), very incomplete
- [ ] [Queries](https://bevy.org/learn/book/storing-data/queries/), currently just a stub
- [ ] [Disabling entities](https://bevy.org/learn/book/storing-data/disabling-entities/), no content, probably the wrong home for it?

Control flow:

- [ ] [Handling Errors](https://bevy.org/learn/book/control-flow/handling-errors/), no content
- [ ] [Events](https://bevy.org/learn/book/control-flow/events/), no content, blocked on event trait overhaul
- [ ] [Observers](https://bevy.org/learn/book/control-flow/observers/), no content, needs reorganization
- [ ] [Hooks](https://bevy.org/learn/book/control-flow/hooks/), no content

Game loop:

- [ ] [Input and windowing](https://bevy.org/learn/book/the-game-loop/input-and-windowing/) should be cut from MVP or given minimal content
- [ ] [Rendering frames](https://bevy.org/learn/book/the-game-loop/rendering-frames/) needs opinions on whether to keep or cut for MVP
- [ ] [Custom loops](https://bevy.org/learn/book/the-game-loop/custom-loops/) needs short content talking about custom runners

Development practices:

- [ ] [Fast compiles](https://bevy.org/learn/book/development-practices/fast-compiles/) port existing content
- [ ] most other sections can / should be cut for MVP

Releasing projects:

- [ ] [creating libraries for Bevy](https://bevy.org/learn/book/releasing-projects/libraries-for-bevy/), should be moved from existing home buried in the bevyengine/bevy/docs
- [ ] cut everything else from MVP

Other chapters (audio, rendering, input etc) are solidly out-of-scope. We would like to add them later, but they should not be written yet, because they will draw resources away from shipping something valuable and achievable.

viridia:
I would cut the Disabling chapter entirely. However, I am biased here: I think the `Disabled` component is a blunt instrument which requires expertise to be able to use well (as opposed to more selective disabling techniques like `Visibility::Hidden`, `Display::None`, layers, `InteractionDisabled`, or whatever Avian does to disable things - these have the benefit of having more predictable consequences).

[...]

PR #2318, opened 2025-12-24
Add book page for Fast Compiles

Based-A:
Per #2193, the Fast Compiles page is just porting over the existing section from the Setup page in the Quick Start guide. This is that, with an introductory paragraph and minor formatting and spelling adjustments.


mockersf:
I would prefer to not copy outdated / invalid claims. On alternative linkers:
- lld is the default linker on linux, nothing needed to do https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/
- let's not throw marketing claims for mold, even on their repo they're not claiming to be 5x faster than lld

alice-i-cecile:
This is fine for now; I'm happy to approve this once CI is passing.

PR #2319, opened 2026-01-03
content: fix broken getting started plugins link

cpu:
The markdown link target was incorrectly named "WininitPlugin" instead of "WinitPlugin", resulting in the following broken link at render time:

markdown glitch_000


mockersf (approved):

cpu:
Thanks for the quick review :bow:

Issue #2320, opened 2026-01-07
Polish 0.18 Release Notes


alice-i-cecile:
You can browse the current release notes on the [`release-0.18.0`](https://github.com/bevyengine/bevy/tree/release-0.18.0) branch.

The list of notes follows, with notes on status. Checked items are at a shippable quality:

- [ ] fullscreen_material.md
   - code snippet or line count comparison would be nice
   - screenshot might be helpful
   - explanation of what fullscreen shaders are used for would be good
- [ ] optional_asset_reader_seek.md
     - add more context on what this enables
     - sync with the latest changes to this work
     - add hyperlinks

## Completed

- [x] generalized_atmosphere.md 
   - needs a photo
- [x] more_standard_widgets.md
    - screenshots needed
- [x] render-assets-diagnostics.md
    - reword to match release note format
    - add context on why this is useful
    - reduce proper nouns, then add links
    - j/k just cut it
- [x] easy_marketing_material.md
    - `EasyScreenRecordPlugin` isn't showing up in the docs
        - Explain that a feature is required
        - Track down subcrate docs and link
- [x] automatic_directional_navigation.md
   - AutoDirectionalNagivator mentioned was only added in https://github.com/bevyengine/bevy/pull/22340, and does not have release-candidate docs to link to
- [x] Userspace glTF Extension Handling
    - consider renaming to explain impact immediately
    - explain what gltf is
    - rename Case Study section
    - just link to the examples
- [x] get_components_mut.md
   - needs more context on *why* this is so cool / useful: scripting / OOP-y patterns
- [x] atmosphere_occlusion.md
   - will need an image
   - writing is good; does not urgently need editing
- [x] camera_controllers.md 
- [x] cargo_feature_collections.md
- [x] computed_node_helper_functions.md 
   -  should be cut
- [x] fallible_interpolation.md
- [x] font_weights.md
   - needs a bit more context for folks on what font weights are
   - merge with opentype_font_features.md
- [x] opentype_font_features.md
   - merge with font_weights.md
- [x] ignore-scroll.md
   - move beside other UI release notes
- [x] per_text_section_picking.md
     - flesh out, adding context about why this is useful and define picking
- [x] radio_group_and_button_widget_improvements.md 
   - merge into other UI release notes
- [x] remove_systems.md 
   - could benefit from a minor editing pass
- [x] ring_primitive.md
    - probably cut, but alternatively shorten dramatically
- [x] short_type_path_asset_processors.md
[...]

PR #2335, opened 2026-01-12
Note required features for screen recording

alice-i-cecile:
This is important, hard to discover information, so I've added a note about the exact incantation needed to enable this new feature.

Part of #2320.

I've also hyperlinked the structs, relying on the dev-docs links. These [aren't showing up on the docs.rs page](https://docs.rs/bevy/0.18.0-rc.2/bevy/?search=screenrecord) at all, I think because they're not part of the default features.

IMO these links are better than nothing, but if y'all disagree I'll remove them.

### DIFF
--- a/content/news/2026-01-12-bevy-0.18/index.md
+++ b/content/news/2026-01-12-bevy-0.18/index.md
@@ -642,13 +642,15 @@ setting it up was relatively involved.
 
 This process has been streamlined, with the new [`EasyScreenshotPlugin`] allowing you to take a screenshot with consistent formatting with a single button press. With its default settings, once you add this plugin to your application, a PNG screenshot will be taken when you press the `PrintScreen` key. You can change the trigger key, or the screenshot format between PNG, JPEG or BMP.
 
-We've taken this one step further, allowing you to record video directly from Bevy, with the new `EasyScreenRecordPlugin`. This plugins add a toggle key, space bar by default, that will toggle screen recording. Recording can also be started and stopped programmatically with the `RecordScreen` messages.
+We've taken this one step further, allowing you to record video directly from Bevy, with the new [`EasyScreenRecordPlugin`]. This plugins add a toggle key, space bar by default, that will toggle screen recording. Recording can also be started and stopped programmatically with the [`RecordScreen`] messages.
 
 Screen recording is currently not supported on Windows due to challenges with video codecs.
 While all dev-tools features are off-by default, screen recording is a bit trickier to activate due to this limitation.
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
> The main concern to using dev-docs here for me is if the enum or the screen record plugin change on main past the 0.18.0 release as far as I understand
>
> I’m personally OK with that to start, but maybe it’s worth writing a caveat about those docs pointing to the dev docs, which may get out of date as time goes on forward

I think that this is likely to be more distracting than useful.

> (Is there something we can do now or for future releases that generate these docs if desired?)

I think we can enable all features when building the docs on docs.rs?

content/news/2026-01-12-bevy-0.18/index.md
```diff
@@ -642,11 +642,15 @@ setting it up was relatively involved.
 
 This process has been streamlined, with the new [`EasyScreenshotPlugin`] allowing you to take a screenshot with consistent formatting with a single button press. With its default settings, once you add this plugin to your application, a PNG screenshot will be taken when you press the `PrintScreen` key. You can change the trigger key, or the screenshot format between PNG, JPEG or BMP.
 
-We've taken this one step further, allowing you to record video directly from Bevy, with the new `EasyScreenRecordPlugin`. This plugins add a toggle key, space bar by default, that will toggle screen recording. Recording can also be started and stopped programmatically with the `RecordScreen` messages.
+We've taken this one step further, allowing you to record video directly from Bevy, with the new [`EasyScreenRecordPlugin`]. This plugins add a toggle key, space bar by default, that will toggle screen recording. Recording can also be started and stopped programmatically with the [`RecordScreen`] messages.
```
it-me-joda, line 645:
```suggestion
We've taken this one step further, allowing you to record video directly from Bevy, with the new [`EasyScreenRecordPlugin`]. This plugin adds a toggle key, space bar by default, that will toggle screen recording. Recording can also be started and stopped programmatically with the [`RecordScreen`] messages.
```

content/news/2026-01-12-bevy-0.18/index.md
```diff
@@ -642,11 +642,15 @@ setting it up was relatively involved.
...
 
 Screen recording is currently not supported on Windows due to challenges with video codecs.
+While all dev-tools features are off-by default, screen recording is a bit trickier to activate due to this limitation.
```
it-me-joda, line 648:
```suggestion
While all dev-tools features are off by default, screen recording is a bit trickier to activate due to this limitation.
```

PR #2354, opened 2026-02-04
Add Events book page content

Based-A:
With the Event rework in 0.17, I thought I'd try and take a stab at filling out a book page for Events. I pretty much just followed the 0.17 release post features, along with double checking the docs pages to make sure that I wasn't missing anything major.

Hope this is in the direction of content that is wanted, let me know if I need to change or re-work anything!

### DIFF
--- a/content/learn/book/control-flow/events.md
+++ b/content/learn/book/control-flow/events.md
@@ -8,11 +8,17 @@ status = 'hidden'
 
 <!-- TBW -->
 
-To make things happen in our Bevy app, we can use an **Event** to activate some logic or functionality on any observing Entities. Events require a **Trigger** that is called by accessing [`World`] (via [`World::trigger`]) to run the Event immediately, or by using [`Commands`] (via [`Commands::trigger`]) to add the Event to the Command Queue.
+While Systems are great for running logic at scheduled updates, many features need to be reactive, not scheduled. Picking up an item, performing an attack, or even hovering the mouse over an object are all actions that wouldn't work very well if they *had* to run on every schedule update. Instead of a System, we can use **Events** to activate some logic or functionality at a specific time or in response to something occurring.
 
-A basic use of an Event would something like this:
+There are three required parts when using an Event: a **Trigger**, an **Observer**, and the `Event` itself.
 
-- Implement the [`Event`] trait on a type:
+- A **Trigger** is the condition that will determine when an `Event` will happen in the `World`.
+- The Entities that will react to our `Event` are known as **Observers**. These Entities "observe" the `World` and will run some functionality in response to our `Event`.
+- Finally, the `Event` is a Rust type that implements the `Event` trait.
+
+A basic use of an `Event` would something like this:
+
+1. Implement the [`Event`] trait on a type:
 
 ```rust
 #[derive(Event)]
@@ -21,7 +27,7 @@ struct Speak {
 }
 ```
 
-- Add an [`Observer`] to the World that will watch for our event:
+2. Add an [`Observer`] to the World that will watch for our event:
 
 ```rust
 // To add the observer immediately:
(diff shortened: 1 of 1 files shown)


Hilpogar (approved):
I was looking for more information about Observers. The reading was very instructive. Thank you

content/learn/book/control-flow/events.md
```diff
@@ -7,3 +7,268 @@ status = 'hidden'
 +++
 
 <!-- TBW -->
+
+To make things happen in our Bevy app, we can use an **Event** to activate some logic or functionality on any observing Entities. Events require a **Trigger** that is called by accessing [`World`] (via [`World::trigger`]) to run the Event immediately, or by using [`Commands`] (via [`Commands::trigger`]) to add the Event to the Command Queue.
```
alice-i-cecile: IMO the introductory sentence of this is too imprecise. I would split it off into its own paragraph, and go into some more detail about the sort of tasks that observers are good and bad at.

content/learn/book/control-flow/events.md
```diff
@@ -7,3 +7,268 @@ status = 'hidden'
...
+[`Event`]: https://docs.rs/bevy/latest/bevy/prelude/trait.Event.html
+[`World`]: https://docs.rs/bevy/latest/bevy/prelude/struct.World.html
+[`World::trigger`]: https://docs.rs/bevy/latest/bevy/prelude/struct.World.html#method.trigger
+[`Commands`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Commands.html
+[`Commands::trigger`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Commands.html#method.trigger
+[`Observer`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Observer.html
+
+## Event Triggers & Observers
+
+Every Event requires a [`Trigger`], represented by default with the `On<Event>` syntax. It's best to think of a Trigger as the call which activates the Event, hence why we use `On`: our code will run `On` our `<Event>`.
```
alice-i-cecile:
```suggestion
Every `Event` requires a [`Trigger`], represented by default with the `On<Event>` syntax. It's best to think of a Trigger as the call which activates the Event, hence why we use `On`: our code will run `On` our `<Event>`.
```

content/learn/book/control-flow/events.md
```diff
@@ -7,3 +7,268 @@ status = 'hidden'
...
+
+Events are triggered for [`Observers`], special entities which are created specifically to watch for the specified event. If a Trigger is the Event's activation call, then the Observers are the ones who listen for the activation call and run their code in response.
+
+Observers are created using the `add_observer` methods on both [`World`] ([`World::add_observer`]) and [`Commands`] ([`Commands::add_observer`]). Observers are detailed further in the dedicated [Observers] page, but it's worth reviewing how they depend on Triggers and Events.
```
alice-i-cecile: IMO we should cut the `Observers` page as part of this PR (and thus not refer to it here).

alice-i-cecile: Observers can also be created by spawning entities with an Observer component directly, but I'm not sure if that's a particularly useful pattern to teach or promote.

Based-A: Yeah that's fair, I wasn't sure how much to get into the topic since Events and Observers are very connected. Would you say to cut the whole paragraph for creating Observers, or just the line referencing the direct page?

Based-A: I did think about including that, but I felt that that would be better left for the Observer book page directly rather than trying to shoehorn it in here. Although like you mentioned above, it might be best to remove the paragraph on Observers entirely and just focus on the Events/Triggers.

alice-i-cecile: Cut the line referencing the direct page, and delete the stub page for Observers in this PR (I think it's a stub still?).

alice-i-cecile: I don't think that a page about events, distinct from observers makes any sense.

We *do* still need a page on `Messages`, which is why these were previously distinct.

PR #2356, opened 2026-02-06
Add Observers section to Events book page

Based-A:
Here is the Observers section for Events, along with a correction for a mistake I made in the EntityEvents section and some re-wordings to make things a bit clearer.


alice-i-cecile (approved):
Excellent :)

PR #2363, opened 2026-02-20
Stars sorting

cookie1170:
# Why
Being able to see star counts on assets, as well as assets with more stars being shown first makes it easier to find quality assets, especially as the number of assets grows.
# What
This PR adds support for fetching stars on GitHub and GitLab for assets in `bevy-assets`, as well as changing the sort to take stars into account.
The new sort order is:
- Assets with a manually assigned order in `bevy-assets`
- Assets that are semver compatible with bevy, sorted by stars
- Assets that aren't semver compatible, sorted by stars
# How
For each asset, the algorithm looks at its link:
- If it's a link to GitHub or GitLab, then fetch the star info from there
- If it's a link to crates.io, or it has a crate and the previous fetch failed; look at its repository field
# Example
Here is the 2D section of assets before the change:

image
And after:

image

### DIFF
--- a/generate-assets/src/github_client.rs
+++ b/generate-assets/src/github_client.rs
@@ -102,19 +102,11 @@ impl GithubClient {
         }
     }
 
-    /// Gets the star count from a github repo
-    ///
-    /// Note that this method requests the whole repository info
-    /// If any other fields from it are needed, this method may be modified as to not make unnecessary requests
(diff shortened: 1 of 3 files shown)


cart:
Just dumping some context I found here (doesn't need to be done in this PR unless you really want to). Codeberg API requires no authentication:

`GET https://codeberg.org/api/v1/repos/guix/guix`. Stars are under the `stars_count` json field in the response.

Docs: <https://codeberg.org/api/swagger#/repository/repoGet>

PR #2365, opened 2026-02-24
Rewrite Commands book page to include examples and different sections.

Based-A:
As I was working through implementing some of the points from #2161 I thought the page content could be expanded and go into detail a bit more. Not sure if I've covered everything, and I did remove some stuff (like Deferred, for example) as I thought it might be a bit out of the scope for this book page.

### DIFF
--- a/content/learn/book/control-flow/commands.md
+++ b/content/learn/book/control-flow/commands.md
@@ -51,11 +51,15 @@ To use `Commands` in your systems, it is easiest to pass it in as a system param
 
 ```rust
 fn my_system(mut commands: Commands) {
-    // Some functionality that modifies the World!
+    // Add a new Resource to track scores with implicit values.
+    commands.insert_resource(Scoreboard{
+        current_score: 0,
+        high_score: 0,
+    });
 }
 ```
 
-While Bevy offers a number of different pre-defined `Commands` to use, it also offers the [`queue`] method to make changes that aren't provided by the default `Commands`. To use `queue`, we create a closure that mutably accesses the `World` and then define what code we want to run.
+While Bevy offers a number of different pre-defined `Commands` to use, it also offers the [`queue`] method to make changes that aren't provided by the default `Commands`. `queue` gives us mutable access to the `World`, which we can use however we want. The most straightforward approach is to construct our modifications within the method itself (like we do below), however we can go one step further and create custom commands by implementing the `Command` trait on a custom struct, or even extend the `Commands` struct with custom traits and methods. We'll go more into this aspect further down the page in [Custom Commands].
 
 ```rust
 // A custom Resource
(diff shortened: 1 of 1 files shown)


content/learn/book/control-flow/commands.md
```diff
@@ -6,60 +6,246 @@ weight = 5
...
+```
 
-## What order do commands get applied in?
+- Triggering Observers
 
-Each system can hold their own copy of [`Commands`] in their local system state.
-When commands are applied, these queues are evaluated as in the same order that the systems were run.
-Within each system, the commands are applied in a first-in-first-out order.
+```rust
+// Trigger an `event` Event for all Observers watching `event`.
```
alice-i-cecile:
```suggestion
// Trigger the `event` Event for all Observers watching for events of a matching type.
```

content/learn/book/control-flow/commands.md
```diff
@@ -6,60 +6,246 @@ weight = 5
...
+
+// Meanwhile this System runs in the `Update` schedule.
+fn update_system(mut commands: Commands) {
+    commands.trigger(UpdateEvent);
+}
+```
+
+In the above example, we have two systems: `insert_observer_system` running in the `Startup` schedule and `update_system` running in the `Update` schedule. Since the `Startup` schedule only runs once, the `add_observer()` command is only ran when our applications is launched. However the `trigger()` command is run everytime the `Update` schedule finishes, meaning that a `trigger()` command is sent to the queue and ran every time `Update` completes.
+
+In addition, if a system with `Commands` is ordered before another system, that system will always see the effects of the `Commands` in the first system. Bevy ensures this occurs by dynamically inserting synchronization points, during which all `Commands` are applied. Each system can hold their own copy of `Commands` in their local system state. When `Commands` are applied, these queues are evaluated as in the same order that the systems were run. Within each system, the `Commands` are applied in a first-in-first-out order.
```
alice-i-cecile: Briefly mentioning apply_deferred would be good here.

content/learn/book/control-flow/commands.md
```diff
@@ -6,60 +6,246 @@ weight = 5
...
+// Spawn a new Entity with a `bundle` of Components.
+commands.spawn(bundle);
 
-When applying changes to a single entity, the [`Commands`] type is transformed into [`EntityCommands`] via [`Commands::entity`].
-While the [`Command`] trait can have arbitrary effect on the [`World`], the [`EntityCommand`] trait is designed to modify a single entity.
+// Despawn the `entity` Entity.
+commands.entity(entity).despawn(); 
+```
 
-If you want to send commands from within a parallel context (such as via [`Query::par_iter_mut`]), [`ParallelCommands`] can be used.
```
alice-i-cecile, line 26: Some mention of this might be good.

content/learn/book/control-flow/commands.md
```diff
@@ -6,60 +6,246 @@ weight = 5
...
 
-## Custom commands
+To use `Commands` in your systems, it is easiest to pass it in as a system parameter. This ensures that every system that needs access to `Commands` gets it and that all `Commands` that are called are placed into the `Command` queue.
 
-Because of their flexible nature, custom commands are a powerful tool for implementing game-specific operations.
-While they may not be as fast or transparent as working with events or observers,
-the arbitrary flexibility can be great for quickly evolving game logic and performing operations atomically.
+```rust
+fn my_system(mut commands: Commands) {
+    // Some functionality that modifies the World!
```
alice-i-cecile: We should improve this example, or at least the comment. Demonstrating something simple but useful would be good.

content/learn/book/control-flow/commands.md
```diff
@@ -6,60 +6,246 @@ weight = 5
...
+}
+```
 
-Writing custom commands is quite simple: create a struct and implement [`Command`] for it.
-If you want to pass in data, add fields to your struct.
-To send a custom command, simply call `commands.queue(CustomCommandStruct { my_data })`.
+While Bevy offers a number of different pre-defined `Commands` to use, it also offers the [`queue`] method to make changes that aren't provided by the default `Commands`. To use `queue`, we create a closure that mutably accesses the `World` and then define what code we want to run.
```
alice-i-cecile: This is a bit weird, because it frames it like you *have* to use a closure, rather than *can* use a closure. Ordinary functions work too!

content/learn/book/control-flow/commands.md
Based-A: I did move it down into its own section, should it be mentioned in the opening? My thought was to move it completely so that it can be explained with more context rather than just a bullet point in the opening section.

alice-i-cecile: Yeah, fully agree: this is not intro material.

PR #2373, opened 2026-03-04
Messages Book Page

Based-A:
After running through Events and Component Lifecycles, I decided to take a peak into Messages. Even though it seems really straightforward, I feel like there is a lot that can be done with Messages, and so a new book page was a good fit. As I've gone through it, it seems like the focus of the page should be (apart from introducing and exploring Messages) how Messages contrast and are used alongside Events. 

I'd like to get some feedback as to whether I'm on the right track with it or whether I've completely misunderstood some of the concepts. As well, if there are any suggestions or recommendations for sections or examples to include I am all ears. Thanks!

### DIFF
--- a/content/learn/contribute/project-information/project-goals.md
+++ b/content/learn/contribute/project-information/project-goals.md
@@ -52,15 +52,15 @@ Goals are tracked on the [Project Goals Board](https://github.com/orgs/bevyengin
 
 Goals can have a variety of "states", which are expressed as "statuses" on the [Project Goals Board](https://github.com/orgs/bevyengine/projects/23/views/1).
 
-- **Proposed**: One or more **SME** or **Maintainer** thinks this Goal is worth considering
-  - Goals enter this state when they are created by an SME, Maintainer, or Project Lead
+- **Proposed**: One or more **SME**, **Maintainer**, or trusted contributor thinks this Goal is worth considering
+  - Goals enter this state when they are created
   - The SMEs and Project Lead discuss what state to transition the Goal to. Ideally Goals do not stay in the Proposed state for long.
 - **Postponed**: We might want to do this later, but we don't have the bandwidth or inclination to invest in it now.
-- **Blocked (Approved)**: We want to do this, one or more SME has agreed to "staff" it once unblocked, and the Project Lead has approved it, but other work is required first
+- **Blocked (Approved)**: We want to do this, one or more SME has agreed to "staff" it once unblocked, and the **Project Lead** and **SME**s have approved it, but other work is required first
   - Goals that are blocked on other Goals or issues should include a "blocked by" relationship on GitHub.
-- **Inactive (Approved)**: We want to do this, one or more **SME** has agreed to staff it, the **Project Lead** has approved it, but there is no active **Working Group**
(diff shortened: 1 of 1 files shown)


content/learn/book/control-flow/messages.md
```diff
@@ -0,0 +1,203 @@
...
+title = "Messages"
+insert_anchor_links = "right"
+[extra]
+weight = 12
+status = 'hidden'
++++
+
+<!-- TBW -->
+
+What do we do when we want to communicate between systems, but we don't need to immediately react to those communications? We use **Messages**! [`Messages`] offer a communication channel between one or more systems. They can be created and read from using system parameters and are stored in the [`Messages<M>`] resource.
```
alice-i-cecile: I think it's more helpful to frame messages as a tool for "batch processing". You can send and read many messages in an efficient way, and then process them all at once. And unlike events, you can even examine the whole queue, in case you want to do things like "only process the most recent".

content/learn/book/control-flow/messages.md
```diff
@@ -0,0 +1,203 @@
...
+        // Once all of our gameplay systems ran in Update, then we can accurately write
+        // a new ScoreboardUpdate message, which will be updated in the next frame.
+        .add_systems(PostUpdate, write_scoreboard_update)
+    ;
+}
+```
+
+In reality, this setup is too simple to be used without further modifications. For example, because all of our schedules are run *every frame*, it wouldn't be practical to update an actual scoreboard like this. However, you should now be able to see how `Messages` can be used and how they are executed.
+
+## Altering Messages
```
alice-i-cecile, line 208: Make sure to cover both the `ResMut<Messages` case and `MessageMutator`.

content/learn/book/control-flow/messages.md
```diff
@@ -0,0 +1,203 @@
...
+
+While it might be tempting to always use `Events` and `Observers` to immediately update your application, `Messages` can be more efficient than using `Events` in the right contexts. Processing a large number of `Messages` in a single batch is one such case. This is because a single `Message` can be *read* by multiple systems in parallel (although *writing* is still sequential). Add in the fact that `Messages` are read at a predictable fixed point and we can start to see that unless we *need* to immediately react to something occuring in our application, using `Messages` is usually preferred.
+
+[`Events`]: https://docs.rs/bevy/latest/bevy/prelude/trait.Event.html
+
+### Using Messages *With* Events
```
alice-i-cecile: This section should be cut. Those examples in the engine are tech debt.

content/learn/book/control-flow/messages.md
```diff
@@ -0,0 +1,203 @@
...
+Bevy implements `Message` for a number of actions that are also `Events`, including [`GamepadEvent`] and other input events, [`AssetEvent`] when loading assets, and even [`AppExit`] when the application closes. These allow you to add some deferred functionality in a `Message` while also using an `Event` to perform more immediate actions.
+
+[`GamepadEvent`]: https://docs.rs/bevy/latest/bevy/input/gamepad/enum.GamepadEvent.html
+[`AssetEvent`]: https://docs.rs/bevy/latest/bevy/prelude/enum.AssetEvent.html
+[`AppExit`]: https://docs.rs/bevy/latest/bevy/prelude/enum.AppExit.html
+
+
+
+
+## Message Lifespan
```
alice-i-cecile, line 295: Make sure to cover fixed update here!

PR #2376, opened 2026-03-11
Initial draft of upstreaming guidance

alice-i-cecile:
There's been a number of upstreaming initiatives at this point, some of which have gone better than others. Similarly, we've had a number of folks ask about "what about upstreaming this crate?". 

Writing out the guidance lets us agree on principles / best practices in general, and helps teach folks what to think about and how they can help.

### DIFF
--- a/content/learn/contribute/project-information/upstreaming.md
+++ b/content/learn/contribute/project-information/upstreaming.md
@@ -37,7 +37,7 @@ However, there are also "bad reasons to upstream", which on their own are not co
 - this crate is popular (good reasons may exist, but this is not a factor on its own)
 - Bevy has no existing solution for this problem
   - the scope of a game engine is large, but not infinite
-  - each feature must justify its value relative to the maintenance/compile time/binary size costs
+  - each feature must justify its value relative to the maintenance / compile time / binary size costs
   - Bevy is intentionally very extensible: allowing users to 
 - the crate author(s) needs help with maintenance
   - when this happens to crates which are essential to the ecosystem function, we should try and find a solution
--- a/content/learn/contribute/project-information/upstreaming.md
+++ b/content/learn/contribute/project-information/upstreaming.md
@@ -73,7 +73,7 @@ deliberation that would be taken to build this functionality from scratch.
 
 Is this a tiny helper crate? Just submit a PR, giving credit!
 A huge project with foundational implications?
-Please make a [Goal](./project-goals.md) and form a [Working Group](./working-groups.md) to keep track of everything.
+Please propose a [Goal](./project-goals.md) and form a [Working Group](./working-groups.md) to keep track of everything.
 Somewhere in between? Talk it over with the relevant experts, then make a PR!
 
 Before you propose upstreaming work, please check in with the author to ask if they're okay with the plan.
--- a/content/learn/contribute/project-information/upstreaming.md
+++ b/content/learn/contribute/project-information/upstreaming.md
@@ -23,7 +23,7 @@ Like with any feature we might add to Bevy, the most important things to conside
 While maintainers will ultimately make the final call on merging code,
 it can be helpful to reflect on these tradeoffs when making or evaluating proposals to upstream something.
 
-Wen thinking about "should we upstream this crate?" in particular, there are both good and bad arguments to both upstream and not upstream, creating a two-by-two matrix of arguments to consider.
+When thinking about "should we upstream this crate?" in particular, there are both good and bad arguments to both upstream and not upstream, creating a two-by-two matrix of arguments to consider.
 The most obvious quadrant is "good reasons to upstream", which includes:
 
 - this work is widely useful to most Bevy users
--- a/content/learn/contribute/project-information/upstreaming.md
+++ b/content/learn/contribute/project-information/upstreaming.md
@@ -122,7 +122,6 @@ More subjectively, they should also:
 - have a pleasant API that feels Bevy-idiomatic
    - makes use of existing ECS primitives where possible: entities, components, resources, events...
    - prefers simple patterns and avoids new abstractions
-   - defaults to relatively public internals
    - uses enums and new-types to encode intent, rather than relying on argument names
    - minimizes use of macros
 - present a solid foundation to build on, by making good architectural decisions
--- a/content/learn/contribute/project-information/upstreaming.md
+++ b/content/learn/contribute/project-information/upstreaming.md
@@ -114,6 +114,11 @@ More subjectively, they should also:
 - have clear and helpful documentation and examples, ideally at the crate, module and API level
 - offer the basic functionality expected by users trying to do the thing the crate was made to do
 - have a pleasant API that feels Bevy-idiomatic
+   - makes use of existing ECS primitives where possible: entities, components, resources, events...
+   - prefers simple patterns and avoids new abstractions
+   - defaults to relatively public internals
+   - uses enums and new-types to encode intent, rather than relying on argument names
+   - minimizes use of macros
 - present a solid foundation to build on, by making good architectural decisions
 
 In-depth evaluations of these factors are *incredibly* valuable, even from rank-and-file contributors.
--- a/content/learn/contribute/project-information/upstreaming.md
+++ b/content/learn/contribute/project-information/upstreaming.md
@@ -35,7 +35,13 @@ The most obvious quadrant is "good reasons to upstream", which includes:
 However, there are also "bad reasons to upstream", which on their own are not compelling:
 
 - this crate is popular (good reasons may exist, but this is not a factor on its own)
+- Bevy has no existing solution for this problem
+  - the scope of a game engine is large, but not infinite
+  - each feature must justify its value relative to the maintenance/compile time/binary size costs
+  - Bevy is intentionally very extensible: allowing users to 
 - the crate author(s) needs help with maintenance
+  - when this happens to crates which are essential to the ecosystem function, we should try and find a solution
+  - upstreaming is not the only possible solution: adding maintainers or creating alternatives is also possible
 - someone important made this crate and thinks it's neat
 - a commercial user or sponsor of Bevy wants us to upstream it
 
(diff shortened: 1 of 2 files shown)


content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,151 @@
...
+Like with any feature we might add to Bevy, the most important things to consider are:
+
+- is this useful?
+- is this well-made?
+- is this worth the maintenance burden?
+
+While maintainers will ultimately make the final call on merging code,
+it can be helpful to reflect on these tradeoffs when making or evaluating proposals to upstream something.
+
+Wen thinking about "should we upstream this crate?" in particular, there are both good and bad arguments to both upstream and not upstream, creating a two-by-two matrix of arguments to consider.
```
atlv24:
```suggestion
When thinking about "should we upstream this crate?" in particular, there are both good and bad arguments to both upstream and not upstream, creating a two-by-two matrix of arguments to consider.
```

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,151 @@
...
+More subjectively, they should also:
+
+- minimize the use of `unsafe`, not have any UB, and not present serious security concerns
+  - `unsafe` may be unavoidable when interacting with performance-critical or non-Rust code
+- have good tests with reasonable coverage (which pass!)
+- be judicious in their use of dependencies
+  - this is particularly true for other crates which rely on `bevy`!
+- have clear and helpful documentation and examples, ideally at the crate, module and API level
+- offer the basic functionality expected by users trying to do the thing the crate was made to do
+- have a pleasant API that feels Bevy-idiomatic
```
Freyja-moth, line 122: May be worth mentioning what qualifies something as bevy-idiomatic so that users wanting to upstream know what to aim for?

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,151 @@
...
+they just need to be a solid foundation to build on.
+
+## Starting the upstreaming process
+
+The level of deliberation involved in upstreaming something should be proportional to the
+deliberation that would be taken to build this functionality from scratch.
+
+Is this a tiny helper crate? Just submit a PR, giving credit!
+A huge project with foundational implications?
+Please make a [Goal](./project-goals.md) and form a [Working Group](./working-groups.md) to keep track of everything.
```
Jondolf: With the current guidelines, your average contributor isn't allowed to create a Goal directly (and doing so would issue you a warning)
```suggestion
Please propose a [Goal](./project-goals.md) and form a [Working Group](./working-groups.md) to keep track of everything.
```

content/learn/contribute/project-information/upstreaming.md
alice-i-cecile:
```suggestion
- have a pleasant API that feels Bevy-idiomatic
   - makes use of existing ECS primitives where possible: entities, components, resources, events...
   - prefers simple patterns and avoids new abstractions
   - defaults to relatively public internals
   - uses enums and new-types to encode intent, rather than relying on argument names
   - minimizes use of macros
```

CorvusPrudens: This could have an entire document of its own! I think what makes something "Bevy-idiomatic" is also a moving target.

Although maybe something like "APIs should be expressed in the vocabulary of Bevy's ECS" would be enough. A touch vague, but slightly more substantial.

alice-i-cecile: Yeah, I wanted to stick to some reasonably tangible and uncontroversial elements here. This one has more subjectivity than most!

That said, a lot of the most serious problems here that I've seen in various ecosystem crates aren't actually about their ECS usage at all: it's about weird idioms translated from non-Rust languages!

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,156 @@
...
+it can be helpful to reflect on these tradeoffs when making or evaluating proposals to upstream something.
+
+When thinking about "should we upstream this crate?" in particular, there are both good and bad arguments to both upstream and not upstream, creating a two-by-two matrix of arguments to consider.
+The most obvious quadrant is "good reasons to upstream", which includes:
+
+- this work is widely useful to most Bevy users
+- this work is needed by another part of the engine itself (e.g. the Editor)
+- users expect game engines to ship with this functionality out of the box
+- this functionality would be much easier to implement or maintain if it was in one of Bevy's own crates
+- this would be a major improvement over Bevy's existing solution
```
Bluefinger, line 33: In some cases, the existing solution is nothing at all, as Bevy does not provide anything here (though of course, that can be intentionally so for reasons later on in the doc).

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,156 @@
...
+- multiple competing solutions exist and we don't want to pick a winner
+  - there are social difficulties here, and care should be taken
+  - Bevy's modular design can help alleviate the effects on innovation
+  - creating shared unopinionated interfaces can help ease switching costs
+  - ultimately, if the functionality is needed, we have to move forward with a choice
+- there are more features that would be nice to have
+- there are minor bugs
+- documentation could be improved
+
+Just like when considering greenfield PRs, projects don't need to be *finished* to be considered for upstreaming:
```
Bluefinger, line 72: I feel there can be neutral or situational reasons listed here that are both bad and good. For example:

* Only crate in the ecosystem that does the thing

Can be either good or bad reason for upstreaming, since just having the one doesn't mean it is upstreamable, but it also makes it more compelling to upstream if it clears the quality bar.

Also, only having the one load bearing crate, but maintainer is struggling presents a risk factor on the ecosystem (though not one that necessarily is solved by upstreaming).

There are probably a couple more reasons that fit this situational good/bad block.

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,156 @@
...
+   2. Usually in its own crate or well-isolated module
+   3. Requires extensive vetting and quality control in the original repo beforehand
+   4. If replacing an existing solution, add the new solution first, then deprecate and remove the old one in follow-up PRs, testing as you go
+   5. Consider cutting questionable features: they can always be added back later
+2. One-bite-at-a-time
+   1. Find small, independently useful elements that can be contributed upstream and open modest PRs for them
+   2. Change the source repo to target `bevy/main`, and gradually delete code to use the now-upstream version
+   3. Continue to add functionality one small feature at a time until all code has been absorbed
+
+The all-at-once strategy works best for monolithic, mature projects (e.g. an entire physics engine),
```
Bluefinger, line 160: Also consider that some upstreaming doesn't *need* the entire third-party crate. A perfectly valid option is to split the parts that Bevy could provide first-party (the core functionality), and then leave third-party to provide the extended functionality that builds on top of the core. Bevy itself might not need the full scope of a third-party crate as first-party, but with a modular API, users can just compose the functionality they need.

Shatur (approved):
I like it!

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,170 @@
...
+- this functionality would be much easier to implement or maintain if it was in one of Bevy's own crates
+- this would be a major improvement over Bevy's existing solution
+
+However, there are also "bad reasons to upstream", which on their own are not compelling:
+
+- this crate is popular (good reasons may exist, but this is not a factor on its own)
+- Bevy has no existing solution for this problem
+  - the scope of a game engine is large, but not infinite
+  - each feature must justify its value relative to the maintenance/compile time/binary size costs
+  - Bevy is intentionally very extensible: allowing users to
```
Freyja-moth, line 41: I think this might be in the wrong place?

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,170 @@
...
+- users expect game engines to ship with this functionality out of the box
+- this functionality would be much easier to implement or maintain if it was in one of Bevy's own crates
+- this would be a major improvement over Bevy's existing solution
+
+However, there are also "bad reasons to upstream", which on their own are not compelling:
+
+- this crate is popular (good reasons may exist, but this is not a factor on its own)
+- Bevy has no existing solution for this problem
+  - the scope of a game engine is large, but not infinite
+  - each feature must justify its value relative to the maintenance/compile time/binary size costs
```
weihnaxbaum:
```suggestion
  - each feature must justify its value relative to the maintenance / compile time / binary size costs
```
Might just be me, but otherwise I'd read this as "maintenance time size cost / maintenance binary size costs / compile time size costs / compile binary size costs".

content/learn/contribute/project-information/upstreaming.md
```diff
@@ -0,0 +1,170 @@
...
+  - `unsafe` may be unavoidable when interacting with performance-critical or non-Rust code
+- have good tests with reasonable coverage (which pass!)
+- be judicious in their use of dependencies
+  - this is particularly true for other crates which rely on `bevy`!
+- have clear and helpful documentation and examples, ideally at the crate, module and API level
+- offer the basic functionality expected by users trying to do the thing the crate was made to do
+- have a pleasant API that feels Bevy-idiomatic
+   - makes use of existing ECS primitives where possible: entities, components, resources, events...
+   - prefers simple patterns and avoids new abstractions
+   - defaults to relatively public internals
```
cart: I'm not sure this is something we should be pushing for, at least not without more qualifiers / descriptions of the types of things that should be public. I think the "pub everything" and "pub most things" mindsets are both generally a mistake.

cart (commented):
Solid work! Just one minor point of contention :)

content/learn/contribute/project-information/upstreaming.md
alice-i-cecile:
```suggestion
```

cart (approved):

content/learn/contribute/project-information/upstreaming.md
jasmine-nominal: @alice-i-cecile I think this was overlooked

PR #2377, opened 2026-03-12
Book: Fix links with missing urls

SpecificProtagonist:
There were some source code links that were missing an url. Zola has a link checker, but it only seems to check whether the urls/paths point at something existent, not whether the the link actually has one?

There are times where a Bevy type is referenced multiple times on the same page, but it's not consistent on whether all of them or only the first one is a link. Do we have guidance on this? If so, it should be added to the contributor's style guide. I've left them alone for now.

There are also a few links I don't know what they're supposed to link to, which I've left as is for a future editing pass.

### DIFF
--- a/content/learn/book/control-flow/change-detection.md
+++ b/content/learn/book/control-flow/change-detection.md
@@ -15,6 +15,9 @@ synchronizing data between two contexts.
 In this section, we'll explore how Bevy tracks changes and how you can leverage this
 feature in your game development workflow.
 
+[`Component`]: https://docs.rs/bevy/latest/bevy/ecs/component/trait.Component.html
+[`Resource`]: https://docs.rs/bevy/latest/bevy/prelude/trait.Resource.html
+
 ## Filtering
 
 You can configure queries to filter out entities unless certain components have been modified.
@@ -41,6 +44,9 @@ fn detect_changed_position(query: Query<(Entity, &Position), Changed<Position>>)
     }
 }
 ```
+
+[`Changed<T>`]: https://docs.rs/bevy/latest/bevy/ecs/query/struct.Changed.html
+[`Added<T>`]: https://docs.rs/bevy/latest/bevy/ecs/query/struct.Added.html
 
 Removing components works differently, see the section below.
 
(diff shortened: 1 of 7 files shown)


content/learn/book/storing-data/queries.md
```diff
@@ -39,7 +39,13 @@ we can combine [`QueryData`] or [`QueryFilter`] types by putting them inside of
 wrapping them with parentheses.
 
 [generic type parameters]: (https://doc.rust-lang.org/book/ch10-01-syntax.html)
+[dependency injection]: https://en.wikipedia.org/wiki/Dependency_injection
 ["unit type"]: https://doc.rust-lang.org/core/primitive.unit.html
+[tuple]: https://doc.rust-lang.org/rust-by-example/primitives/tuples.html
+[`Query<D, F>`]: https://docs.rs/bevy/latest/bevy/ecs/entity/struct.Entity.html
```
Based-A: Wrong link I think? Or I might be missing why `Query<D, F>` should link to the Entities struct?

content/learn/book/storing-data/queries.md
```diff
@@ -16,7 +16,7 @@ We introduced queries briefly in our [introduction](../intro/) to Bevy: if you'r
 
 ## Anatomy of a query
 
-To understand how queries work in a bit more detail, let's take a look at the anatomy of a [`Query`].
```
Based-A, line 19: IMO this link should stay, and link to the Query struct docs page?

SpecificProtagonist: I removed it because the full type is already linked two words later.

content/learn/book/storing-data/queries.md
SpecificProtagonist: Thanks, fixed.

content/learn/book/storing-data/queries.md
Based-A: Makes sense. Might be worth revisiting if the full type needs to be in the book rather than just `Query`, but that's irrelevant to this.

PR #2380, opened 2026-03-14
Systems Book Page Rewrite

Based-A:
Filled out the Systems book page with examples, along with some re-wordings and minor expansions.

### DIFF
--- a/content/learn/book/control-flow/systems.md
+++ b/content/learn/book/control-flow/systems.md
@@ -19,29 +19,29 @@ Each of these is covered in their own chapters throughout this book, but we'll b
 ```rust
 // A simple system accessing a Commands parameter, a Query parameter,
 // and a mutable Resource (ResMut) parameter.
-fn bevy_system(
+fn register_players_system(
     // Our System's System Parameters.
     mut commands: Commands,
-    query: Query<Entity, Without<MyComponent>>,
-    mut res: ResMut<MyResource>
+    player_query: Query<Entity, With<ActivePlayerMarker>>,
+    mut player_resource: ResMut<PlayerListResource>
 ) {
     // Using our parameters in a System.
-    for query_entity in query.iter() {
-        // Insert MyComponent into the Entity returned by the Query.
-        commands.entity(query_entity).insert(MyComponent);
+    for player_entity in player_query.iter() {
+        // Insert a InGameMarker component into the Entity returned by the Query.
+        commands.entity(player_entity).insert(InGameMarker);
         // Add the Entity to our Resource.
         res.entity_field.push(query_entity);
     }
 }
 
-// A simple Component.
+// A simple Component to mark a Player as being in game.
 #[derive(Component)]
-struct MyComponent;
+struct InGameMarker;
 
-// A simple Resource.
+// A simple Resource to track a list of players in the current game.
 #[derive(Resource)]
-struct MyResource {
-    pub entity_field: Vec<Entity>,
+struct PlayerListResource {
+    pub player_list: Vec<Entity>,
 }
 ```
 
@@ -52,16 +52,19 @@ As long as every field on our custom struct is a system parameter, we can access
 ```rust
 // Create a custom system parameter accessing a ResMut resource and a Query.
 #[derive(SystemParam)]
-pub struct CustomSystemParameter{
-    field_1: ResMut<'w, MyResource>,
-    field_2: Query<'w, 's, Entity, With<MyComponent>>,
+pub struct PlayerSystemParameter{
+    commands: Commands<'w, 's>,
+    player_query: Query<'w, 's, Entity, With<ActivePlayerMarker>>,
+    list_of_players: ResMut<'w, PlayerListResource>,
 }
 
-// Use our custom system parameter to add the Entities from our 
-// Query to the ResMut resource.
-fn access_custom_parameter(mut custom_parameter: CustomSystemParameter) {
-    for entity in custom_parameter.field_2.iter() {
-        custom_parameter.field_1.entity_field.push(entity);
+// Use our custom system parameter to replcate the same functionality
+// as the system in the previous example.
+fn register_players_system(mut player_parameter: PlayerSystemParameter) {
+    for player_entity in player_parameter.player_query.iter() {
+        player_parameter.commands.entity(player_entity).insert(InGameMarker);
+        
+        player_parameter.list_of_players.player_list.push(entity);
     }
 }
 ```
--- a/content/learn/book/control-flow/systems.md
+++ b/content/learn/book/control-flow/systems.md
@@ -12,7 +12,7 @@ As the third pillar of the ECS model, systems give us the power to operate over 
 When we create entities with similar components, those components are laid out beside each other in memory.
 This gives us good [cache locality] when operating on them in a batched fashion.
 
-Systems in Bevy are any Rust function where every argument implements the [`SystemParam`] trait.
+A Bevy `System` is any Rust function where every argument implements the [`SystemParam`] trait.
 Some common parameters you'll use and see in Bevy are [`Query`], [`Res`], [`ResMut`], [`MessageReader`], [`MessageWriter`], [`Local`] and [`Commands`].
 Each of these is covered in their own chapters throughout this book, but we'll be using them here to help explain the concepts that directly tie in to systems.
 
@@ -45,16 +45,29 @@ struct MyResource {
 }
 ```
 
-Bevy provides these and other parameters by default, but you can also create your own by implementing the [`SystemParam`] trait on a custom struct.
-Note that the fields on a custom struct must also implement the `SystemParam`  trait, along a couple of other restrictions as well (check the [`SystemParam`] page for more details).
+Bevy allows us to access these parameters individually by default, but we can be more ergonomic. 
+If we have several parameters we know we want to access repeatedly, we can create a custom system parameter by implementing the [`SystemParam`] trait on a custom struct.
+As long as every field on our custom struct is a system parameter, we can access the custom struct as a single system parameter.
 
 ```rust
+// Create a custom system parameter accessing a ResMut resource and a Query.
 #[derive(SystemParam)]
 pub struct CustomSystemParameter{
-    field_1: Res<'w, SomeResource>,
-    field_2: Query<'w, 's, &'static SomeComponent>,
+    field_1: ResMut<'w, MyResource>,
+    field_2: Query<'w, 's, Entity, With<MyComponent>>,
+}
+
+// Use our custom system parameter to add the Entities from our 
+// Query to the ResMut resource.
+fn access_custom_parameter(mut custom_parameter: CustomSystemParameter) {
+    for entity in custom_parameter.field_2.iter() {
+        custom_parameter.field_1.entity_field.push(entity);
+    }
 }
 ```
+
+There are some additional caveats and restrictions that come with the `SystemParam` trait, but those are beyond the scope of this section. 
+If you're interested, you can check the [`SystemParam`] page for more details.
 
 [cache locality]: https://en.wikipedia.org/wiki/Locality_of_reference
 
@@ -72,8 +85,12 @@ pub struct CustomSystemParameter{
 ## Accessing Data In Systems
 
 One of the major benefits of Bevy's system abstraction is that it easily and efficiently ["splits the borrow"] of a `World`.
-Multiple systems can access non-overlapping parts of the world at the same time.
-Systems can even run in parallel if they aren't accessing the same data with the help of schedule executors like [`MultiThreadedExecutor`].
+When a system is ran, the requested data is automatically fetched from the `World`. 
+Other systems are prevented from accessing the requested data if their access would violate the rules of the borrow checker. 
+The prime directive of Rust still applies: accessing `World` data can be mutable *or* shared, but never both at once.
+
+We can see this in action when using multiple systems.
+If two systems access non-overlapping parts of the world they can run at the same time. 
 In the example below, both systems would be able to run in parallel since they do not access the same data.
 
 ```rust
@@ -93,22 +110,55 @@ fn access_resource(custom_resource: Res<CustomResource>) {
 }
 ```
 
-The requested data is automatically fetched from the `World` when the system is run, locking out access to the underlying data to avoid violating the rules of the borrow checker.
-When working within a system, or when running multiple systems at once, the prime directive of Rust applies: access can be mutable *or* shared, but never both at once.
+If two systems do access the same data, as long as one system isn't *mutably* accessing the data then both systems could run in parallel.
+Both systems in the example below access the same data, but neither changes the data.
+These systems can run in parallel.
 
 ```rust
-// This system would not compile because it accesses the same Transform data 
-// in two queries simultaneously.
+// This System immutably accesses the Transform component. 
+fn print_player_and_enemy_locations(
+    player_query: Single<&Transform, With<Player>>,
+    enemy_query: Single<&Transform, With<Enemy>>
+) {
+    println!("Player Location: {:?}", player_query.translation);
+    println!("Enemy Location: {:?}", enemy_query.translation);
+}
+
+// While this System also immutably accesses the same Transform components,
+// meaning it can run alongside `print_player_and_enemy_locations`
+// even though they both access the same data.
+fn compare_player_and_enemy_locations(
+    player_query: Single<&Transform, With<Player>>,
+    enemy_query: Single<&Transform, With<Enemy>>,
+    target_location: Res<GoalLocation>
+) {
+    let player_distance = target_location.0 - player_query.translation;
+    let enemy_distance = target_location.0 - enemy_query.translation;
+    
+    if player_distance < enemy_distance {
+        println!("The Player is closer to the Target!");
+    } else {
+        println!("The Enemy is closer to the Target!");
+    }
+}
+```
+
+Even though we've been talking about multiple systems, single systems also have to abide by the borrow checker when accessing data.
+As an example, `Component` data cannot be accessed both mutably and immutably without using some workarounds (which you can read about in the [Queries book section]). 
+The same concept applies to resource access as well.
+
+```rust
+// This system would not compile because it accesses the 
+// Transform component data both mutably and immutably.
 fn player_and_enemy_access(
     player_query: Single<&Transform, With<Player>>,
-    same_player_query: Single<&Transform, With<Player>>,
-    enemy_query: Single<&Transform, With<Enemy>>
+    enemy_query: Single<&mut Transform, With<Enemy>>
 ) {
     ...
 }
 
-// This system would not compile because it accesses the same data both 
-// mutably and immutably.
+// This system would not compile because it accesses the same resource 
+// data both mutably and immutably.
 fn access_resource(
     custom_resource: Res<CustomResource>,
     mut mut_custom_resource: ResMut<CustomResource>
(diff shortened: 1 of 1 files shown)


content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
 status = 'hidden'
 +++
 
-Systems are the primary way that Bevy programs interact with the [`World`].
-Systems are great for regularly scheduled logic,
-especially when you want to operate over a large number of similar entities in an efficient way.
+**Systems** are the primary way that Bevy applications interact with the [`World`].
+When you need to perform regularly scheduled logic, a [`System`] will usually be the best tool for the job.
+As the third pillar of the ECS model, systems give us the power to operate over a large number of similar entities in an efficient way.
+Since components of a given type are laid out beside each other in memory, we get good [cache locality] when operating on them in a batched fashion.
```
fallible-algebra: There is an argument here that this should be "Entities with the same sets of components" / some other alluding to archetypes but I think this is a generally useful simplification.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+    println!("Custom Resource Value: {:?}", custom_resource.value);
+}
+```
 
-The requested data is automatically fetched from the [`World`] when the system is run,
-locking out access to the underlying data to avoid violating the rules of the borrow checker.
-When working within a system, or when running multiple systems at once, the prime directive of Rust applies:
-access can be mutable or multiple: but never both at once.
+The requested data is automatically fetched from the `World` when the system is run, locking out access to the underlying data to avoid violating the rules of the borrow checker.
+When working within a system, or when running multiple systems at once, the prime directive of Rust applies: access can be mutable or multiple, but never both at once.
```
fallible-algebra:
```suggestion
When working within a system, or when running multiple systems at once, the prime directive of Rust applies: access can be mutable or shared, but never both at once.
```

"Mutable or multiple" scans a bit poorly here, "mutable or shared" is a little better / more 'idiomatic' but I can see space for a third option too.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
 
-## Running systems in schedules
+[local system state]: ../storing-data/local-system-param
+
+[`MultiThreadedExecutor`]: https://docs.rs/bevy/latest/bevy/ecs/schedule/struct.MultiThreadedExecutor.html
+
+## Running Systems In Schedules
+
+Systems are usually repeatedly ran throughout the life of your application.
+We are able to control when systems run and how systems are ordered by placing them into a [`Schedule`].
```
fallible-algebra, line 194: There's an argument to go into a little bit more detail about the motivation behind splitting up the "Game loop" into schedules up here rather than further down, but my own attempts at suggestions got a bit too wordy.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+```rust
+fn main() {
+    App::new()
+        .add_systems(Update, my_system);
+}
+```
 
-There's a great deal of complexity and nuance here: please see the [Game Loop chapter] for a more complete treatment of the topic.
+In the above example, we're running the `my_system` system in the [`Update`] schedule.
+`Update` is one of the standard schedules that Bevy provides, but there are many more to choose from.
```
fallible-algebra:
```suggestion
`Update` is one of the standard schedules that Bevy provides, but there are many more to choose from. Each one represents a different "Stage" of a frame (or other part of the app lifecycle).
```

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+    println!("This is a one-shot system that is ran on demand.");
+}
+
+// System that runs on command and takes a usize input.
+fn one_shot_system_with_input(In(input): In<usize>) {
+    println!("This is a one-shot system ran on demand with input: {}", input);
+}
+```
+
+In many cases though, simply passing in the function via [`World::run_system_cached`] is more convenient.
```
fallible-algebra: It might be the case that `World::run_system_cached` / `World::run_system_once` should be mentioned at the same time as the running-via-systemid method, especially as this has been the more maintainable + type-retaining (you can retrieve output data of systems, not only give inputs) way of running one-shot systems in practice ime (having designed a card game backend with it).

Could do this by briefly mentioning both 'modes' of doing this up near the top of this section in an enumerated list, just so people can quickly parse that there's two ways of doing this. It'll catch people's eyes and prime them for understanding benefits and trade-offs.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+In order to safely mutate our `World`, we have to use **exclusive systems**.
 
-Systems can optionally take input, and produce output, via the `In` and `Out` generics on the various system traits
-such as [`System`].
+Exclusive Systems:
+
+- *Can* access any data in the world.
+- *Cannot* be run in parallel with any other system.
+- *Do not* use [`Commands`], including to spawn or despawn entities or to insert or remove components.
+- Are created by using `&mut World` as the first system parameter.
```
fallible-algebra, line 348: I really like this summary!

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+    // This removes all entities from the World.
+    world.clear_entities();
+    // This adds a custom schedule to be ran in the World.
+    world.add_schedule(MySchedule);
+}
+```
 
+Exclusive systems are useful for operations that require making large or unique changes to your `World`.
+Spawning large numbers of entities at once is one example of this, as there is no additional overhead incurred from *queueing* these changes like there is when using `Commands`.
+They are also extremely useful for unusually complex game logic or control flow, as any data, and can be used to immediately run one-shot systems or schedules.
```
fallible-algebra:
```suggestion
They are also extremely useful for unusually complex game logic or control flow, as any data, and can be used to immediately run one-shot systems (including other exclusive systems) or schedules.
```
Mostly suggesting this to re-emphasise the "top-down recursive" model of one-shot systems architecture. One can imagine a cautious beginner assuming exclusive systems cannot call exclusive systems (though it could still be too niche+complexityfootgun to add to the list up top).

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
-If we call [`World::run_system_once_with`] we can pass in an input value, and return an output value,
-just like any other function.
-The output type is always inferred, but when using system input the first parameter must be [`In<T>`],
-where `T` is any type of input that you want to pass in.
+
+```rust
+fn call_system(mut world: &mut World) {
+    // Call the one-shot system with an input value of 42.
+    let system_value = world.run_system_once_with(one_shot_system_with_input, 42);
+    // After going through the one-shot system, system value will equal 44.
```
fallible-algebra:
```suggestion
    // After going through the one-shot system, system value will equal Ok(44).
```

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+// This System will register and then run two One-shot Systems.
+fn register_one_shot_systems(mut world: &mut World) {
+    // Register two One-shot Systems.
+    let some_system_id = world.register_system(one_shot_system);
+    let another_system_id = world.register_system(one_shot_system_with_input);
+    
+    // Run the first One-shot System. 
+    world.run_system(some_system_id);
+    // Run the second One-shot System.
+    world.run_system_with(another_system_id, 5);
```
fallible-algebra, line 248: Should maybe show (might be too in the weeds to mention directly) how running these returns a result? Even if there's no "output" data due to them being run from systemid.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+
+fn one_shot_system_with_input(In(input): In<usize>) -> usize {
+    // `input` is passed into the system with an `In` generic value.
+    input += 2;
+    // This function returns a usize value, and is thus our `Out` value.
+    input
+}
+```
+
+If we call [`World::run_system_once_with`] we can pass in an input value, and return an output value, just like any other function.
```
fallible-algebra:
```suggestion
If we call [`World::run_system_once_with`] we can pass in an input value, and return an output value, just like any other function, but wrapped in a `Result` in case the conditions of the system parameters weren't met.
```
Another maybe too-detailed suggestion here.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
 The application of this is simplest to understand in the context of one-shot systems.
-If we call [`World::run_system_once_with`] we can pass in an input value, and return an output value,
-just like any other function.
-The output type is always inferred, but when using system input the first parameter must be [`In<T>`],
-where `T` is any type of input that you want to pass in.
+
+```rust
+fn call_system(mut world: &mut World) {
+    // Call the one-shot system with an input value of 42.
+    let system_value = world.run_system_once_with(one_shot_system_with_input, 42);
```
fallible-algebra:
```suggestion
    let system_value: Result<usize, _> = world.run_system_once_with(one_shot_system_with_input, 42);
```

fallible-algebra (commented):
This is such a good section! Very excited to see the work here. Apologies if I've retread some existing revisions here, I'm just picking out what I run into when I see it.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
+    query: Query<&Component>,
+    mut res: ResMut<Resource>
+) {
+    // Using our parameters in a System.
+    for item in query.iter() {
+        let new_entity = commands.spawn(BundleOfComponents).id();
+        
+        res.entity_field.push(new_entity);
+    }
+}
```
Jondolf: `Component` and `Resource` would likely conflict with the trait names in an actual application

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,333 @@ weight = 0
...
 
-That high-performance access pattern is why ECS was originally invented:
-because components of a given type are laid out beside each other in memory,
-we get good [cache locality] when operating on them in a batched fashion.
+Systems in Bevy are any Rust function where every argument implements the [`SystemParam`] trait.
+Some common parameters you'll use and see in Bevy are [`Query`], [`Res`], [`ResMut`], [`MessageReader`], [`MessageWriter`], [`Local`] and [`Commands`].
+Each of these is covered in their own chapters throughout this book, but we'll be using them here to help explain the concepts that directly tie in to systems.
+
+```rust
+// A simple system accessing the Commands, Query, and Resource parameters.
```
Jondolf:
```suggestion
// A simple system accessing the Commands, Query, and ResMut parameters.
```

content/learn/book/control-flow/systems.md
Based-A: Good catch, thanks!

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,366 @@ weight = 0
...
+**Systems** are the primary way that Bevy applications interact with the [`World`].
+When you need to perform regularly scheduled logic, a [`System`] will usually be the best tool for the job.
+As the third pillar of the ECS model, systems give us the power to operate over a large number of similar entities in an efficient way.
+When we create entities with similar components, those components are laid out beside each other in memory.
+This gives us good [cache locality] when operating on them in a batched fashion.
 
-That high-performance access pattern is why ECS was originally invented:
-because components of a given type are laid out beside each other in memory,
-we get good [cache locality] when operating on them in a batched fashion.
+Systems in Bevy are any Rust function where every argument implements the [`SystemParam`] trait.
```
SpecificProtagonist: Grammatical number mismatch

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,366 @@ weight = 0
...
+
+// A simple Resource.
+#[derive(Resource)]
+struct MyResource {
+    pub entity_field: Vec<Entity>,
+}
+```
+
+Bevy provides these and other parameters by default, but you can also create your own by implementing the [`SystemParam`] trait on a custom struct.
+Note that the fields on a custom struct must also implement the `SystemParam`  trait, along a couple of other restrictions as well (check the [`SystemParam`] page for more details).
```
SpecificProtagonist: This mixes implementing in general and deriving – with a custom implementation you are not bound to merely composing existing system params this way. But I don't think we need to talk about manual implementations in the book.

```suggestion
Bevy provides these and other parameters by default, but you can also create your own by bundling existing ones in a custom struct and deriving the [`SystemParam`] trait (check the [`SystemParam`] page for more details).
```

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,366 @@ weight = 0
...
+
+## Accessing Data In Systems
 
-One of the major benefits of Bevy's system abstraction is that it easily and efficiently ["splits the borrow"] of a [`World`],
-allowing users to access non-overlapping parts of the world at the same time within each system,
-and with the help of schedule executors like [`MultiThreadedExecutor`],
-across multiple systems in parallel at the same time.
+One of the major benefits of Bevy's system abstraction is that it easily and efficiently ["splits the borrow"] of a `World`.
+Multiple systems can access non-overlapping parts of the world at the same time.
+Systems can even run in parallel if they aren't accessing the same data with the help of schedule executors like [`MultiThreadedExecutor`].
```
SpecificProtagonist: This makes it sound like you need to deliberately use `MultiThreadedExecutor`, but its the default.
```suggestion
Systems can even run in parallel if they aren't accessing the same data.
```

Also I think it would be better to move the mut⊕shared explanation from the second example to here to replace the "Systems can even run in parallel if they aren't accessing the same data", as it's misleading: Systems can access the same data and not conflict, and can also (since #16885) access only different archetypes and still conflict.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,366 @@ weight = 0
...
+Each standard schedule provides access to a different point of time for a single frame, allowing you to place your systems in relation to the state of the frame.
+You can read more about schedules in the dedicated [Schedules section], but for now you just need to know a couple things:
 
-## One-shot systems
+- By default, systems within a single schedule run in parallel.
+- Systems within a single schedule can also be explicitly ordered to run relative to each other.
+- Each of the standard schedules are evaluated *once per frame*.
+- There are also [`Fixed`] schedules which run at a consistent interval.
+
+[Schedules section]: ../the-game-loop/schedules
```
SpecificProtagonist, line 214: That reminds me, it's not clear to me whether the schedules chapter should be in "control flow" or "the game loop". Nothing for this PR to consider though.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,366 @@ weight = 0
...
+[`App::add_system`]: https://docs.rs/bevy/latest/bevy/app/struct.App.html#method.add_systems
+[`Update`]: https://docs.rs/bevy/latest/bevy/app/struct.Update.html
+[`Fixed`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Fixed.html
+
+## One-shot Systems
 
 Systems can also be run on demand, via a "one-shot" pattern.
 This is an extremely flexible tool, allowing you to execute arbitrary logic on the world in an ergonomic way whenever you please.
-One-shot systems are particularly useful for testing, handling callbacks in UI, creating scripted events or when architecting turn-based games.
+One-shot systems are particularly useful for testing, handling callbacks in UI, creating scripted events, or when architecting turn-based games.
```
SpecificProtagonist: What makes one-shot systems particularly useful for turn-based games?

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,366 @@ weight = 0
...
+    // After going through the one-shot system, system_value will equal `Ok<44>`.
+    println!("System value: {}", system_value);
+}
+
+fn one_shot_system_with_input(In(input): In<usize>) -> usize {
+    // `input` is passed into the system with an `In` generic value.
+    input += 2;
+    // This function returns a usize value, and is thus our `Out` value.
+    input
+}
```
SpecificProtagonist, line 387: The `input` binding needs to be marked `mut` for this to compile (and I think not doing so here is a tiny bit clearer).
```suggestion
fn one_shot_system_with_input(In(input): In<usize>) -> usize {
    // `input` is passed into the system with an `In` generic value.
    // This function returns a usize value, which is thus our `Out` value.
    input + 2
}
```

content/learn/book/control-flow/systems.md
Based-A: Fair point!

content/learn/book/control-flow/systems.md
Based-A: I would imagine it has to do with running the one-shot systems on a specific turn based on different events or situations. But it probably is a bit too broad of a claim to be provided without an example, which is definitely not within the scope of the book.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,415 @@ weight = 0
...
+    field_2: Query<'w, 's, Entity, With<MyComponent>>,
+}
+
+// Use our custom system parameter to add the Entities from our 
+// Query to the ResMut resource.
+fn access_custom_parameter(mut custom_parameter: CustomSystemParameter) {
+    for entity in custom_parameter.field_2.iter() {
+        custom_parameter.field_1.entity_field.push(entity);
+    }
+}
```
SpecificProtagonist, line 68: From the [style guide](https://bevy.org/learn/contribute/helping-out/writing-docs/#contributor-s-style-guide): Prefer game-relevant, descriptive examples and variable names over generic ones like MyEvent.

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,415 @@ weight = 0
...
+
+// A simple Resource.
+#[derive(Resource)]
+struct MyResource {
+    pub entity_field: Vec<Entity>,
+}
+```
+
+Bevy allows us to access these parameters individually by default, but we can be more ergonomic. 
+If we have several parameters we know we want to access repeatedly, we can create a custom system parameter by implementing the [`SystemParam`] trait on a custom struct.
```
SpecificProtagonist:
```suggestion
If we have several parameters we want to access together repeatedly, we can be more ergonomic: We can create a custom system parameter by deriving the [`SystemParam`] trait on a custom struct.
```

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,415 @@ weight = 0
...
+    println!("Enemy Location: {:?}", enemy_query.translation);
+}
+
+// While this System accesses a Resource.
+fn access_resource(custom_resource: Res<CustomResource>) {
+    println!("Custom Resource Value: {:?}", custom_resource.value);
+}
+```
+
+If two systems do access the same data, as long as one system isn't *mutably* accessing the data then both systems could run in parallel.
```
SpecificProtagonist: They can even be parallel if they don't access the same data at all :p
```suggestion
Systems can run in parallel, as long as one system isn't *mutably* accessing the same data as the other system.
```

content/learn/book/control-flow/systems.md
```diff
@@ -6,107 +6,415 @@ weight = 0
...
+    let player_distance = target_location.0 - player_query.translation;
+    let enemy_distance = target_location.0 - enemy_query.translation;
+    
+    if player_distance < enemy_distance {
+        println!("The Player is closer to the Target!");
+    } else {
+        println!("The Enemy is closer to the Target!");
+    }
+}
+```
```
SpecificProtagonist: Since we're talking about mutability, maybe we could use an example that has it, e.g.:
```suggestion
fn damage_player(
    mut player_query: Single<(&Transform, &mut Stats), With<Player>>,
    enemy_query: Single<&Transform, With<Enemy>>,
) {
    let (player_transform, player_stats) = player_query.deref_mut();

    let distance_to_player = enemy_query
        .translation
        .distance(player_transform.translation);

    if distance_to_player < DAMAGE_RADIUS {
        player_stats.health -= 1
    }
}
```

SpecificProtagonist (approved):

PR #2388, opened 2026-03-25
Fill in missing content for states page

alice-i-cecile:
This page was incomplete! We should complete it so we can ship the book!

The hardest bit here was finding a motivating example for `ComputedState`: I decided to lean on the existing [computed_states.rs example](https://github.com/bevyengine/bevy/blob/main/examples/state/computed_states.rs), and use a unifying game state, focusing on the benefits for reducing fragility.

I'm still not content with the flow and framing of this chapter (States and Schedules / States and Entities are bad names), computed / substates come too early. But I'm going to fix that in a follow-up PR for ease of review.

### DIFF
--- a/content/learn/book/control-flow/states.md
+++ b/content/learn/book/control-flow/states.md
@@ -170,7 +170,7 @@ Similar helper components exist: [`DespawnOnEnter`], for when you want to clean 
 
 Sometimes you want a state that you can't directly change, but that is instead *derived* from one or more other states.
 
-In our arcade game, many systems — the HUD, the score display, enemy spawning — need to run during *both* `Playing` and `LevelComplete`, but not during `Intro`.
+In our arcade game, many systems — the HUD, the score display, maybe even player movement — need to run during *both* `Playing` and `LevelComplete`, but not during `Intro`.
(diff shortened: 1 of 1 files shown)


content/learn/book/control-flow/states.md
```diff
@@ -160,12 +160,114 @@ fn spawn_enemy(mut commands: Commands) {
...
+
 [`DespawnOnExit`]: https://docs.rs/bevy/latest/bevy/prelude/struct.DespawnOnExit.html
+[`DespawnOnEnter`]: https://docs.rs/bevy/latest/bevy/prelude/struct.DespawnOnExit.html
+[`DespawnWhen`]: https://docs.rs/bevy/latest/bevy/prelude/struct.DespawnWhen.html
+
+## Computed States: Automatically-Derived States
+
+Sometimes you want a state that you can't directly change, but that is instead *derived* from one or more other states.
+
+In our arcade game, many systems — the HUD, the score display, enemy spawning — need to run during *both* `Playing` and `LevelComplete`, but not during `Intro`.
```
alice-i-cecile:
```suggestion
In our arcade game, many systems — the HUD, the score display, maybe even player movement — need to run during *both* `Playing` and `LevelComplete`, but not during `Intro`.
```

PR #2389, opened 2026-03-25
State chapter editing and reorganization

alice-i-cecile:
The core issues here:

1. The ordering of the content was poor: substates and computed states need to come last
2. The section titles were not very helpful to readers
3. Some important context for how states are used in practice was not communicated clearly enough.

This is the re-org / polish promised in #2388 :)

### DIFF
--- a/content/learn/book/control-flow/states.md
+++ b/content/learn/book/control-flow/states.md
@@ -122,9 +122,6 @@ When a transition does occur, the schedules run in this order:
 2. [`OnTransition`] for the transition
 3. [`OnEnter`] for the new state
 
-For sub-states and computed states, the transitions will cascade. 
-If changing `GameState` causes `ActionState` to be removed, the `OnExit` schedule for the old `ActionState` value will run as well.
-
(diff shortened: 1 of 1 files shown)


viridia (approved):

content/learn/book/control-flow/states.md
```diff
@@ -132,21 +89,62 @@ app.add_systems(Update, update_enemies.run_if(in_state(GameState::Playing)));
...
+This means that systems which will run later in the _same_ tick will still see the _old_ state.
+
+When a transition does occur, the schedules run in this order:
+
+1. [`OnExit`] for the old state
+2. [`OnTransition`] for the transition
+3. [`OnEnter`] for the new state
+
+For sub-states and computed states, the transitions will cascade. 
+If changing `GameState` causes `ActionState` to be removed, the `OnExit` schedule for the old `ActionState` value will run as well.
```
Based-A: This wording is a little vague/confusing. Is it supposed to mean that `OnExit` will still run for `ActionState` even if we change `GameState` to another state? Separately, it's also a little iffy to include it here because it's referencing SubStates before SubStates are talked about.

PR #2392, opened 2026-03-26
Add some docs on asset hot reloading.

andriyDev:
I've left the hot reloading code section blank since I have no experience here.

Feel free to take this, rewrite it entirely, or leave it to rot! I don't mind.

If this isn't the right level of detailed, please let me know! And if there's more questions, I'd love to know what needs to be clarified.


SpecificProtagonist:
Might be worth mentioning `AssetEvent::Modified` as a way of reacting to assets having been changed due to e.g. hot reloading.

Based-A (approved):

PR #2393, opened 2026-03-26
Write testing chapter

alice-i-cecile:
This is the last outstanding chapter that I feel that we *must* have before the book's MVP release 🎉

Testing is important and underrated, and I think that sharing best-practices here will be very effective.

This was a research-heavy chapter, especially with respect to the graphical testing, so please let me know if you spot any errors.
Unfortunately, that section was too important to leave out: other guides on testing in Rust simply don't cover it.

### DIFF
--- a/content/learn/book/development-practices/testing.md
+++ b/content/learn/book/development-practices/testing.md
@@ -425,7 +425,7 @@ Bevy's own plugins are designed to work quite robustly without rendering.
 For example, `bevy_ui` and `bevy_ui_render` are different crates with their own plugins.
 The best way to ensure that your code works without rendering (sometimes called "headlessly") is to create a `headless` feature flag for your application, and then disable plugins which are causing crashes in a feature-flagged cfg block:
 
-```rust, hide-lines=1
+```rust, hide-lines=1,hide_lines=1
 # use bevy::prelude::*;
 let mut app = App::new();
 
@@ -470,7 +470,7 @@ We'll cover the concepts and point you to the right tools, but expect to dig int
 Bevy provides the [`Screenshot`] component for programmatic screen capture.
 Spawn an entity with a `Screenshot` component and attach an observer to handle the result:
 
-```rust, hide-lines=1
+```rust, hide-lines=1,hide_lines=1
 # use bevy::prelude::*;
 fn take_screenshot(mut commands: Commands) {
 commands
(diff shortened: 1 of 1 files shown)


content/learn/book/development-practices/testing.md
```diff
@@ -5,4 +5,735 @@ insert_anchor_links = "right"
...
+- **Set a difference threshold.** Anti-aliasing and floating-point rounding differ across platforms. Allow some pixel deviation rather than demanding exact matches.
+- **Compare per-platform.** Each GPU, and each software renderer will produce slightly different output. Don't compare Linux screenshots against macOS ones.
+
+Bevy uses [Pixel Eagle] for this:
+an external service that stores screenshots per commit and per platform,
+and compares them against `main`.
+If a PR changes rendering output, a bot comments on the PR with a link to the visual diff.
+
+For your own project, you could use Pixel Eagle, or build something simpler
+with an image comparison library like [`image-compare`].
```
IceSentry: I believe pixel eagle uses https://github.com/gfx-rs/nv-flip-rs which has some really good heuristics to detect if an image output visually changed. The algorithm is from nvidia but IIRC it runs on any hardware. It seems worth mentioning it here at least. I also think nv-flip is much more appropriate for games while image-compare is more intended for things like computer vision but I'm not sure about that.

content/learn/book/development-practices/testing.md
```diff
@@ -5,4 +5,735 @@ insert_anchor_links = "right"
...
+
+In the end, **visual regression testing is for unexpected changes, not for correctness.**
+A human still has to verify that the baseline looks right;
+the automation just tells you when something *changed*.
+You decide whether that change is a bug or an improvement.
+
+#### The siren's call of computer vision
```
IceSentry: I'm not sure this section is needed tbh. I feel like anyone reaching the point where they want to use custom built computer vision solution isn't really the target audience of this book. It doesn't really hurt to keep it but to me it's a bit unnecessary and might make it look less approachable for beginners.

alice-i-cecile: My initial reaction was "ooh I bet I could use computer vision for this!", and I suspect a lot of over-ambitious clever hobbyists will have the same response. It's not very practical though, and I felt that it's good to steer people away from it proactively.

PR #2402, opened 2026-03-31
Quick Start Guide Refresh

Based-A:
Mainly formatting and removing/rewriting some outdated sections from the Quick Start Guide. Haven't touched a lot of the core writing, but I added links to the Bevy Book and removed the Fast Compiles section (since it's better off in the Book rather than in the QSG).

I also fixed two spelling mistakes that `typos` was throwing at me when ran over the whole repo.


alice-i-cecile (changes requested):
I would prefer more granular PRs for content that is published. In this case, I would have preferred 3 PRs:

1. Typo fixes
2. Line splits
3. Real edits

For non-trivial changes to content that is published, we need two approvals before merging.

PR #2416, opened 2026-04-02
Rewrite Introduction chapter of Book to improve flow

alice-i-cecile:
The previous Introduction was a quick first draft, assembled before we had any of the other pieces in place.
Now that the MVP of the book content is ready, we can do much better!

The primary issues centered around flow:

- the Book started without any introduction to what Bevy *is* (this was on page 2)
- the links to other learning material were the first thing you could see, immediately guiding users *out* when they were just starting
- the "why / why not use Bevy" sections were in the critical learning path. This is useful material, but fundamentally an aside and can feel like marketing.
- I felt we should very quickly explain the ECS in a non-technical way, right at the start of the book

Additionally:

- we needed a page for "now what?" at the end of the book
- the "how to submit edits" section should use a callout element
- we should mention where to edit the book when asking for quick PRs
- there was a bunch of assorted wording improvements to make


content/learn/book/further-reading/_index.md
```diff
@@ -0,0 +1,39 @@
...
+
+Hungry for more knowledge?
+Bevy has a wide range of different learning materials, targeted at different use cases and learning preferences.
+
+## Documentation Sources
+
+Everyone learns differently, so we offer a variety of complementary paths:
+
+- [Quickstart](/learn/quickstart): Build a simple game through a hands-on tutorial designed for absolute beginners.
+- [The Book](/learn/book): Learn the core concepts that make Bevy work, then explore advanced topics needed to ship a production game.
```
Based-A, line 18: Does it make sense to link to the book if we're already in the book?

Based-A (approved):

content/learn/book/further-reading/_index.md
alice-i-cecile: Kind of? I was thinking we'll link into this page from all sorts of places (like the start and end of Quick Start).

PR #2417, opened 2026-04-02
Book - Handling Input Chapter & Initial Page

Based-A:
Adding an Input Chapter for the Bevy Book. Looking for feedback and critiques before trying to get it merged. 

What this has so far:
- An overview of using input, basic examples for using input, and a soft overview of the input process.
  - Input Device Overview
  - Input Messages 
  - ButtonInput Resources & it's methods
    - Has `Pressed` vs `Just_Pressed` distinction.
  - Input Conditions
- ~~Per-device type overviews (keyboards, mice, gamepads, and touch input).~~
  - Moved to separate pages, to be introduced in a subsequent PR.

What I still want add:
- ~~Windowing Events (has a section heading, haven't filled it out yet).~~
  - Will be moved to its own page, this initial one is too full.
- ~~Picking functionality (has a section heading, haven't filled it out yet).~~
  - ~~Even though it's not "input" per-say, I still feel like it'd be important to at least mention if not go into here.~~
  - Will be moved to a separate page, to be introduced in a later PR.
- ~~Better Examples.~~
  - ~~I placed the current ones in to fill out the page with something, but I don't think they do a good enough job at showcasing a good application.~~
- ~~Input Focus~~
  - ~~Might be better off in a separate page that actually details the windowing and input process?~~
  - Will be better off in a "Input and Windowing" page in the Game Loop chapter.
- ~~Maybe point out (but not necessarily go into detail) about third party crates that expand input functionality?~~
  - ~~Something like `bevy_enhanced_input`, or `leafwing_input_manager`, etc.~~
  - ~~Might be too contentious, not firm on wanting this.~~
  - Might be added as a separate page in the future.


alice-i-cecile (changes requested):
This actually feels like it should be its own chapter on Input :) The "windowing -> runner -> input"  stuff from before belongs here, but "how do I use input" is effectively unelated to the game loop.

Later, we can add stuff like actions to that chapter.

Picking should be its own page under the Input chapter.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,330 @@
...
+}
+// Register a system that will be queued to run.
+fn register_system(mut commands: Commands) {
+    let registered_system = commands.register_system(mouse_click_system);
+}
+```
+
+Alternatively, most input actions have an associated resource that we can directly access via a system parameter.
+Directly accessing the resource can provide us with more information about the action and give us more control over how we respond to it.
+For example, accessing a [`ButtonInput`] resource for a specific input action (`KeyCode` in the example below) provides us with a number of methods that lets us see if the button [has just been pressed], [is currently being pressed], or [if its just been released].
```
alice-i-cecile: Pressed vs just pressed etc is really important and deserves its own section.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+[`ButtonInput`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html
+[has just been pressed]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.just_pressed
+[is currently being pressed]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.pressed
+[if its just been released]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.just_released
+
+## Input Devices
+
+Bevy can read input from the aforementioned devices: gamepads (controllers), keyboards, mice, and touch inputs.
```
alice-i-cecile: Flow here is very weird.

I feel like maybe we should do:

- Bevy can talk to input devices
- ButtonInput
- pressed / just_pressed

Then new chapters for:

- (later) actions, including keybindings
- keyboard
- mouse
- gamepad (inc haptic feedback)
- touch
- picking

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,330 @@
...
+[extra]
+weight = 6
++++
+
+By definition, games are interactive.
+Players will perform some action or reaction, and the game will react and change in return.
+When it comes to video games specifically, we need to receive input from the player through some kind of device.
+Whether it's controllers, keyboards, mice, or even touch screens, Bevy allows us to interact with input from all of these devices.
+
+You can interact with user input in one of two ways: react to input events, or access an input resource.
```
viridia: Instead of "access an input resource" I would say "by polling". The term "access" is vague, while "input resource" is overly narrow (you can also poll other data sources besides resources).

content/learn/book/handling-input/utilizing-input.md
Based-A: Yeah that makes sense. Trying to fit everything into this one page felt like it was condensing a lot. I'll retool this into an Input Chapter then, and move the device specific sections and picking to new pages in separate PRs.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,410 @@
...
+
+By definition, games are interactive.
+Players will perform some action or reaction, and the game will react and change in return.
+When it comes to video games specifically, we need to receive input from the player through some kind of device.
+Whether it's controllers, keyboards, mice, or even touch screens, Bevy allows us to interact with input from all of these devices.
+
+## Interacting with Input Devices
+
+Bevy can read input from the aforementioned devices: gamepads (controllers), keyboards, mice, and touch inputs.
+All of these devices are enabled through a [feature flag].
```
alice-i-cecile: "Each of these"

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
...
+                *player_crouch_state = CrouchState::Crouched,
+            (ButtonState::Unpressed, CrouchState::Crouched) => 
+                *player_crouch_state = CrouchState::Uncrouched,
+            _ => *player_crouch_state = CrouchState::Uncrouched,
+        } 
+    }
+}
+```
+
+Because of the immediate nature of input messages, they also make activating one-shot systems incredibly straightforward.
```
alice-i-cecile: I like this pattern overall, but we need to compare-and-contrast to an ordinary system with a run_if(on_message) run condition.

IMO that pattern is better for lazy player controller input.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
...
+        }
+    }
+}
+// A system that will run when LeftMouseButton is clicked.
+fn mouse_click_system() {
+    println!("The Left Mouse Button was clicked!");
+}
+// Register a system that will be queued to run.
+fn register_system(mut commands: Commands) {
+    let registered_system = commands.register_system(mouse_click_system);
```
alice-i-cecile: We shouldn't be using commands to register one-shot systems when you can just do it in a plugin. Commands only make sense here for dynamic applications.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
...
+}
+```
+
+Because of the immediate nature of input messages, they also make activating one-shot systems incredibly straightforward.
+We can see this in the example below, where every click of the left mouse button will queue a `mouse_click_system` to be run.
+Toggling UI boxes, initiating interactions with NPC characters, and adding a new player when a controller is connected are all situations where one-shot systems could be triggered by input messages.
+
+```rust
+// This system will read input events from `MouseButtonInput`.
+fn run_system_on_click(
```
alice-i-cecile: This pattern needs to be contrasted with observers.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
...
+[`KeyboardInput`]: https://docs.rs/bevy/latest/bevy/input/keyboard/struct.KeyboardInput.html
+[`MouseButtonInput`]: https://docs.rs/bevy/latest/bevy/input/mouse/struct.MouseButtonInput.html
+[`MessageReader`]: https://docs.rs/bevy/latest/bevy/prelude/struct.MessageReader.html
+[`MessageWriter`]: https://docs.rs/bevy/latest/bevy/prelude/struct.MessageWriter.html
+[`MessageMutator`]: https://docs.rs/bevy/latest/bevy/ecs/message/struct.MessageMutator.html
+
+## ButtonInput Resources
+
+Once most input messages are processed, they're placed into an associated [`ButtonInput`] resource that we can directly access via a system parameter.
+Input data is added to a `ButtonInput` via the same systems that process and clear the input `Message` data, meaning we'll only have access to the input data as long as the original input `Message` is still available.
```
alice-i-cecile: We should discuss "what makes an input buttonlike".

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,461 @@
...
+[`MessageReader`]: https://docs.rs/bevy/latest/bevy/prelude/struct.MessageReader.html
+[`MessageWriter`]: https://docs.rs/bevy/latest/bevy/prelude/struct.MessageWriter.html
+[`MessageMutator`]: https://docs.rs/bevy/latest/bevy/ecs/message/struct.MessageMutator.html
+
+## ButtonInput Resources
+
+Once most input messages are processed, they're placed into an associated [`ButtonInput`] resource that we can directly access via a system parameter.
+Input data is added to a `ButtonInput` via the same systems that process and clear the input `Message` data, meaning we'll only have access to the input data as long as the original input `Message` is still available.
+
+Interacting with input data this way provides us with easier access to the state of the button input and gives us more control over how we respond to it. For example, accessing a [`ButtonInput`] resource provides us with a number of methods that will return a `bool` based on if the button [has just been pressed], [is currently being pressed], or [if its just been released].
```
alice-i-cecile: The distinction between pressed / just pressed etc and when you should use each one deserves more words.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
...
+[`clear`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.clear
+[`clear_just_pressed`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.clear_just_pressed
+[`clear_just_released`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.clear_just_released
+[`reset`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.reset
+[`reset_all`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.reset_all
+
+[`release`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.release
+[`release_all`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.release_all
+
+### Pressed Versus Just Pressed
```
alice-i-cecile: Ah, it's down here. I think that this should be up higher!

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
...
+}
+```
+
+This example doesn't quite cover the full extent of a mechanic like this, however it does provide you with an idea of how `pressed`, `just_pressed`, and `just_released` can be effectively used.
+
+[`pressed`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.pressed
+[`just_pressed`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.just_pressed
+[`just_released`]: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.just_released
+
+## Input System Conditions
```
alice-i-cecile: Yeah, this needs a dedicated section (maybe even page) contrasting one-shot systems, observers and systems with run conditions.

content/learn/book/handling-input/utilizing-input.md
```diff
@@ -0,0 +1,461 @@
++++
+title = "Utilizing Input"
```
alice-i-cecile: A bit stiff. Maybe "Using input"? Or perhaps split this page into a few parts? This is getting quite long.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,461 @@
++++
+title = "Utilizing Input"
+insert_anchor_links = "right"
```
alice-i-cecile, line 3: This page / related pages should talk about time + input: it's really important to get right.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+        // Match the input message based on what keyboard key is being pressed.
+        match input.key_code {
+            KeyCode::W => player_location.0.translation.y += 0.1,
+            KeyCode::A => player_location.0.translation.x -= 0.1,
+            KeyCode::S => player_location.0.translation.y -= 0.1,
+            KeyCode::D => player_location.0.translation.x += 0.1,
+            _ => continue;
+        }
+    }
+}
```
cookie1170: that's not how you should use the `KeyboardInput` messages, they're not emitted every frame. They're emitted when the key is first pressed, then every so often if you keep it held after a small delay - that's not what you want for game movement

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+If you know that your game will not need touch input (or keyboard and mouse inputs if you're building a mobile game), you can disable these input devices by turning their feature flag off in your project `Cargo.toml` file.
+See the [Selective Feature Use section] in the Compiling Less Code page for more details.
+
+[Selective Feature Use section]: /learn/book/releasing-projects/compiling-less-code/#more-selective-feature-use
+
+[Winit]: https://crates.io/crates/winit
+[`bevy_winit`]: https://docs.rs/bevy/latest/bevy/winit/index.html
+[feature flag]: https://docs.rs/bevy/latest/bevy/index.html#feature-list
+
+## Input Messages
```
cookie1170: I think this whole section should be lower down - putting it as the first makes it seem more "important" than the `ButtonInput` resources section, even though these messages aren't often used directly, and can cause issues if you forget to check for `repeat`
It should also clarify that the `ButtonInput` resource is usually used for game input, and this is meant more for things like text input (since that's when you _do_ want to respect the repeat events)

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+{% callout(type="info") %}
+
+### What Makes An Input "Button-like"?
+
+When you start looking at each input device, you might wonder what makes an input "button-like"?
+
+It's actually very straightforward.
+To be considered "button-like" in Bevy, the input has to be "press-able".
+This means that Bevy can register the state of the input as either `pressed` or `released`.
+Both of these values are explicitly stored in a [`ButtonState`] enum, which is apart of every "button-like" input `Message` that is sent.
```
cookie1170:
```suggestion
Both of these values are explicitly stored in a [`ButtonState`] enum, which is a part of every "button-like" input `Message` that is sent.
```
nit: a part means part of, apart means separated from

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+```rust
+fn weapon_attack(
+    button_input: Res<ButtonInput<MouseButton>>,
+    mut player_weapon: Single<(&Weapon, &mut WeaponAttackInterval), With<PlayerWeapon>>
+) {
+    if button_input.just_pressed(MouseButton::LeftMouseButton) {
+        // Perform an initial weapon attack.
+        player_weapon.0.attack();
+        // Create a new the timer within `WeaponAttackInterval` that will run.
+        player_weapon.1.from_seconds(1.5, TimerMode::Repeating);
```
cookie1170:
```suggestion
        player_weapon.1.0 = Timer::from_seconds(1.5, TimerMode::Repeating);
```
nit: `Timer::from_seconds` isn't an instance method, it's a constructor

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+        // Progress the WeaponAttackInterval.
+        player_weapon.1.tick(delta_time);
+    }
+    
+    // The initial LeftMouseButton press.
+    if button_input.just_pressed(MouseButton::LeftMouseButton) {
+        // Perform an initial weapon attack.
+        player_weapon.0.attack();
+        // Create a new the timer within `WeaponAttackInterval` that will run.
+        player_weapon.1.from_seconds(1.5, TimerMode::Repeating);
```
cookie1170:
```suggestion
        player_weapon.1.0 = Timer::from_seconds(1.5, TimerMode::Repeating);
```
nit: constructor not instance method

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+```rust
+fn weapon_attack(
+    button_input: Res<ButtonInput<MouseButton>>,
+    mut player_weapon: Single<(&Weapon, &mut WeaponAttackInterval), With<PlayerWeapon>>,
+    time: Res<Time>,
+) {
+    let delta_time = time.delta();
+    
+    // Check the state of the WeaponAttackInterval timer if it's active.
+    if player_weapon.1.is_finished() != true {
```
cookie1170:
```suggestion
    if !player_weapon.1.is_finished() {
```
nit: use `!x` instead of `x != true`

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,436 @@
...
+    if button_input.just_released(MouseButton::LeftMouseButton) {
+        player_weapon.1.set_mode(TimerMode::Once);
+    }
+}
+```
+
+You might have noticed that this example only `ticks` (progresses) the timer when receiving input.
+Since input data is only received each frame, if we experience lag or dropped frames, the weapon timer won't function consistently.
+We won't cover a fix in this example since the intent is to provide you with an idea of how `pressed`, `just_pressed`, and `just_released` can be effectively used.
+If you'd like to see how we can fix this issue, check out the [Timers and Input] section on the Input Patterns page.
```
cookie1170: That's not true, these systems run every `Update`, so it ticks every update

cookie1170 (changes requested):

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+
+[Observers (and the events that trigger them)]: /learn/book/control-flow/events
+
+### Activating One-shot Systems with Input
+
+Continuing with the observer comparison, directly running one-shot systems with input can skip an entire processing step when compared to triggering an observer.
+If we wanted to use an observer, we would have to:
+
+1. Spawn our `Observer` and the `Event` that will trigger it.
+2. Read an input `Message` and trigger our `Observer` in response.
```
cookie1170: I think we should mention somewhere to be careful about repeat events that you get when using the input messages
or maybe we shouldn't recommend using the messages for gameplay logic at all?

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+We also have one more system condition function that relates to input: [`input_toggle_active`].
+This condition uses XOR bitwise logic, taking in a button to check the state of (using `just_pressed`) and a `bool` value to compare against.
+If either the `bool` you pass in or the button you want to check evaluate to true, then the system will run.
+However, if both the `bool` and the evaluated button state are equivalent, then the system will not run.
+
+In the example below, we only want to run the `pause_menu` system if the `Escape` key is pressed.
+We'll pass in a `false` value alongside `KeyCode::Escape` to ensure that `pause_menu` will only run when `KeyCode::Escape` is pressed.
```
cookie1170: that's not how `input_toggle_active` works.
the way that it works is the condition's state gets toggled whenever you press the button you pass it, and if the condition's state is enabled it runs. the boolean argument is just the initial value

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+{% end %}
+
+[`input_just_pressed`]: https://docs.rs/bevy/latest/bevy/input/common_conditions/fn.input_just_pressed.html
+[`input_pressed`]: https://docs.rs/bevy/latest/bevy/input/common_conditions/fn.input_pressed.html
+[`input_just_released`]: https://docs.rs/bevy/latest/bevy/input/common_conditions/fn.input_just_released.html
+[`input_toggle_active`]: https://docs.rs/bevy/latest/bevy/input/common_conditions/fn.input_toggle_active.html
+
+## Input & Time
+
+Input [`Messages`] are sent every frame, which means that its possible for frame drops and lag to introduce gaps where there is no player input being received by the game.
```
cookie1170: this is false. they are _not_ sent every frame, they're sent whenever the buttons are pressed, released, or a repeat event is triggered

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+
+```rust
+fn move_player(
+    input: Res<ButtonInput<KeyCode>>,
+    mut player_transform: Single<&mut Transform, With<Player>>,
+    time: Res<Time>
+){
+    const PLAYER_SPEED: f32 = 60.;
+    
+    if input.just_pressed(KeyCode::W) {
```
cookie1170:
```suggestion
    if input.pressed(KeyCode::W) {
```
this should happen while the button is held, not when it's just pressed

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+        player_transform.translation.x += PLAYER_SPEED * time.delta_secs();
```
cookie1170:
```suggestion
        player_transform.translation.y += PLAYER_SPEED * time.delta_secs();
```
nit: W usually moves on the y axis (in 2d at least), not on the x axis

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+### Utilizing Delta Time with Input
+
+Fixing the discrepancies between frame rate and consistent game systems is a fundamental aspect of game design.
+Games usually don't run at a consistent frame rate, but players do experience a consistent flow of time.
+If the what the game displays and what the player is perceiving don't align, it can lead to an unexpected (and likely negative) experience.
+To reconcile these, we use ["delta time"].
+
+When incorporating delta time into input data, we have to think about what we're trying to adjust.
+If we're trying to adjust player movement, we can just multiply the player's speed by the `delta_secs` value Bevy tracks internally.
+This ensures that the player will be moving at a consistent rate between frames.
```
cookie1170, line 196: we should also mention that you should _not_ multiply impulses, such as a jump or dash, by delta time, it doesn't matter to them how long the frame took, because they only happen once and don't carry between frames

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,314 @@
...
+    button_input: Res<ButtonInput<MouseButton>>,
+    mut player_weapon: Single<(&Weapon, &mut WeaponAttackInterval), With<PlayerWeapon>>,
+    time: Res<Time>,
+) {
+    let delta_time = time.delta_secs();
+    
+    // Check the state of the WeaponAttackInterval timer if it's active.
+    if player_weapon.1.is_finished() != true {
+        // Progress the WeaponAttackInterval.
+        player_weapon.1.tick(delta_time);
```
cookie1170: as i mentioned before, this timer always ticks, since this system is in `Update`, so we don't need to tick it in a separate system

cookie1170 (changes requested):

content/learn/book/handling-input/input-patterns.md
```diff
@@ -76,25 +76,23 @@ Continuing with the observer comparison, directly running one-shot systems with
...
     mut commands: Commands,
-    input_event_reader: MessageReader<MouseButtonInput>
+    button_input: Res<ButtonInput<MouseButton>>,
 ) {
-    for event in input_event_reader {
-        if event.button == MouseButton::Right && 
-            event.state == ButtonState::Pressed {
-            commands.run_system(toggle_weapon_sights);
-        }
+    if button_input.pressed(MouseButton::RightMouseButton) {
```
cookie1170:
```suggestion
    if button_input.just_pressed(MouseButton::Right) {
```
you only wanna toggle the sights when you press the button, not when you keep holding it

content/learn/book/handling-input/input-patterns.md
```diff
@@ -134,12 +132,10 @@ fn main() {
 ```
 
 We also have one more system condition function that relates to input: [`input_toggle_active`].
-This condition uses XOR bitwise logic, taking in a button to check the state of (using `just_pressed`) and a `bool` value to compare against.
-If either the `bool` you pass in or the button you want to check evaluate to true, then the system will run.
-However, if both the `bool` and the evaluated button state are equivalent, then the system will not run.
+This condition taking in a input button to check the state of (using `just_pressed`) and a `bool` value to set the initial state.
```
cookie1170: the wording here still feels a bit unclear to me, you should mention that when using the condition, the system gets toggled every time you press the button, and the bool value is the initial state

content/learn/book/handling-input/input-patterns.md
```diff
@@ -175,16 +171,18 @@ On the other hand, our input systems do allow us to explicitly check the state o
 
 ## Input & Time
 
-Input [`Messages`] are sent every frame, which means that its possible for frame drops and lag to introduce gaps where there is no player input being received by the game.
+Input data can be read every frame, which means that its possible for frame drops and lag to introduce gaps where there is no player input being received by the game.
 This also means that if we're using timers and cooldowns inside of systems that will only run based on input, we have to be aware of the fact that those timers and cooldowns won't function properly unless we manually correct them.
```
cookie1170: the input data gets accumulated in a resource, which is then read from every frame, so frame drops shouldn't cause any issues with that.
and using timers inside of systems that read the `ButtonInput` resource is fine, you just need to make sure that you always tick the timers instead of it depending on input
for example if you have this system that runs in update
```rs
fn my_system(mut timer: ResMut<MyTimer>, input: Res<ButtonInput<MouseButton>>, time: Res<Time>) {
    timer.tick(time.delta();
    if input.pressed(MouseButton::Left) && timer.just_finished() {
        // shoot
    }
}
```
the timer will always tick just fine, since it's always called regardless of if space is pressed
however, if you have this system
```rs
fn my_system(mut timer: ResMut<MyTimer>, input: Res<ButtonInput<MouseButton>>, time: Res<Time>) {
    // the second condition of an and isn't evaluated if the first one is false! 
    if input.pressed(MouseButton::Left) && timer.tick(time.delta()).just_finished() {
        // shoot
    }
}
```
then the timer will only tick _while_ space is pressed

content/learn/book/handling-input/using-input.md
```diff
@@ -81,62 +96,31 @@ fn keyboard_message_mutator(mut keyboard_inputs: MessageMutator<KeyboardInput>)
...
-    for input in movement_input.read() {
-        // Match the input message based on what keyboard key is being pressed.
-        match input.key_code {
-            KeyCode::W => player_location.0.translation.y += 0.1,
-            KeyCode::A => player_location.0.translation.x -= 0.1,
-            KeyCode::S => player_location.0.translation.y -= 0.1,
-            KeyCode::D => player_location.0.translation.x += 0.1,
-            _ => continue;
+    for input in button_inputs.iter() {
+        if input.key_code == KeyCode::Esc {
```
cookie1170:
```suggestion
        if input.key_code == KeyCode::Esc && !input.repeat {
```
You need to make sure that repeat events don't toggle the pause menu again. I think we should also explicitly warn about that somewhere (in the info box?)

cookie1170 (changes requested):

content/learn/book/handling-input/input-patterns.md
Based-A: Yeah I realized that the whole Timer/Tick example was complete bunk after your first reviews. It is very much appreciated, thank you for correcting my mistakes!

content/learn/book/handling-input/input-patterns.md
```diff
@@ -132,10 +132,16 @@ fn main() {
 ```
 
 We also have one more system condition function that relates to input: [`input_toggle_active`].
-This condition taking in a input button to check the state of (using `just_pressed`) and a `bool` value to set the initial state.
+This condition takes in a button to check the state of (using the `just_pressed` method) and a `bool` value.
+When you press the button you passed into `input_toggle_active` for the first time, the system will switch from the initial `bool` value you pass in.
+If you initially passed in `true` to have the system run from it's initialization, `input_toggle_active` will switch to `false` and the system will stop running.
```
cookie1170:
```suggestion
If you initially passed in `true` to have the system run from its initialization, `input_toggle_active` will switch to `false` and the system will stop running.
```
nit: posessive "it" is "its" not "it's"

content/learn/book/handling-input/using-input.md
```diff
@@ -48,11 +48,14 @@ These types of mechanics should instead be accessing the [`ButtonInput` resource
 
 Instead, input messages are best suited for testing and logging input events, tracking text input, and activating systems or logic that don't rely on consistently repeated input.
 
+Additionally, if you do choose to use input messages for some functionality, be wary of the `repeat` field on an input message which indicates that the message will be re-sent.
```
cookie1170: the `repeat` field indicates that it's a repeated event (like when holding down a key), not that it will be resent

cookie1170 (changes requested):

content/learn/book/handling-input/_index.md
SpecificProtagonist, line 1: Hm. Not specific to this chapter: For the multi-page chapters, the introduction page doesn't contain much useful information. I see two things it's attempting to do:
- Summarize what the chapter is about
- Summarize what the problem is/how Bevy is approaching it

The first is already mostly served by the chapter & subchapter names. The second one is on a very superficial level where one has to read the chapter to learn anything useful. The chapter introductions in the [Rust book](https://doc.rust-lang.org/book/) seem to have the same goals but work better as they tend to go into more detail.

In this case the content of the last sentence of this page is just a duplication of the first sentence of the next page. So this page says very little. I would prefer we move the content of the "Using Input" page directly here.

content/learn/book/handling-input/_index.md
```diff
@@ -0,0 +1,13 @@
...
+insert_anchor_links = "right"
+[extra]
+weight = 14
+status = 'hidden'
++++
+
+By definition, games are interactive.
+Players will perform some action or reaction, and the game will react and change in return.
+When it comes to video games specifically, we need to receive input from the player through some kind of device.
+Whether it's controllers, keyboards, mice, or even touch screens, Bevy allows us to interact with input from all of these devices.
```
SpecificProtagonist:
```suggestion
When it comes to video games specifically, we need to receive input from the player through some kind of device:
Controllers, keyboards, mice, or even touch screens.
```
It seems to me like the last sentence part was superfluous: If Bevy didn't allow us to interact with some of these devices, we would either say that or not discuss them in the first place.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
++++
+title = "Using Input"
+insert_anchor_links = "right"
+[extra]
+weight = 1
++++
+
+Bevy can read input data from gamepads (which includes controllers), keyboards, mice, and touch inputs.
+The process for reading and interacting with input data will generally be the same across all devices.
+When the button on a gamepad or a key on a keyboard gets pressed, Bevy uses [Winit] (via [`bevy_winit`], Bevy's conversion tool for Winit) to initially turn the input into a [`Message`].
```
SpecificProtagonist:
```suggestion
When the button on a gamepad or a key on a keyboard gets pressed, Bevy uses [Winit] (via [`bevy_winit`], Bevy's conversion layer for Winit) to initially turn the input into a [`Message`].
```
Tool kinda makes me think of something used during development? Layer might be a bit clearer here.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
++++
+title = "Using Input"
+insert_anchor_links = "right"
+[extra]
+weight = 1
++++
+
+Bevy can read input data from gamepads (which includes controllers), keyboards, mice, and touch inputs.
```
SpecificProtagonist: Gamepads are a subset of controllers, not the other was around. Do we support non-gamepad controllers?

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+While input data from every device will follow the same process that is covered on this page, you should read each devices' page to see how their data is uniquely handled:
+
+- [Keyboard Input](/learn/book/handling-input/keyboard-input)
+- [Mouse Input](/learn/book/handling-input/mouse-input)
+- [Gamepad Input](/learn/book/handling-input/gamepad-input)
+- [Touch Input](/learn/book/handling-input/touch-input)
+
+When you use larger groupings of features (like profiles and collections), all of these devices are enabled in your game by default.
+However we can adjust which ones are enabled by manually enabling their [feature flag].
+If you know that your game will not need touch input (or keyboard and mouse inputs if you're building a mobile game), you can disable these input devices by turning their feature flag off in your project `Cargo.toml` file.
```
SpecificProtagonist: Maybe leave that out – afaik you can only disable keyboard/mouse/touch features if you don't use `bevy_window`.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+title = "Using Input"
+insert_anchor_links = "right"
+[extra]
+weight = 1
++++
+
+Bevy can read input data from gamepads (which includes controllers), keyboards, mice, and touch inputs.
+The process for reading and interacting with input data will generally be the same across all devices.
+When the button on a gamepad or a key on a keyboard gets pressed, Bevy uses [Winit] (via [`bevy_winit`], Bevy's conversion tool for Winit) to initially turn the input into a [`Message`].
+These `Messages` are then processed and placed into a resource (or a component) that we can then read and use for setting up movement, tracking aim, activating abilities, or any other input-based actions you'd like to set up.
```
SpecificProtagonist:
```suggestion
These `Message`s are then processed and placed into a resource (or a component) that we can then read and use for setting up movement, tracking aim, activating abilities, or any other input-based actions you'd like to set up.
```

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+
+Before we cover the general process though, you should know that each device type also has their own unique circumstances to be aware of.
```
SpecificProtagonist:
```suggestion
```
Doesn't add anything not said in the next sentence.

content/learn/book/handling-input/using-input.md
SpecificProtagonist: Additionally, it's only for keyboard input. Holding down a mouse button does not repeatedly fire events.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+Each unique `Message` type will be automatically set up for the enabled devices in your game.
+Bevy will also insert systems in the [`PreUpdate`] schedule to process, store, and eventually clear each `Message` type.
+
+{% callout(type="info") %}
+
+### Using Input Messages
+
+While receiving input messages is the way that Bevy accesses input data, this doesn't mean that you should use input messages in every scenario.
+
+Input messages are only sent when an input is initially activated and then re-sent periodically if the input is still being activated (i.e. if a button is being pressed down or a joystick is continuously being pushed in a direction).
```
SpecificProtagonist: Holding down a mouse button does not repeatedly send messages.

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+## Input Messages
+
+Bevy takes input data from a device and converts it into a [`Message`].
+These messages can be [read like any other message], and will have a unique type for each device: [`KeyboardInput`] for keyboards, [`MouseButtonInput`] for mouse button presses, and so on for each input type.
+Each unique `Message` type will be automatically set up for the enabled devices in your game.
+Bevy will also insert systems in the [`PreUpdate`] schedule to process, store, and eventually clear each `Message` type.
+
+{% callout(type="info") %}
+
+### Using Input Messages
```
SpecificProtagonist: I think this callout should just be part of the introduction of this page, since it's about when to use input `Messages` and then to use `ButtonInput` (this is kinda already indicated by the heading not really fitting – the whole "Input Messages" section *except* this callout is about "Using Input Messages").

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+
+{% end %}
+
+Accessing input data through messages gives us access to all of the regular functionality that messages provide, including [`MessageReader`], [`MessageWriter`], and [`MessageMutator`].
+
+```rust
+// This system reads and prints out all `KeyboardInput` messages.
+fn keyboard_message_reader(keyboard_inputs: MessageReader<KeyboardInput>) {
+    for keyboard_input in keyboard_inputs.read() {
+        info!("{:?}", keyboard_input);
```
SpecificProtagonist:
```suggestion
        info!("{keyboard_input:?}");
```

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,417 @@
...
+    }
+}
+// This system writes a new `KeyboardInput` message.
+fn keyboard_message_writer(
+    mut keyboard_inputs: MessageWriter<KeyboardInput>,
+    window_query: Single<(Entity, &Window)>,
```
SpecificProtagonist:
```suggestion
    window_query: Single<Entity, With<Window>>,
```

SpecificProtagonist (changes requested):
(just a partial review so far)

content/learn/book/handling-input/_index.md
Based-A: I was thinking about this as well, but I would prefer to make each `_index` page a chapter summary (like a "go to _x page_ for _x information_ and _y page_ for _y information_") rather than making it an introduction page. 

When I was first going through the book, I didn't even realize that the `_index` pages existed and had writing on them since they have the option to drop down a list of the individual pages that have the actual information I want. Maybe it's just me being a bit dense, but I can't imagine I'm the only one who would encounter this. I think avoiding placing information in areas that people might not be able to intuitively find should be a higher consideration than potentially having duplicate information.

For this chapter's `_index` page specifically, I am not opposed to changing what's on there (especially if we do want to make the `_index` a summary that links out to each subpage), ~~but any changes I think should be done as part of a separate PR (or multiple if we're changing them to summaries)~~.

content/learn/book/handling-input/using-input.md
Based-A: Not officially I don't think, but I'm sure you'd be able to make your own support if you dug into winit. But that's definitely outside the scope of the book.

content/learn/book/handling-input/using-input.md
Based-A: I think it is worth mentioning since you can explicitly enabled the `keyboard` and `mouse` feature flags (and thus disable `touch` and `gamepad` inputs). Before we had the top level profiles, I went through and individually enabled the features I wanted in my projects, and I do know that I specifically did not enable `gamepad` or `touch` input. 

Granted, I think I could probably re-word this to shift from "disabling" (which isn't really possible with cargo) to "specifically enabling", which might be better for then pointing the reader at the relevant book page.

content/learn/book/handling-input/_index.md
Based-A: Of course I say this before I start making changes to `_index`, ignore that last section

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+// A resource that tells us how many units to spawn on the map initially.
+#[derive(Resource)]
+struct SetupParameters {
+    pub initial_units: i32,
+    ...
+}
+
+// A system reading input data which will call a `PlaceUnitOnMap` event.
+fn trigger_observer(mouse_input: Res<ButtonInput<MouseButton>>, mut commands: Commands) {
+    if mouse_input.just_pressed(MouseButton::LeftMouseButton) {
```
cookie1170:
```suggestion
    if mouse_input.just_pressed(MouseButton::Left) {
```
nit: it's `MouseButton::Left`, not `MouseButton::LeftMouseButton`

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+```rust
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins)
+        // `jump` if the Space key has just been pressed.
+        .add_systems(Update, jump.run_if(input_just_pressed(KeyCode::Space)))
+        // Repeat a weapon attack if LeftMouseButton is currently being pressed.
+        .add_systems(Update, weapon_repeated_attack.run_if(input_pressed(MouseButton::LeftMouseButton)))
+        // Unscope the weapon if RightMouseButton has been released.
+        .add_systems(Update, unscope_weapon.run_if(input_just_released(MouseButton::RightMouseButton)))
```
cookie1170:
```suggestion
        .add_systems(Update, weapon_repeated_attack.run_if(input_pressed(MouseButton::Left)))
        // Unscope the weapon if RightMouseButton has been released.
        .add_systems(Update, unscope_weapon.run_if(input_just_released(MouseButton::Right)))
```
nit: mouse button names don't include `MouseButton`

cookie1170 (approved):

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
++++
+title = "Input Data Patterns"
```
alice-i-cecile: This title feels rather stiff to me.

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+insert_anchor_links = "right"
+[extra]
+weight = 2
++++
+
+Interacting with input data can be done in many forms.
```
alice-i-cecile: Kind of a nothing sentence. We can do better :)

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+This page is meant to illustrate some of the ways you can combine input data with other tools that Bevy provides, along with highlighting some patterns to avoid if possible.
```
alice-i-cecile: Good idea, wording could be improved.

content/learn/book/handling-input/_index.md
```diff
@@ -0,0 +1,30 @@
++++
+title = "Handling Input"
+template = "docs.html"
+insert_anchor_links = "right"
+[extra]
+weight = 14
+status = 'hidden'
++++
+
+By definition, games are interactive.
```
alice-i-cecile:
```suggestion
If a player is to *play* your game, it must be interactive.
```

content/learn/book/handling-input/_index.md
```diff
@@ -0,0 +1,30 @@
...
+Players will perform some action or reaction, and the game will react and change in return.
+When it comes to video games specifically, we need to receive input from the player through some kind of device.
```
alice-i-cecile:
```suggestion
To make this happen, we need to receive input from the player through some kind of device.
```

content/learn/book/handling-input/_index.md
```diff
@@ -0,0 +1,30 @@
...
+In Bevy, we can use input from gamepads, keyboards, mice, or even touch screens in our games and applications.
```
alice-i-cecile:
```suggestion
In Bevy, we can collection input from gamepads, keyboards, mice, or even touch screens.
```

content/learn/book/handling-input/_index.md
```diff
@@ -0,0 +1,30 @@
...
+The process of accessing and using this data is detailed on the [Using Input page](/learn/book/handling-input/using-input), which is where we'd recommend you start.
+
+While input data from every device will follow the same process, you should also read the page of each device relevant to your project to see how their data is uniquely handled:
```
alice-i-cecile: Cut from this PR: add later.

content/learn/book/handling-input/_index.md
```diff
@@ -0,0 +1,30 @@
...
+
+- [Keyboard Input](/learn/book/handling-input/keyboard-input)
+- [Mouse Input](/learn/book/handling-input/mouse-input)
+- [Gamepad Input](/learn/book/handling-input/gamepad-input)
+- [Touch Input](/learn/book/handling-input/touch-input)
+
+When you use larger groupings of features (like profiles and collections), all of these devices are enabled in your game by default.
```
alice-i-cecile: Cut from this PR. I don't think this is worth discussing.

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Reacting To Input
+
+The obvious application of input data is to run systems and update the game based on player input.
```
alice-i-cecile: This is a very weird sentence. I'm not sure what it's trying to say.

Additionally, we should try to avoid words like "obvious" when teaching. They're often discouraging.

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+title = "Input Data Patterns"
+insert_anchor_links = "right"
+[extra]
+weight = 2
++++
+
+Interacting with input data can be done in many forms.
+This page is meant to illustrate some of the ways you can combine input data with other tools that Bevy provides, along with highlighting some patterns to avoid if possible.
+
+## Reacting To Input
```
alice-i-cecile: The framing of this section is... weird. I think this should be a list of the options, followed by a section for each option with pros and cons.

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+## Input & Time
+
+Input data can be accessed every frame, which means that any changes we make based on input data will be applied every frame.
+This isn't an issue for one-shot logic, but if we have some functionality that repeats and needs to be consistent across the player's real-time experience, we begin to run into issues.
+
+Most of what we'll cover here can also be found in more depth over in the [Time and Timers] page within The Game Loop chapter, so if there are concepts that aren't making sense or you just want to read more, be sure to check that page out as well.
+
+[Time and Timers]: /learn/book/the-game-loop/time-and-timers
+
+### Utilizing Delta Time with Input
```
alice-i-cecile: Try to avoid needlessly elaborate words like "utilizing".

content/learn/book/handling-input/using-input.md
```diff
@@ -0,0 +1,396 @@
...
++++
+
+The process for reading and interacting with input data will generally be the same across all devices that Bevy can use.
+When the button on a gamepad or a key on a keyboard gets pressed, Bevy uses [Winit] (via [`bevy_winit`], Bevy's conversion layer for Winit) to initially turn the input into a [`Message`].
+These `Message`s are then processed and placed into a [`ButtonInput`] resource (or a `ButtonInput` component for gamepads) that we can then read and use for setting up movement, tracking aim, activating abilities, or any other input-based actions you'd like to set up.
+
+[Winit]: https://crates.io/crates/winit
+[`bevy_winit`]: https://docs.rs/bevy/latest/bevy/winit/index.html
+
+## Using ButtonInput Versus Input Messages
```
alice-i-cecile: This feels like a "patterns" bit.

alice-i-cecile (changes requested):
Content is broadly good, but this needs another pass on high-level organization. As a reader, the flow is not great, and the division of the topics into these pages feels arbitrary.

Let me know if you'd like me to spend some time chewing on how we might organize this :)

Based-A:
Yes please, I am somewhat wandering in the dark when it comes to organizing the writing. Theoretically everything that is relevant to input is *there*, but I can agree that it needs to be refreshed and probably re-done. I'll do a pass myself and see how it looks, but any suggestions would be welcome and appreciated!

alice-i-cecile:
I was thinking:

1. Intro: a sentence or two of preamble, a tiny example using `ButtonInput<KeyCode>`
2. Buttonlike data flow: system libraries -> messages ->  ButtonInput, pressed / justpressed distinction
3. Patterns for responding to input:  contrasting systems, observers, run conditions and so on
4. (not this PR) Why actions are useful
5. (not this PR) Detailed pages on each input method

content/learn/book/handling-input/input-patterns.md
```diff
@@ -0,0 +1,220 @@
...
+
+Input data can also be used as a conditional check for deciding when a system should run.
+If the input data is only meant to signal that a system should run (instead of being needed inside the system itself), we can use several built-in functions to check whether a system should run.
+These functions will return a boolean based on the state of a specific input button, and require us to pass in a type that can be accessed from a `ButtonInput` type.
+
+```rust
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins)
+        .add_systems(Update, jump.run_if(input_just_pressed(KeyCode::Space)))
```
IQuick143, line 100: Just speaking from gamedev experience:
This kind of input handling is convenient at first, but it effectively becomes tech-debt if your game will ever have any sort of input rebinding or alternative control schemes. So I'm not a huge fan of teaching this pattern. (Had to recently refactor a lot of code that was like this.)

janhohenheim (changes requested):
Great idea, but the chapter as is teaches several antipatterns:
- as @IQuick143 mentions, it teaches you to hardcode keys instead of allowing rebinds 
- Using KeyCode is great in theory, but impossible to keybind since you cannot print the corresponding Key due to a winit limitation. As such, you're setting yourself up for trouble as soon as you need to support keybinds.
- Using delta time times sone input shown here should use a fixed timestep schedule, not update. As soon as you want to use physics through eg Avian, you will run straight into a wall here. Bevy as-is is bad at teaching users about which schedule to put gameplay logic in; let's not reinforce the wrong pattern
- It seems strange to have a chapter about this stuff on a low level since we need a proper input manager on top in practice. Some people write their own, but the clear favorite for most in production is BEI. If this chapter was merged, I would tell beginners to not read it and instead use bevy_enhanced_input. Given that we want to upstream it anyways, this seems like pretending like using raw bevy_input is how things are done in real projects when in reality it is not. A bit like going through the wgpu API instead of teaching about just `StandardMaterial`, you know?

I hope I'm not too harsh here, I appreciate the thought and agree that this stuff is important to have in the book. But right now I think that merging this would do more harm than good by setting beginners up for longterm failure

Based-A:
Definitely not too harsh, I appreciate the feedback and the honesty! If this isn't a good approach for beginners wanting to set up input, then I can agree that this shouldn't be merged. Maybe a different approach could be taken at a less-practical/more-conceptual level, but if there's an entire tool/set of tooling that Bevy doesn't have yet then I guess it would be better to hold off until its implemented? Either way I think I'll bin this PR and restart when an input manager gets created/upstreamed (or something else happens and an input chapter needs to get written) :)

janhohenheim:
Thanks for understanding, that's a relief <3
IMO we should just hurry up and upstream BEI ASAP so we can even start having this conversation, this is not the only place that is suffering under the lack of having any input management (see e.g. our camera controller code or navigation handling in bevy_feathers)

viridia:
I think it really depends on whether we prefer a top-down or bottom-up approach to instruction.

I learned computing bottom-up: first hex math, then assembly language, and only after that did I learn high level languages (Fortran and COBOL in my case). This meant that each level, I understood what was going on at the levels beneath the ones I was working in, leading to a deeper understanding.

However, this is not the most common approach to teaching. Most students learn the simplified, high level language first. This means that they are able to quickly become productive for simple tasks, giving them near-instant gratification. This is especially helpful for students who don't see the point of all the low-level detail, or how it connects to the ultimate end goal they are trying to reach. However, it also carries a risk that their understanding will be shallow - a lot of people who learned Java or Python in school don't really know what a pointer is.

Even for game developers who choose to use an input mapper, knowing the details of how input messaging works at a low level may be useful. Nor do I think that it's wrong to choose not to use an input mapper for a simple game or demo - it's a developer choice, just as the option to localize (or not) is a developer choice.

Based-A:
I think we can have both a high-level "get going after reading this" page and a lower level "this is how this feature actually works" page. But I also don't think that they should be next to each other in the book or really require that you read one and then the next in order to get what you're looking for. Especially after reading a bit more into use cases of BEI vs bevy_input, making a generic "input chapter" that tries to place both the "why" and the "how" on equal footing isn't an enjoyable experience. Unless input is a particular interest for someone, I don't think they'll necessarily care or want to read about how input messages are processed and used; they'll just want to see how they can hook input into their game/app as quickly as possible.

It's actually a thought I've been having for a bit now across the whole Book, where a single monolithic Bevy Book feels like too much to both get through as a reader and too heavy to keep relevant as a writer/maintainer. I almost want to suggest splitting the Book into an "Applied Bevy Book" for practical advice and actual usable code examples and then a "Theoretical Bevy Book" for the concepts that actually dive into how things in the engine work. We sort of already have a version of it with pages like the "Designing Components" vs "Entities Have Components" already in the book. But considering the current book isn't even public yet, I won't be suggesting that we do anything like that yet.

alice-i-cecile:
I agree with @viridia here: I think it's important for us to clearly teach this. But I share @janhohenheim's concerns: without a follow-up chapter and caveats about "how to do it right", I'm very nervous about publishing this and leaving it at that.

Strong, strong agreement on upstreaming BEI ASAP though.

Issue #2325, opened 2026-01-10
`zola serve` with Latest Zola Version (0.22.0) breaks


kfc35:
## What’s Wrong

Zola just came out with a new version, 0.22.0, and it seems to be incompatible with `main` repo.

`zola serve` with Zola Version (0.22.0) results in:
```
zola serve
Building site...
ERROR Failed to serve the site
ERROR TOML parse error at line 28, column 1
   |
28 | extra_syntaxes_and_themes = ["syntaxes"]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^
unknown field `extra_syntaxes_and_themes`, expected one of `highlighting`, `render_emoji`, `external_links_class`, `external_links_target_blank`, `external_links_no_follow`, `external_links_no_referrer`, `external_links_external`, `smart_punctuation`, `definition_list`, `bottom_footnotes`, `lazy_async_image`, `insert_anchor_links`, `github_alerts`
```

It seems the section under [markdown] in config.toml needs to be updated based on the changelog: https://github.com/getzola/zola/blob/master/CHANGELOG.md#0220-2026-01-09

## Possible solutions:
- Update the README to reflect that Zola must be <= 0.21.0.
- Write the necessary changes to config.toml (and possibly elsewhere) to be compatible with 0.22.0 and update the README to reflect the updated Zola version

## Extra Context

PR with the breaking change: https://github.com/getzola/zola/pull/3044

kfc35:
I started the upstream RON syntax support process: https://github.com/shikijs/textmate-grammars-themes/pull/173

ChristopherBiscardi:
I checked out upgrading today. The only thing we need to resolve is regaining our custom syntax highlighting in #2425 

Anyone who wants to explore writing the vscode theme file should do so.

PR #2425, opened 2026-04-07
Update to zola 0.22.1

ChristopherBiscardi:
> [!IMPORTANT]
> 
> This is currently *not mergable* as the new zola syntax highlighting means our custom CSS overrides are not applied.

<details><summary>side-by-side of ron/rust/wgsl examples on the site</summary>

Screenshot 2026-04-06 at 5 31 51 PM
Screenshot 2026-04-06 at 5 31 58 PM
Screenshot 2026-04-06 at 5 34 50 PM

</details>

Notably, zola has produced their own code highlighting solution that in the future causes us to need to write our custom .css syntax theme as a vscode theme instead. However, as far as I can tell, this is not possible until zola releases a new version. The current state is that the CSS overrides we use aren't applied as they were previously when using zola's new giallo.

- zola built a replacement for syntect called giallo: https://github.com/getzola/zola/issues/2758
- There are CSS fixes to code style generation that seem relevant but will fully break any attempt at the css class based highlighting that we have now: https://github.com/getzola/giallo/pull/56
  > The current CSS generation in giallo is broken: getzola/giallo#56 will be included in the next Zola release.
  >
  > [keats](https://github.com/getzola/zola/issues/3075#issuecomment-4189396022)
- We will need to write our theme as a vscode theme to regain styling. I haven't explored actually writing this theme file yet so don't know if there will be regressions or opportunities for enhancements.


ChristopherBiscardi:
I'm also going to mention @Keats , since they commented on the relevant https://github.com/bevyengine/bevy-website/issues/2325 , just in case they're interested in the topic.

Our next steps are to explore writing the vscode theme file to regain our custom syntax highlighting.

---

however, the syntax highlighting is the only blocker we have for upgrading zola. There are no other issues as far as I can tell.

Keats:
> There are no other issues as far as I can tell.

Shouldn't be. 0.23 will have the biggest changes for the template engine but it's not out yet. Since it will contain the fix for giallo CSS though, it might be better to wait for it

ChristopherBiscardi:
People are talking about this again, resulting in #2547 which doesn't solve the theming concern (or pass CI yet). zola 0.23 is also not yet released. However, since other people seem to want to try to push this forward I'll close this draft to let them take over.

PR #2426, opened 2026-04-07
add plausible analytics

### LINKED ISSUE #2334: Privacy friendly, cookie less simple web analytics
I think web analytics could be useful for the website, and give us the following benefits:
- Traffic to the book, and which pages get the most visits
- Devices / browsers loading the examples
- Adoption and growth over time

The constraint we have are:
- No cookie, no GDPR banner
- Not tracking users
- Privacy friendly
- Simple to use

An example of a provider that could work for us is https://plausible.io/

This issue is to collect community feedback, opinions and other options

mockersf:
Fixes #2334

Basic plausible integration

As we get more book content, I would like to get analytics on it

This is using the 30 days trial to try it

### DIFF
--- a/templates/layouts/base.html
+++ b/templates/layouts/base.html
@@ -241,5 +241,10 @@
       }
     }
     </script>
+    <script async src="https://plausible.io/js/pa-Y3chmKZU0FVsmsgC432RR.js"></script>
+    <script>
+      window.plausible=window.plausible||function(){(plausible.q=plausible.q||[]).push(arguments)},plausible.init=plausible.init||function(i){plausible.o=i||{}};
+      plausible.init()
+    </script>
   </body>
 </html>
--- a/templates/layouts/base.html
+++ b/templates/layouts/base.html
@@ -108,6 +108,11 @@
 
     <title>{% if page.extra.public_draft or section.extra.public_draft %}[DRAFT] {% endif -%}{{ page_title }}</title>
     {% block head_extensions %}{% endblock head_extensions %}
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I'm on board to try this. Analytics are useful to have, and if they're privacy-respecting I don't have any strong argument against doing this.

cart (approved):
Something I'm just now noticing about plausible is that the price scales with page views. For us it would not be $9 / month (which is for 10k views).

_Just_ from google search clicks, bevy gets 20k clicks a month. If we assume each of those people clicks 3 links, we're at 30k views. Then lets choose an arbitrary multiplier from other sources of traffic (regular users, reddit, hacker news, bluesky, etc) ... lets say 3 (I suspect it might be more than that). That leaves us at 90k page views per month, which is $19 / month (or $228 / year). This is pretty reasonable I think, but its the type of thing the board should discuss.

In the event that we get a ton of pageviews in a month (ex: a million), we're looking a $69 / month. Also pretty reasonable, especially given that we are "winning" when we hit those numbers.

However in a world where we are _already_ at 1mil / month (which is theoretically possible if you factor in bots), then we're looking at $828 / year, which is a noteworthy, perhaps exorbitant expenditure for our current scale and funding.

Sounds like we need some sort of analytics trial to determine what we'd be paying :)

mockersf:
> Sounds like we need some sort of analytics trial to determine what we'd be paying :)

Yup, I like the 30 days trial to know what traffic we actually get, and what the dashboards look like 👍 

Worst case in 30 days we remove the script

PR #2432, opened 2026-04-12
Mention map_severity helper in Handling Errors page

alice-i-cecile:
This helper was added by @Freyja-moth in https://github.com/bevyengine/bevy/pull/23749.

It's extremely useful, so we should call it out in the book page directly.

### DIFF
--- a/content/learn/book/control-flow/handling-errors.md
+++ b/content/learn/book/control-flow/handling-errors.md
@@ -342,11 +342,24 @@ of the error and ignores/logs/panics accordingly:
 Encountered an error in system `test_programs::update`: No entities fit the query bevy_ecs::system::query::Query<'_, '_, &bevy_camera::camera::Camera>
 ```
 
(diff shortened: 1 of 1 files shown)


Based-A:
Currently the `map_severity` docs link doesn't work, so I think until 0.19 is released it should probably stay out of the book. But once it's out, it'll be a good addition to the book.

ChristopherBiscardi (approved):

ChristopherBiscardi:
> I think until 0.19 is released it should probably stay out of the book

As far as I'm aware, the book targets main/0.19-dev currently, so this makes sense to add

alice-i-cecile:
> Currently the `map_severity` docs link doesn't work, so I think until 0.19 is released it should probably stay out of the book. But once it's out, it'll be a good addition to the book.

Almost none of the links in the chapter work; and `Severity` itself hasn't shipped yet :) The book is written with 0.19 in mind, but it rarely comes up elsewhere.

PR #2433, opened 2026-04-16
Start assets chapter for Book

alice-i-cecile:
Just the introductory page for now, but I still think this is enough to be very useful!

In the future I would like to add:

1. A page on pre-loading assets.
2. A page on asset processing.
3. A page on automated quality control for assets.
4. A page on organizing your assets folder.
5. A page on asset version control.
6. A page on working with custom asset formats.

All of these things are very useful, practical things which only kind of have to do with Bevy at times, and so are a very natural fit for the book.

For now though, this PR should be a useful MVP of this chapter for newcomers to Bevy who don't read the bevy_assets crate docs that I wrote :p

### DIFF
--- a/content/learn/book/assets/bevy's-asset-framework.md
+++ b/content/learn/book/assets/bevy's-asset-framework.md
@@ -108,12 +108,12 @@ Should you change the handle that your sprite holds, or the asset that the handl
 
 Both approaches will change what your sprite looks like, and are valid to do, but the effects differ in an important way:
 
-- mutating the handle changes the [`Image`] asset your sprite entity is pointing to
-  - this only affects the entity in question
-  - this is analogous to replacing a `&` reference
-- mutating the asset changes the underlying data that the handle is pointing to
-  - this affects *all* entities that point to the same asset
-  - this is analogous to mutating the data that your `&` reference is pointing to
+- mutating the handle changes the [`Image`] asset your sprite entity is pointing to.
+  - this only affects the entity you specify.
+  - you can think of it as replacing one `&` reference with a different `&` reference.
+- mutating the asset changes the underlying data that the handle is pointing to.
+  - this affects *all* entities that point to the same asset.
+  - this is like changing the actual data that a `&` reference is pointing to.
 
 To mutate a handle (90% of cases):
 
--- a/content/learn/book/assets/bevy's-asset-framework.md
+++ b/content/learn/book/assets/bevy's-asset-framework.md
@@ -12,7 +12,7 @@ status = 'hidden'
 2. They are often very large: multiple megabytes, when typical data-storing components are measured in bytes.
 
 The first point means that we need tools to dynamically load (and unload) them at runtime.
-This is handled by the [`AssetServer`], 
+This is handled by the [`AssetServer`], which handles the surprisingly complex process of turning a path to an asset we want to use into bytes in RAM that we can stick in a Rust struct.
 
 The second point means that we really don't want to be storing multiple copies of the same asset: RAM and VRAM usage is often a critical limitation for game performance.
 In Bevy, the canonical version of every asset of a type `A` is stored in a matching resource: [`Assets<A>`].
@@ -107,17 +107,17 @@ This allows us to change the asset file during testing and see those changes ref
 
 ## Mutating handles vs mutating assets
 
-Understanding the [`Handle<A>`] / [`Assets<A>`] distinction becomes quite important when when you want to mutate assets.
+Understanding the [`Handle<A>`] / [`Assets<A>`] distinction becomes quite important when you want to mutate assets.
 Should you change the handle that your sprite holds, or the asset that the handle points to?
 
-Both appraoches will change what your sprite looks like, and are valid to do, but the effects differ in an important way:
+Both approaches will change what your sprite looks like, and are valid to do, but the effects differ in an important way:
 
 - mutating the handle changes the [`Image`] asset your sprite entity is pointing to
   - this only affects the entity in question
-  - this is analagous to replacing a `&` reference
+  - this is analogous to replacing a `&` reference
 - mutating the asset changes the underlying data that the handle is pointing to
   - this affects *all* entities that point to the same asset
-  - this is analagous to mutating the data that your `&` reference is pointing to
+  - this is analogous to mutating the data that your `&` reference is pointing to
 
 To mutate a handle (90% of cases):
 
@@ -168,7 +168,7 @@ This behavior is quite useful, as it allows you to dynamically spawn and despawn
 that rely on different asset data without holding onto all of their assets forever.
 That would be, in effect, a memory leak.
 
-However, this behavior can be frustrating when trying to pre-load asssets.
+However, this behavior can be frustrating when trying to pre-load assets.
 Loading all of your assets ahead of time simply won't work if you immediately drop the handles.
 Instead, you need to hold onto them somehow.
 Resources can work well for this, as can "zoos" of loaded entities or scenes that you quickly clone into your game as needed.
@@ -180,7 +180,7 @@ Resources can work well for this, as can "zoos" of loaded entities or scenes tha
 
 {% callout(type="warning") %}
 
-Rendering is one of the most important consumers of assets in Bevy: consuming
+Rendering is one of the most important consumers of assets in Bevy: using the images and models that we load to make pretty pixels.
 However, when rendering data, we need to load it into VRAM on the GPU;
 not ordinary RAM on the CPU.
 As a result, Bevy is configured to unload [`RenderAsset`] data from the CPU by default,
--- a/content/learn/book/assets/bevy's-asset-framework.md
+++ b/content/learn/book/assets/bevy's-asset-framework.md
@@ -156,8 +156,8 @@ fn fade_enemy_image_asset(
 Rendering is one of the most important consumers of assets in Bevy: using the images and models that we load to make pretty pixels.
 In order to do so, we need to load it into VRAM on the GPU;
 not ordinary RAM on the CPU.
-As a result, Bevy is configured to unload [`RenderAsset`] data from the CPU by default,
-in order to save significant amounts of RAM.
+Because of this, some [`RenderAsset`]s may be configured to unload their CPU data,
+saving significant amounts of RAM.
 
 This can present issues when attempting to mutate asset data,
 as the actual data (like the pixel data of an image) is moved to the GPU, even though the asset remains loaded.
--- a/content/learn/book/assets/bevy's-asset-framework.md
+++ b/content/learn/book/assets/bevy's-asset-framework.md
@@ -42,7 +42,7 @@ But before that: let's load our first assets!
 [`Handle<A>`]: https://docs.rs/bevy/latest/bevy/asset/struct.Handle.html
 [`AssetPath`]: https://docs.rs/bevy/latest/bevy/asset/struct.AssetPath.html
 
-## The basics of loading assets
+## The Basics of Loading Assets
 
 Loading an asset is pretty simple:
 
@@ -101,7 +101,7 @@ This allows us to change the asset file during testing and see those changes ref
 [`AssetServer::load_state`]: https://docs.rs/bevy/latest/bevy/asset/struct.AssetServer.html#method.load_state
 [`Handle::id`]: https://docs.rs/bevy/latest/bevy/asset/struct.Handle.html#method.id
 
-## Mutating handles vs mutating assets
+## Mutating Handles vs Mutating Assets
 
 Understanding the [`Handle<A>`] / [`Assets<A>`] distinction becomes quite important when you want to mutate assets.
 Should you change the handle that your sprite holds, or the asset that the handle points to?
@@ -168,7 +168,7 @@ This behavior can be configured by setting [`RenderAssetUsages`] when loading as
 
 {% end %}
 
-## Handles are reference-counted
+## Handles are Reference-Counted
 
 The [`Handle`] type can be thought of as a [smart pointer] with two key properties:
 
--- a/content/learn/book/assets/bevy's-asset-framework.md
+++ b/content/learn/book/assets/bevy's-asset-framework.md
@@ -12,7 +12,7 @@ status = 'hidden'
 2. They are often very large: multiple megabytes, when typical data-storing components are measured in bytes.
 
 The first point means that we need tools to dynamically load (and unload) them at runtime.
-This is handled by the [`AssetServer`], which handles the surprisingly complex process of turning a path to an asset we want to use into bytes in RAM that we can stick in a Rust struct.
+This is handled by the [`AssetServer`], which handles the surprisingly complex process of turning a path of an asset we want to use, into usable data in memory.
 
 The second point means that we really don't want to be storing multiple copies of the same asset: RAM and VRAM usage is often a critical limitation for game performance.
 In Bevy, the single, authoritative version of every asset of a type `A` is stored in a matching resource: [`Assets<A>`].
@@ -80,7 +80,7 @@ in the same folder as the `Cargo.toml` of your binary crate.
 When shipping your game, this should be an `assets` folder in the same folder as the final binary.
 
 This behavior is a reasonable choice for most projects.
-However, this behavior can be overridden by setting the `BEVY_ASSET_ROOT` environment variable for your program. 
+However, this behavior can be overridden by setting the `BEVY_ASSET_ROOT` environment variable for your program or setting `AssetPlugin::file_path`.
 
 {% end %}
 
@@ -88,15 +88,7 @@ Asset loading is, by default, done asynchronously.
 This means that, while `AssetServer::load("bevy_bird.png")` will give you a [`Handle<Image>`]
 that you can use to set up your [`Sprite`], the program will not block and wait for the image to actually load.
 
-If your asset is large enough (or you are loading many assets at once),
-this can take long enough for it to be noticeable.
-Objects will look like they "didn't spawn" or look "invisible",
-only for them to suddenly "pop in".
-
-You can check the progress of asset loading using [`AssetServer::load_state`],
-looking up the asset identifier to provide by calling [`Handle::id`].
-As your project grows, you will probably want to use this method to create a loading screen (or something more sophisticated);
-waiting for all of your assets to load in before allowing the player to actually play the game.
+Because of asynchronous loading, using the handles immediately can result in objects not rendering, sounds not playing, etc. Most games handle this by creating a loading screen while their assets load. Bevy is no different! You can check the progress of asset loading using [`AssetServer::load_state`]. This can be used to create a loading screen (or something more sophisticated).
 
 {% callout(type="info") %}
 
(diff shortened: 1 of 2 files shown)


content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,205 @@
...
+title = "Bevy's Asset Framework"
+insert_anchor_links = "right"
+[extra]
+weight = 1
+status = 'hidden'
++++
+
+**Assets** have two defining qualities that shape Bevy's architecture for working with them:
+
+1. Their values are not compiled into the program. Typically, this means they're stored in the file system on the user's hard drive.
```
andriyDev: This is not true - embedded assets do have their values compiled into the program. I don't think this is a "defining quality".

Maybe we can reframe this to be "values may not be in memory yet"? Though I wonder if that's even necessary. I feel like just saying "the defining quality is assets big".

Another way we can reframe this is as being dynamic. So like part of why we want scenes to be on disk is to allow changing it without recompiling.

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,205 @@
...
+2. They are often very large: multiple megabytes, when typical data-storing components are measured in bytes.
+
+The first point means that we need tools to dynamically load (and unload) them at runtime.
+This is handled by the [`AssetServer`], which handles the surprisingly complex process of turning a path to an asset we want to use into bytes in RAM that we can stick in a Rust struct.
```
andriyDev: I don't like the way this is phrased. It makes it sound like it's just copying bytes into memory. How about: 


```suggestion
This is handled by the [`AssetServer`], which handles the surprisingly complex process of turning a path of an asset we want to use, into usable data in memory.
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,205 @@
...
+
+{% callout(type="info") %}
+
+Bevy looks for assets inside of the `assets` folder.
+During development, this should be located inside your project folder,
+in the same folder as the `Cargo.toml` of your binary crate.
+When shipping your game, this should be an `assets` folder in the same folder as the final binary.
+
+This behavior is a reasonable choice for most projects.
+However, this behavior can be overridden by setting the `BEVY_ASSET_ROOT` environment variable for your program.
```
andriyDev:
```suggestion
However, this behavior can be overridden by setting the `BEVY_ASSET_ROOT` environment variable for your program or setting `AssetPlugin::file_path`.
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,205 @@
...
+
+If your asset is large enough (or you are loading many assets at once),
+this can take long enough for it to be noticeable.
+Objects will look like they "didn't spawn" or look "invisible",
+only for them to suddenly "pop in".
+
+You can check the progress of asset loading using [`AssetServer::load_state`],
+looking up the asset identifier to provide by calling [`Handle::id`].
+As your project grows, you will probably want to use this method to create a loading screen (or something more sophisticated);
+waiting for all of your assets to load in before allowing the player to actually play the game.
```
andriyDev:
```suggestion
Because of asynchronous loading, using the handles immediately can result in objects not rendering, sounds not playing, etc. Most games handle this by creating a loading screen while their assets load. Bevy is no different! You can check the progress of asset loading using [`AssetServer::load_state`]. This can be used to create a loading screen (or something more sophisticated).
```

I don't think we need to specify Handle::id as well, since load_state accepts a &Handle.

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,205 @@
...
+When it is dropped (via replacement, despawning or removing a component, or deleting a resource that stores the handle),
+the counter decreases by one.
+If this causes the number of handles referencing the asset to hit zero, the data is unloaded.
+
+This behavior is quite useful, as it allows you to dynamically spawn and despawn entities
+that rely on different asset data without holding onto all of their assets forever.
+That would be, in effect, a memory leak.
+
+However, this behavior can be frustrating when trying to pre-load assets.
+Loading all of your assets ahead of time simply won't work if you immediately drop the handles.
```
andriyDev:
```suggestion
However, this behavior needs to be considered when trying to pre-load assets.
Loading all of your assets ahead of time simply won't work if you immediately drop the handles, since this tells the asset system you don't need the assets anymore!
```

This originally kinda sounded like it was some critical flaw in the asset system, when IMO it's very consistent behavior.

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,205 @@
...
+
+If you are trying and failing to access asset data, but are *confident* that your handles have not been dropped, you may be running into a different gotcha.
+
+Rendering is one of the most important consumers of assets in Bevy: using the images and models that we load to make pretty pixels.
+However, when rendering data, we need to load it into VRAM on the GPU;
+not ordinary RAM on the CPU.
+As a result, Bevy is configured to unload [`RenderAsset`] data from the CPU by default,
+in order to save significant amounts of RAM.
+
+This can make assets look like they have been dropped, when they've in fact been moved to the GPU,
```
andriyDev: I'm pretty sure this is no longer true. At least meshes they are definitely not removed from Assets - only their data is removed. Maybe this is still true for `Image`? I'm not certain.

andriyDev (changes requested):
I think this is a fantastic overview! Thanks!

content/learn/book/assets/bevy's-asset-framework.md
alice-i-cecile: IMO "it's a thing that you load from disk" is a key part of the mental model for beginners first learning about assets. I'll try to rephrase this though :)

content/learn/book/assets/bevy's-asset-framework.md
alice-i-cecile: Right, we made that suck less! Lemme move this up and frame it as a gotcha about mutating asset data.

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+But before that: let's load our first assets!
+
+[`AssetServer`]: https://docs.rs/bevy/latest/bevy/asset/struct.AssetServer.html
+[`Assets<A>`]: https://docs.rs/bevy/latest/bevy/asset/struct.Assets.html
+[`Mesh3d`]: https://docs.rs/bevy/latest/bevy/pbr/struct.Mesh3d.html
+[`Sprite`]: https://docs.rs/bevy/latest/bevy/sprite/struct.Sprite.html
+[`Handle<A>`]: https://docs.rs/bevy/latest/bevy/asset/struct.Handle.html
+[`AssetPath`]: https://docs.rs/bevy/latest/bevy/asset/struct.AssetPath.html
+
+## The basics of loading assets
```
Based-A:
```suggestion
## The Basics of Loading Assets
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+While loading data as an asset is more complex than simply hardcoding it, doing so unlocks [asset hot reloading](../development-practices/hot-reloading.md).
+This allows us to change the asset file during testing and see those changes reflected in real time.
+
+{% end %}
+
+[`Handle<Image>`]: https://docs.rs/bevy/latest/bevy/asset/struct.Handle.html
+[`AssetServer::load_state`]: https://docs.rs/bevy/latest/bevy/asset/struct.AssetServer.html#method.load_state
+[`Handle::id`]: https://docs.rs/bevy/latest/bevy/asset/struct.Handle.html#method.id
+
+## Mutating handles vs mutating assets
```
Based-A:
```suggestion
## Mutating Handles vs Mutating Assets
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+Should you change the handle that your sprite holds, or the asset that the handle points to?
+
+Both approaches will change what your sprite looks like, and are valid to do, but the effects differ in an important way:
+
+- mutating the handle changes the [`Image`] asset your sprite entity is pointing to
+  - this only affects the entity in question
+  - this is analogous to replacing a `&` reference
+- mutating the asset changes the underlying data that the handle is pointing to
+  - this affects *all* entities that point to the same asset
+  - this is analogous to mutating the data that your `&` reference is pointing to
```
Based-A:
```suggestion
- mutating the handle changes the [`Image`] asset your sprite entity is pointing to.
  - this only affects the entity you specify.
  - you can think of it as replacing one `&` reference with a different `&` reference.
- mutating the asset changes the underlying data that the handle is pointing to.
  - this affects *all* entities that point to the same asset.
  - this is like changing the actual data that a `&` reference is pointing to.
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+
+To mutate a handle (90% of cases):
```
Based-A, line 118:
```suggestion
Most of the time you'll probably want to mutate a handle, which can be done like this:
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+    mut sprite: Single<&mut Sprite, With<Player>>,
+    asset_server: Res<AssetServer>,
+) {
+    // Only this entity's appearance changes;
+    // other sprites are unaffected
+    sprite.image = asset_server.load("new_image.png");
+}
+```
+
+To mutate the underlying asset (10% of cases):
```
Based-A, line 131:
```suggestion
However, if you do want to mutate the underlying asset, you can use a setup like this:
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+This can present issues when attempting to mutate asset data,
+as the actual info things like "pixel data" is moved to the GPU, even though the asset remains loaded.
+This behavior can be configured by setting [`RenderAssetUsages`] when loading assets.
+
+[`RenderAsset`]: https://docs.rs/bevy/latest/bevy/render/render_asset/trait.RenderAsset.html
+[`RenderAssetUsages`]: https://docs.rs/bevy/latest/bevy/asset/struct.RenderAssetUsages.html
+
+{% end %}
+
+## Handles are reference-counted
```
Based-A:
```suggestion
## Handles are Reference-Counted
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -159,6 +151,23 @@ fn fade_enemy_image_asset(
 
 [`Image`]: https://docs.rs/bevy/latest/bevy/render/texture/struct.Image.html
 
+{% callout(type="warning") %}
+
+Rendering is one of the most important consumers of assets in Bevy: using the images and models that we load to make pretty pixels.
+In order to do so, we need to load it into VRAM on the GPU;
+not ordinary RAM on the CPU.
+As a result, Bevy is configured to unload [`RenderAsset`] data from the CPU by default,
```
andriyDev: "by default" is wrong. Maybe it shouldn't be (I'll ask in rendering-dev), but that's not the current behavior.

How about changing this to something like: 

Some [`RenderAsset`]s may be configured to unload their CPU data in order to save significant amounts of RAM.

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+2. It is [reference-counted]: when all of the handles to a given asset are dropped, the asset will be unloaded.
+
+Whenever a handle is created (or cloned), the counter of "how many handles reference this asset" goes up by one.
+When it is dropped (via replacement, despawning or removing a component, or deleting a resource that stores the handle),
+the counter decreases by one.
+If this causes the number of handles referencing the asset to hit zero, the data is unloaded.
+
+This behavior is quite useful, as it allows you to dynamically spawn and despawn entities
+that rely on different asset data without holding onto all of their assets forever.
+That would be, in effect, a memory leak.
```
Based-A:
```suggestion
If the asset data wasn't automatically unloaded, it would essentially become a memory leak.
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -159,6 +151,23 @@ fn fade_enemy_image_asset(
...
+{% callout(type="warning") %}
+
+Rendering is one of the most important consumers of assets in Bevy: using the images and models that we load to make pretty pixels.
+In order to do so, we need to load it into VRAM on the GPU;
+not ordinary RAM on the CPU.
+As a result, Bevy is configured to unload [`RenderAsset`] data from the CPU by default,
+in order to save significant amounts of RAM.
+
+This can present issues when attempting to mutate asset data,
+as the actual info things like "pixel data" is moved to the GPU, even though the asset remains loaded.
```
andriyDev:
```suggestion
as the actual data (like the pixel data of an image) is moved to the GPU, even though the asset remains loaded.
```

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+If this causes the number of handles referencing the asset to hit zero, the data is unloaded.
+
+This behavior is quite useful, as it allows you to dynamically spawn and despawn entities
+that rely on different asset data without holding onto all of their assets forever.
+That would be, in effect, a memory leak.
+
+However, this behavior needs to be considered when trying to pre-load assets.
+Loading all of your assets ahead of time simply won't work if you immediately drop the handles, since this tells the asset system you don't need the assets anymore!
+Instead, you need to hold onto them somehow.
+A resource storing something like a `HashMap<String, Handle<Image>>` can work well for this, but some games choose to create a collection of hidden, loaded entities or scenes that they can quickly clone into your game as needed.
```
Based-A, line 190: Might be too in-depth for an intro page, but mentioning the way that [`bevy_new_2d` does asset tracking](https://github.com/TheBevyFlock/bevy_new_2d/blob/main/src/asset_tracking.rs) could be a very good example of showcasing this behavior.

content/learn/book/assets/bevy's-asset-framework.md
andriyDev: Ah this is much better worded now! It's the right level of "squishy". Nice!

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+[`Handle<A>`]: https://docs.rs/bevy/latest/bevy/asset/struct.Handle.html
+[`AssetPath`]: https://docs.rs/bevy/latest/bevy/asset/struct.AssetPath.html
+
+## The basics of loading assets
+
+Loading an asset is pretty simple:
+
+1. Call `AssetServer::load("bevy_bird.png")`
+2. It gives you a handle to the asset,
+3. You put that handle inside of the component that needed asset data to control appearance, behavior or sounds.
```
Based-A:
```suggestion
1. Call `AssetServer::load("bevy_bird.png")`.
2. It will give you a handle to the asset.
3. Place that handle inside of a component that needs the asset data to control its appearance, behavior, or sounds.
```

alice-i-cecile:
Lint fixes are in #2434.

content/learn/book/assets/bevy's-asset-framework.md
```diff
@@ -0,0 +1,195 @@
...
+When shipping your game, this should be an `assets` folder in the same folder as the final binary.
+
+This behavior is a reasonable choice for most projects.
+However, this behavior can be overridden by setting the `BEVY_ASSET_ROOT` environment variable for your program or setting `AssetPlugin::file_path`.
+
+{% end %}
+
+Asset loading is, by default, done asynchronously.
+This means that, while `AssetServer::load("bevy_bird.png")` will give you a [`Handle<Image>`]
+that you can use to set up your [`Sprite`], the program will not block and wait for the image to actually load.
```
Based-A:
```suggestion
This means that the program will not block and wait for the image to actually load.
You can call `AssetServer::load("bevy_bird.png")` and receive a [`Handle<Image>`] that can be used to set up a [`Sprite`], but you might not immediately see the `"bevy_bird.png"` image in your game.
```

Based-A (commented):
Some minor edits and a couple of rephrasing that clear up some ambiguity imo.

content/learn/book/assets/bevy's-asset-framework.md
alice-i-cecile:
```suggestion
If unused asset data was never unloaded, it would be, in effect, a memory leak.
```

content/learn/book/assets/bevy's-asset-framework.md
alice-i-cecile: Yeah, this needs a whole chapter IMO. Not here!

content/learn/book/assets/bevy's-asset-framework.md
alice-i-cecile: More verbose and less helpful. I don't think we should change these.

content/learn/book/assets/bevy's-asset-framework.md
alice-i-cecile: Same as above: more words, less useful.

content/learn/book/assets/bevy's-asset-framework.md
Based-A: Fair, but I am really not liking the `(90% of cases)` and `(10% of cases)` phrases. They seem very opinionated without any explicitly stated reasons.

andriyDev: I agree with Alice. I initially felt a little uncomfortable with the percentages, but I do agree that our advice should be "you probably don't want to mutate the asset data". Maybe this could be said explicitly, but I don't think that will make it clearer, just make people stress about something that doesn't matter.

andriyDev (approved):

content/learn/book/assets/bevy's-asset-framework.md
andriyDev: To be clear: I like the percentages because it gives a sense of what you should expect in terms of importance.

alice-i-cecile: I want to give very clear guidance to new users here: mutating the handle is the thing they should almost always do. They should think carefully and second-guess themselves before mutating the underlying asset.

Based-A: I agree, but my point is more that the 90%/10% phrasing is arbitrary guidance that doesn't provide any indication of where the 90% ends and the 10% begins. If we explicitly say something like "We recommend only mutating the `Handle<A>` unless you know you need to mutate the asset data itself", it sets a clear recommendation, places the responsibility on the reader to figure out where they might want to mutate the asset itself, and also avoids any ambiguity that arises from using a specific set of values.

PR #2437, opened 2026-04-21
Write a page on the "lifetimes" of assets as well as methods on preloading them.

andriyDev:


andriyDev:
Writing this made me realize how stupid it is that we don't have nice utilities for loading. We have all the tools, we just kinda need to package them up nicely. I'm gonna see if I can just draft something up quickly. It might be easier to just implement this instead of writing a whole doc. Also I was definitely being too difficult trying to satisfy too many use cases.

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+title = "Asset Lifetimes and Preloading"
+insert_anchor_links = "right"
+[extra]
+weight = 2
++++
+
+Assets are automatically removed whenever all their handles have been dropped (with the exception of
+UUID assets). This is done through reference-counting: creating/cloning a handle increments the
+asset's reference count, dropping a handle decrements the asset's reference count, and when the
+reference count hits zero, the asset is removed (aka unloaded).
```
Based-A:
```suggestion
UUID assets). This is done by tracking every reference created for a specific asset. 
Creating or cloning an asset handle increments that asset's reference count, while dropping a handle decrements that asset's reference count. 
When the reference count hits zero, the asset is removed (aka unloaded).
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+        }
+        commands.spawn((
+            Enemy,
+            Sprite::from_image(asset_server.load("enemy.png")),
+        ));
+    }
+}
+```
+
+Put into words, this system spawns an enemy whenever the spawner's timer finishes. If the timer is
```
Based-A:
```suggestion
The `enemy_spawner` system spawns an enemy whenever the spawner's timer finishes. If the timer is
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+[extra]
+weight = 2
++++
+
+Assets are automatically removed whenever all their handles have been dropped (with the exception of
+UUID assets). This is done through reference-counting: creating/cloning a handle increments the
+asset's reference count, dropping a handle decrements the asset's reference count, and when the
+reference count hits zero, the asset is removed (aka unloaded).
+
+While this strategy is straightforward, it can lead to some issues if not considered. Imagine the
```
Based-A:
```suggestion
While this strategy is straightforward, it can cause issues if an asset is unintentionally unloaded before its finished being used. Imagine the
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
++++
+title = "Asset Lifetimes and Preloading"
+insert_anchor_links = "right"
+[extra]
+weight = 2
++++
+
+Assets are automatically removed whenever all their handles have been dropped (with the exception of
```
alice-i-cecile: This exception is probably best handled through an info box explaining the basics of UUID assets. They could also probably use a more friendly name.

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+Put into words, this system spawns an enemy whenever the spawner's timer finishes. If the timer is
+[`TimerMode::Repeating`], this would spawn an enemy at a regular interval.
+
+Now imagine that the player goes and kills all the enemies, despawning them. Since all the enemies
+have been despawned, their [`Sprite`] components will be dropped, and therefore the `"enemy.png"`
+handles will be dropped. Since there are no more handles to this asset, the image corresponding to
+`"enemy.png"` will be removed due to being unused. When the spawner then tries to spawn the enemy
+**it will trigger a reload of the asset**. This can lead to the enemy being invisible for a while
+(thanks to the rendering system gracefully handling these missing asset cases). For less flexible
+systems (e.g., third-party plugins), missing data like this may not have nice handling (some systems
```
alice-i-cecile: Wording nit: this parenthetical could be cleaned up.

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+may **panic** if the asset data isn't available). In addition, you incur the cost of needing to
```
alice-i-cecile: Wording nit: "you have to pay the cost of loading the asset from disk again."

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
++++
+title = "Asset Lifetimes and Preloading"
```
alice-i-cecile:
```suggestion
title = "Loading assets in advance"
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Waiting for Asset Loading
+
+Loading assets is not instantaneous, and it's also not blocking. That means in our
+`load_enemy_assets_in_startup` system, although we called [`AssetServer::load`], the asset may not
+have been loaded yet. So despite the term "preloading", the first few frames of our game may still
+have missing assets - which is not ideal!
+
+To address this, we need to wait for our assets to load **before** starting our game. In essence, we
+need a loading screen! There are all sorts of ways to implement a loading screen (even some that
```
alice-i-cecile: I think that the digression about "other ways to do loading screens" is distracting here. It could be a fun info boxy though!

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+may be controversial to call loading screens, like the infamous "crawl through a narrow gap"
+animation). We will be ignoring all that and focusing on how to actually tell when the assets have
+been loaded.
+
+The fundamental tools for this are [`AssetServer::is_loaded`],
```
alice-i-cecile, line 136: These methods deserve a little section (info box??) that compares and contrasts them.

content/learn/book/assets/lifetimes.md
Based-A:
```suggestion
**it will trigger a reload of the asset**. 
In this case, the enemy might temporarily become invisible while the asset is reloaded (thanks to the rendering system gracefully handling these missing asset cases). 
However, less flexible systems (e.g., third-party plugins) may not handle missing data as gracefully: some systems may **panic** if the asset data isn't available. 
In addition, you incur the cost of needing to
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+        }
+        commands.spawn((
+            Enemy,
+            Sprite::from_image(enemy_assets.sprite.clone()),
+        ));
+    }
+}
+```
+
+In a startup system, we do the initial [`AssetServer::load`], and insert that handle into a
```
Based-A:
```suggestion
In the `load_enemy_assets_in_startup` system, we do the initial [`AssetServer::load`] and insert that handle into a
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+resource. Our `enemy_spawner` code is basically the same, except now we just use the `EnemyAssets`
```
Based-A:
```suggestion
resource. Our `enemy_spawner` system is basically the same, except now we just use the `EnemyAssets`
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+        // In general, we recommend using `is_loaded_with_dependencies`, since other
+        // asset types may have more complex requirements to be usable.
+        asset_server.is_loaded_with_dependencies(&self.sprite)
+        // If we had more handles here, we could just `&&` them all together.
+    }
+}
+```
+
+Now, all we need is to prevent our `enemy_spawner` system from running until `EnemyAssets` is
+loaded. There's a lot of ways to do this, but the most out-of-the-box solution is to use states.
```
alice-i-cecile:
```suggestion
loaded. The most straightforward approach is to use states:
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+
+```rust
+#[derive(Resource, VisitAssetDependencies)]
+struct EnemyAssets {
+    #[dependency]
+    sprite: Handle<Image>,
+}
+```
+
+How does this help us? Well now we entirely replace our `enemy_assets.is_loaded()` method, with just
```
alice-i-cecile: This paragraph is unconvincing to me, pretending to be a beginner. Why is that any better?

Reading it closely, I think the point is that you don't have to write these methods at all? That's important to make clearer, and it would be good to demonstrate nesting.

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+The disadvantage of course is that while there are no enemies, that sprite remains loaded. This can
+use up memory if the sprite turns out to not be needed at all (for example, there are no enemy
+spawners). Despite this, **our standard recommendation is to do this preloading**. This makes
+gameplay logic simpler and more reliable.
+
+## Waiting for Asset Loading
+
+Loading assets is not instantaneous, and it's also not blocking. That means in our
+`load_enemy_assets_in_startup` system, although we called [`AssetServer::load`], the asset may not
+have been loaded yet. So despite the term "preloading", the first few frames of our game may still
```
Based-A:
```suggestion
Loading assets is not instantaneous, and it's also not blocking.
This means that systems will not pause and wait for the assets to load.
In our `load_enemy_assets_in_startup` system from above, we called [`AssetServer::load`] and placed the asset handle in a resource.
However, the asset may not have been fully loaded when the system finishes.
So despite us "preloading" the asset, the first few frames of our game may still
```

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+have missing assets - which is not ideal!
+
+To address this, we need to wait for our assets to load **before** starting our game. In essence, we
+need a loading screen! There are all sorts of ways to implement a loading screen (even some that
+may be controversial to call loading screens, like the infamous "crawl through a narrow gap"
+animation). We will be ignoring all that and focusing on how to actually tell when the assets have
+been loaded.
```
Based-A:
```suggestion
need a loading screen! 
There are all sorts of ways to implement a loading screen, however we will focus on creating a loading screen that finishes when the assets have fully loaded.
```

Probably isn't a good idea to include advice that might be controversial, especially if we aren't going to expand on it.

content/learn/book/assets/lifetimes.md
Based-A:
```suggestion
loaded. There's a lot of ways to do this, but the most out-of-the-box solution is to use [states].
```
I would recommend linking to the States page here.

content/learn/book/assets/lifetimes.md
```diff
@@ -0,0 +1,208 @@
...
+
+The fundamental tools for this are [`AssetServer::is_loaded`],
+[`AssetServer::is_loaded_with_dependencies`], and
+[`AssetServer::is_loaded_with_direct_dependencies`]. These allow you to query the current state of
+loads for a handle. For example, in our previous example, we could implement the following:
```
Based-A:
```suggestion
[`AssetServer::is_loaded_with_direct_dependencies`]. These allow you to query the current load state of an asset handle. 
For example, we could implement an `is_loaded` method for the `EnemyAssets` struct from before:
```

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: I reworded this slightly. I didn't like referencing "unintentionally unloading" since that's the symptom rather than the cause.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Mostly applied this. I made slight changes to this to make it clear. Also the asset literally can't be loaded yet, so that's fine.

content/learn/book/assets/lifetimes.md
andriyDev: Done. I went with the info box!

content/learn/book/assets/lifetimes.md
andriyDev: I put this in an "info box" as flavor. Also to be clear, I don't think this is this is controversial advice - I just think it may be controversial to describe that as a loading screen. Some devs might think "oh we don't have loading screens, we have this animation we play while we load assets", to which I would roll my eyes and say "that's a loading screen with extra steps".

Hopefully this isn't coming across as a criticism of the "crawl through a narrow gap" thing - that is certainly not my intent. This is a perfectly valid strategy and one we should support with open arms. "infamous" is more about how common this pattern is to sneakily hide a loading screen. So if it is coming across as "this pattern bad", we can soften the wording.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

content/learn/book/assets/lifetimes.md
andriyDev: Done.

andriyDev: We don't currently have a states page in the book afaik. I also wanted to link that here. I'd prefer not to link to the docs.rs though.

content/learn/book/assets/lifetimes.md
andriyDev: I'm not sure which one you mean? Either way, this section has been restructured a bit thanks to @Based-A. PTAL

content/learn/book/assets/lifetimes.md
andriyDev: Hopefully now it's a little clearer. I also gave an example in an info box of a nested struct.

Based-A (approved):

alice-i-cecile (approved):

PR #2441, opened 2026-04-28
Add a page describing how to make custom assets.

andriyDev:


alice-i-cecile (approved):
Lovely! There are some more advanced patterns here to talk about later (cross-referencing! non-serializable types like functions!), but this is 100% good enough to ship.

PR #2442, opened 2026-04-29
Create a new asset book page for asset processing.

andriyDev:
Hopefully this gives users a better understanding of what's going on with processing.

I think one thing this page lacks is a concrete example? Hopefully the compression one is sufficient.

### DIFF
--- a/content/learn/book/assets/processing.md
+++ b/content/learn/book/assets/processing.md
@@ -11,14 +11,15 @@ the internet, from your artists, etc), and the "processed", game-ready assets. F
 these are the same thing - you just ship whatever asset files you get. However, some assets need to
 be converted into a form that is more appropriate for use in a game.
 
-This is what asset processing was created to do. When enabled, assets in your `assets` folder are
-automatically processed by the registered processors to produce the game-ready versions of assets.
-Your game will automatically use the processed assets (without needing to change anything else).
-
 As an example, using an ultra-high resolution texture in your game is (probably) not what you want.
 This increases download sizes, slows down loading speeds, and likely reduces rendering performance.
-Usually, these high-quality versions are compressed, downscaled, etc, before being used in a game.
-This final form is our game-ready version.
+This would be your "raw" file. Usually, these high-quality versions are compressed, downscaled, etc,
+before being used in a game. This final form is our game-ready version.
+
+We can enable **asset processing** to perform this conversion. When enabled, assets in your `assets`
+folder are automatically processed by the registered processors to produce the game-ready versions
+of assets. Your game will automatically use the processed assets (without needing to change anything
+else).
 
 [% callout(type="important") %]
 
@@ -32,7 +33,7 @@ as an **import process**: importing a file could import it as another type entir
 
 {% end %}
 
-To enable asset processing, first set the `AssetPlugin::mode`:
+The first step to enable asset processing is to set the `AssetPlugin::mode` accordingly:
 
 ```rust
 fn main() {
@@ -57,7 +58,7 @@ compress and write them (internally) as the `.basisu` file format.
 
 By default, processed assets are written to `imported_assets/` (as opposed to the "regular" assets
 directory of `assets/`). **Do not** check in the processed assets as to your version control. The
-`imported_assets/` directory *should be totally ephemeral*. You should be able to delete it and Bevy
+`imported_assets/` directory _should be totally ephemeral_. You should be able to delete it and Bevy
 will automatically reprocess the assets. In other words, the "source of truth" for your assets
 during development are those in the `assets/` directory.
 
@@ -67,15 +68,17 @@ game should use!
 
 {% end %}
 
-## Writing your own processors
+## Writing your own Asset Processors
 
 Bevy provides some common processors, but there are plenty of game-specific (or even asset-specific)
-processors that may be needed by users. If you need a processor, you can make one!
+processors that may be needed by users. If you need a processor that Bevy doesn't provide, you can
+make one!
 
-The fundamental trait is the [`Process`] trait. Essentially it takes a `&mut dyn Reader` and writes
-the processed asset to a `&mut dyn Writer`. This is the most flexible (if cumbersome) version of
-this interface. This enables implementations to optimize for all sorts of use cases (for example,
-processing small chunks of the reader to avoid needing to load the entire file into memory).
+The [`Process`] trait is the most low-level trait for asset processing. The [`Process`]
+implementation is given a `&mut dyn Reader` and must write the processed asset to a `&mut dyn Writer`.
+This is the most flexible (if cumbersome) version of this interface. Implementing [`Process`]
+directly can allow you to optimize for all sorts of use cases (for example, processing small chunks
+of the reader to avoid needing to load the entire file into memory).
 
 In practice though, most users just need the simplest sort of processing: load the file, change the
 asset in some way (even maybe changing its type), then save it back out. For this, we have the
@@ -114,7 +117,9 @@ This is exactly the same asset loader as discussed in [`Custom Assets`](custom_a
 
 ### AssetTransformer
 
-This trait takes an [`AssetInput`] type and converts it into an [`AssetOutput`] type.
+This trait takes an [`AssetInput`] type and converts it into an [`AssetOutput`] type. How you do
+this conversion is totally up to you! In the example below, we just call the `NewAssetType::new`
+method with the original asset value:
 
 ```rust
 #[derive(TypePath)]
@@ -185,7 +190,7 @@ Just as with [`AssetLoader`]s, your type needs to be "encoded" somehow, whether 
 whatever else. It may be necessary to create a "serializable" version of your asset. This is
 described in more detail in [`Custom Assets`](custom_assets.md).
 
-## Meta files
+## Meta Files
 
 Throughout the description of our [`LoadTransformAndSave`] implementation, we've skipped over the
 various `Settings` associated types. These allow your processor to expose settings that can change
@@ -222,7 +227,7 @@ As mentioned earlier, we can use **any** processor that has been registered, not
 been registered as a default processor for some file extension. This allows you to create processors
 that target individual assets, rather than all assets of a particular extension.
 
-### "Load" meta files
+### "Load" Meta Files
 
 Meta files can also be used to configure settings of [`AssetLoader`]s during loading rather than
 processing. This also adds a way to "opt out" of processing a particular asset, by specifying that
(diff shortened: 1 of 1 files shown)


content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
++++
+title = "Asset Processing"
+insert_anchor_links = "right"
+[extra]
+weight = 4
++++
+
+So far, we've conceptualized assets as just "stuff on disk you want to put in memory". In practice
+though there are two "states" of assets: the "raw" files you acquire during development (from
+the internet, from your artists, etc), and the "processed", game-ready assets. For some assets,
```
Based-A, line 10: Is there any additional context that can be provided for the "processed" assets? Or maybe a better way to differentiate the "raw" versus "processed" assets? An example or real world use case could be handy, but even just something that explains the actual difference between what a "raw" file is versus a "processed" asset.

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+these are the same thing - you just ship whatever asset files you get. However, some assets need to
+be converted into a form that is more appropriate for use in a game.
+
+This is what asset processing was created to do. When enabled, assets in your `assets` folder are
```
Based-A:
```suggestion
We can enable **asset processing** to perform this conversion. When enabled, assets in your `assets` folder are
```

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+This increases download sizes, slows down loading speeds, and likely reduces rendering performance.
+Usually, these high-quality versions are compressed, downscaled, etc, before being used in a game.
+This final form is our game-ready version.
+
+[% callout(type="important") %]
+
+While it is possible to enable asset processing just before publishing to do steps like compression,
+we **strongly** recommend users choose at the beginning of their project whether to use processing
+or not. In general, processing is **not** guaranteed to be transparent. For example, the "game-ready
+version" of a `.gltf` file could actually be a `Scene` asset, which is not the same as the `Gltf`
```
Based-A, line 29: Is there a way for devs to manually handle how their assets are processed or is it strictly handled by Bevy?

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+asset - these have entirely different structures! There is no guarantee (intentionally) that the
+processed asset has the same type as the original asset. Another way to think of asset processing is
+as an **import process**: importing a file could import it as another type entirely.
+
+{% end %}
+
+To enable asset processing, first set the `AssetPlugin::mode`:
```
Based-A:
```suggestion
The first step of enabling asset processing is to set the `AssetPlugin::mode` setting:
```

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+the `asset_processor` feature on `bevy`/`bevy_asset` (this should be **unset** when publishing your
+game).
+
+Now all that's left to do is to add processors to start processing assets. For example, enabling the
+`compressed_image_saver` feature will automatically add an asset processor for `.png` files to
+compress and write them (internally) as the `.basisu` file format.
+
+{% callout(type="warning") %}
+
+By default, processed assets are written to `imported_assets/` (as opposed to the "regular" assets
```
Based-A, line 59: Should this be a separate section rather than a callout? The information feels very key, especially since it touches topics like version control and publishing.

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+will automatically reprocess the assets. In other words, the "source of truth" for your assets
+during development are those in the `assets/` directory.
+
+In contrast, when publishing, you should be publishing your `imported_assets/` directory, not the
+`assets/` directory. `imported_assets/` contains the game-ready assets which is what your published
+game should use!
+
+{% end %}
+
+## Writing your own processors
```
Based-A:
```suggestion
## Writing Your Own Asset Processor
```

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+
+Bevy provides some common processors, but there are plenty of game-specific (or even asset-specific)
+processors that may be needed by users. If you need a processor, you can make one!
```
Based-A:
```suggestion
processors that may be needed by users. If you need a processor that Bevy doesn't provide, you can make one!
```

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+
+The fundamental trait is the [`Process`] trait. Essentially it takes a `&mut dyn Reader` and writes
+the processed asset to a `&mut dyn Writer`. This is the most flexible (if cumbersome) version of
+this interface. This enables implementations to optimize for all sorts of use cases (for example,
+processing small chunks of the reader to avoid needing to load the entire file into memory).
```
Based-A:
```suggestion
The [`Process`] trait is the core of an asset processor. 
`Process` requires providing a `&mut dyn Reader` to read an asset file. 
The processed asset will then be written to a `&mut dyn Writer`.
This is the most flexible (if cumbersome) version of
this interface.
Implementing `Process` can allow you to optimize for all sorts of use cases (for example,
processing small chunks of the reader to avoid needing to load the entire file into memory).
```

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+processor.
+
+### AssetLoader
+
+This is exactly the same asset loader as discussed in [`Custom Assets`](custom_assets.md). Note that
+`LoadTransformAndSave` can only use your asset loader if it is registered.
+
+### AssetTransformer
+
+This trait takes an [`AssetInput`] type and converts it into an [`AssetOutput`] type.
```
Based-A: I think there should be a bit more explanation explicitly written out as to what exactly is happening here. Splitting up the code block below into a step-by-step process might be a good idea instead of just putting inline comments to generally explain what is happening.

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+This meta file will process its asset with the [`LoadTransformAndSave`] processor we created in the
+previous section. It will also do so with the given settings for all three of our stages. In our
+example, the settings were just `()`, so we don't have anything interesting to configure here. But
+we could!
+
+As mentioned earlier, we can use **any** processor that has been registered, not just ones that have
+been registered as a default processor for some file extension. This allows you to create processors
+that target individual assets, rather than all assets of a particular extension.
+
+### "Load" meta files
```
Based-A:
```suggestion
### "Loading" Meta Files
```

content/learn/book/assets/processing.md
```diff
@@ -0,0 +1,273 @@
...
+        Ok(NewAssetLoader::Settings::default())
+    }
+}
+```
+
+Just as with [`AssetLoader`]s, your type needs to be "encoded" somehow, whether through [`serde`] or
+whatever else. It may be necessary to create a "serializable" version of your asset. This is
+described in more detail in [`Custom Assets`](custom_assets.md).
+
+## Meta files
```
Based-A:
```suggestion
## Meta Files
```

Based-A (changes requested):
This is a fun page, lot of really cool info here. I think with a couple of smaller tweaks it'll be a great addition!

content/learn/book/assets/processing.md
andriyDev: In the future I do want a section on publishing, and we will reiterate that there. But I think this is important here since publishing is usually the last step for a user, so they may not see it until then.

content/learn/book/assets/processing.md
andriyDev: I don't like this because it sounds like a verb "oh we're loading meta files", when I want it to be a noun "these are load meta files".

content/learn/book/assets/processing.md
andriyDev: I give an example two paragraphs down. I moved it up so it's right after the initial paragraph though. I also clarified which one was the "raw" file.

content/learn/book/assets/processing.md
andriyDev: Done.

content/learn/book/assets/processing.md
andriyDev: I don't think that's important to mention here. We get to it later on with writing your own processors.

Unless I'm misunderstanding what you mean by "manually handle".

content/learn/book/assets/processing.md
andriyDev: Done (except instead of setting I said accordingly, setting felt redundant).

content/learn/book/assets/processing.md
andriyDev: Done.

content/learn/book/assets/processing.md
andriyDev: Done.

content/learn/book/assets/processing.md
andriyDev: I took inspiration from this, but I didn't like some of the details.

> `Process` requires providing a `&mut dyn Reader`

The user doesn't need to provide this in any way! The user's processor is **given** this by **Bevy**. Nevertheless I think I incorporated your intent some?

content/learn/book/assets/processing.md
andriyDev: I don't want to be very detailed here since it's very situational. There's nothing special about what we're doing in this AssetTransformer IMO. Going step by step will just make it seem "railroad"y I think like "this is how you **must** make a transformer".

I did explain a little more though.

content/learn/book/assets/processing.md
andriyDev: Done.

alice-i-cecile (approved):

content/learn/book/assets/processing.md
Based-A: I do believe my question is actually answered later on, but my initial understanding of this callout was that a developer cannot guarantee that (for example) a `.gltf` file will automatically resolve to a `Scene` asset. I think I was mostly confused by bringing up the `Gltf` asset, as my mind read it as "your asset could wind up as either a `Gltf` or `Scene` asset", but after re-reading it (and knowing the bit later on in the page) I think I know what you're saying here.

Based-A (approved):

content/learn/book/assets/processing.md
Based-A: Fair enough, although my impression of "Load" is still as a verb though. I think its just that every time I see "load" on its own I'm expecting it to be the action and not a descriptor. Maybe something like Custom Load Files would come across better?

PR #2465, opened 2026-05-22
Document common WASM-related dependency compilation error

selvmaya:
I kept running into this problem in various contexts (also outside of bevy, but often in this exact format when doing WASM things).

It seems to be considered a "known issue", but where the solution is easily forgotten. Since we want bevy to be accessible, this makes a good candidate for a troubleshooting note for this part of the quick learn section.

The solution here doesn't just repeat the written fix of adding "wasm_js" as a feature, but shows how to apply this fix in a larger dependency tree like you typically have with bevy games.

I think the voice could maybe use simpler words, but I feel like it does the job.


BenjaminBrienen (approved):

ThierryBerger (approved):

Based-A (approved):

selvmaya:
🥳

Issue #2463, opened 2026-05-22
Harper Integration


fallible-algebra:
[Harper](https://github.com/Automattic/harper) is a deterministic, offline english spelling and grammar checker. It picks up a lot of common mistakes! We should use this in CI + have a configuration so the language server can be used in local development, but not necessarily block PR merges on it.

[`harper-cli`](https://github.com/Automattic/harper/tree/master/harper-cli) is a CI-friendly frontend for harper. It's not very mature, but it's worth looking into if it does enough to be useful.

Based-A:
Is Harper better than the Typos CI step we currently have? Or would it run alongside / encompass what Typos currently does for us?

fallible-algebra:
> Is Harper better than the Typos CI step we currently have? Or would it run alongside / encompass what Typos currently does for us?

TBD. Having a separate typos run that provides clear failure may be valuable. But harper can detect typos on its own and provides really good (or at least very pretty) errors.

PR #2467, opened 2026-05-24
Initial harper integration for the book

fallible-algebra:
Address #2463, but only keep it to the book (for now).

The `harper-cli` tool is still in its infancy, but the underlying harper tooling it relies on is pretty solid. I couldn't get `cargo-binstall` working with this, so it uses another action to fetch the binary instead. Open to opinions on all of this.

Only doing this for the book for now, we should be careful about what files / directories we expand it to (if we choose to).


fallible-algebra:
(Note that this is expected to fail right now, because we've not fixed the errors in the book or properly populated the dictionary)

PR #2459, opened 2026-05-21
Asset Chapter Edits

Based-A:
Various edits across the Assets chapter in the Bevy Book.

PR #2468, opened 2026-05-24
Book: Small Syntax Fix in Callout

GrandmasterB42:
The Asset Processing Section had a small error, having [% callout(..) %] instead of {% callout(..) %}


GrandmasterB42:
Nevermind, this will probably get fixed with #2459

PR #2472, opened 2026-06-07
Mention enabling cranelift breaks variable inspection while debugging

### LINKED ISSUE #19916: Document that Cranelift breaks LLDB variable inspection
The [bevy setup guide](https://bevy.org/learn/quick-start/getting-started/setup/#Cranelift:~:text=To%20activate%20it%20for%20your%20project%2C%20add%20the%20following%20to%20your%20.cargo/config.toml.) recommends using cranelift. In general, it sounds to me (everywhere cranelift is mentioned on the internet) like [cranelift is suggested for debug builds](https://blog.rust-lang.org/inside-rust/2020/11/15/Using-rustc_codegen_cranelift/#:~:text=The%20existing%20backend,for%20debug%20builds) because it builds the code quickly, but for release builds we should [still use LLVM](https://bevy.org/learn/quick-start/getting-started/setup/#Cranelift:~:text=still%20compile%20it%20with%20LLVM.).  

## Problem Description

The current bevy setup docs make it sound like cranelift is clearly the better option for debug builds.  

It took me roughly a year to realize that cranelift is the reason why the debugger is neither working in vscode, nor in rust-rover, nor with plain lldb in the terminal. At some point I just accepted that debugging seems not to work with dynamic linking. But I have now finally verified that the dynamic linking itself is not the issue!

* Dynamic linking with the following `.cargo/config.toml` _works_:  Hitting a breakpoint, stepping, viewing local variables.  

```toml
# https://bevyengine.org/learn/quick-start/getting-started/setup/#installing-os-dependencies
# Faster builds by 30%, they say.
#
# This enables faster compiles for your binary, but builds Bevy and other dependencies with the more-optimized LLVM backend. 
#
# While cranelift is very fast to compile, the generated binaries are not optimized for speed. Additionally, it is generally still immature, so you may run into issues with it. Notably, Wasm builds do not work yet.
#
# When shipping your game, you should still compile it with LLVM.

#[unstable]
#codegen-backend = true
#
#[profile.dev]
#codegen-backend = "cranelift"
#
#[profile.dev.package."*"]
#codegen-backend = "llvm"

# Faster linking by using lld instead of rust compiler
# for Linux
[target.x86_64-unknown-linux-gnu]
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=lld",
  # (Nightly) Make the current crate share its generic instantiations
  "-Zshare-generics=y",
]

# for Windows
[target.x86_64-pc-windows-msvc]
linker = "rust-lld.exe"
``` 

* dynamic linking with the cranelift settings commented back in instead works for breakpoints and stepping but **not for viewing variables**.  `"Variables are not available"`.  

Image

* Static linking is broken in the same way. We can even remove bevy fully and observe the same problematic behavior.  
* Whether the lld settings are present or not does not change anything

## Reproducible Example
```rs
// src/main.rs
fn main() {
    let a = 1;
    // set a breakpoint here and try to view the value of `a`
    let b = 2;
    println!("Hello! {}+{}={}", a, b, a+b);
}
```

```toml
# Cargo.toml
[package]
name = "rustq-bevyerror"
version = "0.1.0"
edition = "2021"

[dependencies]

# No optimization in the dev profile.
[profile.dev]
opt-level = 0
debug = "full"

# Enable a large amount of optimization in the dev profile for dependencies.
[profile.dev.package."*"]
opt-level = 3

```

```toml
# ./.cargo/config.toml
[unstable]
codegen-backend = true

[profile.dev]
# This is the problematic setting
codegen-backend = "cranelift"

[profile.dev.package."*"]
codegen-backend = "llvm"

```

```toml
# rust-toolchain.toml
# This gives access to the latest performance improvements and "unstable" optimizations, including generic sharing.
[toolchain]
channel = "nightly"

```

```bash
$ rustup --version
rustup 1.28.2 (e4f3ad6f8 2025-04-28)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.90.0-nightly (f26e58023 2025-06-30)`
```

[cranerepro.zip](https://github.com/user-attachments/files/21008202/cranerepro.zip)

## Action
I am unsure whether this issue should be reported to cranelift (where is their issue-tracker even? Is it the same as wasmtime?) or lldb.  

But I am not really a user of either, I am only hearing about them because of my use of bevy. As such, I think it would make sense that the bevy setup guide would mention the tradeoff that using cranelift for faster debug builds will break the debugger.

invertedEcho:
Hi, I just spent a few hours trying to figure out why I can't see any variables while debugging.
Turns out that enabling cranelift, as explained in fast-compiles, may improve compile times, but also breaks variable inspection.

There is even an issue open:
https://github.com/bevyengine/bevy/issues/19916

Therefore, it really should be documented that this has negative downsides, so other people won't run into the same issue.

This closes https://github.com/bevyengine/bevy/issues/19916

### DIFF
--- a/content/learn/book/development-practices/fast-compiles.md
+++ b/content/learn/book/development-practices/fast-compiles.md
@@ -118,7 +118,10 @@ Notably, Wasm builds do not work yet.
 When shipping your game, you should still compile it with LLVM.
 
 {% callout(type="caution") %}
-Enabling cranelift is known to break variable inspection while debugging.
+Enabling cranelift is known to break local variable inspection while debugging.
+You can only inspect statics.
+This is caused by the fact that `rustc_codegen_cranelift` is missing DWARF support.
+Also see https://github.com/rust-lang/rustc_codegen_cranelift/issues/166
 {% end %}
 
 
--- a/content/learn/book/development-practices/fast-compiles.md
+++ b/content/learn/book/development-practices/fast-compiles.md
@@ -121,7 +121,7 @@ When shipping your game, you should still compile it with LLVM.
 Enabling cranelift is known to break local variable inspection while debugging.
 You can only inspect statics.
 This is caused by the fact that `rustc_codegen_cranelift` is missing DWARF support.
-Also see https://github.com/rust-lang/rustc_codegen_cranelift/issues/166
+See [`cranelift #166](https://github.com/rust-lang/rustc_codegen_cranelift/issues/166) for more information.
 {% end %}
 
 
--- a/content/learn/book/development-practices/fast-compiles.md
+++ b/content/learn/book/development-practices/fast-compiles.md
@@ -86,9 +86,6 @@ For more information, see [The rustup book: Overrides](https://rust-lang.github.
 
 ## Cranelift
 
-> [!WARNING]
-> Enabling cranelift is known to break variable inspection while debugging.
-
 This uses a new nightly-only codegen that is about 30% faster at compiling than LLVM.
 It currently works best on Linux.
 
@@ -119,6 +116,11 @@ While Cranelift is very fast to compile, the generated binaries are not optimize
 Notably, Wasm builds do not work yet.
 
 When shipping your game, you should still compile it with LLVM.
+
+{% callout(type="caution") %}
+Enabling cranelift is known to break variable inspection while debugging.
+{% end %}
+
 
 ## Generic Sharing
 
--- a/content/learn/book/development-practices/fast-compiles.md
+++ b/content/learn/book/development-practices/fast-compiles.md
@@ -86,6 +86,9 @@ For more information, see [The rustup book: Overrides](https://rust-lang.github.
 
 ## Cranelift
 
+> [!WARNING]
+> Enabling cranelift is known to break variable inspection while debugging.
+
 This uses a new nightly-only codegen that is about 30% faster at compiling than LLVM.
 It currently works best on Linux.
 
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -376,10 +376,6 @@ This section explains how to speed up iterative compiles: the amount of time it 
 
   When shipping your game, you should still compile it with LLVM.
 
-  {% callout(type="caution") %}
-  Enabling cranelift is known to break variable inspection while debugging.
-  {% end %}
-
 </details>
 
 <details>
--- a/content/learn/quick-start/getting-started/setup.md
+++ b/content/learn/quick-start/getting-started/setup.md
@@ -375,6 +375,11 @@ This section explains how to speed up iterative compiles: the amount of time it 
   Notably, Wasm builds do not work yet.
 
   When shipping your game, you should still compile it with LLVM.
+
+  {% callout(type="caution") %}
+  Enabling cranelift is known to break variable inspection while debugging.
+  {% end %}
+
 </details>
 
 <details>


invertedEcho:
I'm not really sure, seems like the docs are duplicated? I can find this section in two different places.

content/learn/quick-start/getting-started/setup.md
```diff
@@ -375,6 +375,11 @@ This section explains how to speed up iterative compiles: the amount of time it
   Notably, Wasm builds do not work yet.
 
   When shipping your game, you should still compile it with LLVM.
+
+  {% callout(type="caution") %}
+  Enabling cranelift is known to break variable inspection while debugging.
```
Based-A: The Quick Start Guide is getting revamped once the Book is released, so I'd ask to remove this part.

content/learn/book/development-practices/fast-compiles.md
```diff
@@ -117,6 +117,11 @@ Notably, Wasm builds do not work yet.
 
 When shipping your game, you should still compile it with LLVM.
 
+{% callout(type="caution") %}
+Enabling cranelift is known to break variable inspection while debugging.
```
Based-A: If you're able to, would you be able to expand on this a bit more? Anything relating to why it breaks debugging would be beneficial here.

Based-A (changes requested):
This is a good note to have, especially since we're looking to eventually wanting to expand debugging in the Book.

content/learn/quick-start/getting-started/setup.md
invertedEcho: Hmm, okay. But the quick start guide is what's currently online at bevy.org, and (at least me) it's the part that I would find, in a normal "research flow".
IMO it would still make sense to add this section, just so no other people have to waste time.
I mean the book doesn't seem available right now, only the quick start guide.

content/learn/book/development-practices/fast-compiles.md
invertedEcho: Let me know if this is fine. I don't have much knowledge here, so I just summarized what the linked issue says.

content/learn/quick-start/getting-started/setup.md
Based-A: Yup, I understand what you're saying, but the book is being made public alongside the 0.19 release, which will happen any day now. If we were still months/weeks out I wouldn't nitpick, but a lot of the content that's currently located in the Quick Start Guide is being removed and placed into the Book. This just prevents adding stuff that would be cut in a couple days anyways.

content/learn/book/development-practices/fast-compiles.md
Based-A: All good, linking out to the issue is great, thanks!

Based-A (approved):

content/learn/quick-start/getting-started/setup.md
invertedEcho: Ah, I see, now it makes sense :+1:

content/learn/book/development-practices/fast-compiles.md
```diff
@@ -117,6 +117,14 @@ Notably, Wasm builds do not work yet.
 
 When shipping your game, you should still compile it with LLVM.
 
+{% callout(type="caution") %}
+Enabling cranelift is known to break local variable inspection while debugging.
+You can only inspect statics.
+This is caused by the fact that `rustc_codegen_cranelift` is missing DWARF support.
+Also see https://github.com/rust-lang/rustc_codegen_cranelift/issues/166
```
alice-i-cecile:
```suggestion
See [`cranelift #166](https://github.com/rust-lang/rustc_codegen_cranelift/issues/166) for more information.
```

PR #2497, opened 2026-06-17
0.19: Add doc links

cart:
New strategy to semi-automate this:

1. Use IDE to find / select / copy every instance of the ``\[\`(\w)+\`\]`` regex in the release notes
2. Paste into new consolidated list
3. Sort and remove dupes (let the IDE do this)
4. Filter out "third party crate links" (and anything else that isn't a Bevy symbol)
3. Re-format the list using multi-cursor and/or find and replace  in the form: ``firefox -new-tab -url `https://dev-docs.bevy.org/bevy/index.html?search=X` -new-tab -url `https://dev-docs.bevy.org/bevy/index.html?search=Y``
4. Run the command to get a new firefox window with every tab searching for the symbol on Bevy's dev-docs
5. Alt-tab through each tab, use `/` to focus on the search bar and then UP / DOWN to navigate through the symbol list to select the correct symbol. Press enter on the correct symbol.
6. Right click a tab and click "select all tabs"
7. Right click a tab and click "bookmark tabs". Save them to a bookmark folder
8. Right click on the bookmark folder and click "copy"
9. Paste into text buffer. It now has a link to every symbol, in an order that matches the initial list.
10. Use IDE multicursor and/or find and replace to reformat original list into ``[`Symbol`]: `
11. Use IDE multicursor in buffer with the doc links to copy each link
12. Use IDE multicursor in buffer with the ``[`Symbol`]: ` list to paste each link next to the symbol.
13. Paste the final links into the release notes!

### DIFF
--- a/content/news/2026-06-14-bevy-0.19/index.md
+++ b/content/news/2026-06-14-bevy-0.19/index.md
@@ -32,11 +32,11 @@ Since our last release a few months ago we've added a _ton_ of new features, bug
 
 ### BSN (Bevy Scene Notation)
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

alice-i-cecile (changes requested):
I don't think we should link to dev-docs. Linking to the release candidate docs is a better choice for now, then we can swap it in a couple days.

PR #2510, opened 2026-06-18
0.19: Release notes for contact shadows

alice-i-cecile:
I had to do lots of research for this one, but I learned a lot! Please double-check my explanations carefully; I've done my best, but this is not my domain.

Notes for reviewers:

1. I found the Bend Studio link explaining contact shadows in the very helpful https://github.com/godotengine/godot-proposals/issues/14141. Thanks Godot team!
2. I've swapped the rather generic "solari improvements" highlight for contact shadows. I think this serves a very similar role, and is much more tangible (and pretty).
3. I've cut the render recovery section from the highlights. While I think this is important as part of our reliability story, its impact will not be appreciated in e.g. a video about Bevy's new release.
4. I would be happy with a static image or especially a slider comparison that demonstrates this effect compellingly (especially on a character), but I don't have one on hand. None of the screenshots from the [original PR](https://github.com/bevyengine/bevy/pull/22382) struck me as worth including here.


kfc35:
If it’s worth mentioning, https://github.com/bevyengine/bevy/blob/v0.19.0/examples/3d/contact_shadows.rs is the runnable example people can use to see how it works in practice

content/news/2026-06-14-bevy-0.19/contact_shadows.mp4
cart, line 1: This file is way too big / needs re-encoding before we can check it in. We generally aim for <3 MB

cart: I'm going to re-record this video because the example has been improved since the video was recorded and it looks like the example wasn't rendered with TAA, making the shadows much noisier than they should be.

cart: I swapped the video for a comparison image as I think it helps people zero in on the differences (and it saves us some page load times)

cart (approved):

PR #2512, opened 2026-06-18
0.19: Small little cool reference don't worry about it

Trashtalk217:
Add a reference to a lesser-known indie game engine.

### DIFF
--- a/content/news/2026-06-14-bevy-0.19/index.md
+++ b/content/news/2026-06-14-bevy-0.19/index.md
@@ -1371,7 +1371,10 @@ For simple assets with no sub-assets, use `SavedAsset::from_asset`:
 
 ```rust
 let main_asset = InlinedBook {
-    lines: vec!["Save me!".to_string(), "Please!".to_string()],
+    lines: vec![
+        "Waiting? So you were waiting for him?".to_string(), 
+        "Well you see-".to_string()
+    ],
 };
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I think this is fun <3

Architector4:
While funny, I feel like this may cause a moment of confusion...

"Wait, Bevy can import as assets stuff from THAT engine?!?"

cart:
I think this is likely to be confusing more than anything else. Easter eggs are fun, but they shouldn't come at the cost of example clarity. This almost reads like a "mistake" to me.

Trashtalk217:
Should I find a better place for it, or is this just a definite no?

alice-i-cecile:
If you can find a good spot for it I'm on board personally.

cart:
> Should I find a better place for it, or is this just a definite no?

Not a _definite_ "no", but I'm doubtful that we'll find something that works well. I think the big problem I have with something like this (generally) is that its a joke that most people won't "get". Just throwing in the word "godot" somewhere isn't much of a joke and has a likely "confusion factor".

Most people won't have the context of "Godot just used 'bevy' in their release blurb" (which could easily have been completely coincidental / unrelated to us). So this is an "in joke" for the few internal Bevy developers that had the conversation, and it likely doesn't improve the release notes experience for its intended audience.

Trashtalk217:
Second try / also last try.

PR #2523, opened 2026-06-20
Clarify wgpu/metal bindless support

DGriffin91:
This is a wgpu limitation, but the way it's currently worded makes it sound like it's a metal limitation. There's a PR for adding support in wgpu https://github.com/gfx-rs/wgpu/pull/9081

MoltenVK and KosmicKrisp also both support using vulkan bindless buffers on metal.


alice-i-cecile (approved):
Ah, TIL! Thanks <3

PR #2530, opened 2026-07-01
Shorten local system param example

Jengamon:
This is just a shortening of the code, in a way that takes the most advantage of the standard library, and is probably more directly usable from the book.


Jengamon:
https://github.com/bevyengine/bevy-website/pull/2173#discussion_r2190535621

PR #2533, opened 2026-07-04
Add `max-width` to images to avoid overflow

### LINKED ISSUE #2528: Profiling book page images aren't formatted correctly
The [profiling book page](https://bevy.org/learn/book/releasing-projects/profiling/) contains several images that are helpful for referencing what the different windows should look like for various profiling tools. However, the images aren't formatted corrected, which results in some or most of the images hidden from view. Opening the images in a new tab does show them fully, however it somewhat defeats the purpose of displaying them on the book page.

Example 1: https://imgur.com/Zq8hjO5

Example 2: https://imgur.com/VH1ZLxR

Not sure if it's a browser issue (I'm using Zen currently, haven't tested on other browsers) or a website issue.

doup:
If the image is bigger than the content column it overflows and the image is cut. Add max-width to `media-content` images to avoid this.

### BEFORE
image

### AFTER
image

---

Fixes #2528

### DIFF
--- a/sass/pages/_content.scss
+++ b/sass/pages/_content.scss
@@ -113,4 +113,8 @@ $content-font-size: 1.22rem;
     padding-left: 0px;
     padding-right: 0px;
   }
+
+  img {
+    max-width: 100%;
+  }
 }


Based-A (approved):
Thanks for fixing this!

PR #2541, opened 2026-07-23
Mention command that can install outdated version of Zola

selvmaya:
I wanted to look at changing some documentation, and needed to once again get zola 0.19.2 installed.

~~[Since we're waiting](https://github.com/bevyengine/bevy-website/pull/2425#issuecomment-4197113752) on zola 0.23 before being able to update it,~~ (update: no longer true!) I thought it'd be good to note down how to install a tool like zola that isn't on crates.io with a specific git tag version, without relying on a system level package manager that has an old version of the binary. This is especially relevant to people like me on systems that, excluding certain cargo installations, use a package manager that is rolling release without keeping old versions around (e.g. Arch Linux).

Avoiding `--locked` is intentional, since the old version of zola includes yanked dependencies, and did not compile for me when I initially used it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The Bevy website is built using the Zola static site engine. In our experience, 
 
 To check out any local changes you've made:
 
-1. [Install Zola](https://www.getzola.org/documentation/getting-started/installation/) version `0.19.2`.
+1. [Install Zola](https://www.getzola.org/documentation/getting-started/installation/) version `0.19.2`. [^1]
 2. Clone the Bevy Website git repo and enter that directory:
    1. `git clone https://github.com/bevyengine/bevy-website.git`
    2. `cd bevy-website`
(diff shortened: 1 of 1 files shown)


Based-A:
If we're putting in Zola installation instructions, I'd prefer a sub-item rather than a footnote. Mostly because we can then also mention that you can get the standalone binary file from the [Github release page](https://github.com/getzola/zola/releases#release-v0.19.2) as well. Zola is self-contained, which makes installing the binary preferable if you don't have Cargo installed.

selvmaya:
I've used your suggestions - sub-item over footnote, and mentioned the release page.

Based-A (approved):

Based-A:
So fun developments, we've moved to website to Zola 0.22.1 with #2548 ! With that, I don't think that the extra install points are needed, since users will be able to stick to Zola's guide on their website. I appreciate you creating this though. If you think it's still worth merging though, let us know.

selvmaya:
Assuming the idea is to again keep up with Zola and/or note down discrepancies, I think we can close this issue c:

alice-i-cecile:
Obsoleted; we updated Zola!

PR #2545, opened 2026-07-28
Cameras Book Page

Based-A:
Last "core" part of the rendering chapter, Cameras! I would appreciate some reviews on this, as information is scattered and buried for the intricacies of cameras in Bevy. I think I have a good understanding and laid out everything relevant, but let me know if I've missed anything.

I also went back and added in links to this page from the Render Pipelines page to better link them together. Might still have to expand and clarify some of the connections, but I think they currently work.


content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+[`Camera`]: https://docs.rs/bevy/latest/bevy/camera/struct.Camera.html
+[`Camera2d`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Camera2d.html
+[`Camera3d`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Camera3d.html
+[`RenderTarget`]: https://docs.rs/bevy/latest/bevy/camera/enum.RenderTarget.html
+[`Projection`]: https://docs.rs/bevy/latest/bevy/prelude/enum.Projection.html
+[`Transform`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Transform.html
+[`Projection::Orthographic`]: https://docs.rs/bevy/latest/bevy/camera/enum.Projection.html#variant.Orthographic
+[`Projection::Perspective`]: https://docs.rs/bevy/latest/bevy/camera/enum.Projection.html#variant.Perspective
+
+## Camera Driven Rendering
```
alice-i-cecile:
```suggestion
## Camera-Driven Rendering
```

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+    ));
+}
+```
+
+For example, when a player moves their mouse or moves a joystick, the `Camera` entity will usually move in response.
+The player's screen ([`RenderTarget`]) and the camera properties ([`Projection`]) are unlikely to have changed, but the rotation or position ([`Transform`]) of the `Camera` in the game world likely has changed.
+This winds up "moving" the screen from the player's perspective.
+
+If you don't set any of these components with explicit values when creating the camera entity, Bevy will supply them with default values.
+[`Transform`] will default to a location of `0, 0, 0` and no rotations, while [`RenderTarget`] will output a window to your computer's primary screen.
```
cookie1170: the game window can be on any screen, it doesn't have to be the primary

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Render Targets
+
+The point of a camera is to view what it sees, which means we need a "target" that the camera view is sent to.
+In Bevy, we refer to this as a [`RenderTarget`], which is an `enum` that describes what "target" the camera view is going to be rendered to.
+Each `RenderTarget` variant holds information about the target destination.
+For example, the [`RenderTarget::Image`] variant holds an [`ImageRenderTarget`] struct, which allows you to write a camera's view into an [`Image`] asset.
+
+It's likely you'll use the [`RenderTarget::Window`] variant the most, as this allows us to render the camera's view onto a user's desktop via a window.
+To do this, the variant requires a [`WindowRef`] enum specifying either the system's primary window, or a specific `Entity` id that represents an alternate window.
```
cookie1170:
```suggestion
To do this, the variant requires a [`WindowRef`] enum specifying either the game's primary window, or a specific `Entity` id that represents an alternate window.
```
systems don't have a primary window, the apps do

content/learn/book/the-renderer/camera-driven.md
alice-i-cecile: This is the whole thesis of the page: it needs to be introduced at the very start.

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+
+If `RenderTarget` describes the destination, then a [`Viewport`] describes where the rendered view goes in that destination.
+Every `Camera` contains a [`Viewport`] struct that defines the actual area within a `RenderTarget` that the `Camera` can render to.
+For example, if the player is playing in windowed mode (i.e. the game window doesn't take up the entire screen), then the `Viewport` represents the area of the screen that is being drawn to.
```
cookie1170: wouldn't the viewport only represent the part of the _window_ (since that's the render target) that the camera is drawing to, which would usually be the entire thing?

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+## Projections
+
+It's important to know how the images produced by each `Camera` should be displayed.
+A fundamental part of this the camera's [`Projection`].
+A `Projection` describes how objects should be translated from the virtual scene into a flattened image.
+Specifically, each projection is a 4x4 matrix that transforms points from view space (where the `Camera` is looking) into clip space (the [`RenderTarget`] that the `Camera` sends the image to).
+Objects that lie outside the bounds of the clip space are not rendered.
+
+[`Projection`] is an `enum` with `Perspective`, `Orthographic`, or `Custom` variants.
+Each variant is a wrapper containing a struct that houses the actually values each projection uses, [`PerspectiveProjection`], [`OrthographicProjection`], or [`CustomProjection`], respectively.
```
cookie1170:
```suggestion
Each variant is a wrapper containing a struct that houses the actual values each projection uses, [`PerspectiveProjection`], [`OrthographicProjection`], or [`CustomProjection`], respectively.
```

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+
+[`PerspectiveProjection`] is the default view for 3D games, making distant objects smaller than closer objects when converted into the clip space.
+You can alter the values in fields like `fov` (field of view), `near` (near clipping plane), and `far` (far clipping plane), among others, to control the projection's behavior.
+
+[`OrthographicProjection`] is the default for 2D games, where objects will stay a consistent size regardless of how far away they are from the camera.
+Like with `PerspectiveProjection`, the `near` and `far` fields adjust the distances that objects will be rendered within.
+However, because we're using `OrthographicProjection`, we don't have fields like `fov` to alter.
+Instead, fields like `viewport_origin` (the center of the `RenderTarget`) and `scale` (to control the size of objects in view) are used to control how the 2D image is shaped.
```
cookie1170: can we also mention scaling for 2d games? (`OrthographicProjection::scaling_mode`) it's something that's easy to get confused by and detrimental if messed up

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+[`CustomProjection`]: https://docs.rs/bevy/latest/bevy/camera/struct.CustomProjection.html
+[`CameraProjection`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html
+[`Projection::custom()`]: https://docs.rs/bevy/latest/bevy/camera/enum.Projection.html#method.custom
+[`CameraProjection::get_clip_from_view()`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html#tymethod.get_clip_from_view
+[`CameraProjection::update()`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html#tymethod.update
+[`CameraProjection::far()`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html#tymethod.far
+
+## Render Layers
+
+**Render Layers** allow you to layer multiple camera views on top of each other in a specified order.
```
alice-i-cecile: I don't think this is the right framing: that's what `Camera.order` does. My understanding is that render layers are about controlling the visibility of objects by group.

cookie1170: i'm pretty sure render layers don't affect the order

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+The point of a camera is to view what it sees, which means we need a "target" that the camera view is sent to.
+In Bevy, we refer to this as a [`RenderTarget`], which is an `enum` that describes what "target" the camera view is going to be rendered to.
+Each `RenderTarget` variant holds information about the target destination.
+For example, the [`RenderTarget::Image`] variant holds an [`ImageRenderTarget`] struct, which allows you to write a camera's view into an [`Image`] asset.
+
+It's likely you'll use the [`RenderTarget::Window`] variant the most, as this allows us to render the camera's view onto a user's desktop via a window.
+To do this, the variant requires a [`WindowRef`] enum specifying either the system's primary window, or a specific `Entity` id that represents an alternate window.
+
+If `RenderTarget` describes the destination, then a [`Viewport`] describes where the rendered view goes in that destination.
+Every `Camera` contains a [`Viewport`] struct that defines the actual area within a `RenderTarget` that the `Camera` can render to.
```
alice-i-cecile: I'm not sure that this is correct? It's an `Option<Viewport>` field.

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+[`CameraProjection::get_clip_from_view()`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html#tymethod.get_clip_from_view
+[`CameraProjection::update()`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html#tymethod.update
+[`CameraProjection::far()`]: https://docs.rs/bevy/latest/bevy/camera/trait.CameraProjection.html#tymethod.far
+
+## Render Layers
+
+**Render Layers** allow you to layer multiple camera views on top of each other in a specified order.
+You can add a [`RenderLayers`] component to an entity to place them on a specific render layer.
+Entities without a `RenderLayers` component are automatically placed on layer `0`, meaning that you don't have to specify a layer unless you need to.
+These also apply for `Camera`s, which means that entities placed on a render layer without a `Camera` will not be rendered.
```
cookie1170: it's worth mentioning that `RenderLayers` is a bitmask, so one entity can be on multiple render layers

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+This is why the "Prepare & Queue" step occurs in-between: we use each `Camera` to figure out what actually needs to be rendered.
+
+Bevy performs [frustum culling](https://en.wikipedia.org/wiki/Hidden-surface_determination#Viewing-frustum_culling) on the supplied data to see what is actually visible to each `Camera`.
+Draw calls are then dispatched for the entities that can be seen, while those that can't are skipped.
+Every `Camera` is evaluated separately, meaning that one `Camera` isn't dependent on another. 
+You can use multiple cameras in your game and not fear one influencing or disturbing the other.
+If desired, you can turn a `Camera` off to prevent rendering what it sees, change individual `Camera` settings to alter the look of its rendered image, or alter the order that each `Camera` is rendered in.
+
+You can use this functionality to enable a large number of features, like enabling split-screen using two separate `Camera`s that each have a [`RenderTarget`] for one half of the display screen.
```
alice-i-cecile: This doesn't seem to be how the split screen example works? https://github.com/bevyengine/bevy/blob/main/examples/3d/split_screen.rs

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+[`RenderTarget::Window`]: https://docs.rs/bevy/latest/bevy/camera/enum.RenderTarget.html#variant.Window
+[`WindowRef`]: https://docs.rs/bevy/latest/bevy/window/enum.WindowRef.html
+[`Viewport`]: https://docs.rs/bevy/latest/bevy/camera/struct.Viewport.html
+
+## Projections
+
+It's important to know how the images produced by each `Camera` should be displayed.
+A fundamental part of this the camera's [`Projection`].
+A `Projection` describes how objects should be translated from the virtual scene into a flattened image.
+Specifically, each projection is a 4x4 matrix that transforms points from view space (where the `Camera` is looking) into clip space (the [`RenderTarget`] that the `Camera` sends the image to).
```
alice-i-cecile: I don't think the clip space parenthetical is right.

content/learn/book/the-renderer/camera-driven.md
cookie1170: this doesn't really explain _what_ render layers are besides "entities placed on a render layer without a camera will not be rendered", but doesn't explain that entities only get rendered by cameras that have overlapping render layers

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+Like with `PerspectiveProjection`, the `near` and `far` fields adjust the distances that objects will be rendered within.
+However, because we're using `OrthographicProjection`, we don't have fields like `fov` to alter.
+Instead, fields like `viewport_origin` (the center of the `RenderTarget`) and `scale` (to control the size of objects in view) are used to control how the 2D image is shaped.
+
+You can also create a custom camera projection by implementing the [`CameraProjection`] trait on a struct and passing it into the [`Projection::custom()`] method.
+`CameraProjection` requires implementing several methods, like [`CameraProjection::get_clip_from_view()`], [`CameraProjection::update()`], and [`CameraProjection::far()`] among others.
+To see an example of this, we'd recommend checking out the [Custom Projection example](https://github.com/bevyengine/bevy/blob/main/examples/camera/custom_projection.rs) located in the Bevy Engine repository.
+
+[`OrthographicProjection`]: https://docs.rs/bevy/latest/bevy/prelude/struct.PerspectiveProjection.html
```
alice-i-cecile: Wrong link!

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+Then, the environment around the player is on a different render layer that is affected by the player's chosen field-of-view value.
+This ensures that the player experiences the game how they want without sacrificing the look or readability of the item they're using.
+
+[`RenderLayers`]: https://docs.rs/bevy/latest/bevy/camera/visibility/struct.RenderLayers.html
+
+## UI in Cameras
+
+`Camera` independence extends to UI elements as well.
+Each camera can have its own UI root, which will render according to the camera's settings.
+If there is only one `Camera` in the `World`, then no configuration is required.
```
alice-i-cecile, line 343: `IsDefaultUiCamera` would be useful to link here.

cookie1170 (changes requested):

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+
+[`PerspectiveProjection`] is the default view for 3D games, making distant objects smaller than closer objects when converted into the clip space.
+You can alter the values in fields like `fov` (field of view), `near` (near clipping plane), and `far` (far clipping plane), among others, to control the projection's behavior.
+
+[`OrthographicProjection`] is the default for 2D games, where objects will stay a consistent size regardless of how far away they are from the camera.
+Like with `PerspectiveProjection`, the `near` and `far` fields adjust the distances that objects will be rendered within.
+However, because we're using `OrthographicProjection`, we don't have fields like `fov` to alter.
+Instead, fields like `viewport_origin` (the center of the `RenderTarget`) and `scale` (to control the size of objects in view) are used to control how the 2D image is shaped.
+
+You can also create a custom camera projection by implementing the [`CameraProjection`] trait on a struct and passing it into the [`Projection::custom()`] method.
```
alice-i-cecile: This is not useful information IMO.

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+
+// Create a new entity with a 3D camera.
+commands.spawn(Camera3d::default());
+```
+
+You can think of the `Camera` components like a window that looks into your game's `World`. 
+Much like how looking out of a real-world window will only display what is outside that specific window, a Bevy `Camera` will only display what's in front of it.
+Unlike a real-world window though, we can alter what is seen through a `Camera` without the player having to physically move their screen.
+
+We can do this because every `Camera` is affected by three important components:
```
alice-i-cecile: This comes together very naturally to define how these are combined to create a camera frustum. Then, you can use the frustum concept later.

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+
+1. A **render target** ([`RenderTarget`]), which is the target (usually an area of the player's screen) that the `Camera` is viewed through.
```
alice-i-cecile: Hmm, I don't think that the render target is well-described here. It feels too concrete, and doesn't capture the possibilities?

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+2. A **camera projection** ([`Projection`]), controlling how the game's visuals are transformed from the game world into the rendered image.
+3. A **scene location** ([`Transform`]), which describes how the `Camera` is oriented and where it's located in the game's `World`.
```
alice-i-cecile: Scene location is a really confusing term here. It's just the camera's position.

alice-i-cecile: The camera's GlobalTransform is actually what matters.

content/learn/book/the-renderer/camera-driven.md
alice-i-cecile: This should be third in the list if anything: it's most obscure. Probably camera transform -> projection, with render target discussed elsewhere?

content/learn/book/the-renderer/camera-driven.md
Based-A: Yeah I was confused by what `Viewport` should be used for. It makes sense for rendering to a window on a screen, but does it also apply for rendering to an `Image` or `ManualTextureView`? The docs.rs page doesn't specify it, so I just kinda assumed that you could use it for non-window targets. Upon doing some more reading though, I'm assuming it's an `Option<>` since it should only be used with the window target?

alice-i-cecile (changes requested):
Lots to fix, but it's a good place to start. Really need to nail the central dogma (core data flow), and then use that throughout to thread the pedagogical structure. How about big idea -> simple one camera case -> tools for working with multiple cameras -> a few worked examples of multiple camera setups?

More suggested additions:

1. Clear color (really easy to footgun yourself with multiple cameras)
2. Viewport-to-world / world-to-viewport + viewport node widget breadcrumb

content/learn/book/the-renderer/camera-driven.md
Based-A: ~~Might've misunderstood the `WindowRef::Primary` variant. If `RenderTarget` isn't specified by the user, it'll default the primary screen? Or is it just the window that is created by Bevy?~~

Edit: scratched the question off, did some more digging and found out it's the window.

content/learn/book/the-renderer/camera-driven.md
Based-A: Yeah, I meant the computer system, not an ECS system. Need to go back and unify the terms.

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,189 @@
...
+If desired, you can turn a `Camera` off to prevent rendering what it sees, change individual `Camera` settings to alter the look of its rendered image, or alter the order that each `Camera` is rendered in.
+
+You can use this functionality to enable a large number of features, like enabling split-screen using two separate `Camera`s that each have a [`RenderTarget`] for one half of the display screen.
+Or, you could create Portal-like objects by setting the `RenderTarget` of a `Camera` to an [`Image`] handle that is then used for an object in the `World`.
+
+### 2D & 3D Rendering
+
+Like we mentioned above, you have to specify whether you want a [`Camera2d`] or a [`Camera3d`] when creating a camera entity.
+Adding this indicates that the camera entity should be included in the [rendering pipeline](/learn/book/the-renderer/render-pipelines) when a render `Schedule` runs.
+Specifically, `Camera2d` will be included in the [`Core2d`] render schedule, while `Camera3d` will be included in the [`Core3d`] schedule.
```
Zeophlite: Each `Camera2d` runs a separate invocation of the `Core2d` schedule.  Same for 3d, and for e.g. spot lights for shadow maps.

Before each schedule run, it adds `CurrentView(camera_entity)` , runs the schedule, then removes it (important to know if your writing your own `CameraRenderGraph`)

Based-A: Ah, that helps a lot, thanks for clarifying that!

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -5,166 +5,261 @@ insert_anchor_links = "right"
...
-[`Transform`] will default to a location of `0, 0, 0` and no rotations, while [`RenderTarget`] will output a window to your computer's primary screen.
-The default value for [`Projection`] varies depending on the camera type. 
-`Camera2d` defaults to [`Projection::Orthographic`], while `Camera3d` uses [`Projection::Perspective`] by default.
+```rust
+// Create a new 3D camera with a custom Perspective projection.
+commands.spawn((
+    Camera3d::default(),
+    Transform::from_xyz(0.0, 0.0, 5.0),
+    Projection::Perspective(
+        PerspectiveProjection{
```
cookie1170:
```suggestion
        PerspectiveProjection {
```
formatting

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -5,166 +5,261 @@ insert_anchor_links = "right"
...
-[`Transform`]: https://docs.rs/bevy/latest/bevy/prelude/struct.Transform.html
-[`Projection::Orthographic`]: https://docs.rs/bevy/latest/bevy/camera/enum.Projection.html#variant.Orthographic
-[`Projection::Perspective`]: https://docs.rs/bevy/latest/bevy/camera/enum.Projection.html#variant.Perspective
+```rust
+// Create a new 2D camera with a custom Orthographic projection.
+commands.spawn((
+    Camera2d::default(),
+    Transform::from_xyz(0.0, 0.0, 5.0),
+    Projection::Orthographic(
+        OrthographicProjection{
```
cookie1170:
```suggestion
        OrthographicProjection {
```
formatting

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -173,7 +268,36 @@ You can place the tool / weapon on one render layer with a fixed field-of-view v
...
 
+An example of a first person camera setup using [`RenderLayers`] is available in the [Bevy Engine examples Camera folder](https://github.com/bevyengine/bevy/blob/main/examples/camera/first_person_view_model.rs).
+
 [`RenderLayers`]: https://docs.rs/bevy/latest/bevy/camera/visibility/struct.RenderLayers.html
+[`RenderLayers::from_layers()`]: https://docs.rs/bevy/latest/bevy/camera/visibility/struct.RenderLayers.html#method.from_layers
+
+### ClearColor
+
+Whenever a `Camera` renders out sequential images or frames, there is a very brief moment when the previous frame is cleared away, but the next frame hasn't finished being written to the `Viewport`.
+What is shown during this brief moment?
```
cookie1170: that has been solved ages ago with double-buffered rendering, `ClearColor` has no relation to this

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -173,7 +268,36 @@ You can place the tool / weapon on one render layer with a fixed field-of-view v
...
+
+If we left the previous frame up, it would cause visual distortions and choppiness since parts of the next frame would be overlaid on top of the previous frame.
+We could just remove the previous frame and let the `Viewport` stay blank?
+It would work, but if the frame rate ever dropped, players might be stuck looking at a partially rendered screen.
```
cookie1170: that's not true, if vsync is on then the last frame will be shown, if it's off then they'll see screen tearing, but `ClearColor`, again, doesn't change this

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -173,7 +268,36 @@ You can place the tool / weapon on one render layer with a fixed field-of-view v
...
+
+Instead, Bevy clears the `Viewport` of the previous frame and displays a single, uniform color.
+You can think of it like a "background" color that sits behind the rendered frames.
+We call this the [`ClearColor`], a [`Resource`] that stores a [`Color`] value that all cameras will default to.
```
cookie1170: the reason for `ClearColor` is that, if the screen isn't cleared before drawing the frame, then the parts that haven't been drawn over will remain on screen, causing glitches like this, known as the "hall of mirrors effect"

https://github.com/user-attachments/assets/00c1de3a-54df-44a8-972e-6240ad93cb3b

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -173,7 +268,36 @@ You can place the tool / weapon on one render layer with a fixed field-of-view v
...
+
+However, each `Camera` is able to have a unique `ClearColor` if desired.
+The `clear_color` field holds a [`ClearColorConfig`] enum, which allows you to pick either the default [`ClearColor`] resource, a custom [`Color`] value, or `None` if you want the camera to just draw on top of whatever is already in the `Viewport`.
+
+You might be wondering why you would ever choose `None` if we just said that not clearing the `Viewport` would cause visual distortions?
+It's because clearing the `Viewport` of a `Camera` that is rendered on top of another `Camera` would (briefly) create a box of `ClearColor` that obstructs the background `Camera`.
+It might not be an issue if the `Viewport` is updated fast enough, but if the frame rate were ever to drop, players would potentially see a section of their screen covered by `ClearColor`.
+Whereas, if `ClearColorConfig` is set to `None`, then no color is used and the `Viewport` remains blank after clearing the previous screen.
```
cookie1170: the real reason you might wanna choose `None` is if you want the background of an overlay camera to be transparent and show what's behind it (or if you intentionally want to recreate the hall of mirrors effect)

cookie1170 (changes requested):

cookie1170:
i would also still like to see the section on `OrthographicProjection::scaling_mode` - it's very important to know

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -151,6 +155,34 @@ commands.spawn((
...
+Scaling objects in an Orthographic view is not a uniform process.
+While the `scale` field controls an overall multiplier on how large or small the projection will be scaled, we still have to determine how we're viewing the game world itself.
+The [`ScalingMode`] enum can be set to one of several variants that determines how objects in view of your camera will scale.
+
+The simplest variant is the [`WindowSize`] variant, where the number of pixels a window has will map 1:1 with the length of world units displayed by the projection.
+This means that a 64x64 image sprite will be displayed as 64 pixels wide and 64 pixels tall.
+However, this doesn't adapt to different screen sizes, as players with different monitor resolutions will have vastly different views of your game.
+
+Alternatively, the [`Fixed`] variant allows you to manually specify the size of the projection.
+By providing a `f32` value for both the `width` and `height` fields, the projection will always display the same region of the game world, regardless of any window resizing.
```
cookie1170, line 170: we should mention this will cause the image to stretch, which isn't usually desirable

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -151,6 +155,34 @@ commands.spawn((
...
+
+The [`AutoMin`] and [`AutoMax`] variants can be used to maintain the aspect ratio of the window, while additionally specifying that the axes can't be smaller than a given minimum or maximum.
```
cookie1170, line 172: "can be used to maintain the aspect ratio of the window" is unclear, we should say that it behaves like `WindowSize` except the size can't be smaller than a minimum or maximum

cookie1170 (changes requested):

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,355 @@
...
+Each `Camera` struct contains several fields for adjusting the camera's properties, like sub views, ordering, and even a toggle for enabling or disabling the camera.
+
+```rust
+// A basic, default camera entity.
+commands.spawn(Camera::default());
+```
+
+However, a `Camera` component doesn't contain enough information on it's own.
+Like we mentioned above, a [`CameraRenderGraph`] component also needs to be present in order for the camera to actually produce a rendered image.
+If you forget to include a `CameraRenderGraph` component, your game will emit an `error` at runtime.
```
cookie1170:
```suggestion
If you forget to include a `CameraRenderGraph` component, your game will emit an error at runtime.
```

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,355 @@
...
+    // Entities without a specified `RenderLayers` component will
+    // be placed on layer 0.
+    commands.spawn((
+        Transform::default(),
+        ...
+    ));
+}
+```
+
+A render layer is just a `bitmask` value, which means we aren't limited to only a single render layer.
```
cookie1170:
```suggestion
A render layer is a bitmask value, which means we aren't limited to only a single render layer.
```

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,355 @@
...
+You can add a `Camera` or other entity to multiple render layers by using the [`RenderLayers::from_layers()`] method and passing in a reference to an array of `usize` values.
+`RenderLayers` also provides a number of methods that correspond to different bitwise operations, allowing you to compare and contrast the render layers of different entities.
+
+```rust
+fn merge_layers(
+    mut layer_query: Query<&mut RenderLayers>,
+) {
+    // Add all entities with a `RenderLayers` component to layer 3.
+    layer_query.iter_mut().for_each(|mut layers| {
+        layers.with(3);
```
cookie1170: This function doesn't modify the `RenderLayers` itself, instead, it consumes the value and returns a new instance

content/learn/book/the-renderer/camera-driven.md
```diff
@@ -0,0 +1,355 @@
...
+## UI in Cameras
+
+`Camera` independence extends to UI elements as well.
+Each camera can have its own UI root, which will render according to the camera's settings.
+If there is only one `Camera` in the `World`, then no configuration is required.
+Simply spawn your `Camera2d` or `Camera3d` and your UI elements will be displayed within the camera's [`Viewport`].
+
+If you start using multiple cameras, you can use the [`UiTargetCamera`] component to indicate a specific camera to render the UI to.
+`UiTargetCamera` is a wrapper around an `Entity` id value for a `Camera` entity.
+Add it to the entity containing the UI elements and your UI will be rendered to the `Camera`s [`RenderTarget`] while also respecting the `Camera`s `Viewport` and scale.
```
cookie1170:
```suggestion
Add it to the entity containing the UI elements and your UI will be rendered to the `Camera`'s [`RenderTarget`] while also respecting the `Camera`'s `Viewport` and scale.
```

cookie1170 (approved):
i'm happy with the contents now, just a few nits from a final pass

PR #2547, opened 2026-07-30
Migrate to zola 0.22

viridia:
Re-write theming config file into new format.
Switch to One Dark / One Light themes.


viridia:
Note: I did this before I learned about #2425 - very similar, but not quite the same. This has some things the other PR does not.

PR #2548, opened 2026-08-01
Zola 0.22 Update

Based-A:
The long-awaited, hotly debated update to Zola 0.22.1.

- Update website CI files with latest Zola versions.
- Remove outdated syntect options.
- Add custom Bevy Syntax theme.
- Add RON grammar file and license.
- Update `config.toml` to include new theme and RON grammar file.


alice-i-cecile (changes requested):
Really happy to finally see this. A few things I noticed.

1. README isn't updated :)
2. syntaxes folder looks like it might be dead now.
3. Padding around the code blocks is waaay too high, see attached.
4. Font size seems to have increased in code blocks
5. I think we're mishandling light-mode / dark-mode support. We don't actually *have* any light-mode support (at least, I couldn't find a toggle), so I think we're likely to just accidentally ship something broken when the user's OS settings are light mode.

P.S. Do we have dead code detection? Can we set that up? This is exactly the sort of thing where you'll end up with random dead bits of CSS etc.

Before:

Image

After:

Image

viridia (approved):

viridia:
Yes we can delete the syntaxes directory.

As far as light mode theme goes, if we don't pick something it will probably pick the default. 

Claude recommends added `error_on_missing_language = true`, although we can add that in a subsequent PR. Specifically, there are places where we have code blocks with language `rust` and other places where we use language `rs`.

viridia:
> P.S. Do we have dead code detection? Can we set that up? This is exactly the sort of thing where you'll end up with random dead bits of CSS etc.

So, while there are a lot of things we can do w/r/t improving link checking and linting, there's no CSS-related tooling for detecting dead code. Claude code can do this, but you don't want this in your CI.

Based-A:
We can set a single theme for the site rather than having a light and dark theme. It would be nice to eventually get the light theme ported over to TextMate, but I don't want to sit down and manually redo the process for a light mode. I'll look into ways to try and automate it, since there are some similarities between TextMate grammar and Syntect, and supporting a light mode would be nice to do eventually.

We don't want `error_on_missing_languages` set to true right now since there are code blocks generated with options that might not be supported anymore? (specifically I was seeing them in the Errors section). Definitely worth investigating, but I don't think it's a blocking change, and I definitely don't think taking Claude's recommendations at face value are in our interests right now with regards to the website.

@alice-i-cecile What pages were you looking at with the increased font sizes and padding? I wonder if those are client-side, since the screenshots I sent in the website-dev chat didn't have those differences. Even now, running the site locally doesn't give me anything like that. Just to note, if you were looking at the 0.19 news post for the "after" `bsn!` macro code, that is an image, not a code block :)

alice-i-cecile:
Oh lol, those *are* images. I can't reproduce those specific concerns locally any more. Must have been a weird local problem.

alice-i-cecile (approved):

PR #2554, opened 2026-08-03
Convert internal links to zola's `@/` syntax for content-relative paths.

viridia:
Previously, the bevy book used primarily absolute (`"/learn/book"`) paths for links between documents, in a few cases using relative paths. Absolute paths were a problem in particular because zola cannot validate them, so there were 4 broken links that went undetected. This PR converts all the intra-document links to use the `@/` (content-relative) syntax, as well as fixing the broken links.

Note that the `@/` path points to the actual file location, not the slugified / trimmed URL that it is transforms into.

Note that the actual conversion was done by a Python script, not by hand.


Based-A (approved):
Neat, didn't know this was required for link checking.

alice-i-cecile (approved):
Annoying that this is Zola specific but we can just automatically migrate this if and when we move off Zola.

PR #2561, opened 2026-08-08
Update git checkout command for Bevy version

rafaelgdgs:
Rename the checkout branch to conform to the naming standards for releases.

It may help new people to copy paste directly without needing to do extra git commands.


ChristopherBiscardi:
This line refers to a tag, not a branch, and thus the current content is correct: https://github.com/bevyengine/bevy/tags

PR #2572, opened 2026-08-26
Book Editorial Pass: Introduction

cart:
I'm starting my read-through and editorial pass of the Bevy Book. My general priority is to add polish, improve clarity and  flow, and simplify where possible.

### DIFF
--- a/content/learn/book/intro/apps-worlds.md
+++ b/content/learn/book/intro/apps-worlds.md
@@ -7,11 +7,11 @@ status = 'hidden'
 +++
 
 The last core concepts every Bevy user needs to understand are used to organize our data.
-A [`World`] stores our ECS data, while an [`App`] stores a [`World`] and controls the outer loop for how our game actually runs.
+A **World** stores our ECS data, while an **App** stores a [`World`] and controls the outer loop for how our game actually runs.
 
 ## Worlds
 
-The world contains all the data that's in your game or application.
+Worlds contain all of the data that's in your game or application.
 Everything we've talked about so far is either data that is *stored* in a [`World`] (entities, components, resources)
 or operations that are *evaluated on* a [`World`] (systems, queries, commands).
 In the database model, the world *is* the database.
--- a/content/learn/book/intro/apps-worlds.md
+++ b/content/learn/book/intro/apps-worlds.md
@@ -7,18 +7,18 @@ status = 'hidden'
 +++
 
 The last core concepts every Bevy user needs to understand are used to organize our data.
-The **World** stores our ECS data, while an **App** stores a `World` and controls the outer loop for how our game actually runs.
+A [`World`] stores our ECS data, while an [`App`] stores a [`World`] and controls the outer loop for how our game actually runs.
 
-## The World
+## Worlds
 
 The world contains all the data that's in your game or application.
-Everything we've talked about so far is either data that is *stored* in a `World` (entities, components, resources)
-or operations that are *evaluated on* a `World` (systems, queries, commands).
+Everything we've talked about so far is either data that is *stored* in a [`World`] (entities, components, resources)
+or operations that are *evaluated on* a [`World`] (systems, queries, commands).
 In the database model, the world *is* the database.
 
-It is possible to have multiple worlds, and it's also possible to run a world completely headlessly, but the most common case by far is to run a single world inside of a single **app**.
+It is possible to have multiple worlds, and it's also possible to have a world without an app, but the most common case by far is to run a single world inside of a single **app**.
 
-## The App
+## Apps
 
 Bevy provides a modular multi-threaded runtime called an [`App`](@/learn/book/the-game-loop/app.md). If you have used web servers before, the basic ideas of an app will probably be familiar: you configure your app with settings and logic, then `run()` it to enter an update loop. It tends to look something like this:
 
@@ -35,7 +35,13 @@ fn main() {
 In most cases, your world will be contained within your app.
 The app is responsible for scheduling and executing your systems, and passing the data in and out of them appropriately.
 
-Apps can be used to combine code that's been broken down into modular, reusable pieces, called [plugins](@/learn/book/modular-architecture/plugins.md).
+Apps can also be configured using modular, reusable pieces, called [plugins](@/learn/book/modular-architecture/plugins.md):
+
+```rust
+App::new()
+    .add_plugins((DefaultPlugins, PlayerPlugin, TreePlugin))
+    .run();
+```
 
 ## Next Steps
 
@@ -43,3 +49,6 @@ With that, you should have everything you need to start [exploring our examples]
 or diving into the rest of the book.
 
 The remaining chapters of this book are nonlinear: you can read or reference them in any order.
+
+[`World`]: https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.World.html
+[`App`]: https://docs.rs/bevy/latest/bevy/app/struct.App.html
--- a/content/learn/book/intro/the-next-three-letters.md
+++ b/content/learn/book/intro/the-next-three-letters.md
@@ -7,14 +7,15 @@ status = 'hidden'
 +++
 
 If Entities, Components, and Systems are the first three concepts, then **Resources**, **Queries**, and **Commands** are the next three.
-These concepts are core to Bevy's ECS (so much so that they're used in the previous section!), but they aren't inherent to the architecture.
+
+These concepts are core to _Bevy_'s ECS, but they aren't inherent to the ECS architecture. This is an important thing to note: ECS as a general concept is a minimal, baseline set of principles. Bevy ECS is our complete _data model_. We've prioritized building something that is practical, simple, and fun to use. That means going beyond the bounds of traditional ECS!
 
 ## Resources
 
 **Resources** are global state singletons.
 Unlike [components](@/learn/book/intro/the-three-letters.md#the-c-components), which are reused across multiple entities, resources are unique. For any given resource type (like `Score` in the example below), there can only be one instance in your game world at a time.
 
-Like components, resources in Bevy are also "just Rust structs" (or enums).
+Like components, resources in Bevy are also "just Rust types":
 
 ```rs
 #[derive(Resource)]
@@ -34,7 +35,7 @@ fn update_score(mut score: ResMut<Score>) {
 
 ## Queries
 
-**Queries** are used to fetch data from the ECS, either in read only mode (like a [`&`](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#references-and-borrowing)), or in mutable access mode (like a [`&mut`](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references)).
+**Queries** are used to fetch data from the ECS, either in read only mode (like a [`&`](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#references-and-borrowing) in Rust), or in mutable access mode (like a [`&mut`](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references)).
 When writing a query, you provide a set of components, and Bevy will fetch all entities that have every requested component.
 The entities fetched may also have other components, but only the matching component types will be returned to read or write.
 
@@ -80,16 +81,16 @@ fn main() {
 {% callout(type="info") %}
 Bevy queries let you update massive amounts of game data in a tight, cache-friendly loop.
 
-Going back to our database analogy, a query is a lot like a [SQL `SELECT` statement](https://www.w3schools.com/sql/sql_select.asp): `SELECT Color, Location from World`
+Going back to our database analogy, a query is a lot like a [SQL SELECT statement](https://www.w3schools.com/sql/sql_select.asp): `SELECT Color, Location from World`
 {% end %}
 
 Queries have a lot more functionality than what's shown here.
-You can request optional components, fetch the `Entity` associated with each item, add query filters, and much more!
+You can request optional components, fetch the [`Entity`] associated with each item, add query filters, and much more!
 Queries are covered in more detail in the [Queries](@/learn/book/storing-data/queries.md) chapter.
 
 ## Commands
 
-**Commands** allow for arbitrary, deferred changes to the ECS `World`.
+**Commands** allow for arbitrary, deferred changes to the ECS [`World`].
 They are mostly used for complex write operations, such as spawning entities (as we saw in the previous chapter).
 Any system can access the command queue by adding a `mut commands: Commands` to the function signature:
 
@@ -105,3 +106,6 @@ fn spawn_entities(mut commands: Commands) {
 A wide range of built-in commands are provided,
 but you can also write custom commands to queue up any ECS logic you might desire.
 You can read about these in more detail in the [Commands](@/learn/book/control-flow/commands.md) chapter.
+
+[`Entity`]: https://docs.rs/bevy/latest/bevy/ecs/entity/struct.Entity.html
+[`World`]: https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.World.html
--- a/content/learn/book/intro/the-next-three-letters.md
+++ b/content/learn/book/intro/the-next-three-letters.md
@@ -8,7 +8,7 @@ status = 'hidden'
 
 If Entities, Components, and Systems are the first three concepts, then **Resources**, **Queries**, and **Commands** are the next three.
 
-These concepts are core to _Bevy_'s ECS, but they aren't inherent to the ECS architecture. This is an important thing to note: ECS as a general concept is a minimal, baseline set of principles. Bevy ECS is our complete _data model_. We've prioritized building something that is practical, simple, and fun to use. That means going beyond the bounds of traditional ECS!
+These concepts are core to _Bevy_'s ECS, but they aren't inherent to the ECS architecture. This is an important thing to note: ECS as a general concept is a minimal, baseline set of principles. We do not consider Bevy ECS to be constrained by these principles. We've prioritized building something that is capable, easy to understand, and fun to use. That means going beyond the bounds of traditional ECS!
 
 ## Resources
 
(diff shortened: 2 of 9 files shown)


content/learn/book/intro/apps-worlds.md
```diff
@@ -7,18 +7,18 @@ status = 'hidden'
 +++
 
 The last core concepts every Bevy user needs to understand are used to organize our data.
-The **World** stores our ECS data, while an **App** stores a `World` and controls the outer loop for how our game actually runs.
+A [`World`] stores our ECS data, while an [`App`] stores a [`World`] and controls the outer loop for how our game actually runs.
 
-## The World
+## Worlds
 
 The world contains all the data that's in your game or application.
```
sk0g: This sentence is a little confusing with the new title, since it immediately switches from plural worlds, to there being a definitive one.

content/learn/book/intro/the-next-three-letters.md
```diff
@@ -80,16 +81,16 @@ fn main() {
 {% callout(type="info") %}
 Bevy queries let you update massive amounts of game data in a tight, cache-friendly loop.
 
-Going back to our database analogy, a query is a lot like a [SQL `SELECT` statement](https://www.w3schools.com/sql/sql_select.asp): `SELECT Color, Location from World`
+Going back to our database analogy, a query is a lot like a [SQL SELECT statement](https://www.w3schools.com/sql/sql_select.asp): `SELECT Color, Location from World`
```
sk0g: Nit, but if you're capitalising SELECT, probably should capitalise FROM too.

content/learn/book/intro/the-three-letters.md
```diff
@@ -105,44 +87,44 @@ fn my_system(mut entities: Query<&mut Location>) {
...
-  - No more speculative rewrites of whole subsystems: gradually optimize the hot loops.
-- Have engine code that looks like library code that looks like game code.
-  - Weird behavior? Check the source!
+But by focusing on ECS as the heart of an engine (rather than a tacked on feature), we get significant benefits:
+
+- **Write fast, scalable code by default**:
+  - Most operations in games are of the form "look at each of these objects and do the same thing to them". Because of better [data locality], ECS architectures are much faster at iterating during these operations.
+  - Non-ECS-by-default engines either don't receive these benefits at all, or require refactoring your data to use special-cased "fast path" APIs. In Bevy, everything is fast by default using a single, unified API.
+- **Engine code, game code, and library code all look the same**:
+  - Because Bevy uses the same ECS API "everywhere" (including the engine itself), there is no line between "engine developer" and "app developer". Want to know how something in the engine works? Check the source! It will use the same patterns you're used to in your app code.
```
cookie1170: i don't think the quotes around 'everywhere' here make sense

content/learn/book/intro/the-three-letters.md
```diff
@@ -63,40 +48,37 @@ enum Color {
     Red,
     Green,
     Blue,
-    Heliotrope,
 }
 
 /// A "marker" component for entities which represents a player.
-/// Since this contains no data, this is more like a tag.
+/// This contains no data and functions as a "tag".
```
sk0g: Would this be clearer if it said "can be used as a tag"? It doesn't inherently function much like anything, just by convention you'd expect that usage.

content/learn/book/intro/the-three-letters.md
sk0g: The quote around everywhere makes this seem not true or only technically true. Is it required? Could something like practically everywhere work?

sk0g: Yeah I commented on the same lol, quotes always make me very suspicious of the claim by default.

content/learn/book/intro/the-next-three-letters.md
```diff
@@ -7,14 +7,15 @@ status = 'hidden'
 +++
 
 If Entities, Components, and Systems are the first three concepts, then **Resources**, **Queries**, and **Commands** are the next three.
-These concepts are core to Bevy's ECS (so much so that they're used in the previous section!), but they aren't inherent to the architecture.
+
+These concepts are core to _Bevy_'s ECS, but they aren't inherent to the ECS architecture. This is an important thing to note: ECS as a general concept is a minimal, baseline set of principles. Bevy ECS is our complete _data model_. We've prioritized building something that is practical, simple, and fun to use. That means going beyond the bounds of traditional ECS!
```
Based-A: "Data model" on its own isn't too clear for someone who doesn't exactly know what that means. I'd suggest either rewording it ("Bevy ECS is our implementation of an ECS model") or expanding on what a "data model" is.

content/learn/book/intro/the-three-letters.md
```diff
@@ -105,44 +87,44 @@ fn my_system(mut entities: Query<&mut Location>) {
...
+Bevy systems use a technique called [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) to access ECS data. Function parameters like [Query](@/learn/book/intro/the-next-three-letters.md#queries) and [Res](@/learn/book/intro/the-next-three-letters.md#resources) will have their data filled in for you automatically!
 {% end %}
 
 Systems usually access entities and their components via [Queries](@/learn/book/intro/the-next-three-letters.md#queries), which will be covered in the next chapter.
 
+Another cool feature: Bevy systems are generally run in parallel automatically. By inspecting the function parameter types, Bevy can automatically determine if it's safe to run two systems concurrently. For example, if you have a system which regenerates character health by modifying a `Health` component, and a different system that manages the characters' mana pool (say, via a `Mana` component), then Bevy knows that these two data sets are _disjoint_ and can be updated at the same time. This is particularly important for optimal utilization of multiple CPU cores.
```
Based-A:
```suggestion
Bevy systems are generally run in parallel automatically. By inspecting the function parameter types, Bevy can automatically determine if it's safe to run two systems concurrently. This is particularly important for optimal utilization of multiple CPU cores.

For example, if you have a system which regenerates character health by modifying a `Health` component, and a different system that manages the characters' mana pool (say, via a `Mana` component), then Bevy knows that these two data sets are _disjoint_ and can be updated at the same time. 
```

content/learn/book/intro/the-three-letters.md
```diff
@@ -105,44 +87,44 @@ fn my_system(mut entities: Query<&mut Location>) {
...
-- Write fast, scalable code by default.
-  - Most operations in games are of the form "look at each of these objects and do the same thing to them".
-  - Because of better [data locality], ECS architectures are much faster at iterating during these operations.
-  - No more speculative rewrites of whole subsystems: gradually optimize the hot loops.
-- Have engine code that looks like library code that looks like game code.
-  - Weird behavior? Check the source!
+But by focusing on ECS as the heart of an engine (rather than a tacked on feature), we get significant benefits:
+
+- **Write fast, scalable code by default**:
+  - Most operations in games are of the form "look at each of these objects and do the same thing to them". Because of better [data locality], ECS architectures are much faster at iterating during these operations.
```
Based-A:
```suggestion
  - Most operations in games take the shape of "look at each of these objects and do the same thing to them". Because of better [data locality], ECS architectures are much faster at iterating during these operations.
```

content/learn/book/intro/the-three-letters.md
```diff
@@ -105,44 +87,44 @@ fn my_system(mut entities: Query<&mut Location>) {
...
-Learning to take advantage of everything a modern ECS has to offer will take time:
-if you want to be able to tackle any data modelling problem that games have to throw at you,
-you need a lot more than just entities, components, and systems.
-Even if you're a veteran game programmer, there will be a learning curve
-as you explore new approaches and master new tools.
 
-But start simple, and add in new concepts as you encounter the problems they're solving.
-With a bit of persistence, you'll be flying in no time!
+Learning to take advantage of _everything_ a modern ECS has to offer will take time, but Bevy's ECS is actually extremely approachable!
+Start simple, and only add in new concepts and patterns when there is a problem to solve. The basics are straightforward, and if you are coming from a "traditional" engine, you can generally start by using the general patterns and structures you are used to.
```
Based-A:
```suggestion
Start simple, and only add in new concepts and patterns when there is a problem to solve. The basics are straightforward, and if you are coming from a "traditional" engine, you can start by using the general patterns and structures you are used to.
```

sass/_vars.scss
```diff
@@ -273,6 +273,7 @@ $color-white: #fff;
   --media-content-blockquote-text-color: var(--color-neutral-69);
   --media-content-button-text-color: var(--color-white);
   --media-content-link-code-color: #e4c151;
+  --media-content-code-bg-color: var(--color-neutral-18);
```
Based-A, line 276: Is adding this back in necessary? I deleted it in #2568 because I believe it was causing issues on code blocks for users who use light mode. Although as long as it isn't in the light mode section I don't think it'll cause the same issues.

content/learn/book/intro/the-three-letters.md
```diff
@@ -105,44 +87,44 @@ fn my_system(mut entities: Query<&mut Location>) {
...
 
+Another cool feature: Bevy systems are generally run in parallel automatically. By inspecting the function parameter types, Bevy can automatically determine if it's safe to run two systems concurrently. For example, if you have a system which regenerates character health by modifying a `Health` component, and a different system that manages the characters' mana pool (say, via a `Mana` component), then Bevy knows that these two data sets are _disjoint_ and can be updated at the same time. This is particularly important for optimal utilization of multiple CPU cores.
+
+## The Database Analogy
+
+ECS is very similar to an "in-memory database", both from an implementation perspective and a logical model perspective. Entities are like the "rows" of a database. The [`Entity`] id is like a "primary key". Components are like "columns". We can run queries on the data in these "tables". If you have experience with databases, you may find this analogy helpful. But you definitely don't need database experience to get the most out of Bevy!
```
Based-A, line 101: I agree with removing the database references throughout the page, but I think the placement here is a little awkward. Maybe a callout at the top of the page instead of a dedicated section might be better?

content/learn/book/intro/apps-worlds.md
```diff
@@ -7,18 +7,18 @@ status = 'hidden'
 +++
 
 The last core concepts every Bevy user needs to understand are used to organize our data.
-The **World** stores our ECS data, while an **App** stores a `World` and controls the outer loop for how our game actually runs.
+A [`World`] stores our ECS data, while an [`App`] stores a [`World`] and controls the outer loop for how our game actually runs.
```
Based-A: I think the bold is actually more applicable than the type here. I think the contributors guide says to bold new concepts when they're introduced, so while the second `World` should definitely be a type, I don't think the first `World` and `App` should be types.

content/learn/book/intro/the-three-letters.md
```diff
@@ -10,44 +10,29 @@ Bevy's architecture centers around its [ECS](https://en.wikipedia.org/wiki/Entit
...
-  - This is similar to an in-memory SQL database or spreadsheet.
-
-We'll reference both conceptual models throughout this chapter.
-For now, let's dig into each of these core elements.
-
 ## The E: Entities
 
 **Entities are objects** in our game world.
-This might include:
+This might includes things like players, enemies, inventories, buffs, trees, buildings, cameras, skyboxes, particle effects, etc.
```
theotherphil:
```suggestion
This includes things like players, enemies, inventories, buffs, trees, buildings, cameras, skyboxes, particle effects, etc.
```

janhohenheim:
I think the way you use "quotes" is what most readers would expect *italics* for ;)

content/learn/book/intro/the-next-three-letters.md
cart: Hmm. My intent here was to create some distance from "ECS" here to make the point that Bevy ECS is unconstrained by ECS tradition or dogma. I do think this needs some clarity and rephrasing. I might also move the "data model" stuff to the "Why ECS? section in the intro"

sass/_vars.scss
cart: Yes, stylistically this is important because it visually distinguishes inline code blocks (currently they just look like text), and currently the layout is "broken" in that code blocks have weird spacing / padding,  but don't have a background taking up that space. Very ugly.

content/learn/book/intro/the-three-letters.md
cart: > Maybe a callout at the top of the page instead of a dedicated section might be better?

Putting it at the end was intentional: this is more like an  "aside", not high-priority context that people need to frontload. I would rather describe ECS by functional first-principles first. They aren't complicated, and they don't require database knowledge to learn.

content/learn/book/intro/the-three-letters.md
cart: > "everywhere"

Hmmm. The goal  was to create suspicion and _build_ credibility by not making wild claims. We don't _actually_ use it everywhere (we write plenty of non-ECS code and libraries). But if a sample size of two doesn't get it, thats a pretty solid indicator that it needs to change.

cart: I think adding a qualifier like "pretty much" or "essentially" makes sense

sass/_vars.scss
Based-A: Ah, that makes sense

cart:
@janhohenheim 
> I think the way you use "quotes" is what most readers would expect italics for ;)

If you're referring to "everywhere", the point wasn't to _stress_ the word, it was to temper the claim. I think quotes are good for that, but it seems that we're 0/3 on "intention conveyance" here :)

content/learn/book/storing-data/designing-components.md
```diff
@@ -30,13 +30,13 @@ Over time, the Bevy community has converged on a few standard pieces of advice f
   - Logic that will only be repeated once generally belongs in systems.
 - **Marker components** (using unit structs) are incredibly valuable for extending your design.
   - Filtering by component presence/absence is (generally) faster and clearer than looping through a list of boolean values.
-    - Example: _Quickly look for "all entities that are a `Tower`", or "all entities that are `Chilled`"._
+    - Example:_Quickly look for "all entities that are a `Tower`", or "all entities that are `Chilled`".
```
Based-A:
```suggestion
    - Example: Quickly look for "all entities that are a `Tower`", or "all entities that are `Chilled`".
```

content/learn/book/intro/the-three-letters.md
```diff
@@ -10,44 +10,29 @@ Bevy's architecture centers around its [ECS](https://en.wikipedia.org/wiki/Entit
...
 {% callout(type="info") %}
-**Note on terminology**: Sometimes, using the word "entity" on its own can be ambiguous. Does it mean the row/id/primary key or does it mean the game object/thing it represents with all its data? In Bevy, entity ids are modeled in the `Entity` type. As a result, `Entity` typically refers to the id, and a lowercase "entity" typically refers to the game object.
+**Note on terminology**: Sometimes, using the word "entity" on its own can be ambiguous. It can mean the identifier (the `Entity` type) or it can mean the whole entity "object", with all of its stored components and behaviors. By convention, `Entity` typically refers to the id, and a lowercase "entity" typically refers to the "whole" game object.
 {% end %}
 
 ## The C: Components
 
 A **component** is a modular piece of data that can be reused across entities in the world.
-In Bevy, components are "just Rust structs" (or enums).
+In Bevy, components are just normal Rust types that implement the [`Component`] trait:
```
mnmaita, line 35:
```suggestion
In Bevy, components are just regular Rust types that implement the [`Component`] trait:
```
For some reason "regular" sounds better to me here.

content/learn/book/intro/the-three-letters.md
```diff
@@ -63,40 +48,37 @@ enum Color {
...
 
-Entities are usually spawned using [Commands](@/learn/book/control-flow/commands.md), which queue up complex work to be done later.
+Entities are usually spawned using [Commands](@/learn/book/control-flow/commands.md), which queue up work to be done later. They can also be spawned immediately using a [`World`] (more on this distinction later!).
 
 ## The S: Systems
 
 Systems interact with and update the data in the ECS.
 Each system is run every frame by default, and repeats in a loop (specifically, in a [Schedule](@/learn/book/the-game-loop/schedules.md)).
-In Bevy, systems are "just Rust functions".
+In Bevy, systems are normal Rust functions.
```
mnmaita, line 78:
```suggestion
In Bevy, systems are regular Rust functions.
```
Same here.

content/learn/book/storing-data/entities-components.md
```diff
@@ -6,7 +6,8 @@ weight = 1
 status = 'hidden'
 +++
 
-**Entities** are the fundamental objects of your game world, whizzing around, storing cameras, being controlled by the player or tracking the state of a button.
+**Entities** are the fundamental objects of your game world: players, cameras, levels, user interfaces ... everything is an entity!
```
mnmaita, line 9:
```suggestion
**Entities** are the fundamental objects of your game world: players, cameras, levels, user interfaces... everything is an entity!
```
Punctuation looks off like this.

content/learn/book/intro/the-three-letters.md
```diff
@@ -105,44 +87,46 @@ fn my_system(mut entities: Query<&mut Location>) {
...
-Learning to take advantage of everything a modern ECS has to offer will take time:
-if you want to be able to tackle any data modelling problem that games have to throw at you,
-you need a lot more than just entities, components, and systems.
-Even if you're a veteran game programmer, there will be a learning curve
-as you explore new approaches and master new tools.
 
-But start simple, and add in new concepts as you encounter the problems they're solving.
-With a bit of persistence, you'll be flying in no time!
+Learning to take advantage of _everything_ a modern ECS has to offer will take time, but Bevy's ECS is actually extremely approachable!
+Start simple, and only add in new concepts and patterns when there is a problem to solve. The basics are straightforward, and if you are coming from a "traditional" engine, you can start by using the general patterns and structures you are used to.
```
mnmaita, line 126: While I do agree 100% with the "start simple" statement, some things aren't really "translatable" from other engines/patterns/structures, so

> you can start by using the general patterns and structures you are used to

Feels a bit incomplete. As an example, it's not as straightforward to update some UI text piece "reactively" yet, which is something people might be used to nowadays. Or needing CSS knowledge to do UI. Or non-composable patterns like class inheritance to do the typical "Animal -> Cat" specializations (which we achieve with composability instead). Over time we saw people struggling because they were "fighting" against the architecture, which means these general patterns don't work well sometimes (or simply don't work).

I have no clue what to say here instead though, so you can just ignore this comment if nothing more fitting comes to mind. Maybe something along the lines of "The basics are straightforward and will take you far, and if you come from other engines this book will cover the equivalences..." (but properly written 😅)?

mnmaita (approved):
Leaving just a few small nits, these are good improvements IMO.

alice-i-cecile (approved):
I'm on board with these; merging.

PR #2578, opened 2026-09-16
Schedules Book Page Updates

### LINKED ISSUE #2445: Add advice/guidance for using `FixedUpdate` versus `Update`
Currently [the Schedules book page](https://bevy.org/learn/book/the-game-loop/schedules/) only links out to the docs.rs page in order to find an explanation for `FixedUpdate` and the `Fixed` schedules. Choosing whether to place game code in either `FixedUpdate` or regular `Update` can affect how that game code execute, and thus it should probably be at least noted (but more likely properly explained) in the book.

Based-A:
The Schedules page in the book is good, giving an overview of the basic functionality of schedules and the default schedules that Bevy provides. However, it's missing a lot of information, like interacting with the Fixed schedules, and arranging Systems inside of schedules. So here's a PR that fixes this.

- Adds a dedicated "Fixed Schedules" section within "The Standard Bevy Schedules" explaining the `FixedMain` schedule and the schedules within it, along with a brief overview of how Fixed schedules work.
- Adds an advice section for how users should use Fixed Schedules.
- Adds a section on arranging Systems in schedules, covering `.chain()`, `.before()`, and `.after()`.
  - After 0.20 lands, it might be worth adding a subsection here covering the "weak" variants of these.
- Edits the existing content:
  - Introduces `Main` earlier on.
  - Rework `Main` versus `FixedMain` to use the concepts of "frame" versus "tick", similar to how it is in the docs.rs.
  - Makes an edit to a code block to hide an extra line. 

Closes #2445


alice-i-cecile (approved):
I think this is worth including. I may reshuffle this around more later, and keep this thinner relative to the section in the Time chapter, but this is a good improvement.

PR #2588, opened 2026-09-19
Improved book query database analogy in Introduction

Mausimux:
I was reading the book (specifically the `Resources, Queries, and Commands` chapter) and found this query database analogy could be slightly improved if it refereed to the previous example components, instead of making up `Color` and `Location`.

Some images to better understand the changes:
- This is the Query usage example, some lines above the changes: image
- This is the info block following that example, which provides a database analogy: image
- And this is my proposed improvement of that database analogy using the same components as the example: image


alice-i-cecile (approved):
This LGTM, but I think it would be easier for people to follow if the systems and the corresponding select statements had their order swapped to start with the simpler form :)

Mausimux:
> ... I think it would be easier for people to follow if the systems and the corresponding select statements had their order swapped to start with the simpler form :)

Agree, but should the same change apply to the Query usage example? Or only to the database analogy?

alice-i-cecile:
Both please :) They should line up.

Based-A (approved):
Nice, I like it!

Mausimux:
> Both please :) They should line up.

Changed the order so now it looks like this: 
image

Mausimux:
Reading the example more carefully, I think it makes more sense to call the struct `Health` instead of `Life`.
I took the liberty of doing that change without asking for opinions, so let me know yours. It's easy to revert back.

It looks like this now:
image

Mausimux:
Also wondering whether the order which the systems are added to the app should be changed too.

alice-i-cecile:
> Also wondering whether the order which the systems are added to the app should be changed too.

Yes please :) Do that then this LGTM.

Mausimux:
Done!
This is ready for merge now 😉

content/learn/book/intro/the-next-three-letters.md
```diff
@@ -48,40 +48,43 @@ struct Poison {
...
+}
 ```
 
 Then, when systems are run as part of a Bevy [app](@/learn/book/the-game-loop/app.md), the engine automatically fetches the requested data, parallelizing work between systems wherever possible:
 
 ```rs
 fn main() {
     App::new()
-        .add_systems(Update, (apply_poison, tick_down_poison))
+        .add_systems(Update, (tick_down_poison, apply_poison))
```
alice-i-cecile:
```suggestion
        .add_systems(Update, (apply_poison, tick_down_poison).chain())
```

alice-i-cecile (approved):

alice-i-cecile:
Decided that we should actually do this right and demonstrate the correct system ordering. Fixed that in the last commit; merging now.

Mausimux:
I agree with the use of `chain` here, but does chain allow for system parallelization? I might be wrong but as far as I understand it it shouldn't.

I which case, should we consider updating the previous paragraph? Right now it states:
> Then, when systems are run as part of a Bevy [app](https://bevy.org/learn/book/the-game-loop/app/), the engine automatically fetches the requested data, parallelizing work between systems wherever possible

Should it, maybe, mention what `chain` is doing?
More importantly, should it clarify whether these two system are able to be parallelized?

